PR #1261, opened 2014-10-28
[API-Design] - Input validation

### LINKED ISSUE #862: Guard against unsafe inputs / sanitize inputs
For classes like Labels, they assume that a string is inputted when it could easily not be.  We should investigate our code base and make sure that arbitrary inputs can generate expected behavior.

bluong:
Right now we have no validation over our public API points meaning that we are currently forced to deal with inputs that we don't expect.

Closes #862


bluong:
Input validation is not top priority .  We'll consider this for another day.

PR #1352, opened 2014-11-19
Module definition for node.js

piotrdonicz:
Hi,
We are trying to use plottable to generate charts in Node.js, to be able to import Plottable we had to change several things:
- add module definition 
- shadow window, document and navigator objects
- add isSvg function (jsdom return  "SVG" as a nodName, so we had to check "svg" and "SVG" strings)

These changes allowed us to require Plottable but still we had to patch several functions to get a html string with valid svg chart. Here https://github.com/BrandwatchLtd/node-plottable you can find a node module that is using Plottable (it's an alpha version with patches).

The biggest issues that we have are related to calculation of the elements size (getBoundingClientRect in jsdom returns 0 values, we created a very simple function to hide overlapping labels and legends are not working because of calculation of the text dimensions), we found some memory leaks related to event listeners (they are patched as well).

 Could you let us know what do you think about this and if there is a plan to make plottable as a node package?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -10,6 +10,7 @@ module.exports = function(grunt) {
       outDir: "build/src/",
       options: {
         target: 'es5',
+        module: 'commonjs',
         noImplicitAny: true,
         sourceMap: false,
         declaration: true,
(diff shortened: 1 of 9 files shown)


bluong:
@piotrdonicz Has there been any updates?

jtlan:
Closing due to lack of activity.

PR #1472, opened 2014-12-18
Move bar label to SVGTypewriter.

endrjuskr:


kentor:
Hi, I stumbled upon this. How would I add bar labels on each bar?

bluong:
Hey @kentor !

As a prior note, when you have questions for how an API works or if you want to make a feature request, please create an issue as that will give us the best paper trail and hopefully respond to you in an expedient manner.

`Plot.Bar`s have a `barLabelsEnabled` API currently that is initially set to false.  Thus, if you set the enabled flag to true through
```typescript
barPlot.barLabelsEnabled(true)
```
then you should be able to see the bar labels.  An example can be found here: http://jsfiddle.net/bluong63/oxj3dff1/ .  Bar labels are sourced from the value attribute of the datum (the y-projection on vertical bar plots / x-projection on horizontal) and formatted through the barLabelFormatter on barPlots.

Also, the way that barLabels are implemented in Plottable is currently not necessarily in a state we would like it to be in (as this pull request suggests) so they may be rearchitected in API in the future.

kentor:
Ah thank you. I will create an issue in the future. I looked at the API docs and saw that method but nothing happened. Turns out the bars weren't wide enough.

PR #1495, opened 2015-01-06
[BarPlot] - Have scale account for bar width

### LINKED ISSUE #1280: Scales in bar plots don't account for bar width
If a bar was placed on the edge of the scale, part of the bar could extend past the edge of the scale, causing that part of the bar to clip.  The scale should be extended to account for the bar.

bluong:
Update scale extents before render based on the width of the bar

[@bluong: 1/29 14:31] EDIT: Will only perform operation if width is manually set.  If width is auto calculated then #1544 should be used.

Fixes #1280


src/components/plots/barPlot.ts
```diff
@@ -43,6 +43,65 @@ export module Plot {
...
+
+    public detach() {
+      this._datasetKeysInOrder.forEach((key) => {
+        var scaleKey = this.getID().toString() + "_" + key;
+        this._xScale._removeExtent(scaleKey, "bar-extent");
+      });
+      return this;
+    }
+
+    public remove() {
```
jtlan: No `super()`?

src/components/plots/barPlot.ts
```diff
@@ -43,6 +43,65 @@ export module Plot {
...
 
+    protected _onDatasetUpdate() {
+      super._onDatasetUpdate();
+      if (this._isAnchored) {
+        this._updateBarExtent();
+      }
+    }
+
+    public detach() {
+      this._datasetKeysInOrder.forEach((key) => {
```
jtlan: `super()`?

jtlan:
If the goal is to make sure the bars don't get drawn beyond the limits of the scale, wouldn't it make more sense to change the math in `_getBarPixelWidth()`?

bluong:
I thought the goal was extending the scale to fit the bar width?

src/components/plots/barPlot.ts
bluong: github's being dumb as the super shhould be there?

jtlan:
The primary goal is to have the bars not extend past the visible range, which could be done multiple ways.

jtlan:
Unit test?

src/components/plots/barPlot.ts
```diff
@@ -43,6 +43,67 @@ export module Plot {
...
+      return this;
+    }
+
+    public remove() {
+      super.remove();
+      this._datasetKeysInOrder.forEach((key) => {
+        var scaleKey = this.getID().toString() + "_" + key;
+        this._xScale._removeExtent(scaleKey, "bar-extent");
+      });
+      return this;
```
jtlan: `remove()` calls `detach()` (see `AbstractComponent`), so you don't need to override this.

test/components/plots/barPlotTests.ts
```diff
@@ -597,5 +597,26 @@ describe("Plots", () => {
...
+      plot.addDataset(simpleDataset)
+          .project("x", "a", xScale)
+          .project("y", "b", yScale)
+          .project("width", 200)
+          .renderTo(svg);
+
+      plot.getAllBars().each(function(selection) {
+        var bar = d3.select(this);
+        assert.operator(+bar.attr("x"), ">=", 0, "all rects right of left edge of scale");
+        assert.operator(+bar.attr("x") + +bar.attr("width"), "<=", 500, "all rects left of right edge of scale");
```
jtlan: maybe use `scale.range()` to avoid the numbers?

src/components/plots/barPlot.ts
```diff
@@ -43,6 +43,63 @@ export module Plot {
...
+      if (this._isAnchored) {
+        this._updateBarExtent();
+      }
+    }
+
+    public detach() {
+      super.detach();
+      this._datasetKeysInOrder.forEach((key) => {
+        var scaleKey = this.getID().toString() + "_" + key;
+        this._xScale._removeExtent(this.getID().toString(), "bar-extent");
```
jtlan: I wonder if `getID()` should just always return strings. There's no point in doing math on an ID...?

jtlan:
[minor] Test could use `range()` instead of magic numbers.

`Plot.Bar` is getting pretty heavy. I wonder if there's some sort of streamlining refactor in our future...

crmorford:
Can I get a hint on how to make this happen? I can't remember what usually causes this.

crmorford:
this doesn't look resolved to me (look at far left/far right bars) http://jsfiddle.net/mkwvpw4x/2/

crmorford:
Just talked offline, passing the fiddle: http://jsfiddle.net/mkwvpw4x/4/

bluong:
Closing PR due to low priority and difficulties

PR #1498, opened 2015-01-06
Update formatter quicktest to not deep-clone data

jtlan:
`.map()` makes so much more sense....


bluong:
New version looks better, but not sure why the previous version didn't work

PR #1499, opened 2015-01-06
Removing unecessary files from the top-level directory

bluong:
A lot of these files are not needed since the Plottable website contains / encapsulates this information better.

index.html is removed so localhost now shows the entire directory instead of index.html.
examples.html - examples can be found on the plottable website
tutorials directory - the tutorial zip can be found on the plottable website
removing our dependency on lodash - in the few places where it's used, it's not needed


jtlan:
Congrats on winning this month's Add/Remove ratio contest.

bluong:
This month?  I think you mean for all of Plottable history :P

PR #1515, opened 2015-01-14
Ordinal scale bands revamp

bluong:
After thinking about Ordinal Scales, it seems to be this:

Bands mode is useful for certain situations, and points mode is useful for certain situation.  However, bands mode scales the domain value to the start of the band, when interacting with the band usually makes more sense when the value is set to the center of the band.  Thus, we need to make a decision.  Does it make more sense for the default to scale to the center of the band (more work for placing the band) or does it make more sense to default to the start of the band (more work for interacting with the band).  Either way, another issue is that regardless of which we choose, we have to think about an easy way to do the other.

The next step of the revamp is likely to hide away setting the rangeType from the user.  We should either hide or show what happens to the scale and right now it seems that scales should be mostly hidden, only exposed to show connections between components.  In this case, we should have components that use an ordinal scale to explicitly set the rangeType as well as possibly the padding values on them explicity.  This is the next step.

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -319,9 +319,8 @@ export module Plot {
         attrToProjector[secondaryAttr] = (d: any, i: number, u: any, m: PlotMetadata) =>
           positionF(d, i, u, m) - widthF(d, i, u, m) * this._barAlignmentFactor;
       } else {
-        var bandWidth = (<Plottable.Scale.Ordinal> <any> secondaryScale).rangeBand();
         attrToProjector[secondaryAttr] = (d: any, i: number, u: any, m: PlotMetadata) =>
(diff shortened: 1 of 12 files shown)


src/components/plots/barPlot.ts
```diff
@@ -319,9 +319,8 @@ export module Plot {
         attrToProjector[secondaryAttr] = (d: any, i: number, u: any, m: PlotMetadata) =>
           positionF(d, i, u, m) - widthF(d, i, u, m) * this._barAlignmentFactor;
       } else {
```
jtlan: We should be able to merge the two halves of this if we only have one mode right?

jtlan:
Looks pretty straightforward. I think we should have only one mode on the scale, and remove the `rangeType()` API point -- is that going to be in this PR?

bluong:
@jtlan Not yet.  This is strictly to change the behavior of bands mode.

PR #1518, opened 2015-01-15
[Axis] - Removing instance methods from the constructor

bluong:


src/components/axes/abstractAxis.ts
```diff
@@ -416,18 +426,19 @@ export module Axis {
...
-          throw new Error("unsupported orientation");
-        }
+        AbstractAxis.verifyAxisOrientation(newOrientationLC);
         this._orientation = newOrientationLC;
         this._invalidateLayout();
         return this;
       }
     }
 
+    private static verifyAxisOrientation(orientation: string) {
```
jtlan: Maybe the lower-case checking can happen here? And then this method can return a lowercased version of the orientation.

src/components/axes/timeAxis.ts
```diff
@@ -234,12 +247,19 @@ export module Axis {
...
+      if (orientation) {
+        Time.verifyTimeAxisOrientation(orientation);
       }
       return super.orient(orientation); // maintains getter-setter functionality
     }
 
+    private static verifyTimeAxisOrientation(orientation: string) {
+      orientation = orientation.toLowerCase();
+      if (["bottom", "top"].indexOf(orientation.toLowerCase()) === -1) {
+        throw new Error(orientation + " is not a supported orientation for TimeAxis - only horizontal orientations are supported");
```
jtlan: "Axis.Time" instead of "TimeAxis". Maybe also change it to "Valid orientations are 'top' and 'bottom' "?

test/components/baseAxisTests.ts
```diff
@@ -196,16 +196,4 @@ describe("BaseAxis", () => {
 
     svg.remove();
   });
-
-  it("default alignment based on orientation", () => {
```
jtlan: Why was this test removed?

jtlan:
Looks good, but why is the test for default alignment now missing?

test/components/baseAxisTests.ts
bluong: Based on the method shuffling, since the default alignment is determined at the anchor stage (I'm trying to go with a convention of things that affect visual details should be done at the anchor step), it's not determined at the constructor step.

We could have this test still exist by setting the yAlignProportion early... I'm not sure if this is the best place to put the setting.

Also, I guess another aspect is that this touches underscored instance variables, which in my mind, means it's a lot more prone to change from source code changes easily to the point of being outdated

jtlan: I agree about touching the underscored variables. Can we test this by calling `renderTo()` on axes in a square `<svg>` and testing where their bounding boxes wind up?

jtlan:
[+1] hold based on figuring out what to do about the test.

bluong:
Test has been added back in.  After thinking about it, the way I had it before was inflexible since it would override the xAlign/yAlign with the default at the render step.  Keeping it the old way (with a few modifications) allows this to be set without overriding but using some sensible default to start with.

PR #1519, opened 2015-01-15
Components that aren't axes/plots removing instance methods from constructor

bluong:


src/components/abstractComponent.ts
```diff
@@ -274,18 +274,29 @@ export module Component {
...
+
+    private static ensureXAlignment(alignment: string): string {
+      alignment = alignment.toLowerCase();
+      if (["left", "center", "right"].indexOf(alignment) === -1) {
+        throw new Error("Unsupported alignment");
+      }
+      return alignment;
+    }
+
+    public static _xAlignmentToProportion(alignment: string): number {
```
jtlan: public _? Could this be `protected`?

src/components/abstractComponent.ts
```diff
@@ -300,18 +311,29 @@ export module Component {
...
+
+    private static ensureYAlignment(alignment: string): string {
+      alignment = alignment.toLowerCase();
+      if (["top", "center", "bottom"].indexOf(alignment) === -1) {
+        throw new Error("Unsupported alignment");
+      }
+      return alignment;
+    }
+
+    public static _yAlignmentToProportion(alignment: string): number {
```
jtlan: Same comment as for `_xAlignmentToProportion`: Could this be made `protected`?

src/components/gridlines.ts
```diff
@@ -23,15 +23,15 @@ export module Component {
...
-      }
-      if (this._yScale) {
-        this._yScale.broadcaster.registerListener(this, () => this._render());
-      }
+    }
+
+    public _anchor(element: D3.Selection) {
+      super._anchor(element);
+      this._xScale.broadcaster.registerListener(this, () => this._render());
+      this._yScale.broadcaster.registerListener(this, () => this._render());
```
jtlan: This is going to cause a bug if someone passes in `null` for either scale.

src/utils/utils.ts
```diff
@@ -344,6 +344,12 @@ export module _Util {
 
       return "#" + rHex + gHex + bHex;
     }
+
+    export function uniqAdd<T>(arr: T[], item: T) {
```
jtlan: "`uniqPush`"?

src/components/abstractComponent.ts
bluong: I don't think there's such thing as protected static... I'll check though

src/components/gridlines.ts
bluong: But the constructor I think specifies that one must exist i think...?  If not I'll change it

src/utils/utils.ts
bluong: Probably better

src/components/abstractComponent.ts
bluong: Welp apparnetly compiler is ok with this.  Will change

PR #1521, opened 2015-01-15
[Drawer] - removing instance methods from constructor

bluong:
Changing from using setter function to directly setting

### DIFF
--- a/src/drawers/rectDrawer.ts
+++ b/src/drawers/rectDrawer.ts
@@ -13,7 +13,7 @@ export module _Drawer {
 
     constructor(key: string, isVertical: boolean) {
       super(key);
-      this.svgElement("rect");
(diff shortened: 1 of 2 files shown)


src/drawers/rectDrawer.ts
```diff
@@ -13,7 +13,7 @@ export module _Drawer {
 
     constructor(key: string, isVertical: boolean) {
       super(key);
-      this.svgElement("rect");
```
jtlan: Honestly, I'm not sure why this method exists anyway. Is there even any point to being able to change the SVG element on the fly?

bluong: We will probably want to revisit Drawers.  I think it turned from something that was simple to something that is complicated and blackboxed.

PR #1522, opened 2015-01-15
[Scale] - Removing instance function calls from constructor

bluong:


src/core/domainer.ts
```diff
@@ -47,7 +47,7 @@ module Plottable {
       if (this._combineExtents != null) {
         domain = this._combineExtents(extents);
       } else if (extents.length === 0) {
-        domain = scale._defaultExtent();
+        domain = Scale.AbstractQuantitative._defaultExtent();
```
jtlan: Not sure about this change. What if we make a `Scale` whose class needs a different `_defaultExtent`, such as a non-modified Log scale?

PR #1523, opened 2015-01-16
[Plot] - Removing these instance methods from constructors

bluong:
By far the most complicated... and hopefully I didn't oversimplify things

### DIFF
--- a/src/components/plots/areaPlot.ts
+++ b/src/components/plots/areaPlot.ts
@@ -19,13 +19,18 @@ export module Plot {
     constructor(xScale: Scale.AbstractQuantitative<X>, yScale: Scale.AbstractQuantitative<number>) {
       super(xScale, yScale);
       _Util.Methods.uniqAdd(this._cssClasses, "area-plot");
-      this.project("y0", 0, yScale); // default
 
       this._animators["reset"] = new Animator.Null();
       this._animators["main"] = new Animator.Base()
                                             .duration(600)
                                             .easing("exp-in-out");
       this._defaultFillColor = new Scale.Color().range()[0];
+
+      this._projections["y0"] = {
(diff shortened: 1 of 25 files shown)


src/components/plots/areaPlot.ts
```diff
@@ -18,14 +18,19 @@ export module Plot {
...
+      this._animators["main"] = new Animator.Base()
+                                            .duration(600)
+                                            .easing("exp-in-out");
       this._defaultFillColor = new Scale.Color().range()[0];
+
+      this._projections["y0"] = {
+        accessor: d3.functor(0),
+        scale: yScale,
+        attribute: "y0"
+      };
```
jtlan: Weren't we going to remove scales from the constructor?

jtlan:
I'm not sure I see the point of replacing the methods. I know there are bugs that get caused by calling instance methods in the constructor, but *not* using them runs the risks of generating other bugs...

src/components/plots/areaPlot.ts
bluong: I believe I am ok with removing scales from the constructor, but since that cleanup has not been made yet, this is what I'm doing for now until that cleanup is made.

The reason why I'm doing this instead of project is because this is significantly more explicit on what is happening to the state of the object:  Items in the constructor should only use the inputs given to them from the constructor; it should not require any inputs from other instance variables, and calling .project does not guarantee this.

bluong:
If there is any code is duplicated, let me know and I'll see what I can do to remove some of the duplication.

The thing about the happenings inside the constructor is that methods should not be called inside that require knowledge of the state of the object.  We should be able to detect the state of the variables pretty explicitly from the inputs to the constructor.  If we can't, I feel that stems from a lack of understanding of our own architecture.

PR #1538, opened 2015-01-20
Enabling compatibility with require for loading d3

### LINKED ISSUE #1525: Correctly add d3 when require.js is present
Currently Plottable expects to find d3 on window but require.js prevents this from happening.  If require is present then need to call require("d3").

bluong:
Fixes #1525


src/reference.ts
```diff
@@ -1,5 +1,7 @@
 /// <reference path="../bower_components/svg-typewriter/svgtypewriter.d.ts" />
 
+/// <reference path="core/init.ts" />
```
jtlan: Does this guarantee that **init.ts** is run first?

bluong: I thought reference.ts defines the order in which the typescript files are compiled / built.  Is this incorrect?

jtlan:
Let's not do this -- it doesn't make sense for us to build a feature to work with a version of D3 we don't support.

bluong:
As of 3.4.10, d3 is always registered in the global namespace, so this isn't really needed. https://github.com/mbostock/d3/pull/1921

PR #1539, opened 2015-01-20
[BarPlot] - Remove checking if data is invalid for barPixelWidth

bluong:
This needed to exist before since we needed to guard for default accessors being invalid.  This is no longer the case, so this somewhat hefty computation can be removed.

QE: We probably want to regression pass over this for specifically bar charts.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -7010,14 +7010,6 @@ var Plottable;
                 }
                 else {
                     var barAccessor = this._isVertical ? this._projections["x"].accessor : this._projections["y"].accessor;
-                    var barAccessorData = d3.set(Plottable._Util.Methods.flatten(this._datasetKeysInOrder.map(function (k) {
-                        var dataset = _this._key2PlotDatasetKey.get(k).dataset;
-                        var plotMetadata = _this._key2PlotDatasetKey.get(k).plotMetadata;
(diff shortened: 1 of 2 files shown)


jtlan:
It's still possible for the data to be undefined though, right?

bluong:
Yes, but right now we have an inconsistent way of dealing with invalid data.  We may want to discuss possible failure modes and how to deal with them.  The way it is being dealt with currently in this function is pretty clunky

jtlan:
Okay, but not sure if this is NOQE. Do we have tests for invalid data?

bluong:
I believe we don't have tests for invalid data, but the plot won't even load if there was even invalid data, same as for pretty much every other plot we have.

crmorford:
Still want to talk to you about what changed more specifically before +1-ing, but the regression pass looks good

bluong:
@crmorford There was some code in a private function that would check if the bar placement data were numbers before doing the automatic width calculation, and this code was run at the render step.

If the bar placement data was not all numbers then this code would crash since the calculation is expecting numbers, however in this case, the bar plot wouldn't even render, so the validation is just costly and doesn't do anything.

crmorford:
okay, GIM

PR #1543, opened 2015-01-21
resetZoom() when scale domains change in Interaction.PanZoom.

### LINKED ISSUE #1388: PanZoomInteraction: resetZoom when scale is changed externally
I know there's a planned rewrite of PZIs, but I've seen at least 3 different implementations of extending PZIs to do this.

Could we add a resetZoom on domain change as a stopgap till interactions are rewritten?

Related: #753 #254

jtlan:
HACKHACK for now until we get to rewriting `Interaction.PanZoom`.

Close #1388.

### DIFF
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -28,7 +28,7 @@ export module Interaction {
       if (yScale) {
         this._yScale = yScale;
         // HACKHACK #1388: self-register for resetZoom()
-        this._xScale.broadcaster.registerListener("pziY" + this.getID(), () => this.resetZoom());
+        this._yScale.broadcaster.registerListener("pziY" + this.getID(), () => this.resetZoom());
       }
     }
 
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -20,18 +20,16 @@ export module Interaction {
      */
     constructor(xScale?: Scale.AbstractQuantitative<any>, yScale?: Scale.AbstractQuantitative<any>) {
       super();
-      if (xScale == null) {
-        xScale = new Plottable.Scale.Linear();
+      if (xScale) {
+        this._xScale = xScale;
+        // HACKHACK #1388: self-register for resetZoom()
+        this._xScale.broadcaster.registerListener("pziX" + this.getID(), () => this.resetZoom());
       }
-      if (yScale == null) {
-        yScale = new Plottable.Scale.Linear();
+      if (yScale) {
+        this._yScale = yScale;
+        // HACKHACK #1388: self-register for resetZoom()
+        this._xScale.broadcaster.registerListener("pziY" + this.getID(), () => this.resetZoom());
       }
-      this._xScale = xScale;
-      this._yScale = yScale;
-      this._zoom = d3.behavior.zoom();
-      this._zoom.x((<any> this._xScale)._d3Scale);
-      this._zoom.y((<any> this._yScale)._d3Scale);
-      this._zoom.on("zoom", () => this._rerenderZoomed());
     }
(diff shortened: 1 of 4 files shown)


jtlan:
I had trouble coming up with a test for this that wasn't tautological. @bluong, any ideas?

bluong:
Unsure if this makes sense, but maybe... 
Do you now a simplificiation of what ._zoom.x / ._zoom.y does?  Theoretically there is some state in there that changes that essentially matches the point of this fix.

Given this...

 if you change the domain of the scale, then you should look for the simple change shown above.

(When you say tautological, I assume you mean that testing if resetZoom() is called when the domain changes is a tautological test.  You should probably go for an effect that's a lot more concrete/simpler to reason like "Does this variable change to what I want? or similar)

src/interactions/panZoomInteraction.ts
```diff
@@ -20,18 +20,16 @@ export module Interaction {
...
+        // HACKHACK #1388: self-register for resetZoom()
+        this._yScale.broadcaster.registerListener("pziY" + this.getID(), () => this.resetZoom());
       }
-      this._xScale = xScale;
-      this._yScale = yScale;
-      this._zoom = d3.behavior.zoom();
-      this._zoom.x((<any> this._xScale)._d3Scale);
-      this._zoom.y((<any> this._yScale)._d3Scale);
-      this._zoom.on("zoom", () => this._rerenderZoomed());
     }
```
bluong: Will there we a problem if this._zoom is not initialized until resetZoom?

jtlan: see `_anchor()`, which calls `resetZoom()`.

src/interactions/panZoomInteraction.ts
```diff
@@ -40,24 +38,33 @@ export module Interaction {
...
+        this._zoom.y((<any> this._yScale)._d3Scale);
+      }
       this._zoom.on("zoom", () => this._rerenderZoomed());
       this._zoom(this._hitBox);
     }
 
     public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
       super._anchor(component, hitBox);
-      this._zoom(hitBox);
+      this.resetZoom();
```
bluong: Is this the initialization step?

test/tests.js
```diff
@@ -6550,6 +6550,24 @@ describe("Interactions", function () {
...
+        it("Resets zoom when the scale domain changes", function () {
+            var xScale = new Plottable.Scale.Linear();
+            var yScale = new Plottable.Scale.Linear();
+            var svg = generateSVG();
+            var c = new Plottable.Component.AbstractComponent();
+            c.renderTo(svg);
+            var pzi = new Plottable.Interaction.PanZoom(xScale, yScale);
+            c.registerInteraction(pzi);
+            var zoomBeforeX = pzi._zoom;
+            xScale.domain([10, 1000]);
```
bluong: Can you enact a zoom before this line? so we can actually see the scale reset

test/tests.js
```diff
@@ -6550,6 +6550,24 @@ describe("Interactions", function () {
...
+            var zoomAfterX = pzi._zoom;
+            assert.notStrictEqual(zoomBeforeX, zoomAfterX, "D3 Zoom was regenerated after x scale domain changed");
+            var zoomBeforeY = pzi._zoom;
+            yScale.domain([10, 1000]);
```
bluong: Same here.

bluong:
Just a test issue

crmorford:
Definitely an improvement, GIM

src/interactions/panZoomInteraction.ts
jtlan: Yes.

PR #1544, opened 2015-01-21
[BarPlot] - Auto width calculation takes border in account

bluong:
Auto width currently makes sure that none of the bars overlap.  Now it makes sure that it doesn't extend over the edge

[@bluong 1/29 14:34] EDIT: If user does not specify the barWidth, then this will be used to gate the barWidths at the edge of the scale.


src/components/plots/barPlot.ts
```diff
@@ -388,7 +388,20 @@ export module Plot {
...
+        if (this._barAlignmentFactor !== 0 && minScaledDatum > 0) {
+          barPixelWidth = Math.min(barPixelWidth, minScaledDatum / this._barAlignmentFactor);
+        }
+        var maxScaledDatum = _Util.Methods.max(scaledData, 0);
+        if (this._barAlignmentFactor !== 1 && maxScaledDatum < barWidthDimension) {
+          var margin = barWidthDimension - maxScaledDatum;
+          barPixelWidth = Math.min(barPixelWidth, margin / (1 - this._barAlignmentFactor));
+        }
+
+        barPixelWidth *= 0.95;
```
jtlan: Is it possible to store off the `0.4` and `0.95` somewhere?

jtlan:
[minor] magic number storage

src/components/plots/barPlot.ts
bluong: Yep.  WIll change

jtlan:
Err, I meant pull them out as member/static variables, something where they're accessible outside of the method....

bluong:
I consdereed pulling them out as member static variables but my first thought is they shouldn't be considered global... Do we want them to be?

crmorford:
Note: Same data, very different appearance
Before:
screen shot 2015-02-05 at 2 06 51 pm
After:
screen shot 2015-02-05 at 2 07 00 pm

That being said, looks good. I just like recording this stuff somewhere.

PR #1546, opened 2015-01-22
Upgrade to typescript-1.4

bluong:
Run
`npm update && npm install`
to update your typescript.

We should regression pass this as this is a library/language upgrade?

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -113,9 +113,9 @@ declare module Plottable {
              * @returns {[{ [key: string]: any }} coppied map.
              */
             function copyMap<T>(oldMap: {
-                [x: string]: T;
+                [key: string]: T;
             }): {
-                [x: string]: T;
+                [key: string]: T;
             };
             function range(start: number, stop: number, step?: number): number[];
             /** Is like setTimeout, but activates synchronously if time=0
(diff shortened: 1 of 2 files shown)


plottable.d.ts
```diff
@@ -113,9 +113,9 @@ declare module Plottable {
              * @returns {[{ [key: string]: any }} coppied map.
              */
             function copyMap<T>(oldMap: {
-                [x: string]: T;
+                [key: string]: T;
             }): {
-                [x: string]: T;
+                [key: string]: T;
```
jtlan: ?? I'm bewildered that TS1.3 didn't do this correctly.

plottable.d.ts
```diff
@@ -3745,7 +3745,7 @@ declare module Plottable {
              */
             _doHover(p: Point): HoverData;
         }
-        class Hover extends AbstractInteraction {
+        class Hover extends Interaction.AbstractInteraction {
```
jtlan: Why does this only happen here and not for `Click extends AbstractInteraction`?

jtlan:
No real problems, just confused by inconsistencies that the upgrade seems to have introduced.

Also, no JS code was changed, so we can probably NOQE this?

plottable.d.ts
bluong: Weird.  This is really inconsistent?

bluong:
Oh didn't notice that no JS code has changed.  Yep should be NOQE

PR #1547, opened 2015-01-23
[PiePlot] PiePlot will throw a warning if it has negative values.

### LINKED ISSUE #1505: Pie plot should throw error when given negative numbers
http://jsfiddle.net/ajxeeh29/
Pie plots try to include negative values, and the chart doesn't break in a very obvious way, which may make debugging difficult.

rcchen:
Close #1505

### DIFF
--- a/src/drawers/arcDrawer.ts
+++ b/src/drawers/arcDrawer.ts
@@ -43,6 +43,11 @@ export module _Drawer {
                           .value(valueAccessor)(data);
 
       drawSteps.forEach(s => delete s.attrToProjector["value"]);
+      pie.forEach((slice) => {
+        if (slice.value < 0) {
+          _Util.Methods.warn("Negative values will not render correctly in a pie chart.");
+        }
+      });
       return super.draw(pie, drawSteps, userMetadata, plotMetadata);
     }
   }
--- a/test/components/plots/piePlotTests.ts
+++ b/test/components/plots/piePlotTests.ts
@@ -166,11 +166,11 @@ describe("Plots", () => {
     });
 
     it("throws warnings on negative data", () => {
-      var message;
+      var message: String;
       var oldWarn = Plottable._Util.Methods.warn;
       Plottable._Util.Methods.warn = (warn) => message = warn;
       piePlot.removeDataset("simpleDataset");
-      negativeDataset = new Plottable.Dataset([{value: -5, value2: 10, type: "A"}, {value: 15, value2: 10, type: "B"}]);
+      var negativeDataset = new Plottable.Dataset([{value: -5}, {value: 15}]);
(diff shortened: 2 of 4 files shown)


src/components/plots/piePlot.ts
```diff
@@ -37,6 +37,14 @@ export module Plot {
         _Util.Methods.warn("Only one dataset is supported in Pie plots");
         return this;
       }
+      var keyAndData = super._getKeyAndData(keyOrDataset, dataset);
+      // Check to see whether data is a Dataset or an array of values
+      var data = "_data" in keyAndData.data ? keyAndData.data._data : keyAndData.data;
+      data.forEach((datum) => {
+        if (datum.value < 0) {
+          _Util.Methods.warn("Negative values will not render correctly in a pie chart.");
```
jtlan: I think this is the wrong place to check this. There's no guarantee that the `value` property of each datum is actually driving the `value` projection. Additionally, even if it was, there's no guarantee that the `project()` call was passed an accessor that returns negative values.

jtlan:
I should probably explain the `project()` process to you in more detail. Come find me.

test/components/plots/piePlotTests.ts
```diff
@@ -164,5 +164,17 @@ describe("Plots", () => {
       });
 
     });
+
+    it("throws warnings on negative data", () => {
+      var message;
```
bluong: I'm surprised tslint didnt complain.

This should be 
var message: string;

test/components/plots/piePlotTests.ts
```diff
@@ -164,5 +164,17 @@ describe("Plots", () => {
...
+      var oldWarn = Plottable._Util.Methods.warn;
+      Plottable._Util.Methods.warn = (warn) => message = warn;
+      piePlot.removeDataset("simpleDataset");
+      negativeDataset = new Plottable.Dataset([{value: -5, value2: 10, type: "A"}, {value: 15, value2: 10, type: "B"}]);
```
bluong: I don't think the value2 part is necessary.  Removing this attribute would mke the test simpler

src/drawers/arcDrawer.ts
```diff
@@ -43,6 +43,11 @@ export module _Drawer {
                           .value(valueAccessor)(data);
 
       drawSteps.forEach(s => delete s.attrToProjector["value"]);
+      pie.forEach((slice) => {
+        if (slice.value < 0) {
+          _Util.Methods.warn("Negative values will not render correctly in a pie chart.");
+        }
+      });
```
bluong: This is fine for now.  Just noting that we will want to handle these types of verification checks in a more first-class way.  This and the similar answers we have are not ideal.

crmorford:
looks good :)

PR #1549, opened 2015-01-23
[Scale.Color] - Correctly parses color strings into rgb

### LINKED ISSUE #1527: Scale.Color produces odd results when used with named colors
Repro: http://jsfiddle.net/mj26qxxs/
Expected: Bars are red and green.
Actual: Bars are black.

Inspecting element reveals this:
screen shot 2015-01-19 at 2 41 15 pm
It appears auto-lightening code doesn't work correctly with named colors ("red", "green", etc).

rcchen:
Fixes #1527

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -113,9 +113,9 @@ declare module Plottable {
              * @returns {[{ [key: string]: any }} coppied map.
              */
             function copyMap<T>(oldMap: {
-                [x: string]: T;
+                [key: string]: T;
             }): {
-                [x: string]: T;
+                [key: string]: T;
             };
             function range(start: number, stop: number, step?: number): number[];
             /** Is like setTimeout, but activates synchronously if time=0
@@ -493,7 +493,7 @@ declare module Plottable {
          *
          * The listeners are called synchronously.
          */
-        class Broadcaster extends PlottableObject {
+        class Broadcaster extends Core.PlottableObject {
             listenable: Listenable;
             /**
              * Constructs a broadcaster, taking the Listenable that the broadcaster will be attached to.
@@ -1422,7 +1422,7 @@ declare module Plottable {
     module Scale {
         module TickGenerators {
             interface TickGenerator<D> {
-                (scale: AbstractQuantitative<D>): D[];
+                (scale: Plottable.Scale.AbstractQuantitative<D>): D[];
             }
             /**
              * Creates a tick generator using the specified interval.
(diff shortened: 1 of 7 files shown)


rcchen:
```
### Fast Compile >>typings/mocha/mocha.d.ts
Using the custom compiler : ./node_modules/typescript/bin/tsc
/Volumes/git/plottable/src/utils/utils.ts(310,7): error TS2322: Type 'Color' is not assignable to type 'HSLColor':
  Property 'h' is missing in type 'Color'.

>> 1 non-emit-preventing type warning  
>> Type errors only.
```

Also we should change darkenColor to leverage the same d3 functions; same commit or no?

bluong:
@rcchen  Just to note that it's nice to make the pull request name to something that's a lot more readable.  I personally like the appraoch of
[{{relevantClassName}}] - {{Issue concerning that class}}

Also, I don't think the fixes tag works in pr titles?

bluong:
@rcchen Feel free to make the darkenColor switch into a separate PR unless it's blocking this PR from going in, just in case that the darkenColor switch takes more review time that would block this fix.

bluong:
@rcchen I think you're using commit amends here or something similar, but in any case, you should make the commit messages a bit more meaningful.  we're not at all strict about commit messages since not too much info can be derived from them, but a little information would be nice in what happened.

plottable.d.ts
```diff
@@ -113,9 +113,9 @@ declare module Plottable {
              * @returns {[{ [key: string]: any }} coppied map.
              */
             function copyMap<T>(oldMap: {
-                [key: string]: T;
+                [x: string]: T;
```
bluong: Seems like didn't upgrade typescript?

rcchen:
```
rchen2-mbp:plottable rchen$ tsc -v
message TS6029: Version 1.4.1.0
```

bluong:
LGTM aside not upgrading typescript

crmorford:
Sorry to ask this after the dev +1...but we could do this for InterpolatedColor also? http://jsfiddle.net/mj26qxxs/2/

bluong:
As discussed InterpolcatedColor has a colorRange API that will accomplish this, separate from this PR

bluong:
Also as discussed, let's get the lightenFactor to a number we want

crmorford:
Looks great! GIM

crmorford:
I want to note that some colors are a little close, but it's a definite improvement on clarity (having 3 white squares, like before, is not good)
before: http://jsfiddle.net/u4k5t8xe/2/
after: http://jsfiddle.net/u4k5t8xe/1/

GIM

src/scales/colorScale.ts
```diff
@@ -77,7 +77,8 @@ export module Scale {
     public scale(value: string): string {
       var color = super.scale(value);
       var index = this.domain().indexOf(value);
-      var modifyFactor = Math.floor(index / this.range().length);
+      var numLooped = Math.floor(index / this.range().length);
+      var modifyFactor = Math.log(numLooped * 1.6 + 1);
```
jtlan: Magic 1.6? Store off as a parameter.

PR #1550, opened 2015-01-24
[ColorScale] - Re-enabling color string names as range values

bluong:
Same as #1549 

Couldn't check out a numeric-named branch so creating a new PR


jtlan:
Should be same as the other one, GIM.

PR #1554, opened 2015-01-27
[Plot] - Adding getAllSelections() API

### LINKED ISSUE #1545: API for getting selection out of Plot.Scatter
Similar to `getAllBars()` on `Plot.Bar`.

bluong:
Close #1545.

### DIFF
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -419,6 +419,14 @@ export module Plot {
       var maxTime = _Util.Methods.max(times, 0);
       this._additionalPaint(maxTime);
     }
+
+    public getAllSelections(): D3.Selection {
+      this._renderArea.selectAll(this._getSelector());
+    }
+
+    protected _getSelector(): string {
+      return "";
+    }
   }
 }
 }
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -431,10 +431,6 @@ export module Plot {
 
       return allSelections;
     }
-
-    protected _getSelector(): string {
-      return "";
-    }
   }
 }
 }
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -421,7 +421,15 @@ export module Plot {
     }
 
     public getAllSelections(): D3.Selection {
-      return this._renderArea.selectAll(this._getSelector());
+      var allSelections = d3.select();
+      allSelections[0] = [];
+      this._getDrawersInOrder().forEach((drawer) => {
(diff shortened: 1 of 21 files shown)


src/components/plots/areaPlot.ts
```diff
@@ -90,6 +90,10 @@ export module Plot {
       attrToProjector["stroke"] = attrToProjector["stroke"] || d3.functor(this._defaultFillColor);
       return attrToProjector;
     }
+
+    protected _getSelector(): string {
+      return ".area";
```
jtlan: This will fail to return the line part of the area?

jtlan:
Can we get the selection method to coordinate with the `Drawer`s somehow? Right now, the drawers sometimes attach classes to the selection they draw, but this implementation adds a separate `_getSelector()` API point on `Plot`, which means the `Drawer` and `Plot` could easily get out of sync because we have the information in two places.

src/components/plots/areaPlot.ts
bluong: Wait I thought we didn't want to return the line part

bluong:
@jtlan I deliberately didn't want to borrow from the Drawer since right now the Drawer's architecture seems very black-boxy and complex, meaning that we may want to change it in the future.  In which case, I wanted this code to be neutral to whatever change we have in Drawers.

I am ok with reverting to the Drawer usage but slightly hesitant

jtlan:
@bluong I meant more that `_getSelector()` could be implemented on the drawer, allowing the Plot to get the selection the drawer is using without duplicating logic.

src/components/plots/abstractPlot.ts
```diff
@@ -419,6 +419,22 @@ export module Plot {
...
+          allSelections[0].push(selection);
+        });
+      });
+
+      return allSelections;
+    }
+
+    protected _getSelector(): string {
+      return "";
+    }
```
jtlan: probably should remove this?

jtlan:
Nice, the `Drawer`-based implementation is much more compact. You left something from the previous implementation in **abstractPlot.ts** though.

jtlan:
@crmorford For testing, maybe create a small test where you try to set the CSS class (using `classed()`) or stylings (using `style()`) of the returned `D3.Selection`.

crmorford:
Could you guys leave me a small example when you add new features / add stuff to the API? I can figure it out this time, but I think it's more efficient that way.

crmorford:
*Works on all standard Bar/Area/Line/Scatter Plots, Clustered Bar, Stacked Bar, Stacked Area, Pie Chart
*Doesn't give unexpected behavior when called on axes/labels/empty plots/gridlines/legend
*Speed better than initial render on large amounts of data

Nice! GIM

PR #1555, opened 2015-01-28
[Axis] Ensure labels are confined to Axis boundaries, fixes #1501

### LINKED ISSUE #1501: Axis labels leaving svg boundaries
example: the scale_date quicktest at http://localhost:9999/quicktests/overlaying/index.html#
screen shot 2015-01-07 at 3 52 31 pm

rcchen:
Close #1501.

### DIFF
--- a/src/components/axes/abstractAxis.ts
+++ b/src/components/axes/abstractAxis.ts
@@ -456,7 +456,6 @@ export module Axis {
 
     protected _hideEndTickLabels() {
       var boundingBox = this._element.select(".bounding-box")[0][0].getBoundingClientRect();
-      var boundingSVG = _Util.DOM.getBoundingSVG(this._element);
       var tickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS);
       if (tickLabels[0].length === 0) {
         return;
@@ -475,12 +474,12 @@ export module Axis {
     protected _hideOverflowingTickLabels() {
       var boundingSVG = _Util.DOM.getBoundingSVG(this._element).getBoundingClientRect();
       var tickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS);
-      if (tickLabels[0].length === 0) {
+      if (tickLabels.empty()) {
         return;
       }
-      tickLabels[0].forEach(function(elem: any) {
-        if (!_Util.DOM.containedInBoundingBox(boundingSVG, elem.getBoundingClientRect())) {
-          d3.select(elem).style("visibility", "hidden");
+      tickLabels.each(function(d: any, i: number) {
+        if (!_Util.DOM.containedInBoundingBox(boundingSVG, this.getBoundingClientRect())) {
+          d3.select(this).style("visibility", "hidden");
         }
       });
     }
--- a/src/components/axes/abstractAxis.ts
+++ b/src/components/axes/abstractAxis.ts
@@ -456,28 +456,33 @@ export module Axis {
 
     protected _hideEndTickLabels() {
       var boundingBox = this._element.select(".bounding-box")[0][0].getBoundingClientRect();
-
-      var isInsideBBox = (tickBox: ClientRect) => {
-        return (
-          Math.floor(boundingBox.left) <= Math.ceil(tickBox.left) &&
-          Math.floor(boundingBox.top)  <= Math.ceil(tickBox.top)  &&
-          Math.floor(tickBox.right)  <= Math.ceil(boundingBox.left + this.width()) &&
-          Math.floor(tickBox.bottom) <= Math.ceil(boundingBox.top  + this.height())
-        );
-      };
-
+      var boundingSVG = _Util.DOM.getBoundingSVG(this._element);
       var tickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS);
       if (tickLabels[0].length === 0) {
         return;
       }
       var firstTickLabel = tickLabels[0][0];
-      if (!isInsideBBox(firstTickLabel.getBoundingClientRect())) {
+      if (!_Util.DOM.containedInBoundingBox(boundingBox, firstTickLabel.getBoundingClientRect())) {
         d3.select(firstTickLabel).style("visibility", "hidden");
       }
       var lastTickLabel = tickLabels[0][tickLabels[0].length-1];
-      if (!isInsideBBox(lastTickLabel.getBoundingClientRect())) {
+      if (!_Util.DOM.containedInBoundingBox(boundingBox, lastTickLabel.getBoundingClientRect())) {
         d3.select(lastTickLabel).style("visibility", "hidden");
       }
+    }
+
+    // Responsible for hiding any tick labels that break out of the bounding SVG
+    protected _hideOverflowingTickLabels() {
+      var boundingSVG = _Util.DOM.getBoundingSVG(this._element).getBoundingClientRect();
+      var tickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS);
+      if (tickLabels[0].length === 0) {
+        return;
+      }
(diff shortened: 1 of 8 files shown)


src/utils/domUtils.ts
```diff
@@ -107,6 +107,28 @@ export module _Util {
       if (boxA.top > boxB.bottom) { return false; }
       return true;
     }
+
+    export function containedInBoundingBox(container: ClientRect, element: ClientRect) {
```
jtlan: I'm surprised we didn't have this function already.

src/utils/domUtils.ts
```diff
@@ -107,6 +107,28 @@ export module _Util {
...
+      return (
+        Math.floor(container.left) <= Math.ceil(element.left) &&
+        Math.floor(container.top)  <= Math.ceil(element.top)  &&
+        Math.floor(element.right)  <= Math.ceil(container.left + container.width) &&
```
jtlan: Could you use `container.right` instead of `container.left + container.width`?

src/utils/domUtils.ts
```diff
@@ -107,6 +107,28 @@ export module _Util {
...
+        Math.floor(element.bottom) <= Math.ceil(container.top  + container.height)
```
jtlan: Could you use `container.bottom` instead of `container.top + container.height`?

src/components/axes/abstractAxis.ts
```diff
@@ -456,30 +456,35 @@ export module Axis {
...
-      var isInsideBBox = (tickBox: ClientRect) => {
-        return (
-          Math.floor(boundingBox.left) <= Math.ceil(tickBox.left) &&
-          Math.floor(boundingBox.top)  <= Math.ceil(tickBox.top)  &&
-          Math.floor(tickBox.right)  <= Math.ceil(boundingBox.left + this.width()) &&
-          Math.floor(tickBox.bottom) <= Math.ceil(boundingBox.top  + this.height())
-        );
-      };
-
+      var boundingSVG = _Util.DOM.getBoundingSVG(this._element);
```
bluong: Variable never seems to be used

src/utils/domUtils.ts
bluong: It was inlined when it needed to be used.

src/utils/domUtils.ts
```diff
@@ -107,6 +107,28 @@ export module _Util {
...
+      );
+    }
+
+    export function getBoundingSVG(elem: D3.Selection) {
```
jtlan: Two things:
1. Please update `isSelectionRemovedFromSVG()` to use this implementation.
2. There's a much faster way to do this:
```Typescript
var svg = elem.node().ownerSVGElement;
return (svg == null) ? elem.node() : svg;
```

src/components/axes/abstractAxis.ts
```diff
@@ -456,30 +456,35 @@ export module Axis {
...
+      if (!_Util.DOM.containedInBoundingBox(boundingBox, lastTickLabel.getBoundingClientRect())) {
         d3.select(lastTickLabel).style("visibility", "hidden");
       }
     }
 
+    // Responsible for hiding any tick labels that break out of the bounding SVG
+    protected _hideOverflowingTickLabels() {
+      var boundingSVG = _Util.DOM.getBoundingSVG(this._element).getBoundingClientRect();
+      var tickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS);
+      if (tickLabels[0].length === 0) {
```
bluong: tickLabels.empty() ?

src/components/axes/abstractAxis.ts
```diff
@@ -456,30 +456,35 @@ export module Axis {
...
+        return;
+      }
+      tickLabels[0].forEach(function(elem: any) {
```
bluong: tickLabels.each()

bluong:
Tests please

src/components/axes/abstractAxis.ts
jtlan: It'd be more semantic to do this:
```Typescript
tickLabels.each(function(d: any, i: number) {
  var elem = this; // inside a D3 loop, "this" refers to the DOM element itself
  ...
  // rest of the implementation...
  ...
});
```

jtlan:
Also, does it make more sense to have the labels confined to the bounding box of the `Axis`, rather than the `<svg>`?

bluong:
@jtlan agreed.

src/components/axes/abstractAxis.ts
```diff
@@ -456,30 +456,34 @@ export module Axis {
...
       var lastTickLabel = tickLabels[0][tickLabels[0].length-1];
-      if (!isInsideBBox(lastTickLabel.getBoundingClientRect())) {
+      if (!_Util.DOM.containedInBoundingBox(boundingBox, lastTickLabel.getBoundingClientRect())) {
         d3.select(lastTickLabel).style("visibility", "hidden");
       }
     }
 
+    // Responsible for hiding any tick labels that break out of the bounding container
+    protected _hideOverflowingTickLabels() {
+      var boundingBox = this._element.select(".bounding-box")[0][0].getBoundingClientRect();
```
jtlan: I think we generally use `selection.node()` instead of `selection[0][0]` to get the element back out.

It might be worth storing off the bounding-box in `AbstractComponent` -- right now this implementation depends on the actual CSS class applied.

src/utils/domUtils.ts
```diff
@@ -107,6 +103,20 @@ export module _Util {
...
+        Math.floor(container.left) <= Math.ceil(element.left) &&
+        Math.floor(container.top)  <= Math.ceil(element.top)  &&
+        Math.floor(element.right)  <= Math.ceil(container.right) &&
+        Math.floor(element.bottom) <= Math.ceil(container.bottom)
+      );
+    }
+
+    export function getBoundingSVG(elem: D3.Selection) {
+      var svg = elem[0][0].ownerSVGElement;
+      return (svg == null) ? elem : svg;
```
jtlan: Actually, there's a problem here. `ownerSVGElement` is also null if `elem` is no longer in the DOM tree, so we'll mistakenly return `elem` even though it's not an `<svg>`. This implementation addresses that:
```Typescript
export function getBoundingSVG(elem: HTMLScriptElement): HTMLScriptElement {
  var ownerSVG = elem.ownerSVGElement;
  if (ownerSVG != null) { 
    return ownerSVG;
  }

  if (elem.nodeName.toLowerCase() === "svg") { // elem itself is an SVG
    return elem;
  }

  return null; // not in the DOM
}
```
I'll also note that the implementation you had above didn't have a consistent return type...

src/utils/domUtils.ts
```diff
@@ -39,11 +39,7 @@ export module _Util {
     }
 
     export function isSelectionRemovedFromSVG(selection: D3.Selection) {
-      var n = (<Node> selection.node());
-      while (n !== null && n.nodeName.toLowerCase() !== "svg") {
-        n = n.parentNode;
-      }
-      return (n == null);
+      return selection !== getBoundingSVG(selection);
```
jtlan: Wouldn't actually do the right thing, due to problems in `getBoundingSVG()`; see below.

test/components/numericAxisTests.ts
```diff
@@ -331,4 +331,40 @@ describe("NumericAxis", () => {
...
+
+    linePlot.attr("x", function(d: any) { return d3.time.format("%x").parse(d.x);}, xScale);
+    linePlot.project("y", "y", yScale);
+    linePlot.addDataset(data);
+
+    var chart = new Plottable.Component.Table([
+      [yAxis, linePlot],
+      [null, xAxis]
+    ]);
+    chart.renderTo(svg);
```
jtlan: We should be able to unit-test the `Axis` functionality without creating an entire chart. We can do this by setting a really wide `formatter` on the `Axis`, and rendering just the `Axis` to the `svg`.

jtlan:
GIL.

crmorford:
Looks good!

PR #1556, opened 2015-01-28
[AbstractAxis] More regular behavior for tick labels appearing/disappearing, fixes #1196

### LINKED ISSUE #1196: Uneven Behavior of Tick Labels Appearing/Disappearing in Negative and Positive
SLATE issue 1111
Plottable repro: http://jsfiddle.net/tkim/k1gxzp1L/3/
(In the screenshot, 500000 shows up on the right when on the left -500000 doesn't, vice versa)
screen shot 2014-10-14 at 11 50 24 am

rcchen:
Close #1196.

### DIFF
--- a/src/components/axes/abstractAxis.ts
+++ b/src/components/axes/abstractAxis.ts
@@ -492,7 +492,7 @@ export module Axis {
       var interval = 1;
 
       // Get the widths of all the labels
-      var visibleTickLabelRects = [];
+      var visibleTickLabelRects: ClientRect[] = [];
       visibleTickLabels.each(function(d: any) {
         var clientRect = this.getBoundingClientRect();
         visibleTickLabelRects.push(clientRect);
@@ -506,10 +506,12 @@ export module Axis {
           // We increment the variable if the next label's width is greater than the previous one's (one-time)
           // This only applies to the bottom and top orientations
           if (visibleTickLabelRects[i].width > visibleTickLabelRects[i-1].width && !hasSkipped &&
-            (this._orientation == "bottom" || this._orientation == "top")) {
+            (this._orientation === "bottom" || this._orientation === "top")) {
             interval += 1;
             hasSkipped = true;
-          } else break;
+          } else {
+            break;
+          }
         } else {
           interval += 1;
         }
@@ -518,7 +520,7 @@ export module Axis {
       // Set any label that is not part of the interval as hidden
       visibleTickLabels.each(function(d: any, i: number) {
         var tickLabel = d3.select(this);
-        if (i % interval != 0) {
+        if (i % interval !== 0) {
           tickLabel.style("visibility", "hidden");
         }
       });
--- a/src/components/axes/abstractAxis.ts
+++ b/src/components/axes/abstractAxis.ts
@@ -488,43 +488,42 @@ export module Axis {
                                     });
       var lastLabelClientRect: ClientRect;
 
-      // Base interval value between labels
-      var interval = 1;
-
       // Get the widths of all the labels
       var visibleTickLabelRects: ClientRect[] = [];
-      visibleTickLabels.each(function(d: any) {
+      visibleTickLabels.each(function() {
         var clientRect = this.getBoundingClientRect();
         visibleTickLabelRects.push(clientRect);
       });
 
       // Figure out the appropriate interval for overlap
       var hasSkipped = false;
-      for (var i = 1; i < visibleTickLabelRects.length; i++) {
-        // Checks to see if there is any overlap between the first element and the current element
-        if (!_Util.DOM.boxesOverlap(visibleTickLabelRects[i], visibleTickLabelRects[0])) {
-          // We increment the variable if the next label's width is greater than the previous one's (one-time)
-          // This only applies to the bottom and top orientations
-          if (visibleTickLabelRects[i].width > visibleTickLabelRects[i-1].width && !hasSkipped &&
-            (this._orientation === "bottom" || this._orientation === "top")) {
-            interval += 1;
-            hasSkipped = true;
+      var interval = 1;
+      // Only do interval modifications for bottom and top orientations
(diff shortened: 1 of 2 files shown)


src/components/axes/abstractAxis.ts
```diff
@@ -488,16 +488,43 @@ export module Axis {
                                     });
       var lastLabelClientRect: ClientRect;
 
-      visibleTickLabels.each(function (d: any) {
+      // Base interval value between labels
+      var interval = 1;
+
+      // Get the widths of all the labels
+      var visibleTickLabelRects: ClientRect[] = [];
+      visibleTickLabels.each(function(d: any) {
```
bluong: If you're not using the d parameter, you should be able to remove it

src/components/axes/abstractAxis.ts
```diff
@@ -488,16 +488,43 @@ export module Axis {
                                     });
       var lastLabelClientRect: ClientRect;
 
-      visibleTickLabels.each(function (d: any) {
+      // Base interval value between labels
+      var interval = 1;
```
bluong: Probably want to move this variable down to the for loop.  It's a bit out of place up here

jtlan:
As a first thought, if you don't need the special context of `this` provided but using the *function* keyword, it might make more sense to use the fat-arrow notation that Typescript provides:
```Typescript
(d: any) => {
  // function body
}
```

src/components/axes/abstractAxis.ts
```diff
@@ -488,16 +488,43 @@ export module Axis {
...
+          } else {
+            break;
+          }
+        } else {
+          interval += 1;
+        }
+      }
+
+      // Set any label that is not part of the interval as hidden
+      visibleTickLabels.each(function(d: any, i: number) {
```
bluong: d I believe is a string?

src/components/axes/abstractAxis.ts
```diff
@@ -488,16 +488,43 @@ export module Axis {
...
+
+      // Figure out the appropriate interval for overlap
+      var hasSkipped = false;
+      for (var i = 1; i < visibleTickLabelRects.length; i++) {
+        // Checks to see if there is any overlap between the first element and the current element
+        if (!_Util.DOM.boxesOverlap(visibleTickLabelRects[i], visibleTickLabelRects[0])) {
+          // We increment the variable if the next label's width is greater than the previous one's (one-time)
+          // This only applies to the bottom and top orientations
+          if (visibleTickLabelRects[i].width > visibleTickLabelRects[i-1].width && !hasSkipped &&
+            (this._orientation === "bottom" || this._orientation === "top")) {
```
bluong: The for loop is done regardless of the orientation.  We probably want to move the orientation check higher for cleaner code and for performance's sake.

bluong:
Fat arrow when possible is preferable

bluong:
@jtlan It seems he needs the this context in both of his changes?  Where are you referring to?

src/components/axes/abstractAxis.ts
```diff
@@ -488,16 +488,43 @@ export module Axis {
...
+            interval += 1;
+            hasSkipped = true;
```
bluong: Im a bit confused at the logic presented here...?

If the boxes don't overlap and this box is bigger than the previous box, then skip? and increase of interval of skipping?

bluong: There doesn't seem to be explicit reasoning on why the box needs to be skipped due to overlap issues.

rcchen: So this was an interesting case that I encountered with a fairly narrow axis that went from 0 to 1 in increments of 0.1. There is no overlap between the 0 and the 0.1 label because the 0 label is smaller, but the 0.1 label and the 0.2 label have an overlap due to their size. The skipping covers this edge case and maintains a consistent interval throughout the rendered axis. 

develop:
image

#1556:
image

rcchen: Sorry, clarification. They're not technically overlapping but are close enough to make visibility an issue (0-2px distance between labels).

src/components/axes/abstractAxis.ts
```diff
@@ -488,14 +488,40 @@ export module Axis {
                                     });
       var lastLabelClientRect: ClientRect;
 
-      visibleTickLabels.each(function (d: any) {
+      // Get the widths of all the labels
+      var visibleTickLabelRects: ClientRect[] = [];
+      visibleTickLabels.each(function() {
         var clientRect = this.getBoundingClientRect();
+        visibleTickLabelRects.push(clientRect);
```
jtlan: Would
```Typescript
var visibleTickLabelRects = visibleTickLabels[0].map(function(label: HTMLScriptElement) { return label.getBoundingClientRect(); });
```
be more semantic?

src/components/axes/abstractAxis.ts
```diff
@@ -488,14 +488,40 @@ export module Axis {
...
+      });
+
+      // Figure out the appropriate interval for overlap
+      var hasSkipped = false;
+      var interval = 1;
+      // Only do interval modifications for bottom and top orientations
+      if (this._orientation === "bottom" || this._orientation === "top") {
+        for (var i = 1; i < visibleTickLabelRects.length; i++) {
+          // Checks to see if there is any overlap between the first element and the current element
+          if (!_Util.DOM.boxesOverlap(visibleTickLabelRects[i], visibleTickLabelRects[0])) {
```
jtlan: Isn't there still a potential problem here, where the first label's rectangle is very small, but later labels overlap each other?

jtlan:
Concerned that it doesn't actually solve the problem in all cases, such as when the first label is small ("0") but later labels are long ("300000").

rcchen:
Need to delete the branch, moving to a different pull request.

rcchen:
New pull request at https://github.com/palantir/plottable/pull/1574

PR #1557, opened 2015-01-28
Add in InterpolatedColorLegend.

### LINKED ISSUE #1528: InterpolatedColor Legend
Grid Plots would be far easier to understand if a 'heat' legend was a feature in Plottable (like this: https://pepperprincess.files.wordpress.com/2012/02/new-heat-scale1.png?w=497)

jtlan:
Modified Scale.InterpolatedColor to allow sampling the backing D3 scale (using ticks()).

Public API points:
```Typescript
/**
 * Creates an InterpolatedColorLegend.
 *
 * The InterpolatedColorLegend consists of a sequence of swatches, showing the
 * associated Scale.InterpolatedColor sampled at various points. Two labels
 * show the maximum and minimum values of the Scale.InterpolatedColor.
 *
 * @constructor
 * @param {Scale.InterpolatedColor} interpolatedColorScale
 * @param {string} orientation (horizontal/left/right).
 * @param {Formatter} The labels are formatted using this function.
 */
constructor(interpolatedColorScale: Scale.InterpolatedColor, orientation?: string, formatter?: (d: any) => string);

remove(): void;

/**
 * Gets the current formatter on the InterpolatedColorLegend.
 *
 * @returns {Formatter} The current Formatter.
 */
formatter(): Formatter;
/**
 * Sets the current formatter on the InterpolatedColorLegend.
 *
 * @param {Formatter} formatter If provided, data will be passed though `formatter(data)`.
 * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
 */
formatter(formatter: Formatter): InterpolatedColorLegend;

/**
 * Gets the orientation of the InterpolatedColorLegend.
 *
 * @returns {string} The current orientation.
 */
orient(): string;
/**
 * Sets the orientation of the InterpolatedColorLegend.
 *
 * @param {string} newOrientation The desired orientation (horizontal/left/right).
 *
 * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
 */
orient(newOrientation: string): InterpolatedColorLegend;
```

Close #1528.

Looks like this: (bounding box shown for clarity)
screen shot 2015-01-28 at 3 53 53 pm
screen shot 2015-01-28 at 3 57 18 pm


bluong:
Is adding this class really a good idea?  A better answer to the problem architectural wise would be to have InterpolatedColorScale inherit from ColorScale, wouldn't it?

bluong:
I think ideally we should only ever need one type of legend.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+  export class InterpolatedColorLegend extends AbstractComponent {
+    private _measurer: SVGTypewriter.Measurers.Measurer;
+    private _wrapper: SVGTypewriter.Wrappers.Wrapper;
+    private _writer: SVGTypewriter.Writers.Writer;
+    private _scale: Scale.InterpolatedColor;
+    private _orientation: String ;
+    private _padding = 5;
+    private _formatter: Formatter;
+
+    private _swatchContainer: D3.Selection;
```
bluong: What's a swatch?

jtlan:
@bluong wrote:
>Is adding this class really a good idea? A better answer to the problem architectural wise would be to have InterpolatedColorScale inherit from ColorScale, wouldn't it?
Not really, since this legend fundamentally looks different from the one that takes in a `ColorScale`. In fact, it probably has more in common with `Axis`, because it visualizes a scale with a continuous input domain.

src/components/interpolatedColorLegend.ts
jtlan: A swatch is a color sample. The term is reasonably common (in paint and color selection) and describes the purpose of the `<rect>`s that are created.

http://en.wiktionary.org/wiki/swatch

bluong:
Am I interpreting an InterpolatedColorLegend incorrectly?  My initial guess on what it would look like visually is simply

Color - Domain Value
Color - Domain Value
...

The only difference between InterpolatedColorLegend and Legend seems to be that continuous range in the colors, but that should be sorted out on the scale's end and not on the Legend's end.

Basically, I don't see an actual functional difference between these two legend types, and adding another class can open up a can or worms surrounding duplication.

jtlan:
The positioning of swatches / icons in this legend is fundamentally different from that of the other `Legend` though. Is there something that unifies them architecturally?

In this case, there weren't even supposed to be distinct swatches, but it's possible for the `Scale.InterpolatedColor` to be linear and have tons of inflection points.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+    public formatter(formatter?: Formatter): any {
+      if (formatter === undefined) {
+        return this._formatter;
+      }
+      this._formatter = formatter;
+      this._invalidateLayout();
+      return this;
+    }
+
+    private static validateOrientation(orientation: string) {
```
bluong: Mainly semantical, but I like the term ensureOrientation rather than validateOrientation since I feel that validateOrientation doesn't sound like a term that would return me a corrected orientation if it passes the validation check.  I feel that ensure essentially tells me that the return value will be a guaranteed good orientation or an error will be thrown.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+      } else {
+        this._orientation = InterpolatedColorLegend.validateOrientation(newOrientation);
+        this._invalidateLayout();
+        return this;
+      }
+    }
+
+    protected _setup() {
+      super._setup();
+      this.classed("legend", true).classed("interpolated-color-legend", true);
```
bluong: Worth it to factor out to static strings?

bluong: Also, not sure how much it matters, but if the user does not want a legend class stapled on, this will force it.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+      var labelWidths = domain.map((d: number) => this._measurer.measure(this._formatter(d)).width);
+
+      var desiredHeight: number;
+      var desiredWidth: number;
+      if (this._isVertical()) {
+        var longestWidth = _Util.Methods.max(labelWidths, 0);
+        desiredWidth = this._padding + textHeight + this._padding + longestWidth + this._padding;
+        desiredHeight = this._padding + numSwatches * textHeight + this._padding;
+      } else {
+        var totalWidth = labelWidths[0] + labelWidths[1];
```
bluong: Minor
totalLabelWidth

bluong: Variable doesn't seem to be used.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+                textRotation: 0
+            };
+
+      var swatchWidth: number;
+      var swatchHeight: number;
+      var swatchX: (d: any, i: number) => number;
+      var swatchY: (d: any, i: number) => number;
+
+      if (this._isVertical()) {
+        var longestTextWidth = text1Width > text0Width ? text1Width : text0Width;
```
bluong: Math.max

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+        swatchWidth = (this.width() - 4 * padding - text0Width - text1Width) / numSwatches;
+        swatchHeight = this.height() - 2 * padding;
+        swatchX = (d: any, i: number) => (padding + text0Width + padding) + i * swatchWidth;
+        swatchY = (d: any, i: number) => padding;
+
+        upperWriteOptions.xAlign = "right";
+        upperLabelShift.x = -padding;
+        lowerWriteOptions.xAlign = "left";
+        lowerLabelShift.x = padding;
+      }
```
bluong: How much of the above can we ternary?

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+
+      this._upperLabel.text(""); // clear the upper label
```
bluong: Would a _writer.clear() or similar accomplish this is in a more highlevel manner?

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,249 @@
...
+      var upperTranslateString = "translate(" + upperLabelShift.x + ", " + upperLabelShift.y + ")";
+      this._upperLabel.attr("transform", upperTranslateString);
+
+      this._lowerLabel.text(""); // clear the lower label
+      this._writer.write(text0, this.width(), this.height(), lowerWriteOptions);
+      var lowerTranslateString = "translate(" + lowerLabelShift.x + ", " + lowerLabelShift.y + ")";
+      this._lowerLabel.attr("transform", lowerTranslateString);
+
+      swatchWidth = Math.max(swatchWidth, 0);
+      swatchHeight = Math.max(swatchHeight, 0);
```
bluong: This correcting shoulnd't be done here.  Mightas well do it where all the other calculations are being done.

Also, Math.abs

bluong:
Ina strict functional manner it looks fine.  Mainly code cleanup.

src/components/interpolatedColorLegend.ts
jtlan: That happens in normal `Legend` though. Most classes forcibly attach a CSS class that identifies them. The user can always remove it.

src/components/interpolatedColorLegend.ts
jtlan: I thought about that, but the behavior is poorly defined if either end of the scale is `undefined`. Is that okay?

src/components/interpolatedColorLegend.ts
jtlan: You'd have `this._isHorizontal() ?` every single line. I tried :(

src/components/interpolatedColorLegend.ts
jtlan: It would; there's a similar #HACKHACK filed in Label.

src/components/interpolatedColorLegend.ts
jtlan: Not abs, if the value is less than zero I have to use zero.

Putting it with the other calculations results in the lines being duplicated.

src/components/interpolatedColorLegend.ts
jtlan: Thanks, nice catch.

src/components/interpolatedColorLegend.ts
bluong: Since we don't have the developer anymore... what do we do about this?

src/components/interpolatedColorLegend.ts
bluong: Really?  While you would have to this._isHorizontal a bunch of lines, the number of lines is significantly reduced

src/components/interpolatedColorLegend.ts
bluong: They can only remove it after the render step though.

src/components/interpolatedColorLegend.ts
bluong: How does this current version differ from Math.max?

src/components/interpolatedColorLegend.ts
jtlan: I'll move the call.

src/components/interpolatedColorLegend.ts
jtlan: You get `undefined` instead of `NaN`, but I guess they're equally bad :( Changing.

src/components/interpolatedColorLegend.ts
jtlan: I think this is easier to read, even though it's more verbose. There's also less risk of a screw-up based on accidentally reversing one of the ternaries. I feel like that's happened to us before (although that's not a sufficient reason to not use ternaries).

src/components/interpolatedColorLegend.ts
jtlan: **svg-typewriter** was transferred to Palantir, so we can edit it. Issue [21](https://github.com/palantir/svg-typewriter/issues/21) on svg-typewriter.

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,261 @@
...
```
bluong: Still want to move this correction upwards probably.  I don't see a good reason to do a calculation and then complete the calculation here.

plottable.css
```diff
@@ -124,6 +124,7 @@ svg.plottable {
 }
 
 .plottable .legend text {
+  fill: #32313F;
```
bluong: Why this change?

src/components/interpolatedColorLegend.ts
```diff
@@ -0,0 +1,277 @@
...
+
+      this._lowerLabel.text(""); // clear the lower label
+      this._writer.write(text0, this.width(), this.height(), lowerWriteOptions);
+      var lowerTranslateString = "translate(" + lowerLabelShift.x + ", " + lowerLabelShift.y + ")";
+      this._lowerLabel.attr("transform", lowerTranslateString);
+
+      this._swatchBoundingBox.attr(boundingBoxAttr);
+
+      swatchWidth = Math.max(swatchWidth, 0);
+      swatchHeight = Math.max(swatchHeight, 0);
```
bluong: Still not sure about these calls needing to be here.  Might as well do all of the calculation in one place

src/scales/interpolatedColorScale.ts
```diff
@@ -14,6 +14,7 @@ export module Scale {
    * By default it generates a linear scale internally.
    */
   export class InterpolatedColor extends AbstractScale<number, string> {
+    protected _d3Scale: D3.Scale.QuantitativeScale;
```
bluong: I assume this is to shadow the super class's variable.  If so... I am not a fan of this convention, if the super class changes this variable name, this class will not be able to detect the change other than through obscure bugs etc.  Probably should cast where applicable.

Also, I don't see why you need to make this change to begin with?  Nothing so far seems to require the fact that this is a QuantitativeScale

plottable.css
jtlan: Sets the `Legend` text to be the same color as used on `Axis` and `Label`. I can get Design to officially sign off on it, but the goal is consistency.

src/scales/interpolatedColorScale.ts
jtlan: It was required when I tried to put a `ticks()` call on `Scale.InterpolatedColor`, but not necessary any more.

bluong:
Still needs tests, otherwise other than the 1 calculation placement issue, looks fine.

src/components/interpolatedColorLegend.ts
jtlan: Just tried to write something that used the `swatchWidth` and `swatchHeight` values set inside the if-block, and got bitten by the negative values. Moving...

src/utils/domUtils.ts
```diff
@@ -107,6 +107,15 @@ export module _Util {
...
     }
+
+    export function boxIsInside(inner: ClientRect, outer: ClientRect) {
+      return (
+        Math.floor(outer.left) <= Math.ceil(inner.left) &&
+        Math.floor(outer.top)  <= Math.ceil(inner.top)  &&
+        Math.floor(inner.right)  <= Math.ceil(outer.right) &&
+        Math.floor(inner.bottom) <= Math.ceil(outer.bottom)
+      );
+    }
```
bluong: Why does this look familiar? :)

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
     svg.remove();
   });
 });
+
+describe("InterpolatedColorLegend", () => {
```
bluong: the name is the file is now super confusing because of the shared namespaces...  We probably want to place this in a separate file.

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
...
+  beforeEach(() => {
+    svg = generateSVG(400, 400);
+    colorScale = new Plottable.Scale.InterpolatedColor();
+  });
+
+  function assertBasicRendering(legend: Plottable.Component.InterpolatedColorLegend) {
+    var scaleDomain = colorScale.domain();
+    var legendElement: D3.Selection = (<any> legend)._element;
+
+    var swatches = legendElement.selectAll(".swatch");
```
bluong: Doesn't have to be done in this PR, but should we have a... getAllSwatches() ?  Seems useful for us and the user.

src/utils/domUtils.ts
jtlan: I wanted to cherry-pick from Roger's branch, but there wasn't a single commit that added just those lines. Now that I think about it, I should have done that...

test/components/legendTests.ts
jtlan: Sounds sort of like the API point on the `Plot`s. Should this be something more general?

test/components/legendTests.ts
jtlan: Sure, I can move it. There's duplication of `Legend` tests too ... #1561 ...

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
...
+
+    var elementBCR = legendElement.node().getBoundingClientRect();
+    assert.isTrue(Plottable._Util.DOM.boxIsInside(swatchContainerBCR, elementBCR),
+                  "swatches are drawn within the legend's element");
+
+    var formattedDomainValues = scaleDomain.map((<any> legend)._formatter);
+    var labels = legendElement.selectAll("text");
+    var labelTexts = labels[0].map((textNode: HTMLScriptElement) => textNode.textContent);
+    assert.deepEqual(labelTexts, formattedDomainValues, "formatter is used to format label text");
+  }
```
bluong: Hmm Not sure about how I feel about having all of these assert statements that test a variety of aspects packaged up into a single function.  We possibly want a unit test structure that's more like have a single category for basic that will run through each one of these aspects but in a more unitized manner.

something like...

describe("basic"...
  it("swatches are in bbox")...
  it("color matches")...
);

bluong: The problem with the current structure is that the tests that are higher up will get priority which is ok if the tests are similar since priority has a less of an effect, but if the tests are not similar then you'll only have information from the first run error.

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
...
+
+    svg.remove();
+  });
+
+  it("re-renders when scale domain updates", () => {
+    var legend = new Plottable.Component.InterpolatedColorLegend(colorScale, "horizontal");
+    legend.renderTo(svg);
+
+    colorScale.domain([0, 85]);
+    assertBasicRendering(legend);
```
bluong: Maybe a better test here isn't to check if the basic rendering has been done but check how the domain change affected the legend?

test/components/legendTests.ts
bluong: Not high priority but definitely something to think about.

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
...
+
+  it("renders correctly when width is constrained (orientation: left)", () => {
+    svg.attr("width", 30);
+    var legend = new Plottable.Component.InterpolatedColorLegend(colorScale, "left");
+    legend.renderTo(svg);
+    assertBasicRendering(legend);
+    svg.remove();
+  });
+
+  it("renders correctly when height is constrained (orientation: left)", () => {
```
bluong: I think the core of my problem is mainly from the term "renders correctly"

While we do this in a bunch of places, I think the term is way too imprecise for what we are really looking for.  Figuring this out could take some time though, so I'm just going to focus on reviewing the mother assertBasicRendering function while thinking about a better way to structure these tests.

test/components/legendTests.ts
jtlan: `assertBasicRendering()` checks that the label values match the domain values. If you assume the scale updates correctly it will implicitly check this, but I agree that it'd be clearer what the test was up to if I checked the labels directly.

test/components/legendTests.ts
```diff
@@ -350,3 +350,183 @@ describe("Legend", () => {
...
+    assert.isTrue(Plottable._Util.DOM.boxIsInside(swatchContainerBCR, boundingBoxBCR),
+                  "bounding box contains all swatches");
+
+    var elementBCR = legendElement.node().getBoundingClientRect();
+    assert.isTrue(Plottable._Util.DOM.boxIsInside(swatchContainerBCR, elementBCR),
+                  "swatches are drawn within the legend's element");
+
+    var formattedDomainValues = scaleDomain.map((<any> legend)._formatter);
+    var labels = legendElement.selectAll("text");
+    var labelTexts = labels[0].map((textNode: HTMLScriptElement) => textNode.textContent);
```
bluong: Why not use the text function?  is it due to wrapping?  If so, I don't think the formatter will pick it up?

jtlan: didn't want to write:
```Typescript
labels[0].map((textNode: HTMLScriptElement) => d3.select(textNode).text());
```
because the d3 selection seems redundant.

crmorford:
Works with more colors than spaces (cool) http://jsfiddle.net/Lvossoxs/4/

crmorford:
GIM. nice work.

PR #1564, opened 2015-01-30
[OrdinalScale] - Remove rangeType

bluong:
OrdinalScales will now always be in "bands" mode.  Changing the rest of the codebase to reflect the changes.

### DIFF
--- a/src/components/axes/categoryAxis.ts
+++ b/src/components/axes/categoryAxis.ts
@@ -137,7 +137,7 @@ export module Axis {
      */
     private _measureTicks(axisWidth: number, axisHeight: number, scale: Scale.Ordinal, ticks: string[]) {
       var wrappingResults = ticks.map((s: string) => {
-        var bandWidth = scale.rangeBand();
+        var bandWidth = scale.fullBandWidth();
         var width  = this._isHorizontal() ? bandWidth  : axisWidth - this._maxLabelTickLength() - this.tickLabelPadding();
         var height = this._isHorizontal() ? axisHeight - this._maxLabelTickLength() - this.tickLabelPadding() : bandWidth;
         return this._wrapper.wrap(this.formatter()(s), this._measurer, width, height);
--- a/src/components/axes/categoryAxis.ts
+++ b/src/components/axes/categoryAxis.ts
@@ -137,7 +137,7 @@ export module Axis {
      */
     private _measureTicks(axisWidth: number, axisHeight: number, scale: Scale.Ordinal, ticks: string[]) {
       var wrappingResults = ticks.map((s: string) => {
-        var bandWidth = scale.fullBandStartAndWidth(s)[1];
+        var bandWidth = scale.rangeBand();
         var width  = this._isHorizontal() ? bandWidth  : axisWidth - this._maxLabelTickLength() - this.tickLabelPadding();
         var height = this._isHorizontal() ? axisHeight - this._maxLabelTickLength() - this.tickLabelPadding() : bandWidth;
         return this._wrapper.wrap(this.formatter()(s), this._measurer, width, height);
--- a/src/scales/ordinalScale.ts
+++ b/src/scales/ordinalScale.ts
@@ -68,15 +68,6 @@ export module Scale {
       return this._d3Scale.rangeBand();
     }
 
-//    public innerPadding(): number {
-//      var d = this.domain();
-//      if (d.length < 2) {
-//        return 0;
-//      }
-//      var step = Math.abs(this.scale(d[1]) - this.scale(d[0]));
-//      return step - this.rangeBand();
-//    }
-
     public fullBandStartAndWidth(v: string) {
       var start = this.scale(v) - this.innerPadding() / 2;
       var width = this.rangeBand() + this.innerPadding();
--- a/src/scales/ordinalScale.ts
+++ b/src/scales/ordinalScale.ts
@@ -64,6 +64,17 @@ export module Scale {
     }
 
     /**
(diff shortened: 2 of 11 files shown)


jtlan:
I think we should consider changing the names of some of the function calls to avoid people coming over from D3. For example, `rangeBand()` now returns a very different value than D3's `rangeBand()`.

bluong:
@jtlan How does it return something different?

bluong:
@jtlan Note that the thing to work on next is the padding suggestions you were referring to.

jtlan:
`rangeBand()` in D3 returns only the width of the band itself, our implementation returns the full step?

jtlan:
>How does it return something different?
Because I can't read ( > _ < )

src/scales/ordinalScale.ts
```diff
@@ -50,76 +49,42 @@ export module Scale {
...
     public rangeBand() : number {
       return this._d3Scale.rangeBand();
     }
 
-    public innerPadding(): number {
-      var d = this.domain();
-      if (d.length < 2) {
-        return 0;
+    public innerPadding(): number;
+    public innerPadding(innerPadding: number): Ordinal;
```
jtlan: JSDoc? We should specify what units are being used here.

src/scales/ordinalScale.ts
```diff
@@ -50,76 +49,42 @@ export module Scale {
...
-          this._outerPadding = outerPadding;
-        }
-        if (innerPadding != null) {
-          this._innerPadding = innerPadding;
-        }
-        this.range(this.range());
-        this.broadcaster.broadcast();
-        return this;
-      }
+    public outerPadding(outerPadding: number): Ordinal {
```
jtlan: This should probably also be a getter-setter.

src/scales/ordinalScale.ts
```diff
@@ -50,76 +49,42 @@ export module Scale {
...
+      this._outerPadding = outerPadding;
+      this.range(this.range());
+      this.broadcaster.broadcast();
```
jtlan: If we make the `range()` setter broadcast, we won't need to do this.

src/components/axes/categoryAxis.ts
```diff
@@ -137,7 +137,7 @@ export module Axis {
      */
     private _measureTicks(axisWidth: number, axisHeight: number, scale: Scale.Ordinal, ticks: string[]) {
       var wrappingResults = ticks.map((s: string) => {
-        var bandWidth = scale.fullBandStartAndWidth(s)[1];
+        var bandWidth = scale.rangeBand();
```
jtlan: I think this change breaks something, becuase `fullBandAndStartWidth(s)[1]` is wider than `rangeBand()`

test/components/categoryAxisTests.ts
```diff
@@ -79,27 +79,6 @@ describe("Category Axes", () => {
...
-    assert.deepEqual(ticksNormalizedPosition, [1.5, 2.5, 3.5], "bands");
-
-    scale.rangeType("points", 1, 0);
-    step = SVG_WIDTH / 4;
-    ticksNormalizedPosition = tickMarks.map((s: any) => +d3.select(s).attr("x1") / step);
-    assert.deepEqual(ticksNormalizedPosition, [1, 2, 3], "points");
-
-    svg.remove();
-  });
-
```
jtlan: No test for "scales correctly"? (the merging of these tests)

test/components/plots/barPlotTests.ts
```diff
@@ -76,32 +76,6 @@ describe("Plots", () => {
...
-        assert.equal(numAttr(bar0, "width"), 150, "bar0 width is correct");
-        assert.equal(numAttr(bar1, "width"), 150, "bar1 width is correct");
-        assert.equal(numAttr(bar0, "x"), 0, "bar0 x is correct");
-        assert.equal(numAttr(bar1, "x"), 300, "bar1 x is correct");
-
-        assert.throws(() => barPlot.barAlignment("blargh"), Error);
-        assert.equal((<any> barPlot)._barAlignmentFactor, 1, "the bad barAlignment didnt break internal state");
-        svg.remove();
-      });
-
```
jtlan: What happened to this? Did we officially stop supporting `barAlignment()`?

jtlan:
Questions around tests and potentially changed functionality (might also be why the build failed...).

test/components/plots/barPlotTests.ts
bluong: Since barAlignment doesn't exist on bands mode ordinal scales, I'm removing this aspect for ordinal scale-backed plots (unless we want it)

src/scales/ordinalScale.ts
bluong: I just added a setter. I am unsure of how the units should be.

src/scales/ordinalScale.ts
bluong: Probably.  Note that these are API points that I have for pure convenience.

src/scales/ordinalScale.ts
bluong: Good poitn.

bluong: It seems that scales don't broadcast on range setting.  Not sure if I want to broadcast for just ordinalScales

jtlan:
Looking good, can we fill in the documentation for all those methods?

bluong:
@jtlan I'm going to fill in the documentation in another pr.  Hopefully that pr is the last one we need.

jtlan:
+1. Follow up with the change to padding.

PR #1571, opened 2015-02-03
Add in origin-querying methods.

### LINKED ISSUE #1559: Method for returning (x, y) of Component relative to svg
It should do so efficiently, because `Interaction`s might need this to be run many times per second (so we shouldn't use `getBoundingClientRect()`).

One way is to have a method that returns the position of the `Component` relative to its parent, then call this recursively. The protected fields `_xOrigin` and `_yOrigin` sort of accomplish this, but don't take into account `Component` alignment, which moves the component but doesn't change `_xOrigin` or `_yOrigin`.

jtlan:
`origin()` returns a `Component`'s position relative to its parent.
`originToSVG()` returns a `Component`'s position relative to the <svg>(by iteratively calling `origin()`).

Close #1559.


src/components/abstractComponent.ts
```diff
@@ -138,18 +138,14 @@ export module Component {
...
       var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
       this._width  = this._isFixedWidth()  ? Math.min(availableWidth , requestedSpace.width)  : availableWidth ;
       this._height = this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.height) : availableHeight;
 
-      var xPosition = this._xOrigin + this._xOffset;
-      var yPosition = this._yOrigin + this._yOffset;
-      xPosition += (availableWidth - this.width()) * this._xAlignProportion;
-      yPosition += (availableHeight - requestedSpace.height) * this._yAlignProportion;
+      this._xOrigin = xOrigin + this._xOffset + (availableWidth - this.width()) * this._xAlignProportion;
+      this._yOrigin = yOrigin + this._yOffset + (availableHeight - requestedSpace.height) * this._yAlignProportion;;
```
bluong: Might want to refactor the parameters xOrigin/yOrigin.  In this context, now we have xOrigin be some computation of xOrigin amongst other parameters, which is semantically confusing.

bluong: I think I'm a fan with changing the definition of xOrigin and yOrigin in this manner, but I'm a bit hesitant on the names this._xOrigin and the parameter xOrigin as they are different concepts with the same name.

The parameter is just essentially the place that the component will start from whereas the instance variable holds the corrected version after accounting for alignment and offsets.

Also, why not this.height() ?

bluong:
Still need to review tests, but aside from that and the above comment, looks good to me.

test/core/componentTests.ts
```diff
@@ -423,6 +423,132 @@ describe("Component behavior", () => {
...
+      assert.strictEqual(origin.y, SVG_HEIGHT - cHeight, "returns correct value (yAlign bottom)");
+
+      c.xAlign("left").yAlign("top");
+      var xOffsetValue = 40;
+      var yOffsetValue = 30;
+      c.xOffset(xOffsetValue);
+      c.yOffset(yOffsetValue);
+      origin = c.origin();
+      assert.strictEqual(origin.x, xOffsetValue, "accounts for xOffset");
+      assert.strictEqual(origin.y, yOffsetValue, "accounts for yOffset");
```
bluong: Consider moving the offset accounting into a separate test and possibly moving the alignment into an inner describe with multiple align tests.

test/core/componentTests.ts
```diff
@@ -423,6 +423,132 @@ describe("Component behavior", () => {
...
+
+      c.xAlign("center").yAlign("center");
+      origin = c.origin();
+      assert.strictEqual(origin.x, (groupWidth - cWidth)/2, "returns correct value (xAlign center)");
+      assert.strictEqual(origin.y, (groupHeight - cHeight)/2, "returns correct value (yAlign center)");
+
+      c.xAlign("right").yAlign("bottom");
+      origin = c.origin();
+      assert.strictEqual(origin.x, groupWidth - cWidth, "returns correct value (xAlign right)");
+      assert.strictEqual(origin.y, groupHeight - cHeight, "returns correct value (yAlign bottom)");
```
bluong: A lot of these tests don't seem all that useful?  My first guess is that all you need is one of the three since the above unit test suite already covers a lot of what this tests for.

test/core/componentTests.ts
```diff
@@ -423,6 +423,132 @@ describe("Component behavior", () => {
...
+      assert.strictEqual(origin.y, SVG_HEIGHT - cHeight, "returns correct value (yAlign bottom)");
+
+      c.xAlign("left").yAlign("top");
+      var xOffsetValue = 40;
+      var yOffsetValue = 30;
+      c.xOffset(xOffsetValue);
+      c.yOffset(yOffsetValue);
+      origin = c.originToSVG();
+      assert.strictEqual(origin.x, xOffsetValue, "accounts for xOffset");
+      assert.strictEqual(origin.y, yOffsetValue, "accounts for yOffset");
```
bluong: Same comments about splitting up the tests

test/core/componentTests.ts
```diff
@@ -423,6 +423,132 @@ describe("Component behavior", () => {
...
+      origin = c.originToSVG();
+      assert.strictEqual(origin.x, (groupWidth - cWidth)/2 + groupXOffset, "returns correct value (xAlign center)");
+      assert.strictEqual(origin.y, (groupHeight - cHeight)/2 + groupYOffset, "returns correct value (yAlign center)");
+
+      c.xAlign("right").yAlign("bottom");
+      origin = c.originToSVG();
+      assert.strictEqual(origin.x, groupWidth - cWidth + groupXOffset, "returns correct value (xAlign right)");
+      assert.strictEqual(origin.y, groupHeight - cHeight + groupYOffset, "returns correct value (yAlign bottom)");
+
+      svg.remove();
```
bluong: Same comments

src/components/abstractComponent.ts
jtlan: `this.height()` makes sense. Changing.

What should we call `this._xOrigin` and `this._yOrigin` then?

test/core/componentTests.ts
jtlan: These calls would have returned the wrong values before my modification to the `this._xOrigin` and `this._yOrigin` values...

jtlan:
Probably no more than a quick regression pass is needed here, @crmorford .

crmorford:
Everything looks fine, you can merge :)

PR #1582, opened 2015-02-05
[OrdinalScale] - Padding logic completely reworked.

bluong:
Now the values are in proportion to the bar width.


jtlan:
Can we adjust the default values for `innerPadding` and `outerPadding` to get the same visual effects as currently?

PR #1586, opened 2015-02-06
[OrdinalScale] - Functionality revamp

### LINKED ISSUE #1558: Create Single consistent mode for Scale.Ordinal
Right now `Scale.Ordinal` supports both "bands" and "points" mode. However, there's a logical expected behavior that the user always wants, so they shouldn't be burdened by thinking about whether a particular `Plot` needs the scale in "bands" mode (`Bar` plots) or "points" (everybody else...).

A single mode with the best behavior of both modes is what's needed.

bluong:
OrdinalScales have been greatly revamped.  Now, the ordinalScale only uses a modified version of the rangeBand function of d3's ordinal scales, where the scale divides the scale up so that each domain value has a band and the scale now scales to the center of the band.

Features added:
 * innerPadding and outerPadding API have been added/reworked on the ordinalScale to be in proportion to the band width rather than the width of the full band

* fullBandStartAndWidth has been replaced with fullBandWidth which is the available space the the band can occupy given there is no inner padding.

* scale now scales to the center of the band rather than the start

Features removed:
* rangeType has been removed since maintaining two rangeTypes has led to architectural confusion and very little benefit.

QE:
* Functionality should be mostly maintained (except for the add/removed endpoints), except that now line plots/scatter plots and the like will now scale to the center of the band rather than the the start.

Fixes #1558


crmorford:
Good! I didn't get to test on IE as thoroughly as I wanted since quicktests and IE weren't playing nice. I tried some jsfiddles though and they all looked fine.

Issue #1490, opened 2015-01-05
Zooming on IE9 causes hover to work incorrectly


jtlan:
Repro: http://jsfiddle.net/wu91q6wy/

Try hovering over the points. Then zoom in with `Ctrl`+`+` and note how the hitbox area for the scatterplot no longer aligns with the points in the plot.

jtlan:
Related to #1491, probably.

PR #1590, opened 2015-02-06
Convert Dispatcher.Mouse to a <svg>-wide Dispatcher.

jtlan:
Only one is created per \<svg\>.
Now detects mouse position by drawing a test rectangle,
which allows it to account for CSS zoom and the IE9 zoom bug (#1490).

AbstractDispatcher to follow.

### DIFF
--- a/src/dispatchers/mouseDispatcher.ts
+++ b/src/dispatchers/mouseDispatcher.ts
@@ -3,11 +3,29 @@
 module Plottable {
 export module Dispatcher {
   export class Mouse {
-    static dispatcherKey = "__Plottable_Dispatcher_Mouse";
+    private static dispatcherKey = "__Plottable_Dispatcher_Mouse";
     private _svg: SVGElement;
     private _measureRect: SVGElement;
     private _lastMousePosition: Point;
     public broadcaster: Core.Broadcaster;
+
+    /**
+     * Get a Dispatcher.Mouse for the <svg> containing elem. If one already exists
+     * on that <svg>, it will be returned; otherwise, a new one will be created.
+     *
+     * @param {SVGElement} elem A svg DOM element.
+     * @return {Dispatcher.Mouse} A Dispatcher.Mouse
+     */
+    public static getDispatcher(elem: SVGElement): Dispatcher.Mouse {
+      var svg = _Util.DOM.getBoundingSVG(elem);
+
+      var dispatcher: Mouse = (<any> svg)[Mouse.dispatcherKey];
+      if (dispatcher == null) {
+        dispatcher = new Mouse(svg);
+        (<any> svg)[Mouse.dispatcherKey] = dispatcher;
+      }
+      return dispatcher;
+    }
 
     /**
      * Creates a Dispatcher.Mouse.
@@ -27,9 +45,34 @@ export module Dispatcher {
       this._lastMousePosition = { x: -1, y: -1 };
       this.broadcaster = new Core.Broadcaster(this);
 
-      svg.addEventListener("mouseover", (e) => this._computeMousePosition(e.clientX, e.clientY));
-      svg.addEventListener("mousemove", (e) => this._computeMousePosition(e.clientX, e.clientY));
-      svg.addEventListener("mouseout", (e) => this._computeMousePosition(e.clientX, e.clientY));
+      svg.addEventListener("mouseover", (e) => this._processMoveEvent(e));
+      svg.addEventListener("mousemove", (e) => this._processMoveEvent(e));
+      svg.addEventListener("mouseout", (e) => this._processMoveEvent(e));
+    }
+
+    /**
+     * Registers a callback to be called whenever the mouse position changes,
+     * or removes the callback if `null` is passed as the callback.
+     *
+     * @param {any} key The key associated with the callback.
+     *                  Key uniqueness is determined by deep equality.
+     * @param {(p: Point) => any} callback A callback that takes the pixel position
+     *                                     in svg-coordinate-space. Pass `null`
+     *                                     to remove a callback.
+     * @return {Dispatcher.Mouse} The calling Dispatcher.Mouse.
+     */
+    public onMouseMove(key: any, callback: (p: Point) => any): Mouse {
+      if (callback === null) { // remove listener if callback is null
+        this.broadcaster.deregisterListener(key);
+      } else {
+        this.broadcaster.registerListener(key, () => { callback(this.getLastMousePosition()); });
+      }
+      return this;
+    }
+
+    private _processMoveEvent(e: MouseEvent) {
+      this._computeMousePosition(e.clientX, e.clientY);
+      this.broadcaster.broadcast();
     }
 
     /**
@@ -63,26 +106,6 @@ export module Dispatcher {
       };
 
       this._lastMousePosition = scaledPosition;
-      this.broadcaster.broadcast();
-    }
-
-    /**
-     * Get a Dispatcher.Mouse for the <svg> containing elem. If one already exists
-     * on that <svg>, it will be returned; otherwise, a new one will be created.
-     *
-     * @param {SVGElement} elem A svg DOM element.
-     *
-     * @return {Dispatcher.Mouse} A Dispatcher.Mouse
-     */
-    public static getDispatcher(elem: SVGElement): Dispatcher.Mouse {
-      var svg = _Util.DOM.getBoundingSVG(elem);
-
-      var dispatcher: Mouse = (<any> svg)[Mouse.dispatcherKey];
-      if (dispatcher == null) {
-        dispatcher = new Mouse(svg);
-        (<any> svg)[Mouse.dispatcherKey] = dispatcher;
-      }
-      return dispatcher;
     }
 
     /**
--- a/src/dispatchers/mouseDispatcher.ts
+++ b/src/dispatchers/mouseDispatcher.ts
@@ -3,7 +3,7 @@
 module Plottable {
 export module Dispatcher {
   export class Mouse {
-    private static dispatcherKey = "__Plottable_Dispatcher_Mouse";
+    private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Mouse";
     private _svg: SVGElement;
     private _measureRect: SVGElement;
     private _lastMousePosition: Point;
@@ -19,10 +19,10 @@ export module Dispatcher {
     public static getDispatcher(elem: SVGElement): Dispatcher.Mouse {
       var svg = _Util.DOM.getBoundingSVG(elem);
 
-      var dispatcher: Mouse = (<any> svg)[Mouse.dispatcherKey];
+      var dispatcher: Mouse = (<any> svg)[Mouse._DISPATCHER_KEY];
       if (dispatcher == null) {
         dispatcher = new Mouse(svg);
-        (<any> svg)[Mouse.dispatcherKey] = dispatcher;
+        (<any> svg)[Mouse._DISPATCHER_KEY] = dispatcher;
       }
       return dispatcher;
     }
@@ -71,7 +71,7 @@ export module Dispatcher {
     }
 
     private _processMoveEvent(e: MouseEvent) {
-      this._computeMousePosition(e.clientX, e.clientY);
+      this._lastMousePosition = this._computeMousePosition(e.clientX, e.clientY);
       this.broadcaster.broadcast();
     }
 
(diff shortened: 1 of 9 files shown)


src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,96 @@
 
 module Plottable {
 export module Dispatcher {
-  export class Mouse extends Dispatcher.AbstractDispatcher {
-    private _mouseover: (location: Point) => any;
-    private _mousemove: (location: Point) => any;
-    private _mouseout: (location: Point) => any;
+  export class Mouse {
+    static dispatcherKey = "__Plottable_Dispatcher_Mouse";
```
bluong: Let's specify if this is public or private

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
 
 module Plottable {
 export module Dispatcher {
-  export class Mouse extends Dispatcher.AbstractDispatcher {
-    private _mouseover: (location: Point) => any;
-    private _mousemove: (location: Point) => any;
-    private _mouseout: (location: Point) => any;
+  export class Mouse {
+    private static dispatcherKey = "__Plottable_Dispatcher_Mouse";
```
bluong: What's the need for the double underscore?  Also, what's the convention for the dispatcher key (Why must it have Plottable and Dispatcher)?

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
...
+    private _svg: SVGElement;
+    private _measureRect: SVGElement;
```
bluong: What's the need for this instance var?

src/dispatchers/mouseDispatcher.ts
bluong: Change var to possibly DISPATCHER_KEY

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
...
+    private _lastMousePosition: Point;
+    public broadcaster: Core.Broadcaster;
 
     /**
-     * Constructs a Mouse Dispatcher with the specified target.
+     * Get a Dispatcher.Mouse for the <svg> containing elem. If one already exists
+     * on that <svg>, it will be returned; otherwise, a new one will be created.
```
bluong: What's the need for the magic behind creating the new one if one isn't made already?

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
...
+      this._measureRect.setAttribute("y", String((clientY - origin.y)/scaleY) );
+      mrBCR = this._measureRect.getBoundingClientRect();
+      var trueCursorPosition = { x: mrBCR.left, y: mrBCR.top };
+
+      var scaledPosition = {
+        x: (trueCursorPosition.x - origin.x) / scaleX,
+        y: (trueCursorPosition.y - origin.y) / scaleY
+      };
+
+      this._lastMousePosition = scaledPosition;
```
bluong: Can we separate the computation from the setting of the _lastMousePosition?  It'll make testing easier and possible coding downstream easier.

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
...
      */
-    public mouseout(): (location: Point) => any;
+    private _computeMousePosition(clientX: number, clientY: number) {
+      // get the origin
+      this._measureRect.setAttribute("x", "0");
+      this._measureRect.setAttribute("y", "0");
+      var mrBCR = this._measureRect.getBoundingClientRect();
+      var origin = { x: mrBCR.left, y: mrBCR.top };
+
+      // caculate the scale
```
bluong: calculate*

src/dispatchers/mouseDispatcher.ts
```diff
@@ -2,110 +2,119 @@
...
+      this._measureRect.setAttribute("x", "100");
+      this._measureRect.setAttribute("y", "100");
+      mrBCR = this._measureRect.getBoundingClientRect();
+      var testPoint = { x: mrBCR.left, y: mrBCR.top };
+
+      var scaleX = (testPoint.x - origin.x) / 100;
+      var scaleY = (testPoint.y - origin.y) / 100;
```
bluong: What's the scale portion for?  css zoom factors?

bluong: Also why does it use magic numbers 100?  percentage reason?

src/interactions/abstractInteraction.ts
```diff
@@ -17,6 +17,34 @@ export module Interaction {
...
+    }
+
+    /**
+     * Checks whether a Component-coordinate-space Point is inside the Component.
+     *
+     * @param {Point} p A Point in Coordinate-space coordinates.
+     *
+     * @return {boolean} Whether or not the point is inside the Component.
+     */
+    protected _isInsideComponent(p: Point) {
```
bluong: Worth it to make a function in svg coordinate space?

src/interactions/abstractInteraction.ts
```diff
@@ -17,6 +17,34 @@ export module Interaction {
...
+
+    /**
+     * Translates an <svg>-coordinate-space point to Component-space coordinates.
+     *
+     * @param {Point} p A Point in <svg>-space coordinates.
+     *
+     * @return {Point} The same location in Component-space coordinates.
+     */
+    protected _translateToComponentSpace(p: Point): Point {
+      var origin = this._componentToListenTo.originToSVG();
```
bluong: Aside - After reading this usage, I thought the return variable would be an SVG but nope it's the origin for the SVG.  While this is an API-breaking change, would svgOrigin be a bad name?

src/interactions/hoverInteraction.ts
```diff
@@ -45,14 +46,20 @@ export module Interaction {
 
     public _anchor(component: Hoverable, hitBox: D3.Selection) {
       super._anchor(component, hitBox);
-      this._dispatcher = new Dispatcher.Mouse(this._hitBox);
+      this._dispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> (<any> this._componentToListenTo)._element.node());
```
bluong: What's with the any casting / casting here?

Is _element private/protected?  I remember @jtlan asking that element be exposed

src/interactions/hoverInteraction.ts
```diff
@@ -45,14 +46,20 @@ export module Interaction {
...
 
-      this._dispatcher.mouseover((p: Point) => {
-        this._componentToListenTo._hoverOverComponent(p);
-        this.handleHoverOver(p);
-      });
+      this._dispatcher.onMouseMove("hover"+this.getID(), (p: Point) => this._handleMouseEvent(p));
```
bluong: Let's space out the concatentation on the left

src/utils/domUtils.ts
```diff
@@ -116,6 +116,17 @@ export module _Util {
         Math.floor(inner.bottom) <= Math.ceil(outer.bottom)
       );
     }
+
+    export function getBoundingSVG(elem: SVGElement): SVGElement {
+      var ownerSVG = elem.ownerSVGElement;
+      if (ownerSVG != null) {
+        return ownerSVG;
+      }
+      if (elem.nodeName.toLowerCase() === "svg") { // elem itself is an SVG
```
bluong: Is the ownerSVGElement of an svg not itself?

test/dispatchers/dispatcherTests.ts
```diff
@@ -80,8 +80,34 @@ describe("Dispatchers", () => {
...
+
+      var md1 = Plottable.Dispatcher.Mouse.getDispatcher(<SVGElement> svg.node());
+      assert.isNotNull(md1, "created a new Dispatcher on an SVG");
+      var md2 = Plottable.Dispatcher.Mouse.getDispatcher(<SVGElement> svg.node());
+      assert.strictEqual(md1, md2, "returned the existing Dispatcher if called again with same <svg>");
+
+      svg.remove();
+    });
+
+    it("getLastMousePosition() defaults to a safe value", () => {
```
bluong: What defines safe?

test/interactions/hoverInteractionTests.ts
```diff
@@ -148,7 +148,7 @@ describe("Interactions", () => {
       assert.deepEqual(currentlyHovered.data, ["left", "right"],
                        "retrieves data corresponding to the current position");
 
-      triggerFakeMouseEvent("mouseout", hitbox, 400, 200);
+      triggerFakeMouseEvent("mouseout", hitbox, 401, 200);
```
bluong: Why the offset?

bluong:
Many many questions

src/dispatchers/mouseDispatcher.ts
jtlan: Double-underscore and named because it's being set as a property on a DOM node, so I want to minimize changes of collision. D3 uses `__data__` for example.

src/dispatchers/mouseDispatcher.ts
jtlan: Scale portion for CSS zoom.  Any number would work as well as 100, so long as it's set as the attribute and used to measure. Using a larger integer should avoid floading-point problems.

src/dispatchers/mouseDispatcher.ts
jtlan: `_measureRect` is used to sample the mouse position.

src/dispatchers/mouseDispatcher.ts
jtlan: We could create a new `Dispatcher.Mouse` every time, but since they're all doing the same calculation (relative to the `<svg>` they're attached to), that would be redundant.

src/interactions/abstractInteraction.ts
jtlan: I'm not sure I understand the question.

src/interactions/abstractInteraction.ts
jtlan: "`svgOrigin`" sounds like it gets the origin *of the `<svg>`*, instead of the origin *relative to* the `<svg>`.

src/interactions/abstractInteraction.ts
bluong: This checks based on the Point being in Coordinate-space.  Would it be handy to make a similar function in svg coordinate space?

src/interactions/hoverInteraction.ts
jtlan: `<any>` casting to get `_element`.

The other casting is because `<Element>` can't be cast to `<SVGElement>` directly, and `.node()` returns `Element` (regardless of what the node actually is). I could add a comment about the cast, which would look mysterious after compilation to js because all the casts vanish.

src/utils/domUtils.ts
jtlan: No, the `ownerSVG` of an `<svg>` node is `null`.

test/dispatchers/dispatcherTests.ts
jtlan: non-null, and values within non-null.

test/interactions/hoverInteractionTests.ts
jtlan: because otherwise the `mouseout` event occurs inside the hitbox, failing to actually trigger a `mouseout` with the new `Dispatcher`.

src/dispatchers/mouseDispatcher.ts
bluong: Understood.

Still please change the var name to be in all caps with underscores

src/dispatchers/mouseDispatcher.ts
bluong: Still would want this

src/dispatchers/mouseDispatcher.ts
bluong: fix typo please.

src/dispatchers/mouseDispatcher.ts
bluong: I would factor it out so that we have a single source of truth.

src/interactions/abstractInteraction.ts
bluong: True.  Holding on my comment until I see another semantical pain point

src/interactions/hoverInteraction.ts
bluong: Would you still want the `_element` to be an API point (aside)?  I remember you saying that it would be useful.

test/dispatchers/dispatcherTests.ts
bluong: Let's make a comment on this somewhere (within the test name probably)

src/dispatchers/mouseDispatcher.ts
jtlan: --> "`_DISPATCHER_KEY`"

src/dispatchers/mouseDispatcher.ts
jtlan: Now returns `scaledPosition`, which is used to set `_lastMousePosition` elsewhere.

PR #1593, opened 2015-02-06
Playground

crmorford:
testing out a new idea: webpage that can generate the code for most simple Plottable charts.

(also some changes to quicktests that were done when convenient)


quicktests/overlaying/playground.html
```diff
@@ -0,0 +1,75 @@
...
+        </div>
+
+        <br>
+
+        <div align="center">
+            <span> width (px): </span>
+            <input type="text" id="svg_w" rows="1" cols="10" value="400"></textarea>
+            <span> height (px): </span>
+            <input type="text" id="svg_h" rows="1" cols="10" value="400"></textarea>
+            <button id="changeSizeButton" onclick="svg_size()"> Change Size </button>
```
rcchen: Probably want to change this to "Update" or something because it's the only way to trigger update behavior with the graph. Would also be cool to add a changeListener or something so the button auto-changes to say "update" when you change parameters for the graph.

jtlan:
Should this be in a different directory than `overlaying`? It's a different type of Quicktest, right?

rcchen:
Closing this for now.

PR #1594, opened 2015-02-06
[Legend] Add prefix to legend entry class names

### LINKED ISSUE #1321: Sanitize applied CSS class names in Legend and HorizontalLegend
Currently, `Legend` and `HorizontalLegend` apply the entry name as a CSS class to each entry. However, not all valid entries are valid CSS classes, such as "2". We should sanitize the CSS class name in these cases.

Repro:
http://jsfiddle.net/Lg0dtvrp/

Expected: "2" circle is black.
Actual: "2" circle is still red.

rcchen:
Fixes #1321. The approach used here is to prefix legend entry class names with the value of LEGEND_ENTRY, which should make all class names semantically correct according to http://stackoverflow.com/questions/448981/what-characters-are-valid-in-css-class-selectors

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2325,6 +2325,10 @@ declare module Plottable {
              */
             static LEGEND_ENTRY_CLASS: string;
             /**
+             * The css class that prepends every legend entry
+             */
+            static LEGEND_ENTRY_PREFIX: string;
+            /**
              * Creates a Legend.
              *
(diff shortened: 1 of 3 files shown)


jtlan:
Not sure we should actually fix #1321 at all, given that the non-valid names are still selectable (as mentioned on the ticket). Cross-check with Slate and see what their feelings on this are?

rcchen:
Deferring to later.

PR #1597, opened 2015-02-06
Fixes #1033 SVG visibility issues with Firefox

### LINKED ISSUE #1033: Setting SVG visibility to hidden does not completely hide on Firefox
Super specific yes.

screen shot 2014-09-12 at 2 47 56 pm

Here is a fiddle of the repro: http://jsfiddle.net/2a8y33q4/

rcchen:
Fixes #1033

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -4269,7 +4269,8 @@ var Plottable;
             };
             AbstractAxis.prototype._hideOverlappingTickLabels = function () {
                 var visibleTickLabels = this._tickLabelContainer.selectAll("." + AbstractAxis.TICK_LABEL_CLASS).filter(function (d, i) {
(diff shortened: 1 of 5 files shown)


bluong:
Can you give a brief description in the PR for what the resolution is?  I'm curious myself.

rcchen:
tl;dr changing the CSS class to inherit instead of visible, and then making sure that selectors are looking for anything that matches "visible" or "inherit" when selecting the pool of ticks that can be deselected from.

rcchen:
This way the visibility of ticks is inherited from the parent SVG.

crmorford:
Looks good! Congrats on your victory against IE.

PR #1602, opened 2015-02-07
[TS1.4] - Utilizing the 'type' feature in Typescript

bluong:
Now that Typescript has added the type feature in Typescript 1.4, we can make use of it: http://blogs.msdn.com/b/typescript/archive/2015/01/16/announcing-typescript-1-4.aspx

The type feature allows for Type aliasing which means that you can alias a type into a different name.

One simple example can be to alias
`type Phrase = string;`
so that you can use the term `phrase` as a replacement for the term `string` throughout the codebase, which is very nice for semantical reasoning.

A more useful example can be to alias
`type InteractionCallback = (data: any) => any;`
so that we can now alias callbacks with hard-to-reason-about signatures to a concise, short term

A more interesting example can be to alias
`type Person = {
  firstName: string;
  lastName: string;
}`
so that we can now encapsulate complex ideas even further.

### DIFF
--- a/src/components/axes/timeAxis.ts
+++ b/src/components/axes/timeAxis.ts
@@ -9,7 +9,7 @@ export module Axis {
    * step - number of intervals between each tick.
    * formatter - formatter used to format tick labels.
    */
-  export interface TimeAxisTierConfiguration {
+  export type TimeAxisTierConfiguration = {
     interval: D3.Time.Interval;
     step: number;
     formatter: Formatter;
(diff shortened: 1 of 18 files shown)


src/components/axes/timeAxis.ts
```diff
@@ -20,9 +20,7 @@ export module Axis {
    * Each configuration will be shown on a different tier.
    * Currently, up to two tiers are supported.
    */
-  export interface TimeAxisConfiguration {
-    tierConfigurations: TimeAxisTierConfiguration[];
-  }
+  export type TimeAxisConfiguration = TimeAxisTierConfiguration[];
```
jtlan: BOSS

jtlan:
Please note the breaking change on the [release changes page](https://github.com/palantir/plottable/wiki/Release-Changes).

src/components/axes/timeAxis.ts
bluong: Yep will note

crmorford:
Can you update the time_tier_config quicktest? This broke it

PR #1611, opened 2015-02-10
[Plot] - getAllSelections() with dataset key params

bluong:
Adding optional parameters to the API point getAllSelections to look like
`getAllSelections(datasetKeys?: string | string[])`

The above API endpoint will allow users to retrieve all of the selections for the specified dataset(s).

NOTE: One possible API would be to change getAllSelections to
`getAllSelections(...datasetKeys?: string[] | string[][])`

This would allow input like
`getAllSelections("a", "b", "c")` instead of forcing to use an array.  I still would want the array API endpoint but this is optional.

NOTE: I'm currently in consideration of a second optional boolean argument `exclude` which will return the selections of all datasets excluding the ones specified by the first datasetKeys argument

### DIFF
--- a/test/components/plots/areaPlotTests.ts
+++ b/test/components/plots/areaPlotTests.ts
@@ -31,7 +31,7 @@ describe("Plots", () => {
       svg = generateSVG(500, 500);
       simpleDataset = new Plottable.Dataset(twoPointData);
       areaPlot = new Plottable.Plot.Area(xScale, yScale);
-      areaPlot.addDataset(simpleDataset)
+      areaPlot.addDataset("sd", simpleDataset)
               .project("x", xAccessor, xScale)
               .project("y", yAccessor, yScale)
               .project("y0", y0Accessor, yScale)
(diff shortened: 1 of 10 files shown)


jtlan:
Wouldn't a different name for the API point, like `getSelections()` or `getSelectionsForDatasets()` make more sense? @crmorford, any opinions?

jtlan:
Also, this implementation merges all the selections -- what if someone wants to use the `Dataset`'s key to filter selections or the effects of `Interaction`s?

jtlan:
Also, I hate that you can't tell who's going to be assigned on github unless you scroll the assignee field into view.

bluong:
`getSelections()` would not be a good API name since it would almost certainly collide with a future name for an API name that would retreive selections that are not all the selections.  I am also not a fan of `getSelectionsForDatasets()` since the name is verbose and I think adding another argument in the current API would actually say how to getSelections more concisely with also less API surface area

If someone wanted to use the `Dataset`'s key to filter the selections then they would just use those keys in this API call.

bluong:
Assigning to @crmorford as we want the API signature to see if we want to load a current API point or write a new one.

test/components/plots/areaPlotTests.ts
```diff
@@ -113,17 +113,54 @@ describe("Plots", () => {
...
+        assert.strictEqual(allAreas.size(), 1, "all areas retrieved");
+        var selectionData = allAreas.data();
+        assert.include(selectionData, newTwoPointData, "new dataset data in selection data");
+
+        svg.remove();
+      });
+
+      it("retrieves correct selections (array arg)", () => {
+        var newTwoPointData = [{ foo: 2, bar: 1 }, { foo: 3, bar: 2 }];
+        var newDataset = new Plottable.Dataset(twoPointData);
```
jtlan: This is never used... (twice more in tests above)

test/components/plots/areaPlotTests.ts
```diff
@@ -113,17 +113,54 @@ describe("Plots", () => {
...
+        assert.strictEqual(allAreas.size(), 1, "all areas retrieved");
+        var selectionData = allAreas.data();
+        assert.include(selectionData, newTwoPointData, "new dataset data in selection data");
+
+        svg.remove();
+      });
+
+      it("skips invalid keys", () => {
+        var newTwoPointData = [{ foo: 2, bar: 1 }, { foo: 3, bar: 2 }];
+        var newDataset = new Plottable.Dataset(twoPointData);
```
jtlan: This is never used (3x more in above tests).

test/components/plots/areaPlotTests.ts
```diff
@@ -113,17 +113,54 @@ describe("Plots", () => {
...
-
-      svg.remove();
+    describe("getAllSelections()", () => {
+
+      it("retrieves all selections with no args", () => {
+        var newTwoPointData = [{ foo: 2, bar: 1 }, { foo: 3, bar: 2 }];
+        var newDataset = new Plottable.Dataset(twoPointData);
+        areaPlot.addDataset("newTwo", new Plottable.Dataset(newTwoPointData));
+        var allAreas = areaPlot.getAllSelections();
+        var allAreas2 = areaPlot.getAllSelections((<any> areaPlot)._datasetKeysInOrder);
```
jtlan: Seems odd to compare it to something else that gets tested later... can we create a ground truth using `d3.select` or the protected `_content` field?

test/components/plots/areaPlotTests.ts
```diff
@@ -113,17 +113,54 @@ describe("Plots", () => {
...
+
+        svg.remove();
+      });
+
+      it("retrieves correct selections (string arg)", () => {
+        var newTwoPointData = [{ foo: 2, bar: 1 }, { foo: 3, bar: 2 }];
+        var newDataset = new Plottable.Dataset(twoPointData);
+        areaPlot.addDataset("newTwo", new Plottable.Dataset(newTwoPointData));
+        var allAreas = areaPlot.getAllSelections("newTwo");
+        assert.strictEqual(allAreas.size(), 1, "all areas retrieved");
```
jtlan: Doesn't confirm that the *right* selection was returned, just *a* selection.

test/components/plots/areaPlotTests.ts
```diff
@@ -113,17 +113,54 @@ describe("Plots", () => {
...
+
+        svg.remove();
+      });
+
+      it("retrieves correct selections (array arg)", () => {
+        var newTwoPointData = [{ foo: 2, bar: 1 }, { foo: 3, bar: 2 }];
+        var newDataset = new Plottable.Dataset(twoPointData);
+        areaPlot.addDataset("newTwo", new Plottable.Dataset(newTwoPointData));
+        var allAreas = areaPlot.getAllSelections(["newTwo"]);
+        assert.strictEqual(allAreas.size(), 1, "all areas retrieved");
```
jtlan: Same criticism as above test -- doesn't check for the *right* selection.

jtlan:
A lot of testing questions. Do we need to test this on all `Plot` types? It seems like since its implemented on `AbstractPlot`, we could test against it there.

bluong:
I think we have 2 options: one less painful than the other.  Either test it on just AbstractPlot or test it against its immediate children.  Unfortunately I didn't think at the time that there was enough architecture that would allow AbstractPlot to work which is why I didn't test it on AbstractPlot.  I'll try looking at it again.

bluong:
After looking at the plotTests, I didn't see enough architecture that could allow me to follow through with implementing the tests on the generic AbstractPlot, since the AbstractPlot has no capability of drawing and since AbstractDrawer also has no capability of drawing.  We should consider building architecture to make testing easier as I think the current way is the safest way to test... but this architecture doesn't exist right now.

test/components/plots/areaPlotTests.ts
bluong: Copy paste devil at work.

test/components/plots/areaPlotTests.ts
bluong: I specifically wanted to assert that it returned the same thing that getAllSelections with the same all dataset keys did, rather than assert the ground truth.  Do you think it should instead assert the ground truth?

The way it is now ties the behavior of the two functions whereas changing it will essentially test each one independently.  It is mainly a  matter of which behavior we want ourselves.

test/components/plots/areaPlotTests.ts
bluong: It checks for the right selection by querying the data underneath it.

test/components/plots/areaPlotTests.ts
bluong: Same as above.

jtlan:
@bluong wrote:
> After looking at the plotTests, I didn't see enough architecture that could allow me to follow through with implementing the tests on the generic AbstractPlot, since the AbstractPlot has no capability of drawing and since AbstractDrawer also has no capability of drawing. We should consider building architecture to make testing easier as I think the current way is the safest way to test... but this architecture doesn't exist right now.

I think the way to test this is to create a "[mock](http://en.wikipedia.org/wiki/Mock_object)": implement a `Drawer` class in the test file that produces some testable behavior, create a `Plot` that uses that `Drawer`, and test that.

crmorford:
http://jsfiddle.net/xL0duLo9/10/
GIL

PR #1613, opened 2015-02-10
[Plot] - Switch order of addDataset parameters

bluong:
After interacting with addDataset and even with writing the doc for this signature, I found that having the optional key parameter as the first parameter has been painful.  In general, I think having the optional parameter as the later input arguments just is good style and is easier to interact with.

API-Breaks: Input parameters of addDataset are swapped.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2617,12 +2617,11 @@ declare module Plottable {
              *
              * A key is automatically generated if not supplied.
              *
-             * @param {string} [key] The key of the dataset.
              * @param {Dataset | any[]} dataset dataset to add.
+             * @param {string} [key] The key of the dataset.
              * @returns {Plot} The calling Plot.
              */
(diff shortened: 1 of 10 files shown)


jtlan:
My concern here is that for everything else that needs a key, the key goes first, then the other argument. Compare: `Broadcaster`s. I think the problem was allowing adding datasets without a key. Do we still want to support that use case?

bluong:
The difference is that the key in this case is optional, meaning that from a signature and a usage perspective, it seems a bit odd in my opinion.

jtlan:
This reversal might mess with a lot of people...

crmorford:
We should talk to @guptaso2 about if changing this is worth introducing a breaking change. We haven't had any complaints, right?

bluong:
Getting second opinion on this matter.

bluong:
We should do this before 1.0 but not now

PR #1614, opened 2015-02-11
Convert Broadcaster to use generics, remove Listenable.

### LINKED ISSUE #1604: Broadcaster is public, and only one per object
Currently, `Broadcaster` requires that a `Listenable` be passed in, and a `Listenable` has the following contract:

```Typescript
export interface Listenable {
  broadcaster: Broadcaster;
}
```

This setup has a number of problems:
* The `broadcaster` field is public.
* Having more than one `Broadcaster` per `Listenable` is difficult, because one has to be named "`broadcaster`".

Instead of defining `Listenable` as a distinct type, we should allow `Broadcaster`s to be able to be attached to arbitrary `Object`s.

jtlan:
Previously: `Broadcasters` had-a `Listenable`, which had-a reference to the same `Broadcaster`. This was silly and prevented the use of multiple
`Broadcasters` on the same object.

Now we can use generic typing to convey more information about what the `Broadcaster` is broadcasting.

Close #1604.

### DIFF
--- a/src/core/broadcaster.ts
+++ b/src/core/broadcaster.ts
@@ -3,50 +3,27 @@
 module Plottable {
 export module Core {
   /**
-   * This interface represents anything in Plottable which can have a listener attached.
-   * Listeners attach by referencing the Listenable's broadcaster, and calling registerListener
-   * on it.
-   *
-   * e.g.:
-   * listenable: Plottable.Listenable;
-   * listenable.broadcaster.registerListener(callbackToCallOnBroadcast)
-   */
-  export interface Listenable {
-    broadcaster: Broadcaster;
-  }
-
-  /**
-   * This interface represents the callback that should be passed to the Broadcaster on a Listenable.
-   *
-   * The callback will be called with the attached Listenable as the first object, and optional arguments
-   * as the subsequent arguments.
-   *
-   * The Listenable is passed as the first argument so that it is easy for the callback to reference the
-   * current state of the Listenable in the resolution logic.
-   */
-  export type BroadcasterCallback = (listenable: Listenable, ...args: any[]) => any;
-
-  /**
-   * The Broadcaster class is owned by an Listenable. Third parties can register and deregister listeners
-   * from the broadcaster. When the broadcaster.broadcast method is activated, all registered callbacks are
-   * called. The registered callbacks are called with the registered Listenable that the broadcaster is attached
-   * to, along with optional arguments passed to the `broadcast` method.
+   * The Broadcaster holds a reference to a "listenable" object.
+   * Third parties can register and deregister listeners from the Broadcaster.
+   * When the broadcaster.broadcast() method is called, all registered callbacks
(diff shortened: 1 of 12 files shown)


src/core/broadcaster.ts
```diff
@@ -3,50 +3,27 @@
...
-    public listenable: Listenable;
+    private _listenable: L;
 
     /**
-     * Constructs a broadcaster, taking the Listenable that the broadcaster will be attached to.
+     * Constructs a broadcaster, taking a "listenable" object to broadcast about.
      *
      * @constructor
-     * @param {Listenable} listenable The Listenable-object that this broadcaster is attached to.
+     * @param {L} listenable The listenable object to broadcast about.
```
bluong: broadcast about sounds a bit odd...  Is that right semantically?

src/core/broadcaster.ts
```diff
@@ -55,10 +32,10 @@ export module Core {
      * If there is already a callback associated with that key, then the callback will be replaced.
      *
      * @param key The key associated with the callback. Key uniqueness is determined by deep equality.
-     * @param {BroadcasterCallback} callback A callback to be called when the Scale's domain changes.
-     * @returns {Broadcaster} this object
+     * @param {(listenable: L, ...args: any[]) => any} callback A callback to be called.
```
bluong: I assume type doesn't work?  Maybe just use an interface here to better inform / encapsulate

src/core/resizeBroadcaster.ts
```diff
@@ -14,12 +14,12 @@ export module Core {
...
    */
   export module ResizeBroadcaster {
-    var broadcaster: Core.Broadcaster;
+    var broadcaster: Core.Broadcaster<any>;
     var _resizing: boolean = false;
 
     function _lazyInitialize() {
       if (broadcaster === undefined) {
-        broadcaster = new Core.Broadcaster(<any> ResizeBroadcaster);
+        broadcaster = new Core.Broadcaster<any>(<any> ResizeBroadcaster);
```
bluong: What is with the casting here

bluong:
Very minor issues.

src/core/broadcaster.ts
jtlan: maybe just "to broadcast"?

src/core/broadcaster.ts
jtlan: Can't define the interface inside the class either, and the generic isn't scoped outside of it.

src/core/broadcaster.ts
bluong: Works for me

src/core/resizeBroadcaster.ts
jtlan: The first `<any>` is the generic. The second `<any>` was a cast that isn't needed anymore.

src/core/broadcaster.ts
bluong: We however can define a generic typed interface at the module level and pass in the L.  Would that work?

PR #1616, opened 2015-02-11
Remove Constructor instance methods

bluong:
QE: Regression pass at most.  This is mainly architectural cleanup.

After some time, we found that there was a pain point in terms of awkward bugs arising from the fact that instance methods are called from the constructor.  The reason is because one can argue that until a constructor is fully finished, the object is not fully constructed and the instance variables are not all fully set yet.  The problem with instance methods is that the many of them tend to assume that the object is in a fully constructed state before executing, which can lead to weird bugs if instance methods have not been initialized before the instance method is called.


crmorford:
This actually broke the clustered_bar quicktest

crmorford:
looks like legends now merge into the middle of a plot, instead of at the top right.

crmorford:
Lets figure out why the no_data no_renderer scales are different, otherwise this looks fine now

bluong:
I'll use this branch in order to make a new branch that will try to make constructors cleaner

PR #1621, opened 2015-02-11
[Plot] - getSelections() API and implementation on RectDrawers

bluong:
Implementaion getSelections very similarly to getBars

I'll roll out the other ones through side branches

PR #1622, opened 2015-02-12
[Plot] - Adding basic implementation of getClosestPlotData

bluong:
getClosestSelection will be a new added API endpoint that, given a point, will find the closest selection within a specified distance to return to the user.


jtlan:
We might want to back up a bit on this for the same reasons as for #1621:
 * Is the `AbstractPlot` implementation appropriate?
 * What about `Drawer`s that don't meet the one-datum-one-element contract? (`Line`, `Area`, etc).

bluong:
My opinion on the one datum one element contract is that since `Line` itself is a complete selection, I want to have the entire `Line`retrieved in this call.  We should have some utility for users to interact with the entire line, and I think this is the best place for it.

However, many users will be interested in the specific points right?  I propose that we should have a Point version of the selection API.  Think of it in this context as getClosestPointSelection.  This should cleanly semantically separate the responsibilities I feel.

src/components/plots/abstractPlot.ts
```diff
@@ -429,6 +435,38 @@ export module Plot {
...
+              closestDatum = drawer._getDatum(selection, pixelPoint);
+              closestPixelPoint = pixelPoint;
+              closestSelection = selection;
+              closestPointDistance = pointDistance;
+            }
+          });
+        });
+      });
+
+      return { data: [closestDatum], pixelPoints: [closestPixelPoint], selection: d3.selectAll(closestSelection) };
```
jtlan: ??? Why is the selection being re-selected? `closestSelection` is already a `D3.Selection`.

PR #1624, opened 2015-02-12
[BarPlot] - Domainer Constructor fixes

bluong:
The barPlot called baseline in its constructor previously which enacted a domainer updating that the constructor does not catch.  Note that one possible issue that is arising is when the scales are attached to the plot, we are instantly updating the scale... even when the plot isn't rendered...

### DIFF
--- a/src/components/plots/abstractXYPlot.ts
+++ b/src/components/plots/abstractXYPlot.ts
@@ -30,9 +30,9 @@ export module Plot {
       this._xScale = xScale;
       this._yScale = yScale;
 
-      AbstractXYPlot.updateScaleDomainer(xScale);
+      AbstractXYPlot._updateScaleDomainer(xScale);
       xScale.broadcaster.registerListener("yDomainAdjustment" + this.getID(), () => this._adjustYDomainOnChangeFromX());
-      AbstractXYPlot.updateScaleDomainer(yScale);
+      AbstractXYPlot._updateScaleDomainer(yScale);
       yScale.broadcaster.registerListener("xDomainAdjustment" + this.getID(), () => this._adjustXDomainOnChangeFromY());
 
       this._autoAdjustXScaleDomain = false;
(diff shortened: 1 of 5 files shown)


src/components/plots/abstractXYPlot.ts
```diff
@@ -139,7 +139,7 @@ export module Plot {
       AbstractXYPlot.updateScaleDomainer(this._yScale);
     }
 
-    private static updateScaleDomainer(scale: Scale.AbstractScale<any, number>) {
+    protected static updateScaleDomainer(scale: Scale.AbstractScale<any, number>) {
```
jtlan: Should this be _underscored? That's our policy for non-public fields right?

src/components/plots/barPlot.ts
```diff
@@ -34,6 +34,12 @@ export module Plot {
       this._animators["baseline"] = new Animator.Null();
       this._isVertical = isVertical;
       this._baselineValue = 0;
+
+      if (isVertical) {
+        Bar._updateDomainerBaseline(yScale, 0, this._plottableID);
+      } else {
+        Bar._updateDomainerBaseline(xScale, 0, this._plottableID);
```
jtlan: Why this change? I understand removing instance calls, but `getID()` should be very safe...

jtlan:
Not sure about removing the `getID()` calls. We don't access `_plottableID` directly because that's not as safe. I understand about removing instance calls in constructors, but removing `getID()` seems like over-applying the principle.

bluong:
There are multiple forces at work I believe that are fighting against getting this to work cleanly...
* The scale's domainer is being affected at the constructor step - I remember reading an `updateScaleExtents` call that is made at the `_anchor` step which hints that we should not be touching the scale until the setup/anchor step... but that is not the case here.
* While getID() currently is really simple (good!), it could randomly change functionality to depend on an instance variable which may not be set yet...  I would prefer not to use the getID() call in the constructor at all, but it is required to get this to work due to...
* The domainer is so black-boxy...  I tried moving these domainer calls to the anchor/setup step and they don't seem to work!  One thing that I've noticed about the domainer is that if you change some settings about it, the scale holding the domainer does not react to it...

PR #1626, opened 2015-02-13
[CategoryAxis] - Correcting textFits parameter for space calculations

### LINKED ISSUE #1511: XAxes overlapping (repro/analysis in progress)
NO repro yet, attempt: http://jsfiddle.net/xL0duLo9/

EDIT: Changing the svg height on the above example also messes the chart up
http://jsfiddle.net/xL0duLo9/6/

Looks like this in Slate: 
screen shot 2015-01-09 at 5 29 20 pm

Only happens for ordinal scales.

EDIT (@bluong) : Repro has been made.  Simplifying the example to ensure it's mainly an axis issue, you see the corresponding image:

screen shot 2015-01-22 at 12 25 28 pm

Corresponding jsfiddle - http://jsfiddle.net/xL0duLo9/4/

NOTE: If you remove any of the Label components, the axes get messed up even more, but if you remove both, it acts fine.

EDIT (@bluong) : Even nicer, simpler repro
http://jsfiddle.net/xL0duLo9/8/

bluong:
The textFits parameter currently determines if the text labels on a categoryAxis can fit there.  However the calculation is... incorrect.

It used to read - If I have 1 wrapped line and I am not not an empty string, then I fit
It should read - If I have 1 wrapped line and I am not an empty string, then I fit

In short, confusing double negative is confusing

Fixes #1511

### DIFF
--- a/test/components/categoryAxisTests.ts
+++ b/test/components/categoryAxisTests.ts
@@ -109,20 +109,14 @@ describe("Category Axes", () => {
     svg.remove();
   });
 
-  it("layout calculated correctly under 0 width/height surrounding component conditions", () => {
+  it("axis should request more space if there's not enough space to fit the text", () => {
     var svg = generateSVG(300, 300);
     var years = ["2000", "2001", "2002", "2003"];
     var scale = new Plottable.Scale.Ordinal().domain(years);
     var axis = new Plottable.Axis.Category(scale, "bottom");
-    var table = new Plottable.Component.Table([
-                                               [null],
-                                               [axis],
-                                               [null],
-                                               [new Plottable.Component.Label("")]
-                                                                                  ]);
-    table.renderTo(svg);
-    assert.closeTo(axis.height(), 44, 2, "height is calculated correctly");
-    assert.strictEqual(axis.width(), 300, "width is calculated correctly");
+    axis.renderTo(svg);
+    var requestedSpace = axis._requestedSpace(300, 10);
+    assert.isTrue(requestedSpace.wantsHeight, "axis should want to ask for more space");
 
     svg.remove();
   });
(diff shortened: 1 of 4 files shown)


jtlan:
Test?

test/components/categoryAxisTests.ts
```diff
@@ -108,4 +108,22 @@ describe("Category Axes", () => {
 
     svg.remove();
   });
+
+  it("layout calculated correctly under 0 width/height surrounding component conditions", () => {
```
jtlan: Is there a test that doesn't involve `Table`? What was the root cause of the bug?

jtlan:
Can we get a more "unit" test?

jtlan:
Do we also want a corresponding test for a vertical `Axis.Category`?

test/components/categoryAxisTests.ts
```diff
@@ -108,4 +108,19 @@ describe("Category Axes", () => {
...
+    var svg = generateSVG(300, 300);
+    var years = ["2000", "2001", "2002", "2003"];
+    var scale = new Plottable.Scale.Ordinal().domain(years);
+    var axis = new Plottable.Axis.Category(scale, "bottom");
+    axis.renderTo(svg);
+    var requestedSpace = axis._requestedSpace(300, 10);
+    assert.isTrue(requestedSpace.wantsHeight, "axis should ask for more space");
+    axis.orient("left");
+    requestedSpace = axis._requestedSpace(10, 300);
+    assert.isTrue(requestedSpace.wantsWidth, "axis should ask for more space");
```
jtlan: Can we distinguish between the two asserts in the message? "axis should ask for more space (vertical/horizontal)"

jtlan:
[minor] Distinguishing `assert`s.

crmorford:
In some cases, the axis seems to be grabbing a lot of extra vertical space. Is that intentional?
The "category_axis_non_string" quicktest is a good example

crmorford:
Verified it fixes slate bug though. Just want to see if you know why it's taking extra space sometimes.

bluong:
So essentially unless I'm mistaken, it's taking up extra space because of the gutter concept, which is defaulted to 15 pixels I believe.

crmorford:
GIL

PR #1628, opened 2015-02-13
Add event listeners on registering callbacks

jtlan:
And remove them when all callbacks are removed.
Made callbacks not trigger if measurement fails.
Made `Dispatcher.Mouse`'s `Broadcaster` private.

### DIFF
--- a/src/dispatchers/mouseDispatcher.ts
+++ b/src/dispatchers/mouseDispatcher.ts
@@ -91,7 +91,7 @@ export module Dispatcher {
 
     private _processMoveEvent(e: MouseEvent) {
       var newMousePosition = this._computeMousePosition(e.clientX, e.clientY);
-      if (newMousePosition.x == null || newMousePosition.y == null) {
+      if (newMousePosition == null) {
         return; // couldn't measure
       }
       this._lastMousePosition = newMousePosition;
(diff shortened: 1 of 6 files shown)


src/dispatchers/mouseDispatcher.ts
```diff
@@ -92,6 +115,11 @@ export module Dispatcher {
       mrBCR = this._measureRect.getBoundingClientRect();
       var testPoint = { x: mrBCR.left, y: mrBCR.top };
 
+      // invalid measurements -- SVG might not be in the DOM
+      if (origin.x === testPoint.x || origin.y === testPoint.y) {
+        return { x: null, y: null };
```
bluong: Would it be cleaner to return `null` instead of a point with null parameters?

bluong:
Just the one item

src/dispatchers/mouseDispatcher.ts
jtlan: Since this is an internal method, it's safe enough to do that.

bluong: Yeah this is definitely safe, but I think it's cleaner to return null, so that the check above is simply
if (newMousePosition == null)

Would there be a situation where x could be null and y isn't and this would be an invalid mousePosition?  We could just simply just report all of this as a null point

PR #1632, opened 2015-02-13
[Table] Fix #1513 with a more equal space request algorithm

### LINKED ISSUE #1513: Table layout dependent on length of labels
screen shot 2015-01-11 at 5 35 38 pm

Repro: http://jsfiddle.net/hsfhs4Lo/

Also happens with horizontal labels: http://jsfiddle.net/hsfhs4Lo/1/

This behavior makes sense if you try to keep some amount of whitespace at the end of each label, but doesn't make that much sense in general. `Label`'s space request algorithm may need to be adjusted.

rcchen:
Fixes #1513


bluong:
We should consider removing / re-establishing the point of the quicktests/html as it doesn't do much for us except for easy access to that direct quicktest, but we don't want to have 2 copies of that quicktest

src/components/table.ts
```diff
@@ -155,23 +155,38 @@ export module Component {
...
           xWeights = _Util.Methods.addArrays(xWeights, colWeights);
-        } else { // Otherwise, divide free space according to the weights
-          xWeights = colWeights;
+        } else {
+          if (d3.sum(colWeights) / colWeights.length !== colWeights[0] ||
+              d3.sum(colWeights) === 0) {
+            // Use colWeights if weights are not equal OR weights are all set to zero (fixed-width)
+            xWeights = colWeights;
+          } else {
+            // If they are all equal, divide free space to make components as equal as possible
```
jtlan: Shouldn't we divide the free space so that the *resulting sizes* are proportional to the weights whenever possible, rather than just if the weights are equal?

jtlan:
Tagging with [-1] so we don't get confused about why this PR is here.

rcchen:
Closing this for now because I want to take a fresh look at this problem.

Issue #1635, opened 2015-02-16
remove ResizeBroadcaster


jtlan:
Goes hand-in-hand with #1634. If there's an API point to force redraws, then should be able to set up the window-resize logic themselves, rather than having the Plottable implementation forced on them.

Be sure to include documentation in the Release Notes on how to set up this behavior again.

PR #1636, opened 2015-02-16
[AbstractComponent] Removing resizeBroadcaster and exposing redraw

### LINKED ISSUE #1634: API point for forcing redraw
When the `<svg>` Plottable is in is resized programmatically, or someone change the CSS stylings in a way that can affect layout (like changing font size), Plottable should redraw, but we are unable to detect these events. Nor would we necessarily want to, since the user has a better sense of when redraws are necessary.

Currently, the user can call `resize()` on the top-level component with no arguments, but that name seems unintuitive. The method name should more clearly reflect what it does.

rcchen:
Fixes #1634 and #1635


rcchen:
Also wrote documentation on how the API changed in https://github.com/palantir/plottable/wiki/Release-Changes-(HEAD)

src/components/abstractComponent.ts
```diff
@@ -226,7 +223,7 @@ export module Component {
      * @param {number} [availableHeight] - the height of the container element
      * @returns {Component} The calling component.
      */
-    public resize(width?: number, height?: number): AbstractComponent {
+    public forceRedraw(width?: number, height?: number): AbstractComponent {
```
jtlan: I was thinking `forceRedraw()` wouldn't take arguments. Plottable infers the size of the `<svg>`, so it shouldn't be necessary to tell it anything. It might make sense to have `resize(width, height)` remain as a separate API point (with the arguments now mandatory).

bluong: We could take advantage of the optional variable signature here.  If any of width or height are set then the rootSVG size has to be set.  Otherwise just redraw.  I dunno if a separate API is necessary unless you think the new API will do something fundamentally different.

src/components/abstractComponent.ts
```diff
@@ -226,7 +223,7 @@ export module Component {
...
       if (!this._isTopLevelComponent) {
         throw new Error("Cannot resize on non top-level component");
       }
```
bluong: Was there a reason for this?

rcchen: Attempting to force a redraw on a nested component might not make sense? That's my best guess for why this was here before.

src/components/abstractComponent.ts
rcchen: Agreed with @bluong here, the only minor difference I have with his proposal is requiring both width and height to be set for the rootSVG size to be set as opposed to one or the other.

bluong:
Please update the release notes.  It looks good to me so far

bluong:
Follow me on the path to be a negative contributor!

crmorford:
GIL

PR #1637, opened 2015-02-16
[NumericAxis] Ticks are now only rendered on a valid domain

### LINKED ISSUE #1578: Ticks outside of the visible domain get rendered
If the user uses a `TickGenerator` that doesn't obey the `Scale`'s domain:
screen shot 2015-02-04 at 3 37 17 pm

http://jsfiddle.net/g2sejn6h/

rcchen:
Fixes #1578


crmorford:
GIL

PR #1638, opened 2015-02-16
Pass arguments to Broadcaster callbacks one at a time

### LINKED ISSUE #1583: Broadcaster.broadcast() doesn't pass arguments correctly
The documentation for the `Broadcaster` callback states:
>```The callback will be called with the attached Listenable as the first object, and optional arguments as the subsequent arguments.```

and the documentation for `broacast` reads:
>`Call all listening callbacks, optionally with arguments passed through.`

But the actual code looks like:
```Typescript
public broadcast(...args: any[]) {
  this._key2callback.values().forEach((callback) => callback(this.listenable, args));
  return this;
}
```
[Source](https://github.com/palantir/plottable/blob/v0.42.0/src/core/broadcaster.ts#L76) (as of this writing).

Note that this passes *an array of all the arguments* as the second argument to the callback, rather than *the arguments separately* as suggested in the documentation. We should actually pass the arguments one at a time, as specified.

jtlan:
... instead of as an array. This change forces us to choose a `this`-context for the callbacks; The `this`-context passed to the callbacks is the `Broadcaster`'s "listenable".

Close #1583.

### DIFF
--- a/src/core/broadcaster.ts
+++ b/src/core/broadcaster.ts
@@ -61,7 +61,7 @@ export module Core {
     public broadcast(...args: any[]) {
       args.unshift(this._listenable);
       this._key2callback.values().forEach((callback) => {
-        callback.apply(this.listenable, args);
+        callback.apply(this._listenable, args);
       });
       return this;
     }
--- a/src/core/broadcaster.ts
+++ b/src/core/broadcaster.ts
@@ -41,6 +41,7 @@ export module Core {
      * Registers a callback to be called when the broadcast method is called. Also takes a key which
      * is used to support deregistering the same callback later, by passing in the same key.
      * If there is already a callback associated with that key, then the callback will be replaced.
+     * The callback will be passed the Broadcaster's "listenable" as the `this` context.
      *
      * @param key The key associated with the callback. Key uniqueness is determined by deep equality.
      * @param {BroadcasterCallback<L>} callback A callback to be called.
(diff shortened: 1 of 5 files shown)


jtlan:
Info about `apply()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

src/core/broadcaster.ts
```diff
@@ -58,7 +59,10 @@ export module Core {
      * @returns {Broadcaster} The calling Broadcaster
      */
     public broadcast(...args: any[]) {
-      this._key2callback.values().forEach((callback) => callback(this._listenable, args));
+      args.unshift(this._listenable);
```
bluong: I'm nervous about this call...  We're directly affecting the array that's being passed in.

jtlan: that's fine. It's generated locally (see [here](https://github.com/palantir/plottable/pull/1638/files#diff-cf6e1d12d6a4046561f8ef977741d1a1R1103)).

bluong: Oh I thought the input was an array being inputted from the user, not a bunch of separate args.  Never mind.

test/core/broadcasterTests.ts
```diff
@@ -53,10 +53,10 @@ describe("Broadcasters", () => {
...
     var g2 = {};
     var g3 = "foo";
-    var cb = (a1: any, rest: any[]) => {
-      assert.equal(listenable, a1, "broadcaster passed through");
-      assert.equal(g2, rest[0], "arg1 passed through");
-      assert.equal(g3, rest[1], "arg2 passed through");
+    var cb = (arg1: any, arg2: any, arg3: any) => {
+      assert.equal(listenable, arg1, "broadcaster passed through");
+      assert.equal(g2, arg2, "g2 passed through");
+      assert.equal(g3, arg3, "g3 passed through");
```
bluong: strictEquals?

bluong:
One minor issue. Outside scope of PR a bit.

crmorford:
regression pass looks fine

PR #1641, opened 2015-02-17
[Plot] - Adding PlotData type

bluong:
I'm mainly having this pull request up so that we can figure out the name of the type as well as the parameters.  Hopefully the parameters are appropriate as discussed.

For now I will base the rest of the needed functions off of this branch for convenience.

### DIFF
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -16,6 +16,12 @@ export module Plot {
     datasetKey: string
   }
 
+  export type PlotData = {
+    data: any[];
+    pixelPoints: Point[];
+    selection: D3.Selection;
+  }
+
   export class AbstractPlot extends Component.AbstractComponent {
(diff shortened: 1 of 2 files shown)


src/components/plots/abstractPlot.ts
```diff
@@ -16,6 +16,12 @@ export module Plot {
     datasetKey: string
   }
 
+  export type PlotData = {
+    data: any[];
+    pixelPoints: Point[];
```
jtlan: `pixelPositions`? That's what it's named in `HoverData`. "`pixels`" would also work.

bluong: We should query QE on this one.  I chose `pixelPoints` over `pixelPositions` since it highlights the fact that points are returned.

crmorford: pixelPoints makes sense!

PR #1642, opened 2015-02-17
[Drawer] - Adding _getDatum to drawers

bluong:
_getDatum is a function that given a `D3.Selection` and a `Point`, it will return the singular datum associated with the point.


jtlan:
I'm not sure I see the point of this method ... what would this be used for?

bluong:
@jtlan As we discussed yesterday, we determined that we wanted to return the datum around a pixel point.  Thus, while this works for rectangle selections very cleanly, the manner in how `_Drawer.Line` retrieves data is very different.

bluong:
The function creation here seems to be a bit misguided.  Closing PR

PR #1643, opened 2015-02-17
[Drawer] - Adding _getPixelPoints to drawers

bluong:
Given a singular selection, returns all the pixel points for the selection.  This one might be the most tricky codewise and controversial


jtlan:
As with the other PR (#1642), can you explain what this would be used for? For example, this method won't work with `Plot.Line`.

bluong:
@jtlan The point of this function is to retrieve the pixel points for the selection.  Under the assumption that a selection has been chosen, we want to retrieve the relevant pixel point as we discussed yesterday correct?  Also, I am unsure how this does not work for `Plot.Line` since the point of making this into an Array was to account for it.

bluong:
The function creation here seems to be a bit misguided.  Closing PR

PR #1647, opened 2015-02-17
Convert Dispatcher.Key to be global

jtlan:
Replaced `Dispatcher.Keypress` with `Dispatcher.KeyEvent`, which registers globally and triggers if a key is pressed anywhere.

`Interaciton.Key` uses this new `Dispatcher.KeyEvent` in conjunction with `Dispatcher.Mouse` to determine where a key event occurs and trigger callbacks only if the key is pressed while the user is moused over the `Component`.


src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Dispatcher {
+  export class KeyEvent {
```
bluong: Not sure about KeyEvent (since Mouse then should be called MouseEvent.... which also leads me to believe that the Event part is redundant)

src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
+    private _connected = false;
```
bluong: Why does this need to be here?

src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    private _keydownBroadcaster: Core.Broadcaster<Dispatcher.KeyEvent>;
+    // private _keyupBroadcaster: Core.Broadcaster<Dispatcher.KeyEvent>;
```
bluong: Remove comment (Also do we need this?)

src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    private _processDownCallback = (e: KeyboardEvent) => this._processKeydown(e);
```
bluong: Do we want a processUpCallback?

bluong: we can try to be a bit consistent with our instance var names.

Mouse has _mouseover _mousemove _mouseout

Maybe _keyDown ?

bluong: Not like it matters too much since it is a private var, but do we want the name to be _processDownCallback rather than like _downCallback ?

src/dispatchers/keyEventDispatcher.ts
bluong: Augh I should have commented about this on mouse...

src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    public static getDispatcher(): Dispatcher.KeyEvent {
+      var dispatcher: KeyEvent = (<any> document)[KeyEvent._DISPATCHER_KEY];
+      if (dispatcher == null) {
+        dispatcher = new KeyEvent();
+        (<any> document)[KeyEvent._DISPATCHER_KEY] = dispatcher;
+      }
+      return dispatcher;
+    }
+
+    constructor() {
```
bluong: Let's put some doc here, or move documentation to the top of the class

src/dispatchers/keyEventDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+      } else {
+        this._connect();
+        this._keydownBroadcaster.registerListener(key,
+          (d: Dispatcher.KeyEvent, e: KeyboardEvent) => callback(e));
+      }
+      return this;
+    }
+
+    private _processKeydown(e: KeyboardEvent) {
+      this._keydownBroadcaster.broadcast(e);
```
bluong: Why is this broadcasting a KeyboardEvent when Mouse doesn't

src/interactions/keyInteraction.ts
```diff
@@ -3,34 +3,31 @@
...
       super._anchor(component, hitBox);
 
-      this._dispatcher.target(this._hitBox);
+      this._positionDispatcher = Dispatcher.Mouse.getDispatcher(
+                                   <SVGElement> (<any> this._componentToListenTo)._element.node()
+                                 );
+      this._positionDispatcher.onMouseMove("Interaction.Key" + this.getID(), (p: Point) => 0); // need to register
+
+      this._keyDispatcher = Dispatcher.KeyEvent.getDispatcher();
+      this._keyDispatcher.onKeydown("Interaction.Key" + this.getID(), (e: KeyboardEvent) => this._handleKeyEvent(e));
```
bluong: Weird that github highlights onKeyDown...?

src/interactions/keyInteraction.ts
```diff
@@ -3,34 +3,31 @@
...
+    private _keyCode2Callback: { [keyCode: string]: () => void; } = {};
 
     public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
       super._anchor(component, hitBox);
 
-      this._dispatcher.target(this._hitBox);
+      this._positionDispatcher = Dispatcher.Mouse.getDispatcher(
+                                   <SVGElement> (<any> this._componentToListenTo)._element.node()
+                                 );
+      this._positionDispatcher.onMouseMove("Interaction.Key" + this.getID(), (p: Point) => 0); // need to register
```
bluong: ???  This seems hacky

test/dispatchers/dispatcherTests.ts
```diff
@@ -212,35 +212,21 @@ describe("Dispatchers", () => {
...
-      $target.simulate("mouseout");
-      $target.simulate("keydown", { keyCode: 80 });
-      assert.isFalse(keyDownCalled, "didn't trigger callback after mousing out of the target");
+  describe("KeyEvent Dispatcher", () => {
+    it("triggers callback on mousedown", () => {
+      var ked = Plottable.Dispatcher.KeyEvent.getDispatcher();
 
-      target.remove();
+      var keyDowned = false;
+      var callback = function() {
```
bluong: () => notation?  Could also just 1line this

src/interactions/keyInteraction.ts
```diff
@@ -3,34 +3,31 @@
...
 
-      this._dispatcher.onKeyDown((e: D3.D3Event) => {
-        if (this._keyCode2Callback[e.keyCode]) {
-          this._keyCode2Callback[e.keyCode]();
-        }
-      });
-      this._dispatcher.connect();
+    private _handleKeyEvent(e: KeyboardEvent) {
+      var p = this._translateToComponentSpace(this._positionDispatcher.getLastMousePosition());
+      if (this._isInsideComponent(p) && this._keyCode2Callback[e.keyCode]) {
```
bluong: Does this make sense?  What happens if a consumer just happens to move their mouse off the component space and then presses the key?  I'm ok with this or not having this enforcement, but we should ask QE for what should happen

bluong:
Mostly looks fine, just a few questions.

src/dispatchers/keyEventDispatcher.ts
jtlan: But there are multiple types of Key events (`keydown`, `keyup`, etc), and I'm rolling up the different types of mouse events. Also serves to prevent confusion between "key" as in keyboard and a dictionary "key".

src/dispatchers/keyEventDispatcher.ts
jtlan: Might be important later, if we want to trigger events on `keyup` events.

src/dispatchers/keyEventDispatcher.ts
jtlan: Events have additional information, such as if the CTRL key was pressed, that might be useful. Should `Dispatcher.Mouse` also send the events?

src/interactions/keyInteraction.ts
jtlan: If you don't, the `Dispatcher.Mouse` never connects. An alternative implementation is to have the callback here cache the last known position, and the callback on `_keyDispatcher` just references the cached value, but then we're duplicating information.

src/dispatchers/keyEventDispatcher.ts
bluong: We should determine if we want this currently...  We shouldn't have lingering comments like this enter the codebase

src/dispatchers/keyEventDispatcher.ts
bluong: One can argue the same for most of Mouse...  I don't want to run into a situation where we have some `Dispatcher`s have `Event` appended to the end and some don't

src/dispatchers/keyEventDispatcher.ts
bluong: I think I'm liking the last piece, not 100% certain

src/interactions/keyInteraction.ts
bluong: This is hinting at the fact where I'm not 100% sure now if we want to remove the connection if 0 callbacks are registered.  I'm heavily not a fan of having "junk" calls that exist to switch on this dispatcher...

src/dispatchers/keyEventDispatcher.ts
jtlan: I'll remove it for now. Minimal touch, just trying to replace the old dispatchers.

src/dispatchers/keyEventDispatcher.ts
jtlan: After I convert `Dispatcher.Abstract` this won't be here.

src/dispatchers/keyEventDispatcher.ts
jtlan: Noted. I'll rename to Key.

src/dispatchers/keyEventDispatcher.ts
jtlan: Maybe `_downCallback` would make more sense.

src/dispatchers/keyEventDispatcher.ts
jtlan: Probably the more general way.

src/interactions/keyInteraction.ts
jtlan: How do you know when it's safe to throw away the `Dispatcher` then?

src/dispatchers/keyDispatcher.ts
```diff
@@ -0,0 +1,79 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Dispatcher {
+  export type KeyCallback = (keyCode: number) => any;
+
+  export class Key {
+    private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
+    private _connected = false;
```
bluong: Can we attach the comment here that the parent class var name is going to be changed or similar?

bluong: a TODO or similar

src/dispatchers/keyDispatcher.ts
```diff
@@ -0,0 +1,79 @@
...
+     * @return {Dispatcher.Key} The calling Dispatcher.Key.
+     */
+    public onKeydown(key: any, callback: KeyCallback): Key {
+      if (callback === null) { // remove listener if callback is null
+        this._keydownBroadcaster.deregisterListener(key);
+        this._disconnect();
+      } else {
+        this._connect();
+        this._keydownBroadcaster.registerListener(key,
+          (d: Dispatcher.Key, e: KeyboardEvent) => callback(e.keyCode));
```
bluong: Did we want to hand the event in too as the second arg?

bluong:
Just noting that the current KeyInteraction now is quite a bit useful, but it has the potential to be more useful.  With the idea of mouseDispatchers being on the KeyInteraction, the callbacks could now take mouse positions and the like.  Way beyond the scope of this pr, but food for thought.

bluong:
Minor at this point.  Since it seems that there are still a few changes that need to be made in keyDispatcher before it makes it into the codebase `this._connected` for one, can you please add TODO comments or so to highlight this, so that we can easily look to see if all these comments are completed before the final pr.

src/dispatchers/keyDispatcher.ts
jtlan: We'll add that feature later.

PR #1650, opened 2015-02-18
[Drawer] - getPixelPoint

bluong:
Adding a function that given a selection / datum / index will return the pixel point associated with the 3.


src/drawers/elementDrawer.ts
```diff
@@ -60,6 +60,15 @@ export module _Drawer {
     public _getSelector(): string {
       return this._svgElement;
     }
+
+    public _getPixelPoint(selection: D3.Selection, datum: any, index: number): Point {
+      switch(this._svgElement) {
+        case "circle":
+          return { x: parseFloat(selection.attr("cx")), y: parseFloat(selection.attr("cy")) };
```
jtlan: It seems like the better way to do this is to subclass `Drawer.Element` to create `Drawer.Circle`, which overrides this.

The `Drawer.Element` default implementation should ... hmm ... return some sort of `getBBox()` or `getBoundingClientRect` based implementation?

jtlan:
Some of these implementations work by parsing attributes, while others work by caching the projectors and applying those. Wouldn't it make more sense if all of the implementations just used cached projectors?

src/drawers/elementDrawer.ts
bluong: The former would require some bit of architectural work beyond the scope of this PR.  I'll get to it, but beyond the scope of this pr.

rcchen:
Maybe add tests if you feel it would be useful in this situation?

PR #1653, opened 2015-02-18
Move AbstractDispatcher to new signature

jtlan:
`Dispatcher.Mouse` and `Dispatcher.Key` now extend from it.

Changed tests for `AbstractDispatcher`.

### DIFF
--- a/src/dispatchers/abstractDispatcher.ts
+++ b/src/dispatchers/abstractDispatcher.ts
@@ -3,88 +3,49 @@
 module Plottable {
 export module Dispatcher {
   export class AbstractDispatcher extends Core.PlottableObject {
-    protected _target: D3.Selection;
-    protected _event2Callback: { [eventName: string]: () => any; } = {};
+    protected _event2Callback: { [eventName: string]: (e: Event) => any; } = {};
+    protected _broadcasters: Core.Broadcaster<AbstractDispatcher>[] = [];
     private _connected = false;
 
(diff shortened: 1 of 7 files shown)


src/dispatchers/abstractDispatcher.ts
```diff
@@ -3,88 +3,49 @@
...
+      return () => callback();
+    }
+
+    protected _setCallback(b: Core.Broadcaster<AbstractDispatcher>, key: any, callback: Function) {
+      if (callback === null) { // remove listener if callback is null
+        b.deregisterListener(key);
+        this._disconnect();
+      } else {
+        this._connect();
+        b.registerListener(key, this._getWrappedCallback(callback));
```
bluong: Why do you have to wrap the callback instead of using
b.registerListener(key, callback);

src/dispatchers/abstractDispatcher.ts
```diff
@@ -3,88 +3,49 @@
...
      */
-    public disconnect() {
-      this._connected = false;
-      if (this._target) {
-        Object.keys(this._event2Callback).forEach((event: string) => {
-          this._target.on(this._getEventString(event), null);
-        });
+    protected _getWrappedCallback(callback: Function): Core.BroadcasterCallback<AbstractDispatcher> {
+      return () => callback();
+    }
```
bluong: staticize?

Issue #1644, opened 2015-02-17
CSS Scaling breaks nearly every interaction


crmorford:
The mouse position isn't reported correctly, which causes everything to break.

jtlan:
Subissues:
* #1497.

crmorford:
Hover fixed, Key fixed

jtlan:
Drag fixed in #1816.

jtlan:
Panzoom fixed in #1896. All done!

Issue #1427, opened 2014-12-08
Interactions can block each other


jtlan:
Repro: http://jsfiddle.net/8oz2nz6p/1/

Expected: User is able to hover over the `Plot.Line`, and select bars on the `Plot.VerticalBar`
Actual: User is unable to select bars.

The issue here is that both `Plot`s generate a hitbox to listen for events; the `Plot.Line`'s hitbox is above the `Plot.Vertical`'s hitbox, and blocks its events.

drglitch:
:+1: for this and #1209 - it also affects hovers.

A workaround is to attach to SVG elements (jQuery or otherwise) right after plot has been added and flushed, instead of using a Hover interaction - this way the hitbox never gets shown and therefore does not block all but topmost item in the Group.

A potential solution (which you seem to be well on way to developing in #1209) is to use similar idea to "line" hover mode in Plot.Bar to figure out what's underneath the x/y of the mouse pointer and dispatch HoverData with appropriate Plot + Dataset ID. This would have the added benefit of being able to make crosshairs for axis tracking much easier.

Another, hackier, solution is to implement above workaround + draw transparent circles over each knot of the Line elem, but this will explode count of svg elems in timeseries-style linecharts.

Looking forward to a fix! :)

jtlan:
All `Interaction`s are powered by `Dispatcher`s now, and hitboxes were removed in #1945, so this should be fixed now.

PR #1654, opened 2015-02-18
Dispatcher revamp

### LINKED ISSUE #837: Plottable.Core.KeyEventListener sets an un-namespaced listener on entire document
this causes a quicktest failure (key_interaction.js, pressing 'a' only resets the domain for the right test, and has no effect on the left, #782) and is a bad idea generally.

steps:
 - [ ] - Namespace it to Plottable so that it doesnt conflict w/ other libs
 - [ ] - namespace it on a per-branch basis to solve the quicktest problem.

### LINKED ISSUE #1490: Zooming on IE9 causes hover to work incorrectly

### LINKED ISSUE #1560: <svg>-wide Dispatcher.Mouse
Convert `Dispatcher.Mouse` to use a single listener attached to the root `<svg>`, which then proxies the events to any `Dispatcher`s/`Interaction`s that are want the information.

Dependent on #1559, so that event positions relative to a `Component` can be inferred.

jtlan:
Moved dispatchers to the new architecture.

This should maintain the existing functionality, but fix the CSS zoom (#1644) and IE9 zoom bugs (#1490) for `Hover` and `Key` interactions. Should also solve #837.

Also puts us on the path to resolving #1427, by removing the need for hitboxes.

Close #837.
Close #1490.
Close #1560.


jtlan:
@bluong @rcchen You can re-review, but the pieces of this have passed Dev CR before.

bluong:
```
This should maintain the existing functionality, but fix the CSS zoom (#1644) and IE9 zoom bugs (#1490) for Hover and Key interactions. Should also solve #837.

Also puts us on the path to resolving #1427, by removing the need for hitboxes.
```

I don't think the issues would get parsed correctly to close.  Maybe do a fixes ___ for each just in case?

bluong:
@jtlan We might want to change the keyDispatcher API to keyDown instead of keydown

jtlan:
@bluong wrote:
> I don't think the issues would get parsed correctly to close.  Maybe do a fixes ___ for each just in case?

Not all of them should be closed, so that's probably fine...?

crmorford:
Haven't played with IE yet, but CSS zooming + hover is working!

rcchen:
@crmorford wrote:
> Haven't played with IE yet

famous last words

bluong:
@jtlan wrote:
> Not all of them should be closed, so that's probably fine...?

Are you manually going to close the issues that should be closed after this goes in?

crmorford:
DRUMROLL PLEASE
http://jsfiddle.net/8brgqrnr/1/
This still works....
In IE...
With CSS scaling (bigger OR smaller than 1)...
And ctrl ++ (or ctrl --)
On every single type of plot

Good Job Justin #1 Gold star

rcchen:
:star: :star: :star: :star: :star:

PR #1656, opened 2015-02-18
Plot data

bluong:
QE: Since Drawers are not exposed to the user, a regression pass should be enough.


src/drawers/elementDrawer.ts
```diff
@@ -60,6 +60,15 @@ export module _Drawer {
     public _getSelector(): string {
       return this._svgElement;
     }
+
+    public _getPixelPoint(datum: any, index: number): Point {
+      switch(this._svgElement) {
+        case "circle":
```
jtlan: This is the wrong way to do this again -- it makes more sense to create `Drawer.Circle` as a subclass of `Drawer.Element`, and have it re-implement `_getPixelPoint()` with the circle-specific logic.

PR #1657, opened 2015-02-18
only use attrToProjector as a cache for finding info for pixel points

bluong:


jtlan:
as in "don't use `this._attrToProjector()` for anything else?

bluong:
Yeah, the this._attrToProjector should only be used to hold the data projectors (since the non instance var attrToProjector undergoes some meddling with the attrToProjector (replacing x/y's with d's and the like)

PR #1658, opened 2015-02-19
[Plot] - _getPlotDatumArray

bluong:
Unsure if this function needs to exist, but it's a private utility function made to grab each PlotDatum on the plot (a PlotDatum is like a singular item from PlotData with a singular point, datum, and a 1-item selection)


src/components/plots/abstractPlot.ts
```diff
@@ -421,16 +427,59 @@ export module Plot {
...
+    public getAllSelections(datasetKeys?: string | string[]): D3.Selection {
+      var datasetKeyArray: string[] = [];
+      if (datasetKeys == null) {
+        datasetKeyArray = this._datasetKeysInOrder;
+      } else if (typeof(datasetKeys) === "string") {
+        datasetKeyArray = [<string> datasetKeys];
+      } else {
+        datasetKeyArray = <string[]> datasetKeys;
+      }
+
```
jtlan: As per my comment on #1659, we shouldn't be special-casing out the `Line` drawer here -- instead we should have `Plot.Line` override this method with a `Line`-specific implementation.

bluong: Err it seems the comment was made several lines below rather than here.  Depends if we want this behavior to happen on specific plots or specific drawers

jtlan:
It might actually be good for this to be public, if people want to implement their own closest-point algorithms.

jtlan:
Would it make more sense to return a `PlotData` with all the data + points in it?

bluong:
This doesn't seem useful as `getAllPlotData` seems to hold this info, closing

PR #1659, opened 2015-02-19
[Plot] - getClosestPlotData()

### LINKED ISSUE #1579: getClosestPoint() for XYPlots
For user-customized interactions, it would help to expose these sorts of methods:

```Typescript
public getClosestPoint(p: Point): ???;
```

So the user can pass mouse movements, clicks, etc in to them. See `getBars()` on `Bar` plots.

It should be possible to implement this on `AbstractXYPlot`.

Questions:
 - What gets returned? Data? Pixel values? `D3.Selection`s?

bluong:
Adding the new API getClosestPlotData that is used to retrieve the PlotData object that is closest to the specified x/y arguments.

getClosestPlotData has the following signature
```typescript
getClosestPlotData(xValue: number, yValue: number, withinValue: number = Infinity): PlotData
```

x/yValue are the specified coordinates to query.  withinValue is an optional argument (defaulting to Infinity) to specify the limit of the closest distance from the point.  If there is no PlotData closer than this value, then an empty PlotData is returned.

getClosestPlotData performs by querying all of the pixel points correlated with each datum value and does a minimum distance check across each pixel point until the minimum distance one is found.  Then the selection + datum is stored alongside with this pixel point information.

Fixes #1579


src/components/plots/abstractPlot.ts
```diff
@@ -445,6 +451,43 @@ export module Plot {
...
+
+      var closestDatum: any = null;
+      var closestSelection = d3.select();
+      var closestPixelPoint: Point = null;
+      var closestPointDistance = withinValue;
+
+      this._getDrawersInOrder().forEach((drawer) => {
+        drawer._getRenderArea().selectAll(drawer._getSelector()).each(function(datum: any, index: number) {
+          var selection = d3.select(this);
+          var pixelPoints = (drawer instanceof _Drawer.Line) ?
```
jtlan: We really shouldn't be doing specialized forking logic like this inside abstract classes. I would prefer that `Plot.Line` override `getClosestData()` -- that seems a more correct use of the class-based inheritance.

src/components/plots/abstractPlot.ts
```diff
@@ -445,6 +451,43 @@ export module Plot {
...
+                            (<any[]> datum).map((lineDatum, lineIndex) => drawer._getPixelPoint(lineDatum, lineIndex)) :
+                            [drawer._getPixelPoint(datum, index)];
```
jtlan: `pixelPoints` is set to some invocation of `_getPixelPoint()`, yet there's more than one element in it? I would expect `_getPixelPoint()` to return only one element.

src/drawers/elementDrawer.ts
```diff
@@ -60,6 +60,15 @@ export module _Drawer {
     public _getSelector(): string {
       return this._svgElement;
     }
+
+    public _getPixelPoint(datum: any, index: number): Point {
+      switch(this._svgElement) {
+        case "circle":
```
jtlan: This is the wrong way to do this again -- it makes more sense to create `Drawer.Circle` as a subclass of `Drawer.Element`, and have it re-implement `_getPixelPoint()` with the circle-specific logic.

jtlan:
What is this blocked on? #1656?

src/components/plots/abstractPlot.ts
bluong: What happens in a situation where an arbitrary plot uses a Line Drawer?  This case would only happen on Drawers that use an array as a datum.  Should we special case line plots with the possibility that other plots that use Line Drawers will be handing in an array as a datum will need to override again?

src/components/plots/abstractPlot.ts
bluong: Notice the array format and thus _getPixelPoints only returns a single point but then gets wrapped into an array

src/drawers/elementDrawer.ts
bluong: ??? This was just cherry-picked logic from the blocked pull request

src/components/plots/abstractPlot.ts
jtlan: If a developer creates a new `Plot` type, they are responsible for determining if they need to reimplement `getClosestData()`.

src/components/plots/abstractPlot.ts
jtlan: Why do that at all then?

src/drawers/elementDrawer.ts
jtlan: I realize that, but it's still wrong. Make another PR to address this specific issue?

src/components/plots/abstractPlot.ts
bluong: Yes, but do we want this data vs datum idea to be at the plot level or on the drawer level?

crmorford:
I'm starting to question if doing this based on just one point for a bar plot makes sense.
in this example
screen shot 2015-02-20 at 11 54 43 am
I was hovering over the bottom half of the far right bar (screenshot makes the mouse invisible...inconvenient)
So even though I'm touching one bar, another one is "closest"

crmorford:
And Line, for example.
screen shot 2015-02-20 at 12 05 25 pm
If you have a small withinValue, you can't get a PlotData returned unless you're close to one of the points (not the straight line inbetween data)

http://jsfiddle.net/6hfn2mcu/4/

crmorford:
Talked to Sonny, this heuristic is too painful. We have to fix for Bar/Line/Area

jtlan:
Would it make more sense to have a different variation of this method for different types of `Plot`s? For example, for bars, "closest" is defined differently.

bluong:
We found this might be the wrong appraoch in dealing with the issue.  We should instead have the plot provide all of the plot data and then have the user determine via an object that can do this calculation.  We might build such an object as a default.  Closing PR

PR #1664, opened 2015-02-20
[GridPlot] - getDrawer now returns a RectDrawer

bluong:
Used to create an ElementDrawer that sets its own svgElement to "rect"


bluong:
Not sure if this should be NOQE, as a brief overlaying glance seems to show no changes

jtlan:
Doesn't rectDrawer have some like label-drawing or bar-specific logic?

bluong:
Only if the plot tells the drawer to draw labels (Also, we were considering moving that label logic out)

crmorford:
Good it look

Issue #1649, opened 2015-02-17
CSS classes with a color name get applied incorrectly on Line/Area if animation is on


jtlan:
![screen shot 2015-02-17 at 3 48 03 pm](https://cloud.githubusercontent.com/assets/1440449/6239810/764b08a0-b6bc-11e4-96dd-f926323f9995.png)
Repro: http://jsfiddle.net/68u6qwnn/1/

It looks like the class is applied as the hex value of the color...???

rcchen:
Fixed by #1665

PR #1665, opened 2015-02-20
[AreaPlot/LinePlot] Properly set class during projections

### LINKED ISSUE #1286: project()-ing class erases existing classes on Plot.Line and Plot.Area
`project()`ing the `"class"` attribute removes the existing `line` and `area` classes on `<path>`s in `Plot.Line` and `Plot.Area`.

Repro:
http://jsfiddle.net/nbaxe7ys/1/
Note the rules in the CSS pane

**Expected**: Line is both pink and bolded
**Actual**: Line is only pink

rcchen:
Fix #1286  and #1649 

Demo fiddle: http://jsfiddle.net/68u6qwnn/2/

### DIFF
--- a/src/drawers/areaDrawer.ts
+++ b/src/drawers/areaDrawer.ts
@@ -78,10 +78,13 @@ export module _Drawer {
       if (attrToProjector["fill"]) {
         this._areaSelection.attr("fill", attrToProjector["fill"]); // so colors don't animate
       }
-      step.animator.animate(this._areaSelection, attrToProjector);
+      if (attrToProjector["class"]) {
+        this._areaSelection.attr("class", attrToProjector["class"]);
+        this._areaSelection.classed(Area.AREA_CLASS, true);
+        delete attrToProjector["class"];
+      }
 
-      // Restore default classes that may have been wiped out by class projectors
-      this._areaSelection.classed(Area.AREA_CLASS, true);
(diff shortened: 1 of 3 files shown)


src/drawers/areaDrawer.ts
```diff
@@ -78,10 +78,13 @@ export module _Drawer {
       if (attrToProjector["fill"]) {
         this._areaSelection.attr("fill", attrToProjector["fill"]); // so colors don't animate
       }
-      step.animator.animate(this._areaSelection, attrToProjector);
+      if (attrToProjector["class"]) {
+        this._areaSelection.attr("class", attrToProjector["class"]);
+        this._areaSelection.classed(Area.AREA_CLASS, true);
+        delete attrToProjector["class"];
```
bluong: Why did the bug happen before?

rcchen: My understanding of this (correct me if I'm wrong) is that the class attribute was only applied to the initial step, and was wiped out by the next step when an animation happens.

bluong: My understanding is that I think a lot of the confusion seems to arise from a sort of grungy architecture that is in the Drawers.  I may or may not look into it to see if it can be cleaned up.

crmorford:
Looks good for both bugs

PR #1667, opened 2015-02-20
[NumericAxis] Address floating point issues with formatters

### LINKED ISSUE #1510: Scientific formatter is using very small numbers instead of 0 (178a)
screen shot 2015-01-09 at 3 26 08 pm
Ask me for the link to this, it's internal. I'll work on a repro.

rcchen:
Fixes #1510

### DIFF
--- a/src/utils/formatters.ts
+++ b/src/utils/formatters.ts
@@ -118,7 +118,12 @@ module Plottable {
      */
     export function siSuffix(precision = 3) {
       verifyPrecision(precision);
-      return (d: any) => d3.format("." + precision + "s")(d);
+      return (d: any) => {
+        var multiplier = Math.pow(10, precision);
+        var val = Math.round(d * multiplier) / multiplier;
+        return d3.format("." + precision + "s")(val === -0 ? -val : val);
+      }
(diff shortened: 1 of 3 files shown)


src/utils/formatters.ts
```diff
@@ -118,7 +118,11 @@ module Plottable {
      */
     export function siSuffix(precision = 3) {
       verifyPrecision(precision);
-      return (d: any) => d3.format("." + precision + "s")(d);
+      return (d: any) => {
+        var multiplier = Math.pow(10, precision);
+        var val = Math.round(d * multiplier) / multiplier;
+        return d3.format("." + precision + "s")(val === -0 ? -val : val); // fix for -0.00
```
bluong: Why do we need to do this rounding?  How does this help the situation marked in the issue?

jtlan: I think it's because the d3 scientific formatter is capable of showing small numbers.

jtlan:
Now that I think about it, we should probably **[wontfix]** this one. The scientific formatter is working as intended, which is showing `precision` digits of a number, regardless of how small it is. In this case, I think the tick generator is the one causing problems.

PR #1686, opened 2015-02-26
[Plot] - Adding exclude argument to getAllSelections

bluong:
The new signature for getAllSelections after this pull request should be
```typescript
public getAllSelections(datasetKeys?: string | string[], exclude = false): D3.Selection
```

An exclude argument is added as a parameter that defaults to false that can be set to true if the user wants to retrieve all selections excluding the ones specified in datasetKeys.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2679,7 +2679,7 @@ declare module Plottable {
              * If not provided, all selections will be retrieved.
              * @returns {D3.Selection} The retrieved selections.
              */
-            getAllSelections(datasetKeys?: string | string[]): D3.Selection;
+            getAllSelections(datasetKeys?: string | string[], exclude?: boolean): D3.Selection;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


jtlan:
Is this useful? Couldn't the user just use [`Array.Prototype.Filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)?

bluong:
getAllSelections returns a D3.Selection which has no correlation with the datasets that it is representing.

Alternatively, the user could do this.datasetOrder().filter... but that seems a lot less convenient instead of querying the exclusion directly.  This seems like a very trivial convenient enhancement.  That said, I labelled it p3 I think appropriately

jtlan:
@crmorford, let us know if this seems reasonable.

crmorford:
Cool GIL http://jsfiddle.net/xL0duLo9/11/
wait to merge til hover is in

crmorford:
Accidentally clicked close & comment...my bad

PR #1687, opened 2015-02-26
[Plot] - Adding generateProjectors() API

bluong:
With the idea of the `project` call, users of Plottable are able to set the attributes of the plot through the use of accessors and scales.  However, the exact way that Plottable computes these attributes is hidden from the user.

For example, consider the call
`plot.project("x", "x", xScale)`
It is hidden from the user that in order to compute the x attribute of the rendered plot the fact that the "x" attribute is queried per datum and scaled with the xScale to find the pixel point that the datum will render to.

This is incredibly important for plots like `Plot.Scatter` to function correctly as there must be a way to map from a datum to a pixel position.

This pull request is an attempt at higher transparency between the user and the plot, letting the user query the following API endpoint to find out how attributes computed on a plot.

This is incredibly important for plots like `Plot.Line` where it is incredibly difficult to see how a datum maps to a single point on a line plot.

This can also reveal the fills on the `Plot.Bar` rectangles without rendering them.

The signature is the following:
```typescript
public generateProjectors(datasetKey: string): AttributeToAppliedProjector
```

Example:  On a `Plot.Scatter` it would be nice to know the pixel positions of each point just from the data / without needing to look through the visuals

http://jsfiddle.net/bluong63/ka5oag2j/2/

Example: `Plot.Bar` example

http://jsfiddle.net/bluong63/pb9rorzv/1/

### DIFF
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -209,15 +209,15 @@ export module Plot {
     }
 
     /**
-     * Generates a map from a projected attribute to the function to calculate the value of that attribute
+     * Generates a dictionary mapping an attribute to a function that calculate that attribute's value
      * in accordance with the given datasetKey.
      *
      * Note that this will return all of the data attributes, which may not perfectly align to svg attributes
      *
-     * @param {datasetKey} the key of the dataset to generate the map for
-     * @returns {AttributeToAppliedProjector} A map from attributes to functions to calculate that attribute
+     * @param {datasetKey} the key of the dataset to generate the dictionary for
+     * @returns {AttributeToAppliedProjector} A dictionary mapping attributes to functions
      */
-    public generateAppliedProjectors(datasetKey: string): AttributeToAppliedProjector {
+    public generateProjectors(datasetKey: string): AttributeToAppliedProjector {
       var attrToProjector = this._generateAttrToProjector();
       var plotDatasetKey = this._key2PlotDatasetKey.get(datasetKey);
       var plotMetadata = plotDatasetKey.plotMetadata;
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -218,7 +218,15 @@ export module Plot {
      * @returns {AttributeToAppliedProjector} A map from attributes to functions to calculate that attribute
      */
     public generateAppliedProjectors(datasetKey: string): AttributeToAppliedProjector {
-      return null;
+      var attrToProjector = this._generateAttrToProjector();
+      var plotDatasetKey = this._key2PlotDatasetKey.get(datasetKey);
+      var plotMetadata = plotDatasetKey.plotMetadata;
+      var userMetadata = plotDatasetKey.dataset.metadata();
+      var attrToAppliedProjector: AttributeToAppliedProjector = {};
+      d3.entries(attrToProjector).forEach((keyValue: any) => {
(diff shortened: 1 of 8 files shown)


jtlan:
Shouldn't be returning an underscored type. Can we name the type to something more friendly, if we want to expose it?

src/components/plots/abstractPlot.ts
```diff
@@ -208,6 +208,27 @@ export module Plot {
       return h;
     }
 
+    /**
+     * Generates a map from a projected attribute to the function to calculate the value of that attribute
+     * in accordance with the given datasetKey.
+     *
+     * Note that this will return all of the data attributes, which may not perfectly align to svg attributes
```
jtlan: Can we work with QE / Docs to come up with a clearer explanation of what this does?

bluong:
Moving up in priority since this is pretty necessary to have closestPoint idea working

src/components/plots/abstractPlot.ts
```diff
@@ -208,6 +208,27 @@ export module Plot {
...
+     * @param {datasetKey} the key of the dataset to generate the dictionary for
+     * @returns {AttributeToAppliedProjector} A dictionary mapping attributes to functions
+     */
+    public generateProjectors(datasetKey: string): AttributeToAppliedProjector {
+      var attrToProjector = this._generateAttrToProjector();
+      var plotDatasetKey = this._key2PlotDatasetKey.get(datasetKey);
+      var plotMetadata = plotDatasetKey.plotMetadata;
+      var userMetadata = plotDatasetKey.dataset.metadata();
+      var attrToAppliedProjector: AttributeToAppliedProjector = {};
+      d3.entries(attrToProjector).forEach((keyValue: any) => {
```
jtlan: Is there a reason we do this rather than using `Object.Keys`?

src/components/plots/abstractPlot.ts
```diff
@@ -208,6 +208,27 @@ export module Plot {
...
+    /**
+     * Generates a dictionary mapping an attribute to a function that calculate that attribute's value
+     * in accordance with the given datasetKey.
+     *
+     * Note that this will return all of the data attributes, which may not perfectly align to svg attributes
+     *
+     * @param {datasetKey} the key of the dataset to generate the dictionary for
+     * @returns {AttributeToAppliedProjector} A dictionary mapping attributes to functions
+     */
+    public generateProjectors(datasetKey: string): AttributeToAppliedProjector {
```
jtlan: Will there be a refactor later to have the internal parts of the code use this method?

jtlan:
Small questions.

src/components/plots/abstractPlot.ts
bluong: Possibly, especially after an eventual Drawers refactor, I don't think it's quite possible now.

src/components/plots/abstractPlot.ts
bluong: Object.Keys works as well.  I didn't want to go through the annoyance of isOwnProperty as well as I wanted to iterate on both key/value pair cleanly

bluong: Just in general, if there's a simple d3 utility function I'll tend toward it since I assume that d3 handles any nasty small tidbits (like isOwnProperty)

crmorford:
AH QE+1'd the wrong PR. I need a nap. Sorry!

crmorford:
GIL, sorry this took a billion years

PR #1688, opened 2015-02-26
implement getAllPlotData

bluong:
TODO:
- [x] Write tests
- [x] Discuss if getAllSelections should be removed


bluong:
getAllSelections may be discussed at a later date before v1.0

crmorford:
Pixel point for pie plot is NaN http://jsfiddle.net/80rtooed/1/

test/drawers/arcDrawerTests.ts
```diff
@@ -20,7 +20,7 @@ describe("Drawers", () => {
         var radius = 75;
         var angle = Math.PI / 4 + ((Math.PI * index) / 2);
         var expectedX = radius * Math.sin(angle);
-        var expectedY = radius * Math.cos(angle);
+        var expectedY = -radius * Math.cos(angle);
```
jtlan: ? Why is this necessary?

bluong: Y cartesian coordinates work in the opposite direction of pixel coordinates

jtlan: :(

crmorford:
pixelPoints don't appear to be at the right spot. 
http://jsfiddle.net/80rtooed/4/
type 'n' to add a value, type 'a' to append a circle at plot.getAllPlotData().pixelPoints[0]

bluong:
@crmorford I noticed that you used the getForeground() call to add the circle onto the foreground of the pie plot.  However, the pixelPoint is in relation to the renderArea of the pie plot which is offsetted to be at the center of the component, thus the pixelPoint is in correct but just in the wrong context.  I do note that this usage is not unusual however with unusual resulting behavior.

What do you think we should do?

crmorford:
Solid http://jsfiddle.net/80rtooed/7/

PR #1690, opened 2015-02-27
Use valueOf() when comparing to account for Date objects

### LINKED ISSUE #1689: Time scale locks up if domain is set to two equal values
http://jsfiddle.net/vhqwqwry/2/

Due to `Date` object equality checking.

jtlan:
Close #1689.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -1595,7 +1595,8 @@ var Plottable;
                     return [d - Domainer._PADDING_FOR_IDENTICAL_DOMAIN, d + Domainer._PADDING_FOR_IDENTICAL_DOMAIN];
                 }
             }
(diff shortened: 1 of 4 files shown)


bluong:
Do we want a test?  +1 otherwise

jtlan:
I can add one, but I think `Domainer`s need a major overhaul.

crmorford:
I don't actually see any behavior change from develop on this test. http://jsfiddle.net/ck1psjL3/9/

jtlan:
@crmorford Setting the domain explicitly shuts off autodomaining, even if data changes. Try calling `xScale.autoDomain()` after adding data -- it doesn't work on **develop**.

PR #1691, opened 2015-02-27
Upgrading grunt-mocha-phantomjs

bluong:
PhantomJS results sometimes in a cyclical structure error when attempting to use assert.includeMembers in mocha.  Upgrading should solve the issue.


jtlan:
Does this use Phantom 2.0?

bluong:
I don't think it's 2.0, still with major number 1 I believe

PR #1700, opened 2015-02-28
Factor out ClientToSVGTranslator.

jtlan:
Factored the position-translation logic out into a separate class,
so it can be reused.

### DIFF
--- a/src/utils/clientToSVGTranslator.ts
+++ b/src/utils/clientToSVGTranslator.ts
@@ -1,0 +1,71 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module _Util {
+  export class ClientToSVGTranslator {
+    private static _TRANSLATOR_KEY = "__Plottable_ClientToSVGTranslator"
+    private _svg: SVGElement;
+    private _measureRect: SVGElement;
+
+    public static getTranslator(elem: SVGElement): ClientToSVGTranslator {
+      var svg = _Util.DOM.getBoundingSVG(elem);
+
+      var translator: ClientToSVGTranslator = (<any> svg)[ClientToSVGTranslator._TRANSLATOR_KEY];
+      if (translator == null) {
+        translator = new ClientToSVGTranslator(svg);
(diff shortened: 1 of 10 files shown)


src/utils/clientToSVGTranslator.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (translator == null) {
+        translator = new ClientToSVGTranslator(svg);
+        (<any> svg)[ClientToSVGTranslator._TRANSLATOR_KEY] = translator;
+      }
+      return translator;
+    }
+
+    constructor(svg: SVGElement) {
+      this._svg = svg;
+      this._measureRect = <SVGElement> <any>document.createElementNS(svg.namespaceURI, "rect");
```
bluong: Curious, what's with the multiple casting here

src/utils/clientToSVGTranslator.ts
```diff
@@ -0,0 +1,71 @@
...
+      this._measureRect.setAttribute("class", "measure-rect");
+      this._measureRect.setAttribute("style", "opacity: 0; visibility: hidden;");
+      this._measureRect.setAttribute("width", "1");
+      this._measureRect.setAttribute("height", "1");
+      this._svg.appendChild(this._measureRect);
```
bluong: Curious how much cleaner would this be with using d3?

bluong:
Minor

src/utils/clientToSVGTranslator.ts
jtlan: `createElementNS()` supposedly returns an `Element`, even though in this case it returns an `SVGElement`. Can't cast directly from `Element` to `SVGElement`.

src/utils/clientToSVGTranslator.ts
jtlan: Possibly somewhat, but I wanted to make this work independently of D3.

jtlan:
Purely a regression pass here; visible effects would be on `Interaction.Key` and `Interaction.Hover`.

PR #1704, opened 2015-03-02
[AreaPlot/LinePlot] Disable padding by default

### LINKED ISSUE #747: AreaPlot / LinePlot should default to no padding on xScale

rcchen:
Fixes #747

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2979,6 +2979,8 @@ declare module Plottable {
                     y: number;
                 };
             };
+            protected _updateXDomainer(): void;
+            protected _updateYDomainer(): void;
             _hoverOverComponent(p: Point): void;
             _hoverOutComponent(p: Point): void;
             _doHover(p: Point): Interaction.HoverData;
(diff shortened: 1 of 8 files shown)


bluong:
Builds failed.  Please fix

rcchen:
I ended up removing a number of tests because they seemed to be irrelevant given the new default for padding on area/line plots. Let me know if this should not the case.

crmorford:
Broke barLabelsEnabled (base_animator quicktest)

crmorford:
Not sure if I like taking away the y-drecton padding, either. We should ask Adam (I'll do that)

crmorford:
Adam recommends not padding next to an axis, but padding up to the edge if there is no axis. On a typical area chart, this would mean the left side is padded and the right side isn't. We should think about if it's possible to know this, and if that's what we want to do.

rcchen:
@jtlan is this even possible? I don't think plot elements are knowledgeable about what other plot elements exists to their immediate four sides...

bluong:
@crmorford @rcchen This is certainly not possible / this is not a realm Plottable should dig into.  What we should do instead is come up with a sensible default (I am personally of a fan of using the default where the plot is standalone) and then let the user mess with the paddings as appropriate (the way they should need to mess with the padding should be not difficult).

crmorford:
What does "the default where the plot is standalone" mean?

rcchen:
Closing out for now until we figure out what the best way to deal with defaults is.

PR #1708, opened 2015-03-03
[ScatterPlot] - Symbol Support + Symbol Generators

### LINKED ISSUE #1674: Plot.Scatter support for arbitrary symbol shapes
It'd be good to have the ability to draw arbitrary symbols, instead of circles, for a `Plot.Scatter`. One question is what the API for this looks like? One symbol type per `Plot`? Is it `project()`-ed the same way as any other attribute? Is this a new type of `Plot` entirely?

bluong:
The shapes used in Scatter Plots are now symbols, where symbols are path elements that should be generated using a symbol generator of some sort.  As a result, d3's symbol generator has been leveraged to provide a default way to generate symbols.  Users are free to generate their own SymbolGenerator.

A SymbolGenerator is defined as a function that takes in the radius of the total space for the symbol and returns a string which is the SVG Path string that defines the symbol.

SymbolGenerators can be inserted into a ScatterPlot like so:

```typescript
scatterPlot.project("symbol", symbolGenerator());
```

To ease the use of these symbols, d3's symbol generator has been used and wrapped into a Plottable Util for ease of use.  Its usage can be seen below:

```typescript
scatterPlot.project("symbol", Plottable.SymbolGenerators.d3Symbol("triangle-up"));
```

API-Breaks: Default animators on scatter plots "circles" and "circles-reset" have been renamed to "symbols" and "symbols-reset" respectively.

Fixes #1674


src/components/plots/scatterPlot.ts
```diff
@@ -18,37 +18,23 @@ export module Plot {
       this.classed("scatter-plot", true);
       this._defaultFillColor = new Scale.Color().range()[0];
 
-      this.animator("circles-reset", new Animator.Null());
-      this.animator("circles", new Animator.Base()
+      this.animator("symbols-reset", new Animator.Null());
+      this.animator("symbols", new Animator.Base()
```
jtlan: We'll have to note this as an API-breaking change, I think.

jtlan:
What is the signature of the function that the user is supposed to supply to `project("symbol", ...)`?

test/components/plots/scatterPlotTests.ts
```diff
@@ -20,32 +20,39 @@ describe("Plots", () => {
...
-      assert.closeTo(parseFloat(c2.attr("cy")), 20, 0.01, "second circle cy is correct");
+      var symbols = plot.getAllSelections();
+      var c1 = d3.select(symbols[0][0]);
+      var c2 = d3.select(symbols[0][1]);
+      var c1Position = d3.transform(c1.attr("transform")).translate;
+      var c2Position = d3.transform(c2.attr("transform")).translate;
+      assert.closeTo(parseFloat(c1Position[0]), 0, 0.01, "first symbol cx is correct");
+      assert.closeTo(parseFloat(c1Position[1]), 20, 0.01, "first symbol cy is correct");
+      assert.closeTo(parseFloat(c2Position[0]), 11, 0.01, "second symbol cx is correct");
+      assert.closeTo(parseFloat(c2Position[1]), 20, 0.01, "second symbol cy is correct");
```
jtlan: maybe it would be easier to do these tests with `getBBox()` or `getBoundingClientRect()`?

bluong: It would be less fragile, but I'm unsure about its accuracy.  I'll go try it out

bluong:
Updated PR description

src/utils/symbolGenerators.ts
```diff
@@ -0,0 +1,14 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+
+  export type SymbolGenerator = (datum: any, index: number) => string;
+
+  export module SymbolGenerators {
+
+    export function d3Symbol(symbolType: string | ((datum: any, index: number) => string)) {
```
jtlan: Maybe add some TSDoc for this function?

jtlan:
Docs.

crmorford:
http://jsfiddle.net/80rtooed/6/

crmorford:
Looks great!

crmorford:
On quicktests page, a lot of scatter plots look smaller
screen shot 2015-03-04 at 2 23 58 pm
screen shot 2015-03-04 at 2 23 49 pm

jtlan:
@bluong I've been thinking about this, but would it make sense for the `SymbolGenerator` to return a `<g>` or something with the symbol inside? This might allow us to do something like use the actual `<circle>` element, or images.

src/utils/symbolGenerators.ts
```diff
@@ -0,0 +1,40 @@
...
+     * https://github.com/mbostock/d3/wiki/SVG-Shapes#symbol
+     *
+     * Note that since D3 symbols compute the path strings by knowing how much area it can take up instead of
+     * knowing its dimensions, the total area expected may be off by some constant factor.
+     *
+     * @param {string | ((datum: any, index: number) => string)} symbolType Accessor for the d3 symbol type
+     * @returns {SymbolGenerator} the symbol generator for a D3 symbol
+     */
+    export function d3Symbol(symbolType: string | ((datum: any, index: number) => string)) {
+      return d3.svg.symbol().type(symbolType).size(Math.pow(50, 2));
```
jtlan: Can we store off this `50` somewhere?

bluong:
Union types!

src/utils/symbolGenerators.ts
```diff
@@ -0,0 +1,65 @@
...
+            break;
+          case "cross":
+            sizeFactor = 20/9;
+            break;
+          case "diamond":
+            sizeFactor = 2 * Math.tan(Math.PI / 6);
+            break;
+          case "triangle-up":
+          case "triangle-down":
+            sizeFactor = Math.sqrt(3);
```
jtlan: Can we add some comments explaining all these rescales?

bluong: They shouldn't need to go away.  Now that these numbers have been calculated, we shouldn't have any bounding issues from D3

jtlan:
This is fine for now, but please work with Design to come up with our own `<path>`s for symbols so we don't have to rely on D3's generators.

crmorford:
Can you merge in develop so I can checkout the overlaying quicktests?

bluong:
Release notes pic: 
screen shot 2015-03-13 at 2 00 40 pm

PR #1711, opened 2015-03-03
[Plots] Implement RectanglePlot, reimplement GridPlot to support non-ordinal scales

### LINKED ISSUE #1709: Implement Plot.Rectangle
Plot that allows arbitrary rectangles to be drawn based on (x1, y1) and (x2, y2) that together define a rectangle. This will eventually lead into Histogram (see #1701) as well as a rearchitected GridPlot that allows non-ordinal scales to be used (see #1619)

rcchen:
Still needs unit tests and whatnot, but I wanted to put this out there to kick off a conversation if there is any to be had about the implementation of RectanglePlot (for example, should this be called RectanglePlot?)

Proof of concept with TimeScale and LinearScale/OrdinalScale: http://jsfiddle.net/usv4ombr/2/

Fixes #1709

### DIFF
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -48,8 +48,8 @@ export module Plot {
     }
 
     /**
-     * @param {string} attrToSet One of ["x", "y", "x1", "y1", "x2", "y2", "fill"]. If "fill" is used,
-     * the data should return a valid CSS color. "x1" and "y1" can act as aliases to "x" and "y"
+     * @param {string} attrToSet One of ["x", "y", "x2", "y2", "fill"]. If "fill" is used,
+     * the data should return a valid CSS color.
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
       attrToSet = attrToSet === "x1" ? "x" : attrToSet;
@@ -64,6 +64,11 @@ export module Plot {
           super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
           });
+        } 
+        else if (scale instanceof Scale.AbstractQuantitative) {
+          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m));
+          });
         }
       }
 
@@ -74,6 +79,11 @@ export module Plot {
           });
           super.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        } 
+        else if (scale instanceof Scale.AbstractQuantitative) {
+          super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["y"].accessor(d, i, u, m));
           });
         }
       }
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -54,21 +54,18 @@ export module Plot {
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
       super.project(attrToSet, accessor, scale);
 
-      // Use x to determine x1 by default
       if (attrToSet === "x") {
         super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
           return scale.scale(this._projections["x"].accessor(d, i, u, m));
         });
       }
 
-      // Use y to determine y1 by default
       if (attrToSet === "y") {
         super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
           return scale.scale(this._projections["y"].accessor(d, i, u, m));
         });
       }
 
-      // Situation where x is defined but x2 is not defined
       if (attrToSet === "x" && this._projections["x2"] === undefined) {
         if (scale instanceof Scale.Category) {
           super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
@@ -80,7 +77,6 @@ export module Plot {
         }
       }
 
-      // Situation where y is defined but y2 is not defined
       if (attrToSet === "y" && this._projections["y2"] === undefined) {
         if (scale instanceof Scale.Category) {
           super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -53,39 +53,50 @@ export module Plot {
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
       super.project(attrToSet, accessor, scale);
+
+      // Use x to determine x1 by default
+      if (attrToSet === "x") {
+        super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["x"].accessor(d, i, u, m));
+        });
+      }
+
+      // Use y to determine y1 by default
+      if (attrToSet === "y") {
+        super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["y"].accessor(d, i, u, m));
+        });
+      }
+
+      // Situation where x is defined but x2 is not defined
+      if (attrToSet === "x" && this._projections["x2"] === undefined) {
+        if (scale instanceof Scale.Category) {
+          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
+          });
+          super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        }
+      }
+
+      // Situation where y is defined but y2 is not defined
+      if (attrToSet === "y" && this._projections["y2"] === undefined) {
+        if (scale instanceof Scale.Category) {
+          super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["y"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
+          });
+          super.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["y"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        }
+      }
+
       if (attrToSet === "fill") {
         this._colorScale = this._projections["fill"].scale;
       }
+
       return this;
-    }
-
-    protected _generateAttrToProjector() {
-      var attrToProjector = super._generateAttrToProjector();
-
-      // Copy each of the different projectors
-      var x1Attr = attrToProjector["x"];
-      var y1Attr = attrToProjector["y"];
-      var x2Attr = attrToProjector["x2"];
-      var y2Attr = attrToProjector["y2"];
-
-      // Adjust the xScale if it is ordinal
-      if (this._xScale instanceof Scale.Category) {
-        attrToProjector["width"] = () => (<Scale.Category> this._xScale).rangeBand();
-        attrToProjector["x"] = (d, i, u, m) => x1Attr(d, i, u, m) - attrToProjector["width"](d, i, u, m) / 2;
-      } else {
-        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
-      }
-
-      // Adjust the yScale if it is ordinal
-      if (this._yScale instanceof Scale.Category) {
-        attrToProjector["height"] = () => (<Scale.Category> this._yScale).rangeBand();
-        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m) / 2;
-      } else {
-        attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
-        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m);
-      }
-
-      return attrToProjector;
     }
 
     protected _generateDrawSteps(): _Drawer.DrawStep[] {
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -2,7 +2,7 @@
 
 module Plottable {
 export module Plot {
-  export class Grid extends AbstractXYPlot<string,string> {
+  export class Grid extends Rectangle<any, any> {
     private _colorScale: Scale.AbstractScale<any, string>;
 
     /**
@@ -12,18 +12,23 @@ export module Plot {
      * grid, and the datum can control what color it is.
      *
      * @constructor
-     * @param {Scale.Category} xScale The x scale to use.
-     * @param {Scale.Category} yScale The y scale to use.
+     * @param {Scale.AbstractScale} xScale The x scale to use.
+     * @param {Scale.AbstractScale} yScale The y scale to use.
      * @param {Scale.Color|Scale.InterpolatedColor} colorScale The color scale
      * to use for each grid cell.
      */
-    constructor(xScale: Scale.Category, yScale: Scale.Category, colorScale: Scale.AbstractScale<any, string>) {
+    constructor(xScale: Scale.AbstractScale<any, any>, yScale: Scale.AbstractScale<any, any>,
+      colorScale: Scale.AbstractScale<any, string>) {
       super(xScale, yScale);
       this.classed("grid-plot", true);
 
-      // The x and y scales should render in bands with no padding
-      xScale.innerPadding(0).outerPadding(0);
-      yScale.innerPadding(0).outerPadding(0);
+      // The x and y scales should render in bands with no padding for category scales
+      if (xScale instanceof Scale.Category) {
+        xScale.innerPadding(0).outerPadding(0);
+      }
+      if (yScale instanceof Scale.Category) {
+        yScale.innerPadding(0).outerPadding(0);
+      }
 
       this._colorScale = colorScale;
       this.animator("cells", new Animator.Null());
@@ -56,14 +61,30 @@ export module Plot {
 
     protected _generateAttrToProjector() {
       var attrToProjector = super._generateAttrToProjector();
-      var xStep = (<Scale.Category> this._xScale).rangeBand();
-      var yStep = (<Scale.Category> this._yScale).rangeBand();
-      attrToProjector["width"]  = () => xStep;
-      attrToProjector["height"] = () => yStep;
-      var xAttr = attrToProjector["x"];
-      var yAttr = attrToProjector["y"];
-      attrToProjector["x"] = (d, i, u, m) => xAttr(d, i, u, m) - xStep / 2;
-      attrToProjector["y"] = (d, i, u, m) => yAttr(d, i, u, m) - yStep / 2;
+
+      // Copy each of the different projectors
+      var x1Attr = attrToProjector["x"];
+      var y1Attr = attrToProjector["y"];
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Adjust the xScale if it is ordinal
+      if (this._xScale instanceof Scale.Category) {
+        attrToProjector["width"] = () => (<Scale.Category> this._xScale).rangeBand();
+        attrToProjector["x"] = (d, i, u, m) => x1Attr(d, i, u, m) - attrToProjector["width"](d, i, u, m) / 2;
+      } else {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+      }
+
+      // Adjust the yScale if it is ordinal
+      if (this._yScale instanceof Scale.Category) {
+        attrToProjector["height"] = () => (<Scale.Category> this._yScale).rangeBand();
+        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m) / 2;
+      } else {
+        attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m);
+      }
+
       return attrToProjector;
     }
 
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -48,23 +48,13 @@ export module Plot {
     }
 
     /**
-     * @param {string} attrToSet One of ["x", "y", "fill"]. If "fill" is used,
-     * the data should return a valid CSS color.
+     * @param {string} attrToSet One of ["x", "y", "x1", "y1", "x2", "y2", "fill"]. If "fill" is used,
+     * the data should return a valid CSS color. "x1" and "y1" can act as aliases to "x" and "y"
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
+      attrToSet = attrToSet === "x1" ? "x" : attrToSet;
+      attrToSet = attrToSet === "y1" ? "y" : attrToSet;
       super.project(attrToSet, accessor, scale);
-
-      if (attrToSet === "x") {
-        super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
-          return scale.scale(this._projections["x"].accessor(d, i, u, m));
-        });
-      }
-
-      if (attrToSet === "y") {
-        super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
-          return scale.scale(this._projections["y"].accessor(d, i, u, m));
-        });
-      }
 
       if (attrToSet === "x" && this._projections["x2"] === undefined) {
         if (scale instanceof Scale.Category) {
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -64,8 +64,7 @@ export module Plot {
           super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
           });
-        } 
-        else if (scale instanceof Scale.AbstractQuantitative) {
+        } else if (scale instanceof Scale.AbstractQuantitative) {
           super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m));
           });
@@ -80,8 +79,7 @@ export module Plot {
           super.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
           });
-        } 
-        else if (scale instanceof Scale.AbstractQuantitative) {
+        } else if (scale instanceof Scale.AbstractQuantitative) {
           super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m));
           });
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -52,35 +52,35 @@ export module Plot {
      * the data should return a valid CSS color.
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
-      attrToSet = attrToSet === "x1" ? "x" : attrToSet;
-      attrToSet = attrToSet === "y1" ? "y" : attrToSet;
       super.project(attrToSet, accessor, scale);
 
-      if (attrToSet === "x" && this._projections["x2"] === undefined) {
+      if (attrToSet === "x") {
         if (scale instanceof Scale.Category) {
-          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          this.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
           });
-          super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          this.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
           });
-        } else if (scale instanceof Scale.AbstractQuantitative) {
-          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+        }
+        if (scale instanceof Scale.AbstractQuantitative) {
+          this.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["x"].accessor(d, i, u, m));
           });
         }
       }
 
-      if (attrToSet === "y" && this._projections["y2"] === undefined) {
+      if (attrToSet === "y") {
         if (scale instanceof Scale.Category) {
-          super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          this.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
           });
-          super.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          this.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
           });
-        } else if (scale instanceof Scale.AbstractQuantitative) {
-          super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+        }
+        if (scale instanceof Scale.AbstractQuantitative) {
+          this.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
             return scale.scale(this._projections["y"].accessor(d, i, u, m));
           });
         }
--- a/src/components/plots/rectanglePlot.ts
+++ b/src/components/plots/rectanglePlot.ts
@@ -8,8 +8,8 @@ export module Plot {
      * Constructs a RectanglePlot.
      *
      * A RectanglePlot consists of a bunch of rectangles. The user is required to
-     * project the top left corner of each rectangle (x1, y1) and the bottom right
-     * corner of each rectangle (x2, y2)
+     * project the left and right bounds of the rectangle (x1 and x2 respectively)
+     * as well as the bottom and top bounds (y1 and y2 respectively)
      *
      * @constructor
      * @param {Scale.AbstractScale} xScale The x scale to use.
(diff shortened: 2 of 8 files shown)


bluong:
Remember to run `grunt test` before submitting to avoid random failing builds.

bluong:
On purely a visual front, this looks good so far.

bluong:
The first thing I notice is that x,y,width,height have been changed to x1, x2, y1, y2.  Do we want this to be consistent across all things extending from RectanglePlot?  I'm unsure if this 4-projectedAttribute set would work for a standard bar plot

bluong:
On the GridPlot emulating example, why is y projected but not y2?

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+     * A RectanglePlot consists of a bunch of rectangles. The user is required to
+     * project the top left corner of each rectangle (x1, y1) and the bottom right
+     * corner of each rectangle (x2, y2)
+     *
+     * @constructor
+     * @param {Scale.Ordinal} xScale The x scale to use.
+     * @param {Scale.Ordinal} yScale The y scale to use.
+     */
+    constructor(xScale: Scale.AbstractScale<any, any>, yScale: Scale.AbstractScale<any, any>,
+        colorScale: Scale.AbstractScale<any, string>) {
```
bluong: I don't like having the colorScale in the constructor.  We are planning to remove these args from the constructor anyways so...

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+      super(xScale, yScale);
+
+      // Category scales should not have padding by default
+      if (xScale instanceof Scale.Ordinal) {
+        (<Scale.Ordinal> xScale).innerPadding(0).outerPadding(0);
+      }
+      if (yScale instanceof Scale.Ordinal) {
+        (<Scale.Ordinal> yScale).innerPadding(0).outerPadding(0);
+      }
```
bluong: Err we should be careful here since this works on gridplots but not barPlots or any of the other plots (unless we want them to have 0 padding)

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      this._colorScale = colorScale;
+      this.classed("rectangle-plot", true);
+    }
+
+    public addDataset(keyOrDataset: any, dataset?: any) {
+      super.addDataset(keyOrDataset, dataset);
+      return this;
+    }
```
bluong: function seems to not do anything

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+     * @param {string} attrToSet One of ["x", "y", "fill"]. If "fill" is used,
+     * the data should return a valid CSS color.
+     */
+    public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
+      super.project(attrToSet, accessor, scale);
+      if (attrToSet === "fill") {
+        this._colorScale = this._projections["fill"].scale;
+      }
+      return this;
+    }
```
bluong: Remove this since we shouldn't need a reference to the colorScale i think

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+        return y1Attr(d, i, u, m) - offset;
+      };
+
+      return attrToProjector;
+    }
+
+    protected _generateDrawSteps(): _Drawer.DrawStep[] {
+      return [{attrToProjector: this._generateAttrToProjector(), animator: this._getAnimator("cells")}];
+    }
```
bluong: "cells" seems to be a copy paste remnant.  Overwrite with something appropriate?

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+      var y2Attr = attrToProjector["y2"];
+
+      // Infer width, with an exception for ordinal scales
+      if (x2Attr === undefined) {
+        attrToProjector["width"] = () => (<Scale.Ordinal> this._xScale).rangeBand();
+      } else {
+        attrToProjector["width"] = (d, i, u, m) => {
+          return Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+        };
+      }
```
bluong: Not sure if need to special case like this  Also I'm not a fan of the check

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      // Adjust x to respect the width for ordinal scales
+      attrToProjector["x"] = (d, i, u, m) => {
+        var offset = 0;
+        if (x2Attr === undefined) {
+          offset = attrToProjector["height"](d, i, u, m) / 2;
+        }
+        return x1Attr(d, i, u, m) - offset;
+      };
```
bluong: Above

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      // Adjust y to respect the height, with an exception for ordinal scales
+      attrToProjector["y"] = (d, i, u, m) => {
+        var offset = attrToProjector["height"](d, i, u, m);
+        if (y2Attr === undefined) {
+          offset = offset / 2;
+        }
+        return y1Attr(d, i, u, m) - offset;
+      };
```
bluong: Above

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+
+
+      // Infer height, with an exception for ordinal scales
+      if (y2Attr === undefined) {
+        attrToProjector["height"] = () => (<Scale.Ordinal> this._yScale).rangeBand();
+      } else {
+        attrToProjector["height"] = (d, i, u, m) => {
+          return Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+        };
+      }
```
bluong: Above

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Constructs a RectanglePlot.
+     *
+     * A RectanglePlot consists of a bunch of rectangles. The user is required to
+     * project the top left corner of each rectangle (x1, y1) and the bottom right
+     * corner of each rectangle (x2, y2)
+     *
+     * @constructor
+     * @param {Scale.Ordinal} xScale The x scale to use.
+     * @param {Scale.Ordinal} yScale The y scale to use.
```
bluong: Modify doc

bluong:
Another thing to note is how fast do we want the raw implementation of Plot.Rectangle?  I can imagine that the other plots that use rectangles can use this plot as a parent plot

src/components/plots/gridPlot.ts
```diff
@@ -12,18 +12,23 @@ export module Plot {
...
-     * @param {Scale.Ordinal} xScale The x scale to use.
-     * @param {Scale.Ordinal} yScale The y scale to use.
+     * @param {Scale.AbstractScale} xScale The x scale to use.
+     * @param {Scale.AbstractScale} yScale The y scale to use.
      * @param {Scale.Color|Scale.InterpolatedColor} colorScale The color scale
      * to use for each grid cell.
      */
-    constructor(xScale: Scale.Ordinal, yScale: Scale.Ordinal, colorScale: Scale.AbstractScale<any, string>) {
+    constructor(xScale: Scale.AbstractScale<any, any>, yScale: Scale.AbstractScale<any, any>,
+      colorScale: Scale.AbstractScale<any, string>) {
```
bluong: Probably keep on the same line?  NBD

rcchen: It was > 140 characters, TSLint doesn't like it

src/components/plots/gridPlot.ts
```diff
@@ -55,15 +60,32 @@ export module Plot {
...
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Adjust the xScale if it is ordinal
+      if (this._xScale instanceof Scale.Ordinal) {
+        attrToProjector["width"] = () => (<Scale.Ordinal> this._xScale).rangeBand();
+        attrToProjector["x"] = (d, i, u, m) => x1Attr(d, i, u, m) - attrToProjector["width"](d, i, u, m) / 2;
+      } else {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+      }
```
bluong: Is there a way to incorporate attrToProjector["x2"] in both cases?  Not sure if I'm a fan of disallowing different projections based on the class of the scale

src/components/plots/gridPlot.ts
```diff
@@ -55,15 +60,32 @@ export module Plot {
...
+
+      // Adjust the yScale if it is ordinal
+      if (this._yScale instanceof Scale.Ordinal) {
+        attrToProjector["height"] = () => (<Scale.Ordinal> this._yScale).rangeBand();
+        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m) / 2;
+      } else {
+        attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+        attrToProjector["y"] = (d, i, u, m) => y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m);
+      }
```
bluong: Same as above.

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,61 @@
...
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Define width if both x1 and x2 are defined
+      if (x1Attr !== undefined && x2Attr !== undefined) {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+        attrToProjector["x"] = (d, i, u, m) => {
+          return x1Attr(d, i, u, m) < x2Attr(d, i, u, m) ? x1Attr(d, i, u, m) : x2Attr(d, i, u, m);
+        };
+      }
```
bluong: Not sure if you need this undefined checking.  Verification is good, but we don't have a good consistent story about verification.

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,61 @@
...
+      var x1Attr = attrToProjector["x"];
+      var y1Attr = attrToProjector["y"];
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Define width if both x1 and x2 are defined
+      if (x1Attr !== undefined && x2Attr !== undefined) {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+        attrToProjector["x"] = (d, i, u, m) => {
+          return x1Attr(d, i, u, m) < x2Attr(d, i, u, m) ? x1Attr(d, i, u, m) : x2Attr(d, i, u, m);
```
bluong: Math.min?

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      // Define height if both y1 and y2 are defined
+      if (y1Attr !== undefined && y2Attr !== undefined) {
+        attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+        attrToProjector["y"] = (d, i, u, m) => {
+          return y1Attr(d, i, u, m) > y2Attr(d, i, u, m) ? y1Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m) :
+                                                           y2Attr(d, i, u, m) - attrToProjector["height"](d, i, u, m);
+        };
+      }
```
bluong: Same as above.

src/components/plots/rectanglePlot.ts
rcchen: It wasn't happy with Plottable._Util.Methods.min for some reason, Math.min is disallowed right?

src/components/plots/gridPlot.ts
bluong: We could increase the character limit... but nah.  Maybe put an arg on each line...?  Again NBD

src/components/plots/gridPlot.ts
```diff
@@ -55,15 +60,32 @@ export module Plot {
     }
 
     protected _generateAttrToProjector() {
+
```
bluong: I personally don't like the line here, but NBD

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,56 @@
...
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Define width if both x1 and x2 are defined
+      if (x1Attr !== undefined && x2Attr !== undefined) {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+        attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m));
+      }
+
+      // Define height if both y1 and y2 are defined
```
bluong: This manufacturing of these different attrs might want to be appropriately done at the drawer level.

src/components/plots/gridPlot.ts
bluong: What we might want to do for this project class is correct x1/y1/x2/y2 and then hand these off to the Drawer to compute what width/height/x/y should be

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,50 @@
...
+      var y1Attr = attrToProjector["y"];
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+      attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+
+      attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m));
+      attrToProjector["y"] = (d, i, u, m) => Math.max(y1Attr(d, i, u, m), y2Attr(d, i, u, m)) - attrToProjector["height"](d, i, u, m);
+
```
bluong: Might want to set width/height/x/y on the drawer level

bluong:
Yep I misinterpreted the RectanglePlot and read a lot of GridPlot as RectanglePlot.  We should still move a lot of the width/height/x1/x2/etc. ideas into the drawer as I think it's more appropritate for that manufacturing to exist there.  Feel free to mess with the RectDrawer

src/components/plots/gridPlot.ts
rcchen: Discussed offline, filed in #1716

bluong:
The history is looking all weird...?  commits were said to be submitted after I commented on some changes

rcchen:
I had to do a rebase then a force push because of the Ordinal -> Category refactor.

bluong:
Careful with doing force pushes.  Using them will tend to mess up history.  It should be extremely rare when you'll need to use them.

src/components/plots/gridPlot.ts
```diff
@@ -56,14 +61,30 @@ export module Plot {
...
+      var y1Attr = attrToProjector["y"];
+      var x2Attr = attrToProjector["x2"];
+      var y2Attr = attrToProjector["y2"];
+
+      // Adjust the xScale if it is ordinal
+      if (this._xScale instanceof Scale.Category) {
+        attrToProjector["width"] = () => (<Scale.Category> this._xScale).rangeBand();
+        attrToProjector["x"] = (d, i, u, m) => x1Attr(d, i, u, m) - attrToProjector["width"](d, i, u, m) / 2;
+      } else {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
```
bluong: On a curious note, would it be possible to not need to specify x2 for quantitative scale but have more fine-grained control if they do specify it?

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,54 @@
...
+      if (x2Attr !== undefined) {
+        attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
+        attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m));
+      }
+
+      if (y2Attr !== undefined) {
+        attrToProjector["height"] = (d, i, u, m) => Math.abs(y2Attr(d, i, u, m) - y1Attr(d, i, u, m));
+        attrToProjector["y"] = (d, i, u, m) => Math.max(y1Attr(d, i, u, m), y2Attr(d, i, u, m)) - attrToProjector["height"](d, i, u, m);
+      }
```
bluong: Not sure why this undefined check is needed

test/testReference.ts
```diff
@@ -29,12 +29,13 @@
 ///<reference path="components/plots/linePlotTests.ts" />
 ///<reference path="components/plots/areaPlotTests.ts" />
 ///<reference path="components/plots/barPlotTests.ts" />
+///<reference path="components/plots/clusteredBarPlotTests.ts" />
```
bluong: Why this move?

src/components/plots/rectanglePlot.ts
rcchen: It ends up throwing an exception on line 37/42 or whatever and won't continue execution past the super call (https://github.com/palantir/plottable/pull/1711/files#diff-85d9beec29bb331c98aa7092fd9d30d8R63), where x2/y2 actually are defined

test/testReference.ts
rcchen: Alphabetization, but now that I look at it again, a bunch of the others are also misalphabetized..

src/components/plots/gridPlot.ts
rcchen: Not specifying x2 for quantitative scales has no inherent meaning though, unless drawing 1px lines is a valid use case..?

test/testReference.ts
bluong: I think the ordering is in compile order (barPlot must go before things that extend from barPlot)

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,50 @@ export module Plot {
...
+
+      // Use y to determine y1 by default
+      if (attrToSet === "y") {
+        super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["y"].accessor(d, i, u, m));
+        });
+      }
+
+      // Situation where x is defined but x2 is not defined
+      if (attrToSet === "x" && this._projections["x2"] === undefined) {
```
jtlan: The comment is sort of redundant, since what it says can be determined by examining the next line.

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,50 @@ export module Plot {
...
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
+          });
+          super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        }
+      }
+
+      // Situation where y is defined but y2 is not defined
+      if (attrToSet === "y" && this._projections["y2"] === undefined) {
```
jtlan: Ditto here.

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,50 @@ export module Plot {
...
+      // Use y to determine y1 by default
+      if (attrToSet === "y") {
+        super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["y"].accessor(d, i, u, m));
+        });
+      }
+
+      // Situation where x is defined but x2 is not defined
+      if (attrToSet === "x" && this._projections["x2"] === undefined) {
+        if (scale instanceof Scale.Category) {
```
jtlan: I notice there's no "`else`" block. Does it just blow up if only "x" is set but not "x2" an the scale is not `Category`?

test/components/plots/rectanglePlotTests.ts
```diff
@@ -0,0 +1,60 @@
...
+              .project("x", "x", xScale)
+              .project("y", "y", yScale)
+              .project("x2", "x2", xScale)
+              .project("y2", "y2", yScale)
+              .renderTo(svg);
+      VERIFY_CELLS((<any> rectanglePlot)._renderArea.selectAll("rect"));
+      svg.remove();
+    });
+
+    it("renders correctly when data is set after construction", () => {
```
jtlan: It doesn't seem like this test should be necessary.

src/components/plots/gridPlot.ts
rcchen: It currently blows up, I just need to make sure this general architecture is ok before moving on to other scales.

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,46 @@ export module Plot {
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
       super.project(attrToSet, accessor, scale);
+
+      if (attrToSet === "x") {
+        super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["x"].accessor(d, i, u, m));
+        });
+      }
```
bluong: We should follow the scatter plot model, and essentially

if attrToSet is "x" then change attrToSet to "x1"

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,46 @@ export module Plot {
...
+
+      if (attrToSet === "y") {
+        super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+          return scale.scale(this._projections["y"].accessor(d, i, u, m));
+        });
+      }
```
bluong: Same as above

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,59 @@
...
+      return new _Drawer.Rect(key, true);
+    }
+
+    protected _generateAttrToProjector() {
+      var attrToProjector = super._generateAttrToProjector();
+
+      // Copy each of the different projectors.
+      // In the event that x1/y1 are not present, x/y will be directly leveraged
+      var x1Attr = attrToProjector["x1"] ? attrToProjector["x1"] : attrToProjector["x"];
+      var y1Attr = attrToProjector["y1"] ? attrToProjector["y1"] : attrToProjector["y"];
```
bluong: Let's have subclasses consistent to map to x1/y1

src/components/plots/gridPlot.ts
```diff
@@ -48,23 +53,46 @@ export module Plot {
...
+
+      if (attrToSet === "y" && this._projections["y2"] === undefined) {
+        if (scale instanceof Scale.Category) {
+          super.project("y1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["y"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
+          });
+          super.project("y2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["y"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        }
```
bluong: While the above seems very sensible, I usually see this adjustments made at the generateAttrToProjector step.  That said, I'm not 100% certain where this adjustment should be made.

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,59 @@
...
+      attrToProjector["y"] = (d, i, u, m) => Math.max(y1Attr(d, i, u, m), y2Attr(d, i, u, m)) - attrToProjector["height"](d, i, u, m);
+
+      // Clean up the attributes projected onto the SVG elements
+      delete attrToProjector["x1"];
+      delete attrToProjector["y1"];
+      delete attrToProjector["x2"];
+      delete attrToProjector["y2"];
+
+      return attrToProjector;
+    }
```
bluong: Something to consider, possibly not in this PR, but something that could be applied to all RectanglePlots

I suggest the statement that seems to work well between histograms and bar charts
If x1 === x2 then some sensible width should be used.

bluong:
Few issues

src/components/plots/gridPlot.ts
rcchen: Needs to be the other way around, because AbstractXY plot leverages x/y directly

src/components/plots/rectanglePlot.ts
rcchen: See concern about mapping to x/y due to AbstractXYPlot

bluong:
Would it be possible/make sense if we have x/x2 rather than x1/x2?  It seems we are aliasing x1 to x and vice versa in generateAttrToProjector / project.

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
     }
 
     /**
-     * @param {string} attrToSet One of ["x", "y", "fill"]. If "fill" is used,
+     * @param {string} attrToSet One of ["x", "y", "x2", "y2", "fill"]. If "fill" is used,
      * the data should return a valid CSS color.
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
+      attrToSet = attrToSet === "x1" ? "x" : attrToSet;
```
bluong: ? I thought with our discussion, x would map to x1/x2 if the scale is category.  Now we have x1 map to x?  This also is a bit confusing since RectanglePlot wants to be in terms of x1/x2 to projecting on the x attribute is not useful for RectanglePlot.

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
...
+          });
+          super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        } else if (scale instanceof Scale.AbstractQuantitative) {
+          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m));
+          });
+        }
+      }
```
bluong: With our prior discussion, wouldn't this be more like if attrToSet is x and we have an instance of CategoryScale then we project onto x1/x2?

bluong: With what we discussed, there's no longer any need for checking if the x2 projection exists

bluong: Also, the fact that we have a bunch of super calls to project seems  a bit hacky.  Normal codeflow seems to be some validation before going to the super call to call once, not query using this many codepaths

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,58 @@
...
+     * as well as the bottom and top bounds (y1 and y2 respectively)
+     *
+     * @constructor
+     * @param {Scale.AbstractScale} xScale The x scale to use.
+     * @param {Scale.AbstractScale} yScale The y scale to use.
+     */
+    constructor(xScale: Scale.AbstractScale<X, any>, yScale: Scale.AbstractScale<Y, any>) {
+      super(xScale, yScale);
+      this.classed("rectangle-plot", true);
+    }
```
bluong: I expected here to be a project call that just simply does a translation from x1 -> x and y1 -> y

src/components/plots/rectanglePlot.ts
```diff
@@ -0,0 +1,58 @@
...
+    protected _getDrawer(key: string) {
+      return new _Drawer.Rect(key, true);
+    }
+
+    protected _generateAttrToProjector() {
+      var attrToProjector = super._generateAttrToProjector();
+
+      // Copy each of the different projectors.
+      var x1Attr = attrToProjector["x1"];
+      var y1Attr = attrToProjector["y1"];
```
bluong: Similar to how scatterPlot used to implement attrToProjector, we should probably pull the x/y attrToProjectors and then rename them to use x1/x2

https://github.com/palantir/plottable/blob/master/src/components/plots/scatterPlot.ts#L44

rcchen: x1/x2 are reliant on the projection of x for the calculations to be accurate I believe

https://github.com/palantir/plottable/pull/1711/files#diff-85d9beec29bb331c98aa7092fd9d30d8R62

rcchen:
Summary of things from discussion earlier today:

* If you want to use a CategoryScale with a GridPlot, you pass in values for "x" and "y". These values will be internally translated into "x1", "x2", "y1", "y2"
* If you want to use a QuantitativeScale with a GridPlot, you pass in values for "x", "x2", "y", and "y2". The values for "x" and "y" will be internally translated into "x1" and "y1".
* A RectanglePlot expects to receive the following values from any of its subclasses: "x", "x1", "x2", "y", "y1", "y2". All of these values must be present for a RectanglePlot to be constructed successfully. "x" and "y" are used by the AbstractXYPlot in order to figure out what values correspond to what rectangles, and "x1/2" and "y1/2" are used to define boundaries of each of the rectangles. 
* There are no guarantees that "x" and "x1" actually map to the same values. Case in point: category axes. 

I think that covers the main points. @bluong can you verify that this is what should be expected? I think there's a bunch of back and forth right now on things that I haven't done a great job of explaining unfortunately.

src/components/plots/rectanglePlot.ts
rcchen: Why? x1 is not necessarily going to be x.

Example: CategoryScale, where x1 is mapped to x - rangeBand()/2

src/components/plots/gridPlot.ts
rcchen: What do you suggest to avoid the super.project calls? I think we had a discussion earlier about this and we decided that this was the way to go because x1/x2 projections had to be created, and they also had to be mapped correctly, and this was the most straightforward way to do it...?

bluong:
I'll go into each point and then re-review accordingly
```
* If you want to use a CategoryScale with a GridPlot, you pass in values for "x" and "y". These values will be internally translated into "x1", "x2", "y1", "y2"
```
Yep, this seems correct.  However, does this mean that we project onto x and x1 or just x1?
```
* If you want to use a QuantitativeScale with a GridPlot, you pass in values for "x", "x2", "y", and "y2". The values for "x" and "y" will be internally translated into "x1" and "y1".
```
Note that x1 is the same as x here.
* A RectanglePlot expects to receive the following values from any of its subclasses: "x", "x1", "x2", "y", "y1", "y2". All of these values must be present for a RectanglePlot to be constructed successfully. "x" and "y" are used by the AbstractXYPlot in order to figure out what values correspond to what rectangles, and "x1/2" and "y1/2" are used to define boundaries of each of the rectangles. 
* There are no guarantees that "x" and "x1" actually map to the same values. Case in point: category axes.

rcchen:
Yeah, I think the fundamental misunderstanding was with what "x" and "x1" actually meant, where "x" was the actual value and "x1" was the left bound in pixel space. Is everything cleared up now?

src/components/plots/gridPlot.ts
bluong: Can we just loop it back to a this.project?

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
...
-     * @param {string} attrToSet One of ["x", "y", "fill"]. If "fill" is used,
+     * @param {string} attrToSet One of ["x", "y", "x2", "y2", "fill"]. If "fill" is used,
      * the data should return a valid CSS color.
      */
     public project(attrToSet: string, accessor: any, scale?: Scale.AbstractScale<any, any>) {
+      attrToSet = attrToSet === "x1" ? "x" : attrToSet;
+      attrToSet = attrToSet === "y1" ? "y" : attrToSet;
       super.project(attrToSet, accessor, scale);
+
+      if (attrToSet === "x" && this._projections["x2"] === undefined) {
```
bluong: As per our previous discussion, I don't think we need this x2 projection branch anymore.

Something along the lines of

if (attrToSet is x and the scale is Category instanced)
  project onto x1/x2

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
...
+        if (scale instanceof Scale.Category) {
+          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) - scale.rangeBand() / 2;
+          });
+          super.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        } else if (scale instanceof Scale.AbstractQuantitative) {
+          super.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m));
```
bluong: For this case, since x/x1 are the same thing for quantitative scales

if (attrToSet is x and scale is Quantitative instanced)
project onto x1 the same way

src/components/plots/rectanglePlot.ts
bluong: Got it.  ReactanglePlot knows no correlation between x and x1/x2

src/components/plots/rectanglePlot.ts
bluong: Now the definition of x1/x2/x are a bit more clear, these lines make a bit more sense

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
     }
 
     /**
-     * @param {string} attrToSet One of ["x", "y", "fill"]. If "fill" is used,
+     * @param {string} attrToSet One of ["x", "y", "x2", "y2", "fill"]. If "fill" is used,
```
bluong: Doesn't mention x1/y1

rcchen: x1/y1 aren't set here though, just x/y

src/components/plots/gridPlot.ts
```diff
@@ -43,28 +48,49 @@ export module Plot {
...
+          });
+          this.project("x2", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m)) + scale.rangeBand() / 2;
+          });
+        }
+        if (scale instanceof Scale.AbstractQuantitative) {
+          this.project("x1", (d: any, i: number, u: any, m: Plot.PlotMetadata) => {
+            return scale.scale(this._projections["x"].accessor(d, i, u, m));
+          });
+        }
```
bluong: If attrToSet is x1 and the scale is quantitative, x should be projected right?

bluong:
Everything looks fine except for the two notes I just mentioned

src/components/plots/gridPlot.ts
bluong: x1/y1 are set here through if the scale is quantitative correct?

src/components/plots/gridPlot.ts
rcchen: Same as above, GridPlot is not responsible for directly projecting x1

src/components/plots/gridPlot.ts
rcchen: GridPlot only takes x/y/x2/y2. If it were quantitative you would be projecting x/x2, which internally gets converted into an x1

test/components/plots/rectanglePlotTests.ts
```diff
@@ -0,0 +1,45 @@
...
+    ];
+
+    var VERIFY_CELLS = (cells: D3.Selection) => {
+      assert.equal(cells[0].length, 5);
+      cells.each(function(d: D3.Selection, i: number) {
+        var cell = d3.select(this);
+        assert.closeTo(Number(cell.attr("height")), 50, 0.5, "Cell height is correct");
+        assert.closeTo(Number(cell.attr("width")), 50, 0.5, "Cell width is correct");
+        assert.closeTo(Number(cell.attr("x")), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
+        assert.closeTo(Number(cell.attr("y")), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
```
bluong: You can use the "+" instead of Number(), just a bit cleaner

crmorford:
Seems odd that the default color isn't the same blue we use for everything else http://jsfiddle.net/fvsufoww/16/

crmorford:
I made this http://jsfiddle.net/fvsufoww/18/
and I want it to work like this: http://jsfiddle.net/fvsufoww/19/ (with a time axis and cat axis) How do i make that happen?

crmorford:
I think we should have no default padding, i'm used to grid plots taking up the whole space http://jsfiddle.net/fvsufoww/21/

rcchen:
@crmorford give me a couple of minutes, biking to acquire food right now

rcchen:
Re: default padding, CategoryAxis does not have the concept of a *2 attribute (x2 or y2), it only has the concept of the x/y attribute. See http://jsfiddle.net/fvsufoww/23/

rcchen:
Re: time and cat axis, http://jsfiddle.net/fvsufoww/24/

bluong:
@rcchen Note that @crmorford is using RectanglePlot not GridPlot

rcchen:
@bluong RectanglePlot has no native support for CategoryAxis by design.

crmorford:
AH okay my bad. Is Rectangle Axis meant to be used alone at all? Or just as an ancestor of Grid Plot + future plots.

rcchen:
@crmorford it's mostly intended as an ancestor. You can use RectanglePlot directly to draw really arbitrary rectangles all over the place but if you're actually trying to do anything there should be a plot for that (eventually)

crmorford:
Got it, okay. Looks good to me :) For release notes, make it sure clear
* How projections have changed for grid plot
* How to ensure there isn't undesired padding

rcchen:
Got it, will do re: release notes.

PR #1723, opened 2015-03-09
[WIP] selection-box using Component as a Layer

jtlan:
This implementation uses a `Component`, `SelectionBoxLayer`, that occupies the entire cell. The selection box drawn by the `Component` does not occupy the whole cell.

<del>Compare to implementation in #1724.</del> Decided to go with this one.

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -123,39 +123,39 @@ svg.plottable {
 }
 
 /* SelectionBox */
-.plottable .selection-box .selection-area {
+.plottable .selection-box-layer .selection-area {
   fill: black;
   fill-opacity: 0.03;
   stroke: #CCC;
 }
 
-.plottable .selection-box.x-movable .selection-area {
+.plottable .selection-box-layer.x-movable .selection-area {
   cursor: ew-resize;
 }
-.plottable .selection-box.y-movable .selection-area {
+.plottable .selection-box-layer.y-movable .selection-area {
   cursor: ns-resize;
 }
-.plottable .selection-box.x-movable.y-movable .selection-area {
+.plottable .selection-box-layer.x-movable.y-movable .selection-area {
   cursor: move;
 }
 
-.plottable .selection-box.x-resizable .selection-edge-lr {
+.plottable .selection-box-layer.x-resizable .selection-edge-lr {
   cursor: ew-resize;
 }
-.plottable .selection-box.y-resizable .selection-edge-tb {
+.plottable .selection-box-layer.y-resizable .selection-edge-tb {
   cursor: ns-resize;
 }
 
-.plottable .selection-box.x-resizable.y-resizable .selection-corner-tl {
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-tl {
   cursor: nwse-resize;
 }
-.plottable .selection-box.x-resizable.y-resizable .selection-corner-tr {
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-tr {
   cursor: nesw-resize;
 }
-.plottable .selection-box.x-resizable.y-resizable .selection-corner-bl {
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-bl {
   cursor: nesw-resize;
 }
-.plottable .selection-box.x-resizable.y-resizable .selection-corner-br {
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-br {
   cursor: nwse-resize;
 }
 /* /SelectionBox */
--- a/plottable.css
+++ b/plottable.css
@@ -120,8 +120,45 @@ svg.plottable {
   fill: aliceblue;
   opacity: .7;
   stroke: #C0C8FF;
-
 }
+
+/* SelectionBox */
+.plottable .selection-box .selection-area {
+  fill: black;
+  fill-opacity: 0.03;
+  stroke: #CCC;
+}
+
+.plottable .selection-box.x-movable .selection-area {
+  cursor: ew-resize;
+}
+.plottable .selection-box.y-movable .selection-area {
+  cursor: ns-resize;
+}
+.plottable .selection-box.x-movable.y-movable .selection-area {
+  cursor: move;
+}
+
+.plottable .selection-box.x-resizable .selection-edge-lr {
+  cursor: ew-resize;
+}
+.plottable .selection-box.y-resizable .selection-edge-tb {
+  cursor: ns-resize;
+}
+
+.plottable .selection-box.x-resizable.y-resizable .selection-corner-tl {
+  cursor: nwse-resize;
+}
+.plottable .selection-box.x-resizable.y-resizable .selection-corner-tr {
+  cursor: nesw-resize;
+}
+.plottable .selection-box.x-resizable.y-resizable .selection-corner-bl {
+  cursor: nesw-resize;
+}
+.plottable .selection-box.x-resizable.y-resizable .selection-corner-br {
+  cursor: nwse-resize;
+}
+/* /SelectionBox */
 
 .plottable .legend text {
   fill: #32313F;
--- a/plottable.css
+++ b/plottable.css
@@ -128,36 +128,6 @@ svg.plottable {
   fill-opacity: 0.03;
   stroke: #CCC;
 }
-
-.plottable .selection-box-layer.x-movable .selection-area {
-  cursor: ew-resize;
-}
-.plottable .selection-box-layer.y-movable .selection-area {
-  cursor: ns-resize;
-}
-.plottable .selection-box-layer.x-movable.y-movable .selection-area {
-  cursor: move;
-}
-
-.plottable .selection-box-layer.x-resizable .selection-edge-lr {
-  cursor: ew-resize;
-}
-.plottable .selection-box-layer.y-resizable .selection-edge-tb {
-  cursor: ns-resize;
-}
-
-.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-tl {
-  cursor: nwse-resize;
-}
-.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-tr {
-  cursor: nesw-resize;
-}
-.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-bl {
-  cursor: nesw-resize;
-}
-.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-br {
-  cursor: nwse-resize;
-}
 /* /SelectionBox */
 
 .plottable .legend text {
--- a/src/components/selectionBoxLayer.ts
+++ b/src/components/selectionBoxLayer.ts
@@ -5,7 +5,7 @@ export module Component {
   export class SelectionBoxLayer extends AbstractComponent {
     private _box: D3.Selection;
     private _boxArea: D3.Selection;
-    private _dismissed = true;
+    private _boxVisible = false;
     private _boxBounds: Bounds = {
       topLeft: { x: 0, y: 0 },
       bottomRight: { x: 0, y: 0 }
@@ -53,14 +53,12 @@ export module Component {
         topLeft: topLeft,
         bottomRight: bottomRight
       };
-
-      this._dismissed = false;
       this._render();
       return this;
     }
 
     public _doRender() {
-      if (!this._dismissed) {
+      if (this._boxVisible) {
         var t = this._boxBounds.topLeft.y;
         var b = this._boxBounds.bottomRight.y;
         var l = this._boxBounds.topLeft.x;
@@ -70,12 +68,32 @@ export module Component {
           x: l, y: t, width: r - l, height: b - t
         });
         this._content.node().appendChild(this._box.node());
+      } else {
+        this._box.remove();
       }
     }
 
-    public dismissBox() {
-      this._dismissed = true;
-      this._box.remove();
+    /**
+     * Gets whether the box is being shown.
+     *
+     * @return {boolean} Whether the box is showing.
+     */
+    public boxVisible(): boolean;
+    /**
+     * Shows or hides the selection box.
+     *
+     * @param {boolean} show Whether or not to show the box.
+     * @return {SelectionBoxLayer} The calling SelectionBoxLayer.
+     */
+    public boxVisible(show: boolean): SelectionBoxLayer;
+    public boxVisible(show?: boolean) {
+      if (show == null) {
+        return this._boxVisible;
+      }
+
+      this._boxVisible = show;
+      this._render();
+      return this;
     }
   }
 }
--- a/src/components/selectionBoxLayer.ts
+++ b/src/components/selectionBoxLayer.ts
@@ -5,20 +5,11 @@ export module Component {
   export class SelectionBoxLayer extends AbstractComponent {
     private _box: D3.Selection;
     private _boxArea: D3.Selection;
-    private _boxEdgeT: D3.Selection;
-    private _boxEdgeB: D3.Selection;
-    private _boxEdgeL: D3.Selection;
-    private _boxEdgeR: D3.Selection;
-    private _boxCornerTL: D3.Selection;
-    private _boxCornerTR: D3.Selection;
-    private _boxCornerBL: D3.Selection;
-    private _boxCornerBR: D3.Selection;
-
     private _boxBounds: Bounds = {
       topLeft: { x: 0, y: 0 },
       bottomRight: { x: 0, y: 0 }
     };
-    private _boxEdgeWidth = 2;
+
 
     constructor() {
       super();
@@ -30,27 +21,6 @@ export module Component {
 
       this._box = this._content.append("g").classed("selection-box", true).remove();
       this._boxArea = this._box.append("rect").classed("selection-area", true);
-
-      var createLine = () => this._box.append("line").style({
-                               "opacity": 0,
-                               "stroke": "pink",
-                               "stroke-width": this._boxEdgeWidth
-                             });
-      this._boxEdgeT = createLine().classed("selection-edge-tb", true);
-      this._boxEdgeB = createLine().classed("selection-edge-tb", true);
-      this._boxEdgeL = createLine().classed("selection-edge-lr", true);
-      this._boxEdgeR = createLine().classed("selection-edge-lr", true);
-
-      var createCorner = () => this._box.append("circle")
-                                   .attr("r", this._boxEdgeWidth / 2)
-                                   .style({
-                                     "opacity": 0,
-                                     "fill": "pink"
-                                   });
-      this._boxCornerTL = createCorner().classed("selection-corner-tl", true);
-      this._boxCornerTR = createCorner().classed("selection-corner-tr", true);
-      this._boxCornerBL = createCorner().classed("selection-corner-bl", true);
-      this._boxCornerBR = createCorner().classed("selection-corner-br", true);
     }
 
     /**
@@ -70,7 +40,21 @@ export module Component {
       if (newBounds == null) {
         return this._boxBounds;
       }
-      this._boxBounds = newBounds;
+
+      var topLeft: Point = {
+        x: Math.min(newBounds.topLeft.x, newBounds.bottomRight.x),
+        y: Math.min(newBounds.topLeft.y, newBounds.bottomRight.y)
+      };
+      var bottomRight: Point = {
+        x: Math.max(newBounds.topLeft.x, newBounds.bottomRight.x),
+        y: Math.max(newBounds.topLeft.y, newBounds.bottomRight.y)
+      };
+      this._boxBounds = {
+        topLeft: topLeft,
+        bottomRight: bottomRight
+      };
+
+      this._content.node().appendChild(this._box.node());
       this._render();
       return this;
     }
@@ -84,87 +68,6 @@ export module Component {
       this._boxArea.attr({
         x: l, y: t, width: r - l, height: b - t
       });
-
-      this._boxEdgeT.attr({
-        x1: l, y1: t, x2: r, y2: t
-      });
-      this._boxEdgeB.attr({
-        x1: l, y1: b, x2: r, y2: b
-      });
-      this._boxEdgeL.attr({
-        x1: l, y1: t, x2: l, y2: b
-      });
-      this._boxEdgeR.attr({
-        x1: r, y1: t, x2: r, y2: b
-      });
-
-      this._boxCornerTL.attr({ cx: l, cy: t });
-      this._boxCornerTR.attr({ cx: r, cy: t });
-      this._boxCornerBL.attr({ cx: l, cy: b });
-      this._boxCornerBR.attr({ cx: r, cy: b });
-
-      this._content.node().appendChild(this._box.node());
-    }
-
-    public getEdges(p: Point): String[] {
-      var edges: String[] = [];
-
-      var t = this._boxBounds.topLeft.y;
-      var b = this._boxBounds.bottomRight.y;
-      var l = this._boxBounds.topLeft.x;
-      var r = this._boxBounds.bottomRight.x;
-      var he = this._boxEdgeWidth / 2;
-
-      if (l - he <= p.x && p.x <= r + he) {
-        if (t - he <= p.y && p.y <= t + he) {
-          edges.push("top");
-        }
-        if (b - he <= p.y && p.y <= b + he) {
-          edges.push("bottom");
-        }
-      }
-
-      if (t - he <= p.y && p.y <= b + he) {
-        if (l - he <= p.x && p.x <= l + he) {
-          edges.push("left");
-        }
-        if (r - he <= p.x && p.x <= r + he) {
-          edges.push("right");
-        }
-      }
-
-      return edges;
-    }
-
-    /**
-     * Gets the edge width of the box.
-     *
-     * @return {number} The edge width of the box.
-     */
-    public edgeWidth(): number;
-    /**
-     * Sets the edge width of the box.
-     *
-     * @param {number} width The desired edge width.
-     * @return {SelectionBoxLayer} The calling SelectionBoxLayer.
-     */
-    public edgeWidth(width: number): SelectionBoxLayer;
-    public edgeWidth(width?: number): any {
-      if (width == null) {
-        return this._boxEdgeWidth;
-      }
-      if (width < 0) {
-        throw new Error("Edge width cannot be negative.");
-      }
-      this._boxEdgeWidth = width;
-      this._boxEdgeT.style("stroke-width", this._boxEdgeWidth);
-      this._boxEdgeB.style("stroke-width", this._boxEdgeWidth);
-      this._boxEdgeL.style("stroke-width", this._boxEdgeWidth);
-      this._boxEdgeR.style("stroke-width", this._boxEdgeWidth);
-      this._boxCornerTL.attr("r", this._boxEdgeWidth / 2);
-      this._boxCornerTR.attr("r", this._boxEdgeWidth / 2);
-      this._boxCornerBL.attr("r", this._boxEdgeWidth / 2);
-      this._boxCornerBR.attr("r", this._boxEdgeWidth / 2);
     }
 
     public dismissBox() {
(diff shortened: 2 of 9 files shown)


bluong:
In comparisons to the comments made in #1724, I'll make the assumption that they are two solutions to the same question and criticize accordingly.

tl;dr This seems to be the quick, sorta clean way for a plot to have a SelectionBoxLayer or so on top of it to draw a single SelectionBox in.

bluong:
The good thing about this route so far seems to be that it is correct and it is seemingly a clean way for us to add SelectionBoxes into Plottable.  Specific aspects that I like are that it is clean and decently understandable.

bluong:
However, the thing that seems a bit concerning is if the architectural implanting of a SelectionBox layer makes sense.  As stated in the other PR, having SelectionBoxes themselves be complete "Entities" seems like a more Plottable approach to the problem but at the same time, we don't have nice definitions for these "Entities" and how to use them effectively.

Also, the approach of having multiple SelectionBoxes in this approach is a bit not as clear as opposed to the other approach (Based on the approach you wrote here, this seems to be not as much of an issue.  It seems to be more of an API question than an implementation question)

jtlan:
Decided to go with #1724.

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+
+module Plottable {
+export module Component {
+  export class SelectionBoxLayer extends AbstractComponent {
+    private _box: D3.Selection;
+    private _core: D3.Selection;
+    private _edgeT: D3.Selection;
+    private _edgeB: D3.Selection;
+    private _edgeL: D3.Selection;
+    private _edgeR: D3.Selection;
```
bluong: Would edge selection arrays encapsulate this better?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+    private _cornerTL: D3.Selection;
+    private _cornerTR: D3.Selection;
+    private _cornerBL: D3.Selection;
+    private _cornerBR: D3.Selection;
```
bluong: Same as above for corner arrays

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    private _topLeft: Point = { x: 0, y: 0 };
+    private _bottomRight: Point = { x: 0, y: 0 };
```
bluong: NBD, how different of an answer is this to _cornerTL / _cornerBR information

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    private _detectionRadius = 2;
```
bluong: As said from the other PR, if we want to properly decouple interactions from the component itself, I'm unsure about having this variable here.  That said, we might just need to rename this variable.

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+      super._setup();
+
+      this._box = this._content.append("g").classed("selection-box-layer", true).remove();
+      this._core = this._box.append("rect").classed("selection-area", true);
+
+      var createLine = () => this._box.append("line").style({
+                               "opacity": 0,
+                               "stroke": "pink",
+                               "stroke-width": 2 * this._detectionRadius
+                             });
```
bluong: Why is the stroke pink?  Do we need a stroke here?

bluong: Does opacity 0 make the line invisible?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+      this._edgeB = createLine().classed("selection-edge-tb", true);
+      this._edgeL = createLine().classed("selection-edge-lr", true);
+      this._edgeR = createLine().classed("selection-edge-lr", true);
+
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._detectionRadius)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
```
bluong: Same as comments on line

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      var createLine = () => this._box.append("line").style({
+                               "opacity": 0,
+                               "stroke": "pink",
+                               "stroke-width": 2 * this._detectionRadius
+                             });
+      this._edgeT = createLine().classed("selection-edge-tb", true);
+      this._edgeB = createLine().classed("selection-edge-tb", true);
+      this._edgeL = createLine().classed("selection-edge-lr", true);
+      this._edgeR = createLine().classed("selection-edge-lr", true);
```
bluong: Not sure if we want to default to true here.  I think it should be default to false initially and then have the interactions set this to true.

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._detectionRadius)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
+      this._cornerTL = createCorner().classed("selection-corner-tl", true);
+      this._cornerTR = createCorner().classed("selection-corner-tr", true);
+      this._cornerBL = createCorner().classed("selection-corner-bl", true);
+      this._cornerBR = createCorner().classed("selection-corner-br", true);
```
bluong: Same as above about setting the classes to true at the interaction anchoring step

src/components/selectionBoxLayer.ts
bluong: The above instance variables I think should be renamed to probably be more inlined with the fact that this is a container that contains a single box.

this._topLeft is currently not too semantically correct for querying the optLeft of the selectionBox

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+      this._edgeL.attr({
+        x1: 0, y1: 0, x2: 0, y2: h
+      });
+      this._edgeR.attr({
+        x1: w, y1: 0, x2: w, y2: h
+      });
+
+      this._cornerTR.attr({ cx: w, cy: 0 });
+      this._cornerBL.attr({ cx: 0, cy: h });
+      this._cornerBR.attr({ cx: w, cy: h });
```
bluong: Is cx/cy default to 0,0?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,116 @@
...
+      this._detectionRadius = radius;
+      this._edgeT.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeB.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeL.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeR.style("stroke-width", 2 * this._detectionRadius);
+      this._cornerTL.attr("r", this._detectionRadius);
+      this._cornerTR.attr("r", this._detectionRadius);
+      this._cornerBL.attr("r", this._detectionRadius);
+      this._cornerBR.attr("r", this._detectionRadius);
+    }
```
bluong: As said above, we might need to think of a rename for this idea or try to see if we can more properly decouple the idea of interactions and components

src/components/selectionBoxLayer.ts
jtlan: No, since I need to address each of them individually. I could make a `Selection` or use an array and just assume the correct ordering, but that would make later code less clear as to what is actually happening.

src/components/selectionBoxLayer.ts
jtlan: I'm not sure what you mean...?

src/components/selectionBoxLayer.ts
jtlan: It needs to be invisible, but non-`hidden` so that cursor events will still occur over it.

src/components/selectionBoxLayer.ts
jtlan: That would require the `Interaction` to reach in and touch these edges. Semantically, all these classes do is identify the edges. The presence/absence of cursor is controlled by adding/removing the `x-resizable` and `y-resizable` classes to the main `Component`.

src/components/selectionBoxLayer.ts
jtlan: Yeah. I should probably set it on `_cornerTL` for completeness.

src/components/selectionBoxLayer.ts
bluong: I was curious on what's the difference between this _topLeft and _cornerTL.attr("x"), _cornerTL.attr("y")  .  The main reason why I ask is if we could cut down on some variables due to multiple sources of truth.

bluong: I'm also not sure of these instance variable names as well since they correlate with the actual box but not the layer

src/components/selectionBoxLayer.ts
bluong: Oh I see.  So this just applies a general class for users to style the box, but the stylings specific to interactions in the css file can be enabled by enabled that css class accordingly

src/components/selectionBoxLayer.ts
bluong: Up to you (Not setting it directly has a miniscule performance boost)

src/components/selectionBoxLayer.ts
jtlan: They're the upper-left and lower-right of the box, as expressed relative to the origin of the `Component`.

src/components/selectionBoxLayer.ts
jtlan: --> "_edgeWidth"

src/components/selectionBoxLayer.ts
jtlan: Well, for styling the box themselves they would style `selection-area`. This class allows them to style the cursor to be displayed on hover-over.

bluong: ^ Under the assumption that the interaction (like DragInteraction) handler has to class the edge with resizable on anchoring correct?

src/components/selectionBoxLayer.ts
bluong: Do we need the stroke to be pink?  Not that I'm against the color or anything :p but we should let the stroke be its default unless we have a reason for specifying the color

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    private _boxBounds: Bounds = {
+      topLeft: { x: 0, y: 0 },
+      bottomRight: { x: 0, y: 0 }
+    };
+    private _boxEdgeWidth = 2;
+
+    constructor() {
+      super();
+      this.classed("selection-box", true);
```
bluong: selection-box-layer correct?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._boxEdgeWidth / 2)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
+      this._boxCornerTL = createCorner().classed("selection-corner-tl", true);
+      this._boxCornerTR = createCorner().classed("selection-corner-tr", true);
+      this._boxCornerBL = createCorner().classed("selection-corner-bl", true);
+      this._boxCornerBR = createCorner().classed("selection-corner-br", true);
```
bluong: Not a huge deal, but do we want selection-stuff or selection-box-stff for these classes?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param {Bounds} newBounds The desired bounds of the box.
+     * @return {SelectionBoxLayer} The calling SelectionBoxLayer.
+     */
+    public bounds(newBounds: Bounds): SelectionBoxLayer;
+    public bounds(newBounds?: Bounds): any {
+      if (newBounds == null) {
+        return this._boxBounds;
+      }
+
```
bluong: NBD but I don't care for the line here

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
+      this._boxCornerTL = createCorner().classed("selection-corner-tl", true);
+      this._boxCornerTR = createCorner().classed("selection-corner-tr", true);
+      this._boxCornerBL = createCorner().classed("selection-corner-bl", true);
+      this._boxCornerBR = createCorner().classed("selection-corner-br", true);
+    }
+
+    public bounds(): Bounds;
```
bluong: Let's put the getter documentation

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+     * @param {number} width The desired edge width.
+     * @return {SelectionBoxLayer} The calling SelectionBoxLayer.
+     */
+    public edgeWidth(width: number): SelectionBoxLayer;
+    public edgeWidth(width?: number): any {
+      if (width == null) {
+        return this._boxEdgeWidth;
+      }
+      if (width < 0) {
+        throw new Error("Detection width cannot be negative.");
```
bluong: Probably "Edge width..."

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+      this._boxCornerBL.attr({ cx: l, cy: b });
+      this._boxCornerBR.attr({ cx: r, cy: b });
+
+      this._content.node().appendChild(this._box.node());
+    }
+
+    /**
+     * Gets the edge width of the box.
+     *
+     * @return {number}
```
bluong: Let's put some doc on the return

src/core/interfaces.ts
```diff
@@ -52,13 +52,6 @@ module Plottable {
     wantsHeight: boolean;
   }
 
-  export type _PixelArea = {
-    xMin: number;
-    xMax: number;
-    yMin: number;
-    yMax: number;
-  }
```
bluong: Oh was this just never used?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+      }
+      this._boxEdgeWidth = width;
+      this._boxEdgeT.style("stroke-width", this._boxEdgeWidth);
+      this._boxEdgeB.style("stroke-width", this._boxEdgeWidth);
+      this._boxEdgeL.style("stroke-width", this._boxEdgeWidth);
+      this._boxEdgeR.style("stroke-width", this._boxEdgeWidth);
+      this._boxCornerTL.attr("r", this._boxEdgeWidth / 2);
+      this._boxCornerTR.attr("r", this._boxEdgeWidth / 2);
+      this._boxCornerBL.attr("r", this._boxEdgeWidth / 2);
+      this._boxCornerBR.attr("r", this._boxEdgeWidth / 2);
```
bluong: Probably want to affect the stroke-width of the box itself as well

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+      this._boxEdgeT = createLine().classed("selection-edge-tb", true);
+      this._boxEdgeB = createLine().classed("selection-edge-tb", true);
+      this._boxEdgeL = createLine().classed("selection-edge-lr", true);
+      this._boxEdgeR = createLine().classed("selection-edge-lr", true);
+
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._boxEdgeWidth / 2)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
```
bluong: The "fill" "pink" seems unecessary.

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,141 @@
...
+    protected _setup() {
+      super._setup();
+
+      this._box = this._content.append("g").classed("selection-box", true).remove();
+      this._boxArea = this._box.append("rect").classed("selection-area", true);
+
+      var createLine = () => this._box.append("line").style({
+                               "opacity": 0,
+                               "stroke": "pink",
+                               "stroke-width": this._boxEdgeWidth
```
bluong: The "stroke" "pink" seems unecessary.

src/components/selectionBoxLayer.ts
jtlan: They're inside something classed `selection-box-layer`, so that should be enough. The class name is mostly to identify them anyway.

src/components/selectionBoxLayer.ts
jtlan: This controls only the detection area, not the actual visual width, which is set with CSS.

src/components/selectionBoxLayer.ts
jtlan: Want to avoid cases where a browser sets the default to `"none"` instead of `"#000000"`

src/components/selectionBoxLayer.ts
jtlan: see above.

src/components/selectionBoxLayer.ts
bluong: Considering that the API is named edgeWidth, the box itself visually should be affected no?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      return edges;
+    }
+
+    /**
+     * Gets the edge width of the box.
+     *
+     * @return {number} The edge width of the box.
+     */
+    public edgeWidth(): number;
```
bluong: To be consistent, boxEdgeWidth?

src/components/selectionBoxLayer.ts
bluong: Should we go to a standardized 000000 ?

src/components/selectionBoxLayer.ts
bluong: What do you think about this @crmorford ?

src/components/selectionBoxLayer.ts
jtlan: `"pink"` is shorter, and we can be cheeky about it.

"What's going on here?"
"Invisible pink ____"

jtlan:
Going to move the `Interaction`-related logic to a sublcass.

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,78 @@
...
+export module Component {
+  export class SelectionBoxLayer extends AbstractComponent {
+    private _box: D3.Selection;
+    private _boxArea: D3.Selection;
+    private _boxBounds: Bounds = {
+      topLeft: { x: 0, y: 0 },
+      bottomRight: { x: 0, y: 0 }
+    };
+
+
```
bluong: Remove doublenewline

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,78 @@
...
+    constructor() {
+      super();
+      this.classed("selection-box-layer", true);
+    }
+
+    protected _setup() {
+      super._setup();
+
+      this._box = this._content.append("g").classed("selection-box", true).remove();
```
bluong: Why is the box appended then removed?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,78 @@
...
+      var bottomRight: Point = {
+        x: Math.max(newBounds.topLeft.x, newBounds.bottomRight.x),
+        y: Math.max(newBounds.topLeft.y, newBounds.bottomRight.y)
+      };
+      this._boxBounds = {
+        topLeft: topLeft,
+        bottomRight: bottomRight
+      };
+
+      this._content.node().appendChild(this._box.node());
```
bluong: This call seems a bit odd.  Mainly because if you call the bounds setter multiple times, it seems to call appendChild successive times?

test/components/selectionBoxLayerTests.ts
```diff
@@ -0,0 +1,72 @@
...
+      x: 100,
+      y: 100
+    };
+    var bottomRight: Plottable.Point = {
+      x: 300,
+      y: 300
+    };
+
+    function assertCorrectRendering(expectedTL: Plottable.Point, expectedBR: Plottable.Point, msg: string) {
+      var selectionBox = svg.select(".selection-box");
```
bluong: Should the selection-box class be variableized?

test/components/selectionBoxLayerTests.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    svg.remove();
+  });
+
+  it("dismiss()", () => {
+    var svg = generateSVG();
+    var sbl = new Plottable.Component.SelectionBoxLayer();
+    sbl.renderTo(svg);
+
+    var selectionBox = svg.select(".selection-box");
```
bluong: Same as above

test/components/selectionBoxLayerTests.ts
```diff
@@ -0,0 +1,72 @@
...
+    assert.isTrue(selectionBox.empty(), "initilizes without box in DOM");
```
bluong: initializes

Also, do we necessarily want this behavior?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,78 @@
...
+      var r = this._boxBounds.bottomRight.x;
+
+      this._boxArea.attr({
+        x: l, y: t, width: r - l, height: b - t
+      });
+    }
+
+    public dismissBox() {
+      this._box.remove();
+    }
```
bluong: If we have a dismiss, should we not also have a command to bring the box back?

bluong:
The cleanup was a lot, but I think it's heading toward a cleaner direction.

src/components/selectionBoxLayer.ts
jtlan: So the box doesn't show until someone sets the bounds.

src/components/selectionBoxLayer.ts
jtlan: Calling `appendChild()` with a node that is already appended doesn't cause any ill effects (it appears to be idempotent).

test/components/selectionBoxLayerTests.ts
jtlan: By default, the box shouldn't show until the user specifies the bounds. That said, we should make sure that `bounds()` can be called before rendering.

src/components/selectionBoxLayer.ts
jtlan: `bounds()`? You can use `box.bounds(box.bounds()) to force it to show again.

test/components/selectionBoxLayerTests.ts
jtlan: Not sure it's that important, I mostly added it for legibility in the DOM inspector.

plottable.css
```diff
@@ -120,8 +120,45 @@ svg.plottable {
...
+}
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-tr {
+  cursor: nesw-resize;
+}
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-bl {
+  cursor: nesw-resize;
+}
+.plottable .selection-box-layer.x-resizable.y-resizable .selection-corner-br {
+  cursor: nwse-resize;
 }
```
bluong: Do we still want these css classes since we're moving this functionality?

plottable.css
```diff
@@ -120,8 +120,45 @@ svg.plottable {
...
   opacity: .7;
   stroke: #C0C8FF;
+}
+
+/* SelectionBox */
+.plottable .selection-box-layer .selection-area {
+  fill: black;
+  fill-opacity: 0.03;
+  stroke: #CCC;
+}
```
bluong: Want to sync a designer to see what the selection box should look like?  I'm unsure about a gray? box being the selection area

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,82 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Component {
+  export class SelectionBoxLayer extends AbstractComponent {
+    private _box: D3.Selection;
+    private _boxArea: D3.Selection;
+    private _dismissed = true;
```
bluong: Since the boxBounds are set to 0,0 -> 0,0, should it be _dismissed initially?

src/components/selectionBoxLayer.ts
bluong: Why can't we just have the _box just be appended?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,82 @@
...
+          x: l, y: t, width: r - l, height: b - t
+        });
+        this._content.node().appendChild(this._box.node());
+      }
+    }
+
+    public dismissBox() {
+      this._dismissed = true;
+      this._box.remove();
+    }
```
bluong: I still think we should have a parallel in dismissBox for showing the box

bluong: Also, does it make sense that this can only be called after render?

bluong: Under the assumption that the above 2 statements are true/make sense, I think we could move this._content.node().appendChild(this._box.node()); to not be done in the render step but either in the setup or this other API point

test/components/selectionBoxLayerTests.ts
bluong: Since we have initial bounds at 0,0 -> 0,0 why can't we just show a box that has those dimensions instead of waiting for bounds to be called?

plottable.css
jtlan: This was what Design mentioned when I talked to them. I'll set up a second sync.

plottable.css
jtlan: Good catch. I'll move it to the next PR.

src/components/selectionBoxLayer.ts
jtlan: Its initial dismiss status is independent of the bounds.

src/components/selectionBoxLayer.ts
jtlan: You can call this before anchoring because `remove()` is also idempotent.

bluong: However, this._box is initialized at the setup state.  Before then, it is null.  Thus, you'll run into a NPE right?

src/components/selectionBoxLayer.ts
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public bounds(): Bounds;
+    /**
+     * Sets the bounds of the box, and draws the box.
+     *
+     * @param {Bounds} newBounds The desired bounds of the box.
+     * @return {SelectionBoxLayer} The calling SelectionBoxLayer.
+     */
+    public bounds(newBounds: Bounds): SelectionBoxLayer;
+    public bounds(newBounds?: Bounds): any {
```
bluong: bounds seems to be doing a bit too much to me.  It for one sets the bounds of the box, and it also makes the box appear.

Can we have bounds just set the bounds value and have a separate call to make _dismissed true?

src/components/selectionBoxLayer.ts
bluong: Can we have the dismiss status be completely independent of the bounds?

jtlan: It always is.

src/components/selectionBoxLayer.ts
jtlan: Then it would be in the DOM, and therefore selectable.

src/components/selectionBoxLayer.ts
jtlan: Good catch. I'll move this to the render call.

test/components/selectionBoxLayerTests.ts
```diff
@@ -0,0 +1,70 @@
...
+      topLeft: topLeft,
+      bottomRight: bottomRight
+    }), Error, "can set bounds before anchoring");
+
+    sbl.boxVisible(true);
+    sbl.renderTo(svg);
+
+    function assertCorrectRendering(expectedTL: Plottable.Point, expectedBR: Plottable.Point, msg: string) {
+      var selectionBox = svg.select(".selection-box");
+      var bbox = (<SVGLocatable> <any> selectionBox.node()).getBBox();
```
bluong: Curious about the casting here?  Is there a way we can remove the anycast?

jtlan: No, you can't cast `Element` (the return type of `node()`) directly to `SVGLocatable`.

PR #1724, opened 2015-03-09
[WIP] selection-box implementation using Entity.

jtlan:
**DO NOT MERGE**

This implementation creates a new module, `Entity`. Entities have a visual representation on the screen but do not perform layout the way `Component`s do.

Compare to implementation in #1723

### DIFF
--- a/src/components/selectionBoxContainer.ts
+++ b/src/components/selectionBoxContainer.ts
@@ -1,0 +1,28 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Component {
+  export class SelectionBoxContainer extends AbstractComponent {
+    private _selectionBoxes: Entity.SelectionBox[] = [];
+
+    constructor() {
+      super();
+      this.classed("selection-box-container", true);
+    }
+
+    protected _setup() {
+      super._setup();
+      this._selectionBoxes.push(new Plottable.Entity.SelectionBox(this._content));
+    }
+
+    public getSelectionBoxes() {
+      return this._selectionBoxes;
+    }
+
+    public dismissAll() {
+      this._selectionBoxes.forEach((b) => b.dismiss());
+      return this;
+    }
+  }
+}
+}
--- a/src/entities/selectionBox.ts
+++ b/src/entities/selectionBox.ts
@@ -1,0 +1,125 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Entity {
+  export class SelectionBox {
+    private _parentNode: Element;
+    private _box: D3.Selection;
+    private _core: D3.Selection;
+    private _edgeT: D3.Selection;
+    private _edgeB: D3.Selection;
+    private _edgeL: D3.Selection;
+    private _edgeR: D3.Selection;
+    private _cornerTL: D3.Selection;
+    private _cornerTR: D3.Selection;
+    private _cornerBL: D3.Selection;
+    private _cornerBR: D3.Selection;
+
+    private _topLeft: Point = { x: 0, y: 0 };
+    private _bottomRight: Point = { x: 0, y: 0 };
+
+    private _detectionRadius = 2;
+
+    constructor(parent: D3.Selection) {
+      this._parentNode = parent.node();
+      this._box = parent.append("g").classed("selection-box", true).remove();
+      this._core = this._box.append("rect").classed("selection-area", true);
+
+      var createLine = () => this._box.append("line").style({
+                               "opacity": 0,
+                               "stroke": "pink",
+                               "stroke-width": 2 * this._detectionRadius
+                             });
+      this._edgeT = createLine().classed("selection-edge-tb", true);
+      this._edgeB = createLine().classed("selection-edge-tb", true);
+      this._edgeL = createLine().classed("selection-edge-lr", true);
+      this._edgeR = createLine().classed("selection-edge-lr", true);
+
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._detectionRadius)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
+      this._cornerTL = createCorner().classed("selection-corner-tl", true);
+      this._cornerTR = createCorner().classed("selection-corner-tr", true);
+      this._cornerBL = createCorner().classed("selection-corner-bl", true);
+      this._cornerBR = createCorner().classed("selection-corner-br", true);
+    }
+
+    public classed(cssClass: string): boolean;
+    public classed(cssClass: string, addClass: boolean): SelectionBox;
+    public classed(cssClass: string, addClass?: boolean): any {
+      if (addClass == null) {
+        if (cssClass == null) {
+          return false;
+        } else {
+          return this._box.classed(cssClass);
+        }
+      } else {
+        if (cssClass != null) {
+          this._box.classed(cssClass, addClass);
+        }
+        return this;
+      }
+    }
+
+    public setBounds(topLeft: Point, bottomRight: Point) {
+      this._topLeft = topLeft;
+      this._bottomRight = bottomRight;
+      var w = this._bottomRight.x - this._topLeft.x;
+      var h = this._bottomRight.y - this._topLeft.y;
+
+      this._core.attr({
+        width: w, height: h
+      });
+
+      this._edgeT.attr({
+        x1: 0, y1: 0, x2: w, y2: 0
+      });
+      this._edgeB.attr({
+        x1: 0, y1: h, x2: w, y2: h
+      });
+      this._edgeL.attr({
+        x1: 0, y1: 0, x2: 0, y2: h
+      });
+      this._edgeR.attr({
+        x1: w, y1: 0, x2: w, y2: h
+      });
+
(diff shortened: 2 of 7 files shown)


bluong:
I'm assuming that when you mean "Compare to implementation in #1723", you mean that this and the other pull request are two different solutions to solving how to do selection boxes?

bluong:
Given the above assumption, here's my 2 cents on the idea of Entities.  Based on the implementation of the SelecionBox container, one thing that seems interesting is that the SelectionBoxContainer class is near empty, which spurs the question of... do we actually need a SelectionBoxContainer class?
Specifically, would it be too far to say that we just need an empty cell that we can just stuff these entities in?

bluong:
Also, another thing to note from a semantic/philosophic standpoint, this idea seems to align more closely to the idea of Plottable being an idea of plopping a bunch of objects together to make a chart... under the assumption that this is done consistently and cleanly.  We should verify the bounds / definitions of entities if we want to go with this idea.

bluong:
Essentially, this seems like an interesting idea that we want to delve into to see if this is the right direction, but if we go along with this idea, I am unsure about how much time it would take to nail down all the specifics of Entities

bluong:
Fun thought question.  Are components that don't care about the space they are in... `Entities`?  They don't care about the space they are in / don't perform layout maneuvers.

src/components/selectionBoxContainer.ts
```diff
@@ -0,0 +1,28 @@
...
+      return this._selectionBoxes;
+    }
+
+    public dismissAll() {
+      this._selectionBoxes.forEach((b) => b.dismiss());
+      return this;
+    }
+  }
+}
+}
```
bluong: Docs Docs Docs

src/components/selectionBoxContainer.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    protected _setup() {
+      super._setup();
+      this._selectionBoxes.push(new Plottable.Entity.SelectionBox(this._content));
+    }
+
+    public getSelectionBoxes() {
+      return this._selectionBoxes;
+    }
```
bluong: Super future consideration - getEntities residing on a Component level?

src/components/selectionBoxContainer.ts
```diff
@@ -0,0 +1,28 @@
...
+    private _selectionBoxes: Entity.SelectionBox[] = [];
+
+    constructor() {
+      super();
+      this.classed("selection-box-container", true);
+    }
+
+    protected _setup() {
+      super._setup();
+      this._selectionBoxes.push(new Plottable.Entity.SelectionBox(this._content));
```
bluong: Hmm, this._content seems to be protected...  Is using this._content appropriate?

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    private _topLeft: Point = { x: 0, y: 0 };
+    private _bottomRight: Point = { x: 0, y: 0 };
+
+    private _detectionRadius = 2;
+
+    constructor(parent: D3.Selection) {
+      this._parentNode = parent.node();
+      this._box = parent.append("g").classed("selection-box", true).remove();
+      this._core = this._box.append("rect").classed("selection-area", true);
```
bluong: _core seems like a vague name...  Does it describe the selectionBox itself?

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+    private _edgeR: D3.Selection;
+    private _cornerTL: D3.Selection;
+    private _cornerTR: D3.Selection;
+    private _cornerBL: D3.Selection;
+    private _cornerBR: D3.Selection;
+
+    private _topLeft: Point = { x: 0, y: 0 };
+    private _bottomRight: Point = { x: 0, y: 0 };
+
+    private _detectionRadius = 2;
```
bluong: Going as much as we can about decoupling interactions and visual items, is an idea of a _detectionRadius digging a bit far into the realm of interaction logic?  I'm thinking a rename here is all we need since this seems to be determining the stroke widths of the lines?

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+    constructor(parent: D3.Selection) {
+      this._parentNode = parent.node();
+      this._box = parent.append("g").classed("selection-box", true).remove();
+      this._core = this._box.append("rect").classed("selection-area", true);
+
+      var createLine = () => this._box.append("line").style({
+                               "opacity": 0,
+                               "stroke": "pink",
+                               "stroke-width": 2 * this._detectionRadius
+                             });
```
bluong: pink seems like an odd color

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+module Plottable {
+export module Entity {
+  export class SelectionBox {
+    private _parentNode: Element;
+    private _box: D3.Selection;
+    private _core: D3.Selection;
+    private _edgeT: D3.Selection;
+    private _edgeB: D3.Selection;
+    private _edgeL: D3.Selection;
+    private _edgeR: D3.Selection;
```
bluong: Do we want to define each edge or have edges in a SelectionArray

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+    private _cornerTL: D3.Selection;
+    private _cornerTR: D3.Selection;
+    private _cornerBL: D3.Selection;
+    private _cornerBR: D3.Selection;
```
bluong: Same as above.

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+      var createCorner = () => this._box.append("circle")
+                                   .attr("r", this._detectionRadius)
+                                   .style({
+                                     "opacity": 0,
+                                     "fill": "pink"
+                                   });
+      this._cornerTL = createCorner().classed("selection-corner-tl", true);
+      this._cornerTR = createCorner().classed("selection-corner-tr", true);
+      this._cornerBL = createCorner().classed("selection-corner-bl", true);
+      this._cornerBR = createCorner().classed("selection-corner-br", true);
```
bluong: Do we want these classes to be on by default?  I'm thinking that it is switched off by default and DragInteraction for example would turn these on

src/entities/selectionBox.ts
bluong: Does opacity 0 make it invisible?  If so, why do we need stroke?

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return this;
+      }
+    }
+
+    public setBounds(topLeft: Point, bottomRight: Point) {
+      this._topLeft = topLeft;
+      this._bottomRight = bottomRight;
+      var w = this._bottomRight.x - this._topLeft.x;
+      var h = this._bottomRight.y - this._topLeft.y;
```
bluong: Do we want to do any validations (Math.abs stuff)

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+      this._edgeR.attr({
+        x1: w, y1: 0, x2: w, y2: h
+      });
+
+      this._cornerTR.attr({ cx: w, cy: 0 });
+      this._cornerBL.attr({ cx: 0, cy: h });
+      this._cornerBR.attr({ cx: w, cy: h });
+
+      this._box.attr("transform", "translate("+this._topLeft.x+", "+this._topLeft.y+")");
```
bluong: Let's space out the concatentation

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+      this._edgeL.attr({
+        x1: 0, y1: 0, x2: 0, y2: h
+      });
+      this._edgeR.attr({
+        x1: w, y1: 0, x2: w, y2: h
+      });
+
+      this._cornerTR.attr({ cx: w, cy: 0 });
+      this._cornerBL.attr({ cx: 0, cy: h });
+      this._cornerBR.attr({ cx: w, cy: h });
```
bluong: Are attrs 0 by default?

src/entities/selectionBox.ts
```diff
@@ -0,0 +1,125 @@
...
+      this._detectionRadius = radius;
+      this._edgeT.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeB.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeL.style("stroke-width", 2 * this._detectionRadius);
+      this._edgeR.style("stroke-width", 2 * this._detectionRadius);
+      this._cornerTL.attr("r", this._detectionRadius);
+      this._cornerTR.attr("r", this._detectionRadius);
+      this._cornerBL.attr("r", this._detectionRadius);
+      this._cornerBR.attr("r", this._detectionRadius);
+    }
```
bluong: I would prefer that this idea be either on the interaction level or in an abstract layer inbetween.  Possibly renaming is usable here too...

src/components/selectionBoxContainer.ts
jtlan: That's why it's [WIP] :/

PR #1731, opened 2015-03-10
Add a <rect> to Component background so it can be colored.

### LINKED ISSUE #1725: Add a rect element inside of render-area to allow styling background
Currently it's practically impossible to style the background content of a chart (inside the axis). Inside of .render-area we could add a rect element to make this much easier.

Below is the area we would like to have control over.

image

jtlan:
Select and color using:
```CSS
.plottable .background-fill {
  fill: pink;
}
```

Additional selectors can be inserted before "`.background-fill`" to narrow coloring to specific `Component`s.

Close #1725.

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -43,6 +43,10 @@ svg.plottable {
   display : block; /* SVGs must be block elements for width/height calculations to work in Firefox. */
 }
 
+.plottable .background-fill {
+  fill: none;
+}
+
 .plottable .bounding-box {
   /* Invisible pink bounding-box to allow for collision testing */
   fill: pink;
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -78,6 +78,11 @@ export module Component {
       this._cssClasses = null;
 
       this._backgroundContainer = this._element.append("g").classed("background-container", true);
+      this._boxes.push(
+        this._backgroundContainer.append("rect")
+                                 .classed("background-fill", true)
+                                 .attr("width", this.width()).attr("height", this.height())
+      );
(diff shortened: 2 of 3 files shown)


plottable.css
```diff
@@ -43,6 +43,10 @@ svg.plottable {
   display : block; /* SVGs must be block elements for width/height calculations to work in Firefox. */
 }
 
+.plottable .background-fill {
```
bluong: We should probably think of what to name this background rectangle as it easily has more capability can just coloring the fill

src/components/abstractComponent.ts
```diff
@@ -78,6 +78,11 @@ export module Component {
       this._cssClasses = null;
 
       this._backgroundContainer = this._element.append("g").classed("background-container", true);
+      this._boxes.push(
+        this._backgroundContainer.append("rect")
+                                 .classed("background-fill", true)
+                                 .attr("width", this.width()).attr("height", this.height())
```
bluong: Should we add a foreground rectangle?

bluong: Would using _addBox here work?

plottable.css
jtlan: I guess? Affecting the stroke seems like a mistake though.

bluong:
Just the naming is left.

crmorford:
.background-fill sounds fine

crmorford:
http://jsfiddle.net/n3swy1tf/

jtlan:
Picture for the release:

screen shot 2015-03-12 at 8 47 36 pm

PR #1732, opened 2015-03-10
Add content() getter on Component.

jtlan:

PR #1733, opened 2015-03-11
Refactor Interaction.Drag

jtlan:
Now uses `Dispatcher`.

Removed subclasses, since we'll be adding that functionality using `SelectionBox`.

`Dispatcher.Touch` to be added to `Interaction.Drag` shortly.


bluong:
Just noting that the first thing I noticed is a net -1370 lines of code.  Does this still retain the same functionality?  If so... nice code cleaning! :+1:

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+    private _mouseDispatcher: Plottable.Dispatcher.Mouse;
+    private _dragOrigin: Point;
+    private _dragStartCallback: (p: Point) => any;
+    private _dragCallback: (start: Point, end: Point) => any;
+    private _dragEndCallback: (start: Point, end: Point) => any;
+
+    public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
+      super._anchor(component, hitBox);
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(
+                                <SVGElement> (<any> this._componentToListenTo)._element.node()
```
bluong: Based on prior discussions, should this be on content instead?  I'm a tiny bit concerned at the any casting, but I'm just curious on why the attachment is done at the element level

jtlan:
> Just noting that the first thing I noticed is a net -1370 lines of code. Does this still retain the same functionality?

Sadly, no, because it removes the `Dragbox` classes and descendants...

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+      this._mouseDispatcher.onMouseMove("Interaction.Drag" + this.getID(),
+        (p: Point, e: MouseEvent) => this._doDrag(p, e));
+      this._mouseDispatcher.onMouseUp("Interaction.Drag" + this.getID(),
+        (p: Point, e: MouseEvent) => this._endDrag(p, e));
+    }
+
+    private _translateAndConstrain(p: Point) {
+      var translatedP = this._translateToComponentSpace(p);
+      var constrainedX = Math.min(Math.max(0, translatedP.x), this._componentToListenTo.width());
+      var constrainedY = Math.min(Math.max(0, translatedP.y), this._componentToListenTo.height());
```
bluong: Did I place a clamp call in utils at this point?  If not, we might want to move this to a Util as this may be used elsewhere

As a reference, I wrote the util as
clamp(value, a, b)
  var minValue = min(a, b)
  var maxValue = max(a, b)
  if value is less than min, return min
  if value is greater than max, return max
  return value

The way you have it also I think should work / is clean

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    private _startDrag(p: Point, e: MouseEvent) {
+      var translatedP = this._translateToComponentSpace(p);
+      if (this._isInsideComponent(translatedP)) {
+        e.preventDefault();
+        this._dragging = true;
+        this._dragOrigin = translatedP;
+        if (this._dragStartCallback) {
+          this._dragStartCallback(this._dragOrigin);
+        }
```
bluong: I am unsure about what we did for the other interactions...

But would it be cleaner to have this._dragStartCallback initialize to an empty function?  It would remove the verification check

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+
+    private _doDrag(p: Point, e: MouseEvent) {
+      if (this._dragging) {
+        if (this._dragCallback) {
```
bluong: Same as above

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+          var constrainedP = this._translateAndConstrain(p);
+          this._dragCallback(this._dragOrigin, constrainedP);
+        }
+      }
+    }
+
+    private _endDrag(p: Point, e: MouseEvent) {
+      if (this._dragging) {
+        this._dragging = false;
+        if (this._dragEndCallback) {
```
bluong: Same as above

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+        }
+      }
+    }
+
+    /**
+     * Gets the callback that is called when dragging starts.
+     *
+     * @returns {(start: Point) => void} The callback called when dragging starts.
+     */
+    public onDragStart(): (start: Point) => void;
```
bluong: void?

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+        return this;
+      }
+    }
+
+    /**
+     * Gets the callback that is called during dragging.
+     *
+     * @returns {(start: Point, end: Point) => void} The callback called during dragging.
+     */
+    public onDrag(): (start: Point, end: Point) => void;
```
bluong: void -> any?

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,134 @@
...
+        return this;
+      }
+    }
+
+    /**
+     * Gets the callback that is called when dragging ends.
+     *
+     * @returns {(start: Point, end: Point) => void} The callback called when dragging ends.
+     */
+    public onDragEnd(): (start: Point, end: Point) => void;
```
bluong: void -> any

test/testUtils.ts
```diff
@@ -16,39 +16,54 @@ function getSVGParent(): D3.Selection {
...
+function triggerFakeMouseEvent(type: string, target: D3.Selection, relativeX: number, relativeY: number) {
+  var clientRect = target.node().getBoundingClientRect();
+  var xPos = clientRect.left + relativeX;
+  var yPos = clientRect.top + relativeY;
+  var e = <MouseEvent> document.createEvent("MouseEvents");
+  e.initMouseEvent(type, true, true, window, 1,
+                    xPos, yPos,
+                    xPos, yPos,
+                    false, false, false, false,
+                    1, null);
```
bluong: Augh all this setting looks painful.  Is it possible to take advantage of defaults / optionals to not set all of this manually?

test/testUtils.ts
```diff
@@ -16,39 +16,54 @@ function getSVGParent(): D3.Selection {
...
+  fakeTouchList.item = (index: number) => fakeTouchList[index];
+  e.touches = <TouchList> fakeTouchList;
+  e.targetTouches = <TouchList> fakeTouchList;
+  e.changedTouches = <TouchList> fakeTouchList;
+
+  e.altKey = false;
+  e.metaKey = false;
+  e.ctrlKey = false;
+  e.shiftKey = false;
+  target.node().dispatchEvent(e);
```
bluong: Are these just copy+paste / unchanged items from old code that git thinks is a diff?  If so... git diff just highly confused me...

test/interactions/dragInteractionTests.ts
```diff
@@ -0,0 +1,140 @@
...
+      y: SVG_HEIGHT
+    };
+    var outsidePointNeg = {
+      x: -SVG_WIDTH / 2,
+      y: -SVG_HEIGHT / 2
+    };
+    var constrainedNeg = {
+      x: 0,
+      y: 0
+    };
```
bluong: If we have this constrain idea as a util, you could use the util instead of needing to define another var

bluong:
Aww, but at the same time, I don't see the Dragbox code etc to be even close to a 1000+ lines added

test/interactions/dragInteractionTests.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      var hitbox = c.hitBox();
+      triggerFakeMouseEvent("mousedown", hitbox, startPoint.x, startPoint.y);
+      assert.isTrue(startCallbackCalled, "callback was called on beginning drag (mousedown)");
+      assert.deepEqual(receivedStart, startPoint, "was passed the correct point");
+
+      startCallbackCalled = false;
+      triggerFakeMouseEvent("mousedown", hitbox, outsidePointPos.x, outsidePointPos.y);
+      assert.isFalse(startCallbackCalled, "does not trigger callback if drag starts outside the Component (positive) (mousedown)");
+      triggerFakeMouseEvent("mousedown", hitbox, outsidePointNeg.x, outsidePointNeg.y);
```
bluong: Not sure if you need these outsidePointPos/Neg variables.  You could add a width / subtract a width or so instead and this would lose a variable

test/interactions/dragInteractionTests.ts
```diff
@@ -0,0 +1,140 @@
...
+      assert.isFalse(startCallbackCalled, "does not trigger callback if drag starts outside the Component (negative) (mousedown)");
+
+      assert.strictEqual(drag.onDragStart(), startCallback, "retrieves the callback if called with no arguments");
+      drag.onDragStart(null);
+      assert.isNull(drag.onDragStart(), "removes the callback if called with null");
```
bluong: Do we want this capability

test/interactions/dragInteractionTests.ts
```diff
@@ -0,0 +1,140 @@
...
+      triggerFakeMouseEvent("mousemove", hitbox, outsidePointPos.x, outsidePointPos.y);
+      assert.deepEqual(receivedEnd, constrainedPos, "dragging outside the Component is constrained (positive) (mousemove)");
+      triggerFakeMouseEvent("mousemove", hitbox, outsidePointNeg.x, outsidePointNeg.y);
+      assert.deepEqual(receivedEnd, constrainedNeg, "dragging outside the Component is constrained (negative) (mousemove)");
+
+      assert.strictEqual(drag.onDrag(), moveCallback, "retrieves the callback if called with no arguments");
+      drag.onDrag(null);
+      assert.isNull(drag.onDrag(), "removes the callback if called with null");
+      svg.remove();
+    });
```
bluong: Same statements as onDragStart

test/interactions/dragInteractionTests.ts
```diff
@@ -0,0 +1,140 @@
...
+      assert.deepEqual(receivedEnd, constrainedPos, "dragging outside the Component is constrained (positive) (mouseup)");
+      triggerFakeMouseEvent("mousedown", hitbox, startPoint.x, startPoint.y);
+      triggerFakeMouseEvent("mouseup", hitbox, outsidePointNeg.x, outsidePointNeg.y);
+      assert.deepEqual(receivedEnd, constrainedNeg, "dragging outside the Component is constrained (negative) (mouseup)");
+
+      assert.strictEqual(drag.onDragEnd(), endCallback, "retrieves the callback if called with no arguments");
+      drag.onDragEnd(null);
+      assert.isNull(drag.onDragEnd(), "removes the callback if called with null");
+      svg.remove();
+    });
```
bluong: same statements as onDragStart

bluong:
Mainly some code cleanup issues.  Looks to be functionally correct.

src/interactions/dragInteraction.ts
jtlan: It actually bubbles up to the `<svg>`, so it doesn't matter. Anything will do.

bluong: Would it make sense semantically to not use _element but to use like content?  I would like to move away from using any casting but not for the tradeoff of loss in semantics

src/interactions/dragInteraction.ts
jtlan: Wouldn't work if someone tried to remove the callback by passing in `null`.

src/interactions/dragInteraction.ts
jtlan: Yeah, returning `any` is better. Nice catch.

test/testUtils.ts
jtlan: I think you have to pass them. THis is code that already existed, I just moved it up.

test/interactions/dragInteractionTests.ts
jtlan: I think it's useful, if the `Interction`'s behavior needs to be modified.

src/interactions/dragInteraction.ts
bluong: What's the difference between removing a callback and using a callback that does nothing

src/interactions/dragInteraction.ts
jtlan: `_content` is also protected, although we can use `content()` after #1732 merges.

src/interactions/dragInteraction.ts
jtlan: Well, all it would do is move the `null`-check to the assignment methods; we'd have to restore a nothing-callback if someone passed in `null` to `onDragStart`, for example.

bluong: I'm thinking of not allowing `null` to be used as an input argument so that we don't need to do any of these checks

jtlan: How do you remove the callback then?

src/interactions/dragInteraction.ts
bluong: If you think that we can use content, can you place a HACKHACK about the anycasting referencing #1732 ?

src/interactions/dragInteraction.ts
bluong: Deciding to stick with allowing null being set as a callback.  Will look back if this becomes an issue

src/interactions/dragInteraction.ts
```diff
@@ -0,0 +1,132 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Interaction {
+  export class Drag extends AbstractInteraction {
+    private _dragging = false;
+    private _mouseDispatcher: Plottable.Dispatcher.Mouse;
```
bluong: Do we plan on adding touchDispatcher in this side branch?  If so, add a TODO

bluong:
Probably should update the sidebranch and push it to get a better diff

PR #1737, opened 2015-03-13
Legend no longer adds classes to entries based on their data.

### LINKED ISSUE #1321: Sanitize applied CSS class names in Legend and HorizontalLegend

jtlan:
Instead, users should do it themselves.

Close #1321.


jtlan:
```Javascript
legend.content().selectAll("." + Plottable.Component.Legend.LEGEND_ENTRY_CLASS)
                .each(function(d) {
                  d3.select(this).classed(mySanitizingFunction(d), true);
                });
```

jtlan:
@crmorford, can you make sure the supplied code can be used to effect the previous functionality?

PR #1738, opened 2015-03-13
[Hotfix][SymbolGenerators] - d3 symbol type verification

bluong:
Throws an error if an invalid symbol type is given.  Default sizeFactor (only will be used after the verification)

### DIFF
--- a/src/utils/symbolGenerators.ts
+++ b/src/utils/symbolGenerators.ts
@@ -52,6 +52,9 @@ module Plottable {
           case "triangle-down":
             sizeFactor = Math.sqrt(3);
             break;
+          default:
+            sizeFactor = 1;
+            break;
         }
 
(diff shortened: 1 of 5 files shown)


src/utils/symbolGenerators.ts
```diff
@@ -52,14 +52,24 @@ module Plottable {
...
             break;
+          default:
+            sizeFactor = 1;
+            break;
         }
 
         return sizeFactor * Math.pow(SYMBOL_GENERATOR_RADIUS, 2);
       };
 
+      function ensureSymbolType(symTypeString: string) {
```
jtlan: Should this be static?

bluong: It's specific to d3.  We could make it static, but it's mainly only useful for this single symbolGeneartor

PR #1742, opened 2015-03-17
[LinePlot] - getAllPlotData retrieves 1 line node per dataset

bluong:
getAllPlotData loops through each datum to retrieve a selection, but linePlot just needs to loop through each dataset to get the selection


jtlan:
Before, did it just list the same `<line>` element more than once in the `Selection`? This change makes the behavior inconsistent with the one-to-one `Plot`s, but I think the new behavior makes more sense.

bluong:
Yes.

Due to the fact that LinePlots is overriding the behavior of getAllPlotData, it sure makes it inconsistent with one-to-one `Plots`.  However, with the old implementation, you would get 3 line element nodes when calling getAllPlotData on a linePlot with 3 datum points which doesn't make all that much sense. (selection.size() === 3)

PR #1743, opened 2015-03-17
[Plot] - getClosestPlotData()

bluong:
Using the simplistic idea that each datum is associated with a single point, getClosestPlotData will retrieve the closest datum point using this information.  Note that no knowledge of shapes is used to get a simplistic implementation.  If users want to implement their own closestSelection algorithm, they can use generateProjectors to aid them


src/components/plots/abstractPlot.ts
```diff
@@ -510,6 +510,44 @@ export module Plot {
...
+      var plotData = this.getAllPlotData(datasetKeys);
+      plotData.pixelPoints.forEach((pixelPoint: Point, index: number) => {
+        var distance = _Util.Methods.pointDistance(pixelPoint, queryPoint);
+        if (distance < closestDistance) {
+          closestDistance = distance;
+          closestIndex = index;
+        }
+      });
+
+      return {data: [plotData.data[closestIndex]],
```
jtlan: Will fail if closestIndex is `null`, for example if `withinValue` is too small.

src/components/plots/linePlot.ts
```diff
@@ -95,6 +95,30 @@ export module Plot {
       return ["x", "y"];
     }
 
+    protected _getClosestPlotData(queryPoint: Point, datasetKeys: string[], withinValue = Infinity) {
```
jtlan: How come `Plot.Line` needs a different implementation?

src/utils/utils.ts
```diff
@@ -325,6 +325,10 @@ export module _Util {
 
       return "#" + rHex + gHex + bHex;
     }
+
+    export function pointDistance(p1: Point, p2: Point) {
+      return Math.sqrt(Math.pow(p2.y - p1.y, 2) - Math.pow(p2.x - p1.x, 2));
```
jtlan: We could just use squared distances, saving some computation time. I also wonder if direct multiplcation is faster than `Math.pow()`; Can we test that?

src/components/plots/linePlot.ts
bluong: On normal plot there is a 1to1 mapping between an html element and a datum.  This is not true for line plots

test/components/plots/linePlotTests.ts
```diff
@@ -226,6 +226,25 @@ describe("Plots", () => {
...
+
+      it("retrieves correct data",() => {
+        var dataset3 = [
+          { foo: 0, bar: 1 },
+          { foo: 1, bar: 0.95 }
+        ];
+        linePlot.addDataset("d3", dataset3);
+
+        var lineData = linePlot.getClosestPlotData({x: 490, y: 300});
+        //HACKHACK selection count mirrors datum count
```
jtlan: Is this fixed by #1742? Tag it as such if so.

src/components/plots/linePlot.ts
jtlan: Couldn't you use this implementation for `AbstractPlot`?

bluong: Note that the implementation here uses plotData.selection to mark the closestSelection.  This does not work in the abstractPlot case since there could be multiple selections in plotData.selection whereas in linePlot there is only the line.  That is why the abstractPlot uses the closestIndex idea.

PR #1744, opened 2015-03-17
[MouseDispatcher] - Add in `wheel` event

### LINKED ISSUE #1735: Encapsulate 'wheel' event to a Dispatcher
Similar to #1662 this event should be rolled up into a Dispatcher to use for Interactions like ZoomInteraction

bluong:
For usage of interactions like PanZoomInteraction, the consumer may want to scroll (using the mouse wheel) the plot to zoom in / out or similar.  That interaction should be associated with a WheelDispatcher, which will be the dispatcher associated with that scrolling event.

Closes #1735


jtlan:
I think you'll want to add this to the `Mouse` dispatcher somehow. Knowing where the `wheel` event took place is actually important (for zooming in based on where the cursor is, for example).

bluong:
I'll try the appraoch of appending this event to the Mouse dispatcher

src/dispatchers/mouseDispatcher.ts
```diff
@@ -11,9 +11,11 @@ export module Dispatcher {
     private _moveBroadcaster: Core.Broadcaster<Dispatcher.Mouse>;
     private _downBroadcaster: Core.Broadcaster<Dispatcher.Mouse>;
     private _upBroadcaster: Core.Broadcaster<Dispatcher.Mouse>;
+    private _wheelBraodcaster: Core.Broadcaster<Dispatcher.Mouse>;
```
bluong: whoops -.-;

jtlan:
Where's `onWheel`?

src/dispatchers/mouseDispatcher.ts
```diff
@@ -116,14 +123,31 @@ export module Dispatcher {
...
      * calls broadcast() on the supplied Broadcaster.
      */
-    private _measureAndBroadcast(e: MouseEvent, b: Core.Broadcaster<Dispatcher.Mouse>) {
+    private _measureAndBroadcast(e: MouseEvent, b: Core.Broadcaster<Dispatcher.Mouse>,
+                                 otherBroadcastData: any[] = []) {
       var newMousePosition = this.translator.computePosition(e.clientX, e.clientY);
       if (newMousePosition != null) {
         this._lastMousePosition = newMousePosition;
-        b.broadcast(this.getLastMousePosition(), e);
+        b.broadcast.apply(b, otherBroadcastData.concat(this.getLastMousePosition(), e));
```
jtlan: Kind of annoying that we have to re-`apply()` here.

bluong: The old implementation made the assumption that it will only ever broadcast the mousePosition, but there has to be some adjustment in order to broadcast any other information (like the changes in the deltaY of the mousewheel)

jtlan:
Nice job dodging all the testing potholes here.

jtlan:
Going to NOQE, since I suspect the unit tests already cover the ground that QE would cover (in particular, Sauce Labs covers browser-based issues).

PR #1745, opened 2015-03-17
[InterpolatedColorLegend] - symbolGenerator API

bluong:
As on Legend, adding the symbolGenerator API to InterpolatedColorLegend defaulting to a square symbol

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2469,6 +2469,19 @@ declare module Plottable {
             protected _setup(): void;
             _requestedSpace(offeredWidth: number, offeredHeight: number): _SpaceRequest;
             _doRender(): void;
+            /**
+             * Gets the SymbolGenerator of the interpolated color legend, which dictates how
+             * the symbol in each entry is drawn.
+             *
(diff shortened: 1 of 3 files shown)


jtlan:
... Do we need this feature? I don't think it makes sense on `InterpolatedColorLegend`.

crmorford:
I can't think of a use case

rcchen:
Agreed, I don't think it works in this context.

bluong:
Mainly thinking of adding this to be consistent to how Legends work.  That said, I'm not pushing this too hard, just thought it would be apt to do work on this now since Legends just got symbols.  Here's a JSFiddle for proof of concept.  If we still don't want this, I'll just close the PR.

http://jsfiddle.net/bluong63/ys5514na/

rcchen:
Oh that's actually kind of cool...except for the grey rectangle surrounding it.

bluong:
We could always remove the grey rectangle etc.

bluong:
Seems like squares should be enough for our purposes.  Closing PR

PR #1749, opened 2015-03-20
Add Interaction.Pointer.

jtlan:
Similar to the old `Interaction.Mousemove`, produces pixels whenever the pointer (mouse or touch), enters the `Component`, moves, or exits the `Component`.

### DIFF
--- a/src/interactions/pointerInteraction.ts
+++ b/src/interactions/pointerInteraction.ts
@@ -1,0 +1,105 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Interaction {
+  export class Pointer extends Interaction.AbstractInteraction {
+    private _mouseDispatcher: Dispatcher.Mouse;
+    private _touchDispatcher: Dispatcher.Touch;
+    private _overComponent = false;
+    private _pointerEnterCallback: (p: Point) => any;
+    private _pointerMoveCallback: (p: Point) => any;
+    private _pointerExitCallback: (p: Point) => any;
+
+    public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
+      super._anchor(component, hitBox);
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._mouseDispatcher.onMouseMove("Interaction.Pointer" + this.getID(), (p: Point) => this._handlePointerEvent(p));
+
+      this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._touchDispatcher.onTouchStart("Interaction.Pointer" + this.getID(), (p: Point, e: TouchEvent) => this._handlePointerEvent(p));
+    }
+
+    private _handlePointerEvent(p: Point) {
+      var translatedP = this._translateToComponentSpace(p);
+      if (this._isInsideComponent(translatedP)) {
+        var wasOverComponent = this._overComponent;
+        this._overComponent = true;
+        if (!wasOverComponent && this._pointerEnterCallback) {
+          this._pointerEnterCallback(translatedP);
+        }
+        if (this._pointerMoveCallback) {
+          this._pointerMoveCallback(translatedP);
+        }
+      } else if (this._overComponent) {
+        this._overComponent = false;
+        if (this._pointerExitCallback) {
+          this._pointerExitCallback(translatedP);
+        }
+      }
+    }
+
+    /**
+     * Gets the callback called when the pointer enters the Component.
+     *
+     * @return {(p: Point) => any} The current callback.
+     */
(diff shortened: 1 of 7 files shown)


src/interactions/pointerInteraction.ts
```diff
@@ -0,0 +1,105 @@
...
+    private _pointerMoveCallback: (p: Point) => any;
+    private _pointerExitCallback: (p: Point) => any;
+
+    public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
+      super._anchor(component, hitBox);
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._mouseDispatcher.onMouseMove("Interaction.Pointer" + this.getID(), (p: Point) => this._handlePointerEvent(p));
+
+      this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._touchDispatcher.onTouchStart("Interaction.Pointer" + this.getID(), (p: Point, e: TouchEvent) => this._handlePointerEvent(p));
```
bluong: Don't need the e argument here

src/interactions/pointerInteraction.ts
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public onPointerEnter(): (p: Point) => any;
+    /**
+     * Sets the callback called when the pointer enters the Component.
+     *
+     * @param {(p: Point) => any} callback The callback to set.
+     * @return {Interaction.Pointer} The calling Interaction.Pointer.
+     */
+    public onPointerEnter(callback: (p: Point) => any): Interaction.Pointer;
+    public onPointerEnter(callback?: (p: Point) => any): any {
```
bluong: Should this callback be typed for our semantic needs?

src/interactions/pointerInteraction.ts
```diff
@@ -0,0 +1,105 @@
...
+      if (callback == null) {
```
bluong: How do we cancel out a callback?  I'm personally a fan of using an empty function but convention throughout the library says to let null cancel out the callback.  Let's go with the convention.

src/interactions/pointerInteraction.ts
```diff
@@ -0,0 +1,105 @@
...
+      var translatedP = this._translateToComponentSpace(p);
+      if (this._isInsideComponent(translatedP)) {
+        var wasOverComponent = this._overComponent;
+        this._overComponent = true;
+        if (!wasOverComponent && this._pointerEnterCallback) {
+          this._pointerEnterCallback(translatedP);
+        }
+        if (this._pointerMoveCallback) {
+          this._pointerMoveCallback(translatedP);
+        }
```
bluong: So this allows for both enter and move to execute in the same action?  I think this is fine.  I'm just wondering if that's intended behavior.

bluong:
Mainly a few questions

test/interactions/pointerInteractionTests.ts
```diff
@@ -0,0 +1,147 @@
...
+      assert.deepEqual(lastPoint, { x: SVG_WIDTH/2, y: SVG_HEIGHT/2 }, "was passed correct point (touch)");
+
+      callbackCalled = false;
+      triggerFakeTouchEvent("touchstart", c.content(), SVG_WIDTH/4, SVG_HEIGHT/4);
+      assert.isTrue(callbackCalled, "callback on moving inside Component (touch)");
+      assert.deepEqual(lastPoint, { x: SVG_WIDTH/4, y: SVG_HEIGHT/4 }, "was passed correct point (touch)");
+
+      callbackCalled = false;
+      triggerFakeTouchEvent("touchstart", c.content(), 2*SVG_WIDTH, 2*SVG_HEIGHT);
+      assert.isFalse(callbackCalled, "not called when moving outside of the Component (touch)");
```
bluong: is touchstart used for moving around the touch area?

src/interactions/pointerInteraction.ts
jtlan: That's sort of how the browser behaves if the user enters a DOM element; a `mouseover` and `mousemove` are both sent. From the context of a user who is only interested in pointer-move events, this makes sense.

src/interactions/pointerInteraction.ts
jtlan: I thought about it, but it wouldn't be any more descriptive than just showing the relatively-simple signature.

src/interactions/pointerInteraction.ts
jtlan: My mistake, I should have used `===` here.

src/interactions/pointerInteraction.ts
bluong: I guess another thing for Typing these things is that if we change the type signature we only have to change it in one place.  SIngle source of truth etc.

PR #1752, opened 2015-03-21
[Component] - Replace `merge` with `above`/`below`

### LINKED ISSUE #1037: Ability to set ordering of Components in Group
Essentially, allowing Z-ordering of `Component`s within a group. This is particularly useful if you have multiple plots merged together and wish to highlight a particular plot by bringing it to the front.

bluong:
Replacing the `merge` API with an `above`/`below` API so users can make sense of where they are placing their components in regards to layering them on top of each other.

Fixes #1037

### DIFF
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -423,7 +423,7 @@ export module Component {
      * @param {Component} c The component to merge in.
      * @returns {ComponentGroup} The relevant ComponentGroup out of the above four cases.
      */
-    public merge(c: AbstractComponent, below = true): Component.Group {
+    public _merge(c: AbstractComponent, below: boolean): Component.Group {
       var cg: Component.Group;
       if (this._isSetup || this._isAnchored) {
         throw new Error("Can't presently merge a component that's already been anchored");
@@ -440,11 +440,11 @@ export module Component {
     }
 
     public above(c: AbstractComponent): Component.Group {
-      return this.merge(c, false);
+      return this._merge(c, false);
     }
 
     public below(c: AbstractComponent): Component.Group {
-      return this.merge(c, true);
+      return this._merge(c, true);
     }
 
     /**
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -427,14 +427,14 @@ export module Component {
     }
 
     /**
-     * Merges this Component below another Component, returning a
+     * Merges this Component above another Component, returning a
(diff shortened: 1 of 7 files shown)


jtlan:
Whoa ... mind blown. Get @crmorford's opinion on this?

src/components/abstractComponent.ts
```diff
@@ -410,35 +410,57 @@ export module Component {
...
+     *
+     * @param {Component} c The component to merge in.
+     * @returns {ComponentGroup} The relevant ComponentGroup out of the above four cases.
+     */
+    public above(c: AbstractComponent): Component.Group {
+      return this._merge(c, false);
+    }
+
+    /**
+     * Merges this Component above another Component, returning a
```
jtlan: Wait, when we have `a.below(b)`, does `a` wind up below `b`, or above it? This description suggests that `a` winds up above `b`, which seems counterintuitive.

src/components/abstractComponent.ts
```diff
@@ -410,35 +410,57 @@ export module Component {
...
       } else {
-        cg = new Plottable.Component.Group([this, c]);
+        var mergedComponents = below ? [this, c] : [c, this];
+        cg = new Plottable.Component.Group(mergedComponents);
         return cg;
       }
     }
 
     /**
+     * Merges this Component below another Component, returning a
```
jtlan: Wait, when we have `a.above(b)`, does `a` wind up above `b`, or below it? This description suggests that `a` winds up below `b`, which seems counterintuitive.

src/components/abstractComponent.ts
bluong: `a` should end up below `b`  I think the description I placed is incorrect

bluong: Yep I accidentally swapped the two docs somehow -.-;

jtlan:
Can you fix the description above the constructor in `Group` as well? It'd also be good to add a note about what the ordering is if you pass an array to the constructor.

src/components/componentGroup.ts
```diff
@@ -8,8 +8,11 @@ export module Component {
      * Constructs a GroupComponent.
      *
      * A GroupComponent is a set of Components that will be rendered on top of
-     * each other. When you call Component.merge(Component), it creates and
-     * returns a GroupComponent.
+     * each other. When you call Component.above(Component) or Component.below(Component),
+     * it creates and returns a GroupComponent.
```
jtlan: `Group` instead of `GroupComponent`, here and twice above.

PR #1754, opened 2015-03-23
Make hitbox optional to realize advantage of new Dispatchers.

jtlan:
Required for CSS-based cursor-changing to work on `Interaction.Drag`.

Not doing this on a side branch since I want some QE review too (just making sure all the `Interaction`s still work).

After everything is converted, we won't need `hitbox`es at all.


bluong:
Why is this required for cursor-changing to work?

Also,
> After everything is converted, we won't need `hitbox`es at all.

Why not just remove all the hitbox creation logic after so?  How does this benefit us currently?

jtlan:
> Why is this required for cursor-changing to work?
> Why not just remove all the hitbox creation logic after so? How does this benefit us currently?

The hitbox covers the entire `Component`, preventing any CSS cursor-changing effects in `content` from taking effect. Right now we can't remove the hitbox logic entirely because `PanZoom` still needs it.

test/interactions/hoverInteractionTests.ts
```diff
@@ -65,12 +65,12 @@ describe("Interactions", () => {
       });
 
       testTarget.registerInteraction(hoverInteraction);
-      hitbox = (<any> testTarget)._element.select(".hit-box");
+      target = testTarget.content();
```
bluong: Shouldn't the target be testTarget.hitBox()?  Why is the target changed?

test/core/componentTests.ts
```diff
@@ -244,7 +244,7 @@ describe("Component behavior", () => {
     svg.remove();
   });
 
-  it("hitboxes are created iff there are registered interactions", () => {
+  it("hitboxes are created iff there are registered interactions (that require hitboxes)", () => {
```
bluong: I don't think the parens are necessary

test/core/componentTests.ts
```diff
@@ -259,36 +259,26 @@ describe("Component behavior", () => {
     svg.remove();
     svg = generateSVG();
 
+    // registration before achoring
```
bluong: anchoring*

src/interactions/abstractInteraction.ts
```diff
@@ -18,6 +18,10 @@ export module Interaction {
       this._hitBox = hitBox;
     }
 
+    public _requiresHitbox() {
```
bluong: This endpoint will be removed eventually correct?  Can you place a ticket on eventually removing thehitbox idea and reference that issue here?

test/interactions/hoverInteractionTests.ts
bluong: Is this because HoverInteraction no longer will have a hitbox

test/interactions/interactionTests.ts
```diff
@@ -97,21 +97,21 @@ describe("Interactions", () => {
       ki.on(bCode, bCallback);
       component.registerInteraction(ki);
 
-      var $hitbox = $((<any> component)._hitBox.node());
+      var $target = $((<any> component).content().node());
```
bluong: We don't need to anycast anymore here

bluong:
We just need documentation on what is the purpose of this pr as it seems to me to be essentially a hack workaround over our current clunky architecture

test/interactions/hoverInteractionTests.ts
jtlan: yes.

test/core/componentTests.ts
jtlan: Well, if they didn't require hitboxes, then one isn't created. The parens help designate exactly what functionality is being tested.

src/interactions/abstractInteraction.ts
jtlan: yeah, will do.

test/interactions/interactionTests.ts
jtlan: Right.

test/core/componentTests.ts
bluong: The way I'm reading 
> "hitboxes are created iff there are registered interactions (that require hitboxes)"
is hitboxes are required for registered interactions and that the hitbox requirement is sort of an afterthought

whereas
> "hitboxes are created iff there are registered interactions that require hitboxes"

to me reads as only those interactions that require will have the creation of hitboxes.

jtlan: Oh, I see you meant to remove the parens, not the content inside them.

src/components/abstractComponent.ts
```diff
@@ -342,7 +342,7 @@ export module Component {
       // pushed to this._interactionsToRegister and registered during anchoring. If after, they are
       // registered immediately
       if (this._element) {
-        if (!this._hitBox) {
+        if (!this._hitBox && interaction._requiresHitbox()) {
             this._hitBox = this._addBox("hit-box");
             this._hitBox.style("fill", "#ffffff").style("opacity", 0); // We need to set these so Chrome will register events
         }
```
bluong: I can't comment under this code for some reason -.-;;;  but should we need to change the code to incorporate the 
`interaction._anchor(this, this._hitBox);` below this line.

Maybe 
```
if (requiresHitbox)
  if (hitbox does not exist)
    create hitbox
  anchor hitbox
```
?

bluong:
Just the one question

src/components/abstractComponent.ts
jtlan: `_anchor()` needs to be called whether or not `_hitBox` is null. I'll fix the signature.

bluong: I think it's been decided to do the signature adjustment after hitBox is removed.

PR #1758, opened 2015-03-23
deprecating HoverInteraction

bluong:
It has been remade into PointerInteraction


jtlan:
Probably better to do something like "`Interaction.Hover` is deprecated; use `Interaction.Pointer` in conjunction with `getClosestPlotData()` instead." ?

PR #1759, opened 2015-03-24
[ScrollInteraction] - Initial Implementation

bluong:
This is on the road to being able to zoom into a component.  Once we have a capability to nicely magnify a scale or so, we just need a solid example to derive from to have zooming enabled.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -4092,3 +4092,25 @@ declare module Plottable {
         }
     }
 }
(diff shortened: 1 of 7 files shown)


bluong:
I'm closing this pull request until the zoom idea is more realized on the scale level.  This implementation is likely 60-70% complete but it's not too useful for ideas like zooming yet.

PR #1763, opened 2015-03-25
Allow disabling of constraint functionality.

jtlan:
Important for panning to feel intuitive.

Moved all `assert` statements corresponding to constraint to a new test.

### DIFF
--- a/src/interactions/dragInteraction.ts
+++ b/src/interactions/dragInteraction.ts
@@ -75,17 +75,21 @@ export module Interaction {
     }
 
     /**
-     * Returns whether or not this Interaction constrains drag end-points
-     * to inside its Component (default true).
+     * Returns whether or not this Interaction constrains Points passed to its
+     * callbacks to lie inside its Component.
      *
      * @return {boolean} Whether or not the Interaction.Drag constrains.
      */
     public constrain(): boolean;
     /**
-     * Sets whether or not this Interaction constrains drag end-points
-     * to inside its Component.
+     * Sets whether or not this Interaction constrains Points passed to its
+     * callbacks to lie inside its Component.
      *
-     * @param {boolean} doConstrain Whether or not to constrain the end-point.
+     * If true, when the user drags outside of the Component, the closest Point
+     * inside the Component will be passed to the callback instead of the actual
+     * cursor position.
+     *
+     * @param {boolean} doConstrain Whether or not to constrain Points.
      * @return {Interaction.Drag} The calling Interaction.Drag.
      */
     public constrain(doConstrain: boolean): Drag;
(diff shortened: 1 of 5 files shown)


src/interactions/dragInteraction.ts
```diff
@@ -32,6 +33,10 @@ export module Interaction {
 
     private _translateAndConstrain(p: Point) {
       var translatedP = this._translateToComponentSpace(p);
+      if (!this._constrain) {
```
bluong: Should we use the getter here instead of the instance var?

src/interactions/dragInteraction.ts
```diff
@@ -70,6 +75,29 @@ export module Interaction {
     }
 
     /**
+     * Returns whether or not this Interaction constrains drag end-points
```
bluong: When you say end-points, I'm unsure about if this also talks about the point given during the drag sequence rather than the onDragStart/onDragEnd sequences

jtlan: Yeah. Is there a better way to emphasize that?

test/interactions/dragInteractionTests.ts
```diff
@@ -156,6 +145,49 @@ describe("Interactions", () => {
       assert.deepEqual(receivedStart, startPoint, "was passed the correct starting point");
       assert.deepEqual(receivedEnd, endPoint, "was passed the correct current point");
 
+      assert.strictEqual(drag.onDragEnd(), endCallback, "retrieves the callback if called with no arguments");
+      drag.onDragEnd(null);
+      assert.isNull(drag.onDragEnd(), "removes the callback if called with null");
+      svg.remove();
+    });
+
+    it("constrain()", () => {
```
bluong: We are definitely going to need to revisit our test structure, because I can see why this test is huge and judging from how our tests are set up, it actually makes sense in terms of our infrastructure... which is not a good thing.

bluong:
Just a documentation issue

src/interactions/dragInteraction.ts
bluong: drag points?

test/interactions/dragInteractionTests.ts
jtlan: I could make this a `describe()` and have sub-`describe()`s for the touch and mouse cases.

bluong: This is somewhat exactly what I think we should heading towards.  Having a single `it` for an API point, especially one that has many cases of usage is very hard to manage...

bluong: Don't worry about this for now unless you want to set a precedent.

Issue #1639, opened 2015-02-17
Pan Zoom min/max Zoom Amount


crmorford:
(Slate 1721)
Prevent users from zooming too far in or out

crmorford:
* API approved by @dbt 
* If I'm zoomed in the max amount, I should still be able to pan
* If I'm panned to the smaller or upper limit, I should still be able to zoom
* Works on iPad
* Reasonable default settings
* Min/Max values change on data update (if I add an outlier, it should be taken into account)
* Can still explicitly change domain to whatever I want
* Should be able to be turned off

jtlan:
Closed in #2379.

PR #1764, opened 2015-03-25
[ScaleDomainTransformers] - Adding translate and magnify

bluong:
Mainly to aid for simulating Pan behavior and Zoom behavior, this requires that the scale domains to be affected appropriately such that Interactions that want to use the above behavior can query these domain changes to happen to work.

For Pan, there is the allowance of the user to translate the domain of the scale by a given amount.

For Zoom, there is the allowance of the user to magnify the domain of the scale by a given amount given a value in range space.

Name is completely up to debate.  I just thought of something appropriate at the time.


src/utils/scaleDomainTransformers.ts
```diff
@@ -0,0 +1,18 @@
...
+module Plottable {
+
+  export module ScaleDomainTransformers {
+
+    export function translate<D>(scale: Scale.AbstractQuantitative<D>, delta: number) {
+      var translateTransform = (rangeValue: number) => scale.invert(rangeValue - delta);
+      return scale.range().map(translateTransform);
+    }
+
+    export function magnify<D>(scale: Scale.AbstractQuantitative<D>, magnifyAmount: number, centerValue: number) {
```
jtlan: Should this and `translate()` be static?

jtlan: Is this value in data-space or pixel-space?

src/utils/scaleDomainTransformers.ts
```diff
@@ -0,0 +1,18 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+
+  export module ScaleDomainTransformers {
+
+    export function translate<D>(scale: Scale.AbstractQuantitative<D>, delta: number) {
```
jtlan: Is `delta` in data-space or pixel space?

jtlan:
Some questions that would be answered by docs...

bluong:
All the query values are in pixel space.  I'll doc this up

jtlan:
The semantics of the API suggest that the methods modify the scale, whereas in reality they do not. If I call `translate(scale, amount)`, I would expect the scale to get translated, rather than receiving a domain.

bluong:
I am very much a fan of having a constructive API as that is a lot easier to work with.  That said, the semantics are indeed strange.

Should we try to keep the constructive API and rename the API or go with the current name and make it destructive?

bluong:
I made the API destructive, but I'm still unsure about having a constructive vs destructive API.  Respond on what you think is best @jtlan

jtlan:
I can see some value in an implementation that only uses pure functions; it might make it easier to set zooming and panning limits.

Speaking of, what's your strategy for #1639?

bluong:
There are two ways that I can think about for accomplishing #1639 

1) One is to essentially put more arguments in these API points to have these transformers inherently have the capability to do do min/max.

2) The other is to put these configuration points on the implementation of a panning/zooming.

Advantage to 1) is that it keeps all the configuration in one place so that usage of this transformer can be moved around with hopefully not that much trouble.

Advantage to 2) is that it keeps the transformer lightweight.

Above all being said, I'd like to think that the min/max problem should hopefully not be an issue in terms of how to affect the domain.  For panning, if the translateAmount tells the domain's min to go below the enforced minimum, then it will pan to the enforced minimum.

Zooming is a bit trickier, but what could be done is essentially, if the magnifyAmount tells the domain's min to go above the enforced maximum, then it will zoom to the maximum (enforced minimum on the max end)

bluong:
Playing around with Panning behavior that react with these functions.  I am unsure if the functions should know about the limits and calculate the new domains accordingly.  The reason why seems to be mainly because different interactions will use these limits differently it seems (KeyInteraction with Panning may not want to hit the limits while DragInteraction usually will want to)

bluong:
Labelling BLOCKER since I have an implementation that should be enough to satisfy issues regarding Panning.

bluong:
Trying to implement zoom stuff in the background and finding that having these functions be pure is pretty important.  Pure functions are also just easier to work with.

jtlan:
@bluong
> For panning, if the translateAmount tells the domain's min to go below the enforced minimum, then it will pan to the enforced minimum.

What happens to the other end of the domain?

jtlan:
Unit tests...?

bluong:
> What happens to the other end of the domain?

Let me try to sum it up in an example.  If I were to keeping panning left on an axis, I should eventually hit the limit.  If I hit the limit, then the scale's domain should stop translating on both ends

test/utils/scaleDomainTransformersTests.ts
```diff
@@ -0,0 +1,98 @@
+///<reference path="../testReference.ts" />
+
+var assert = chai.assert;
+
+describe("ScaleDomainTransformers", () => {
+
+  describe("translate", () => {
+
+    it("translation on arbitrary value", ()=> {
```
jtlan: What does this test name mean?

test/utils/scaleDomainTransformersTests.ts
```diff
@@ -0,0 +1,98 @@
...
+      assert.deepEqual(translatedDomain, [5, 105], "domain translated by correct amount");
+    });
+
+  });
+
+  describe("magnify", () => {
+
+    describe("magnifyAmount", () => {
+
+      it("magnify on arbitrary value",() => {
```
jtlan: Once again, why "arbitrary"?

test/utils/scaleDomainTransformersTests.ts
```diff
@@ -0,0 +1,98 @@
...
+        assert.deepEqual(translatedDomain, [0, 100], "domain magnified by correct amount");
+      });
+
+      it("magnify on 0",() => {
+        var domain = [0, 100];
+        var range = [0, 200];
+        var scale = new Plottable.Scale.Linear();
+        scale.domain(domain).range(range);
+        var translatedDomain = Plottable.ScaleDomainTransformers.magnify(scale, 0, 100);
+        assert.deepEqual(translatedDomain, [50, 50], "domain magnified by correct amount");
```
jtlan: Isn't this potentially dangerous?

jtlan:
Any tests for weird inputs? Like:
* Translate by 0
* Translate by negative value
* Magnify by negative value

Last one could be spooky.

test/utils/scaleDomainTransformersTests.ts
```diff
@@ -0,0 +1,98 @@
...
+        var range = [0, 200];
+        var scale = new Plottable.Scale.Linear();
+        scale.domain(domain).range(range);
+        var translatedDomain = Plottable.ScaleDomainTransformers.magnify(scale, 0.90, 200);
+        assert.deepEqual(translatedDomain, [10, 100], "domain magnified and correctly centered");
+      });
+
+    });
+
+    describe("centerValue data space", () => {
```
jtlan: What's this mean?

src/utils/scaleDomainTransformers.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Returns a magnified domain of the input scale with a magnification of the input magnifyAmount
+     * in range space with the center point as the input centerValue, also in range space
+     *
+     * @param {Scale.AbstractQuantitative<D> scale The input scale whose domain is being magnified
+     * @param {number} magnifyAmount The amount to magnify
+     * @param {number} centerValue The center point of the magnification
+     * @param {boolean} centerInDomainSpace Specifying if the domain is in domain space (default = false)
```
jtlan: I think this parameter is confusing:
* Description seems to be incorrect
* Adds additional flexibility that probably won't be used (most `Interaction`s originate in pixel-space
* If the user is interesting in recovering the flexibility, they could just call `invert()` on their pixel value.

test/utils/scaleDomainTransformersTests.ts
bluong: I just wanted to do a standard translation on a nonspecial value

test/utils/scaleDomainTransformersTests.ts
bluong: What would be an alternative expected behavior?

test/utils/scaleDomainTransformersTests.ts
bluong: This is mainly if the user of this function would rather have the centerValue be thought of in data space.  That said, given your statement below, maybe it would be better to just have the user scale before inputting to keep a simpler API.

jtlan:
Let's back up a bit here. The simplest thing to build would probably be something like:
```
scale.zoom(args)
scale.translate(args)
```
and have those be state-changing methods. Maybe the signatures should look something like this:
```Typescript
public translate(distance: pixel): Scale
public zoom(zoomLevel: number, center: pixel): Scale
```
Where we have also done this:
```Typescript
export type pixel = number;
```

This makes it clear that `distance` and `center` are in pixels. the use of the name `zoomLevel` also makes it clear what the convention is: passing in values greater than 1 zooms in, and less than 1 zooms out.

bluong:
@jtlan Yup that implementation I think would be the least changing to our API that has to be solidified soon.  I'll close this PR and see if I can transfer the logic to the scale

PR #1765, opened 2015-03-25
[Interaction.DoubleClick] - Removing Click hierarchy

bluong:
As `Interaction.Click` is going to be rearchitected, we should shelve off `Interaction.DoubleClick` to use the old architecture to retain functionality.


bluong:
If we want this PR to go through, I am unsure if this is NOQE.  A regression pass would be all that is needed

PR #1767, opened 2015-03-25
Add in DragBoxLayer.

jtlan:


plottable.css
```diff
@@ -120,19 +120,32 @@ svg.plottable {
...
+.plottable .drag-box-layer.x-resizable.y-resizable .drag-corner-tr {
+  cursor: nesw-resize;
+}
+.plottable .drag-box-layer.x-resizable.y-resizable .drag-corner-bl {
+  cursor: nesw-resize;
+}
+.plottable .drag-box-layer.x-resizable.y-resizable .drag-corner-br {
+  cursor: nwse-resize;
+}
+/* /DragBoxLayer */
```
bluong: I remember you hearing concerns about using these invisible targets.  Is this the way to go about it or should we just set the cursor based on pixel position ourselves?

src/components/selectionBoxLayer.ts
```diff
@@ -3,7 +3,7 @@
 module Plottable {
 export module Component {
   export class SelectionBoxLayer extends AbstractComponent {
-    private _box: D3.Selection;
```
bluong: Should we have a getter here or not necessary?

bluong:
I haven't read through all of this yet, but when we document this on the releases.  We should highlight the fact that this new component is just us subclassing a component and mashing it against an Interaction to achieve a common behavior.  This is important since users can build use something like this as a guide for how to subclass and build their own Interactable Components

plottable.css
jtlan: I thought about that. The advantage of using CSS is that people could specify custom cursors.

src/components/selectionBoxLayer.ts
jtlan: Probably not? You could find it using `content().select(...)`.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+        this._boxCornerBL.attr("r", this._boxEdgeWidth / 2);
+        this._boxCornerBR.attr("r", this._boxEdgeWidth / 2);
+      }
+
+      /**
+       * Gets whether or not the drag box is resizable.
+       *
+       * @return {boolean} Whether or not the drag box is resizable.
+       */
+      public resizable(): boolean;
```
bluong: We probably want designations on where the box can resize (vertical/horizontal granularity, possibly even straight-up cardinal granularity)

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+       */
+      public edgeWidth(): number;
+      /**
+       * Sets the edge width of the drag box.
+       *
+       * @param {number} width The desired edge width.
+       * @return {DragBoxLayer} The calling DragBoxLayer.
+       */
+      public edgeWidth(width: number): DragBoxLayer;
+      public edgeWidth(width?: number): any {
```
bluong: Now that we are in interactive land, we can think about these names in terms of interactivity possibly.  edgeWidth to me still sounds like the edge with of the actual box (stroke).  What's a better name?

src/components/interactive/dragBoxLayer.ts
jtlan: I sent an email out about this. I'll be (re-)adding `XDragBox` and `YDragBox`.

src/components/interactive/dragBoxLayer.ts
jtlan: `detectionRadius`?

bluong: Let's ask @crmorford to be sure, but yes `detectionRadius` might just come back :P

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+      private _dragInteraction: Interaction.Drag;
+      private _boxEdgeT: D3.Selection;
+      private _boxEdgeB: D3.Selection;
+      private _boxEdgeL: D3.Selection;
+      private _boxEdgeR: D3.Selection;
+      private _boxCornerTL: D3.Selection;
+      private _boxCornerTR: D3.Selection;
+      private _boxCornerBL: D3.Selection;
+      private _boxCornerBR: D3.Selection;
+      private _boxEdgeWidth = 6;
```
bluong: In accordance to my previous comment, these instance vars now only serve the purpose of interactivity and thus should be named accordingly, since _boxEdgeT is not actually the top visual edge of the box but mainly a handle placed there for dragging.

Maybe _boxDetectionEdgeT?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+
+      private _xResizable = false;
+      private _yResizable = false;
+      private _grabbedEdges: _EdgeIndicator;
```
bluong: Note to self.  Is this variable useful?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+      private _boxEdgeL: D3.Selection;
+      private _boxEdgeR: D3.Selection;
+      private _boxCornerTL: D3.Selection;
+      private _boxCornerTR: D3.Selection;
+      private _boxCornerBL: D3.Selection;
+      private _boxCornerBR: D3.Selection;
+      private _boxEdgeWidth = 6;
+
+      private _xResizable = false;
+      private _yResizable = false;
```
bluong: Just making sure.  So even the new XDragBoxes etc.  will start off as not resizable?

jtlan: That's correct. That aligns with the current behavior in **develop**.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+          });
+        });
+
+        this._dragInteraction.onDragEnd((s: Point, e: Point) => {
+          if (s.x === e.x && s.y === e.y) {
+            this.boxVisible(false);
+          }
+        });
+
+        this.registerInteraction(this._dragInteraction);
```
bluong: This constructor is... really heavy.  While it is rightfully heavy in terms of what must be set up (I think?), we probably want to at least split these complicated parts into static functions so that it's a lot easier to read

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+                                       "opacity": 0,
+                                       "fill": "pink"
+                                     });
+        this._boxCornerTL = createCorner().classed("drag-corner-tl", true);
+        this._boxCornerTR = createCorner().classed("drag-corner-tr", true);
+        this._boxCornerBL = createCorner().classed("drag-corner-bl", true);
+        this._boxCornerBR = createCorner().classed("drag-corner-br", true);
+      }
+
+      private _getEdges(p: Point) {
```
bluong: Why does _getEdges return a bunch of booleans?

bluong: Also rename possibly to _getSurroundingEdges

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,237 @@
...
+      constructor() {
+        super();
+        this.classed("drag-box-layer", true);
+
+        this._dragInteraction = new Interaction.Drag();
+
+        this._dragInteraction.onDragStart((s: Point) => {
+          this.boxVisible(true);
+
+          this._grabbedEdges = this._getEdges(s);
```
bluong: _grabbedEdges seems to not be used outside the constructor.  Why need an instance var?

jtlan: It's referenced in two different callbacks, so I need to hold a reference to it to persist it. Its value is initialized in the callback to `onDragStart()` and referenced in the callback to `onDragMove()`. Or are you saying it should be local to the constructor?

src/components/interactive/dragBoxLayer.ts
jtlan: It used to return `string[]`, but it is a lot easier to remove edges from this object rather than remove a string from an array.

src/components/interactive/dragBoxLayer.ts
bluong: local to the constructor seems to make the most sense to me

bluong: I guess _grabbedEdges might be a weird name which could be the main cause for my concern.  When I saw this._grabbedEdges, I assumed that the variable contained the actual selection edges, not boolean indicator information

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,239 @@
...
+          this.bounds({
+            topLeft: topLeft,
+            bottomRight: bottomRight
+          });
+        });
+
+        this._dragInteraction.onDragEnd((s: Point, e: Point) => {
+          if (s.x === e.x && s.y === e.y) {
+            this.boxVisible(false);
+          }
```
bluong: Do you think this is necessary?  It's a really hard task to achieve in the first place

src/components/interactive/dragBoxLayer.ts
jtlan: maybe just "grabbed"?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,239 @@
...
+            this.bounds({
+              topLeft: s,
+              bottomRight: s
+            });
+            this._grabbedEdges.bottom = true;
+            this._grabbedEdges.right = true;
+          }
+          var bounds = this.bounds();
+          this._initTL = bounds.topLeft;
+          this._initBR = bounds.bottomRight;
```
bluong: Same with _initTL/BR with _grabbedEdges

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,239 @@
...
+    export class DragBoxLayer extends Component.SelectionBoxLayer {
+      private _dragInteraction: Interaction.Drag;
+      private _boxEdgeT: D3.Selection;
+      private _boxEdgeB: D3.Selection;
+      private _boxEdgeL: D3.Selection;
+      private _boxEdgeR: D3.Selection;
+      private _boxCornerTL: D3.Selection;
+      private _boxCornerTR: D3.Selection;
+      private _boxCornerBL: D3.Selection;
+      private _boxCornerBR: D3.Selection;
```
bluong: I still think we should have a detection idea on these instance vars

src/components/interactive/dragBoxLayer.ts
bluong: ifEdgesGrabbed or similar?

jtlan: I thought the semantics of `grabbed` were good, since then I could write:
```Typescript
if (grabbed.top) {
  ...
}
```

src/components/interactive/dragBoxLayer.ts
jtlan: If you click, it dismisses the box.

src/components/interactive/dragBoxLayer.ts
jtlan: Yeah, local probably reduces clutter a bit.

src/components/interactive/dragBoxLayer.ts
bluong: Let's go with grabbed and see how that works

src/components/interactive/dragBoxLayer.ts
bluong: Oh got it

src/components/interactive/dragBoxLayer.ts
bluong: Still think this

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,243 @@
...
+          }
+          if (!this._xResizable) {
+            grabbedEdges.left = false;
+            grabbedEdges.right = false;
+          }
+
+          if (!this.boxVisible() ||
+              (!grabbedEdges.top && !grabbedEdges.bottom &&
+               !grabbedEdges.left && !grabbedEdges.right)
+             ) {
```
bluong: Not a big deal, but should the endparen be moved up one line?  It'll just a save us a line

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,243 @@
...
+
+        var createCorner = () => this._box.append("circle")
+                                     .style({
+                                       "opacity": 0,
+                                       "fill": "pink"
+                                     });
+        this._boxCornerTL = createCorner().classed("drag-corner-tl", true);
+        this._boxCornerTR = createCorner().classed("drag-corner-tr", true);
+        this._boxCornerBL = createCorner().classed("drag-corner-bl", true);
+        this._boxCornerBR = createCorner().classed("drag-corner-br", true);
```
bluong: For X/YDragBox are you going to keep the corners but instead unclass them?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,243 @@
...
+        if (this._isSetup) {
+          this._boxEdgeT.style("stroke-width", this._detectionRadius * 2);
+          this._boxEdgeB.style("stroke-width", this._detectionRadius * 2);
+          this._boxEdgeL.style("stroke-width", this._detectionRadius * 2);
+          this._boxEdgeR.style("stroke-width", this._detectionRadius * 2);
+          this._boxCornerTL.attr("r", this._detectionRadius);
+          this._boxCornerTR.attr("r", this._detectionRadius);
+          this._boxCornerBL.attr("r", this._detectionRadius);
+          this._boxCornerBR.attr("r", this._detectionRadius);
+        }
```
bluong: This code should probably be moved to a function like
applyDetectionRadius or similar.  As a result of this function doing all the work, there's an awkward this.detectionRadius(this._detectionRadius) up above.

Also, if you want to be ultra safe, you can stick with this._render but that is up to you as this is indeed more efficient

src/components/interactive/dragBoxLayer.ts
bluong: Or you could instead say

if either of xResize or yResize is false, skip this step

(or if both are true, then do this step)

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,243 @@
...
+            });
+            // a new box is effectively "grabbed" at the bottom-right corner
+            grabbedEdges.bottom = true;
+            grabbedEdges.right = true;
+          }
+
+          this.boxVisible(true);
+          var bounds = this.bounds();
+          _initTL = bounds.topLeft;
+          _initBR = bounds.bottomRight;
```
bluong: Is _initTL and _initBR different from bounds.topLeft/bottomRight at any sensible time?

src/components/interactive/dragBoxLayer.ts
bluong: Actually does it make sense in this context (I click, drag around, and end at the same spot).  Should the box disappear?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,243 @@
...
+            x1: l, y1: t, x2: l, y2: b
+          });
+          this._boxEdgeR.attr({
+            x1: r, y1: t, x2: r, y2: b
+          });
+
+          this._boxCornerTL.attr({ cx: l, cy: t });
+          this._boxCornerTR.attr({ cx: r, cy: t });
+          this._boxCornerBL.attr({ cx: l, cy: b });
+          this._boxCornerBR.attr({ cx: r, cy: b });
```
bluong: What do we do about the corners in the x/yDragBox case?

src/components/interactive/dragBoxLayer.ts
bluong: Should be have the X/YDragBox be configuration points or sublcasses of DragBox?

bluong:
Mainly questions about how X/YDragBox will be implemented based on this parent class and other behavior

src/components/interactive/dragBoxLayer.ts
jtlan: As in, put a "selected" property on it?

src/components/interactive/dragBoxLayer.ts
jtlan: The corners will be kept, but they don't produce any cursor changes unless both properties are set.

src/components/interactive/dragBoxLayer.ts
jtlan: `_initTL` and `_initBR` are the `bounds()` values only when the drag is initialized. I can rename them so that this is clearer.

src/components/interactive/dragBoxLayer.ts
jtlan: I'll be overriding bounds() so that on `XDragBoxLayer`, the y-values are automatically set, and vice-versa for `YDragBoxLayer`.

src/components/interactive/dragBoxLayer.ts
bluong: Oh I meant like
_boxDetectionEdgeT etc.

Cause this is not the _box's EdgeT (that would be the top drawn line on the svg element), but instead the DetectionEdgeT (the invisible line of some thickness used for detection)

src/components/interactive/dragBoxLayer.ts
bluong: Do we want to run through the program logic even if only one (or none) are set?

src/components/interactive/dragBoxLayer.ts
bluong: Let's rename that I think I see why

src/components/interactive/dragBoxLayer.ts
bluong: Oh I see what you're getting at.  But regardless, what does this do for corners

src/components/interactive/dragBoxLayer.ts
jtlan: I'll be making them subclasses.

src/components/interactive/dragBoxLayer.ts
jtlan: @crmorford says the box should disappear. Makes sense, since it would have zero width and height, right?

bluong: Would it be a good idea to have this.boxVisible(false) be tied to bounds with start/endPoint the same?  I'm mainly curious if this is a DragBox or a SelectionBox issue now.

src/components/interactive/dragBoxLayer.ts
jtlan: They'll still be there, but they won't do anything.

src/components/interactive/dragBoxLayer.ts
jtlan: We still have to create them.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+    export class DragBoxLayer extends Component.SelectionBoxLayer {
+      private _dragInteraction: Interaction.Drag;
+      private _detectionEdgeT: D3.Selection;
+      private _detectionEdgeB: D3.Selection;
+      private _detectionEdgeL: D3.Selection;
+      private _detectionEdgeR: D3.Selection;
+      private _detectionCornerTL: D3.Selection;
+      private _detectionCornerTR: D3.Selection;
+      private _detectionCornerBL: D3.Selection;
+      private _detectionCornerBR: D3.Selection;
```
bluong: boxDetectionEdge?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+      constructor() {
+        super();
+        this.classed("drag-box-layer", true);
+
+        this._dragInteraction = new Interaction.Drag();
+        this._setUpCallbacks();
+        this.registerInteraction(this._dragInteraction);
+      }
+
+      private _setUpCallbacks() {
```
bluong: _setupCallbacks?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+              (!grabbedEdges.top && !grabbedEdges.bottom &&
+               !grabbedEdges.left && !grabbedEdges.right)
+             ) {
+            this.bounds({
+              topLeft: s,
+              bottomRight: s
+            });
+            // a new box is effectively "grabbed" at the bottom-right corner
+            grabbedEdges.bottom = true;
+            grabbedEdges.right = true;
```
bluong: What happens if we start dragging and go in the top left direction

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+
+        var createCorner = () => this._box.append("circle")
+                                     .style({
+                                       "opacity": 0,
+                                       "fill": "pink"
+                                     });
+        this._detectionCornerTL = createCorner().classed("drag-corner-tl", true);
+        this._detectionCornerTR = createCorner().classed("drag-corner-tr", true);
+        this._detectionCornerBL = createCorner().classed("drag-corner-bl", true);
+        this._detectionCornerBR = createCorner().classed("drag-corner-br", true);
```
bluong: Shouldn't the above be wrapped up in a this._xResizable && this._yResizable

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+            x1: l, y1: t, x2: l, y2: b
+          });
+          this._detectionEdgeR.attr({
+            x1: r, y1: t, x2: r, y2: b
+          });
+
+          this._detectionCornerTL.attr({ cx: l, cy: t });
+          this._detectionCornerTR.attr({ cx: r, cy: t });
+          this._detectionCornerBL.attr({ cx: l, cy: b });
+          this._detectionCornerBR.attr({ cx: r, cy: b });
```
bluong: Should this only be wrapped up in _xResiable && _yResiable?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+        this._detectionRadius = r;
+        if (this._isSetup) {
+          this._detectionEdgeT.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeB.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeL.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeR.style("stroke-width", this._detectionRadius * 2);
+          this._detectionCornerTL.attr("r", this._detectionRadius);
+          this._detectionCornerTR.attr("r", this._detectionRadius);
+          this._detectionCornerBL.attr("r", this._detectionRadius);
+          this._detectionCornerBR.attr("r", this._detectionRadius);
```
bluong: I think this should be wrapped up in a setDetectionRadius function

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+                                     .style({
+                                       "opacity": 0,
+                                       "fill": "pink"
+                                     });
+        this._detectionCornerTL = createCorner().classed("drag-corner-tl", true);
+        this._detectionCornerTR = createCorner().classed("drag-corner-tr", true);
+        this._detectionCornerBL = createCorner().classed("drag-corner-bl", true);
+        this._detectionCornerBR = createCorner().classed("drag-corner-br", true);
+
+        this.detectionRadius(this._detectionRadius);
```
bluong: This setter placement here seems odd.  I think you should actually call the calls you want to make directly.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,327 @@
...
+        if (this._isSetup) {
+          this._detectionEdgeT.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeB.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeL.style("stroke-width", this._detectionRadius * 2);
+          this._detectionEdgeR.style("stroke-width", this._detectionRadius * 2);
+          this._detectionCornerTL.attr("r", this._detectionRadius);
+          this._detectionCornerTR.attr("r", this._detectionRadius);
+          this._detectionCornerBL.attr("r", this._detectionRadius);
+          this._detectionCornerBR.attr("r", this._detectionRadius);
+        }
```
bluong: End with a this._render?

src/components/interactive/dragBoxLayer.ts
jtlan: I think this is better, since it translates to "set up callbacks" instead of "setup callbacks".

src/components/interactive/dragBoxLayer.ts
jtlan: `grabbedEdges` only indicates which edges the user initially grabbed. If the user drags through to the other side the `bounds()` call will correct for it.

src/components/interactive/dragBoxLayer.ts
jtlan: No, because then when someone sets the properties to `true` we have to go create the elements. If the property is `false` the presence of the elements doesn't do anything bad.

src/components/interactive/dragBoxLayer.ts
jtlan: See above comment.

src/components/interactive/dragBoxLayer.ts
jtlan: But `detectionRadius()` already sets the detection radius.

src/components/interactive/dragBoxLayer.ts
jtlan: Maybe we should just set the detection radius as part of rendering?

src/components/interactive/dragBoxLayer.ts
jtlan: ? We removed "box" previously in favor of this though. Since it's a private var I don't think it matters that much...

src/components/interactive/dragBoxLayer.ts
bluong: But if we start draggin and go to the top left, isn't the grabbedEdges top and left?

bluong:
Mainly the one question

src/components/interactive/dragBoxLayer.ts
jtlan: But not the edges initially grabbed by the user. I can rename this variable to make it clearer.

src/components/interactive/dragBoxLayer.ts
jtlan: Better to have them independent; originally `bounds()` would force the box visible, but there might be cases that you'd want to set the bounds silently or something.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,322 @@
...
+        this._dragInteraction.onDragStart((s: Point) => {
+          grabbedEdges = this._getEdges(s);
+          if (!this._yResizable) {
+            grabbedEdges.top = false;
+            grabbedEdges.bottom = false;
+          }
+          if (!this._xResizable) {
+            grabbedEdges.left = false;
+            grabbedEdges.right = false;
+          }
```
bluong: I assume this is for xDragBox/yDragBox

Instead, shouldn't xDragBox/yDragBox overwrite this._getEdges

src/components/interactive/dragBoxLayer.ts
bluong: Since this occurs after the check above, wouldn't this run into a scenario where yResizable and right is set to true?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,322 @@
...
+            this.bounds({
+              topLeft: s,
+              bottomRight: s
+            });
+            // a new box is effectively "grabbed" at the bottom-right corner
+            grabbedEdges.bottom = true;
+            grabbedEdges.right = true;
+          }
+
+          this.boxVisible(true);
```
bluong: This only needs to be set if !this.boxVisible();

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,322 @@
...
+            this._dragStartCallback(bounds);
+          }
+        });
+
+        this._dragInteraction.onDrag((s: Point, e: Point) => {
+          if (grabbedEdges.bottom) {
+            bottomRight.y = e.y;
+          } else if (grabbedEdges.top) {
+            topLeft.y = e.y;
+          }
```
bluong: If `grabbedEdges.bottom` is `true` but `e` ends up to be above `s`, the bottomRight var will now point to the coordinate in the topLeft corner, wouldn't it?

src/components/interactive/dragBoxLayer.ts
```diff
@@ -0,0 +1,322 @@
...
+
+          if (grabbedEdges.right) {
+            bottomRight.x = e.x;
+          } else if (grabbedEdges.left) {
+            topLeft.x = e.x;
+          }
```
bluong: Same as above?

src/components/interactive/dragBoxLayer.ts
jtlan: Yes, but it's only the point that was initially in the bottom-right (hence the original name `_initBR`).

src/components/interactive/dragBoxLayer.ts
jtlan: Not quite, since later this method sets `grabbedEdges.right` and `grabbedEdges.bottom` to true if no edges were grabbed, to start a new box.

bluong: Yes but that also sounds incorrect to me since for an XDragBox, grabbedEdges.bottom should be set to false right?

src/components/interactive/dragBoxLayer.ts
bluong: Maybe put the `init` in front?  My bad if I told you to remove it (I don't remember at this point)

PR #1768, opened 2015-03-25
[ClickInteraction] - Refactor to new Interaction API

### LINKED ISSUE #1684: Convert Interaction.Click to use Dispatchers

bluong:
`Interaction.Click` responds to clicks within a component using a callback set through `onClick`

API-Breaks:
The `callback` setter on `Interaction.Click` has been reworked and renamed into an `onClick` getter/setter API.

Fixes #1684

### DIFF
--- a/src/interactions/clickInteraction.ts
+++ b/src/interactions/clickInteraction.ts
@@ -5,6 +5,7 @@ export module Interaction {
   export class Click extends AbstractInteraction {
 
     private _mouseDispatcher: Plottable.Dispatcher.Mouse;
+    private _touchDispatcher: Plottable.Dispatcher.Touch;
     private _clickCallback: (p: Point) => any;
 
     public _anchor(component: Component.AbstractComponent, hitBox: D3.Selection) {
@@ -13,6 +14,10 @@ export module Interaction {
       this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> component.content().node());
       this._mouseDispatcher.onMouseDown("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseDown(p));
       this._mouseDispatcher.onMouseUp("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseUp(p));
+
+      this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> component.content().node());
+      this._touchDispatcher.onTouchStart("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseDown(p));
+      this._touchDispatcher.onTouchEnd("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseUp(p));
     }
 
     private _handleMouseDown(p: Point) {
--- a/src/interactions/clickInteraction.ts
+++ b/src/interactions/clickInteraction.ts
@@ -28,8 +28,8 @@ export module Interaction {
     }
 
     private _handleClickUp(p: Point) {
-      if (this._clickedDown && this._isInsideComponent(p) && (this.onClick() != null)) {
-        this.onClick()(p);
+      if (this._clickedDown && this._isInsideComponent(p) && (this._clickCallback != null)) {
+        this._clickCallback(p);
       }
       this._clickedDown = false;
     }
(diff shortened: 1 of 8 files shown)


src/interactions/clickInteraction.ts
```diff
@@ -3,41 +3,58 @@
...
+    private _handleMouseDown(p: Point) {
+      if (this._isInsideComponent(p)) {
+        this._clickedDown = true;
+      }
     }
 
-    protected _listenTo(): string {
-      return "click";
+    private _handleMouseUp(p: Point) {
+      if (this._clickedDown && this._isInsideComponent(p) && (this.onClick() != null)) {
```
jtlan: I think `this._clickCallback` would be clearer than `this.onClick() != null`.

src/interactions/clickInteraction.ts
```diff
@@ -3,41 +3,58 @@
...
-        this._callback({x: x, y: y});
-      });
+
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> component.content().node());
+      this._mouseDispatcher.onMouseDown("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseDown(p));
+      this._mouseDispatcher.onMouseUp("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseUp(p));
+
+      this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> component.content().node());
+      this._touchDispatcher.onTouchStart("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseDown(p));
+      this._touchDispatcher.onTouchEnd("Interaction.Click" + this.getID(), (p: Point) => this._handleMouseUp(p));
```
jtlan: Should we rename the called methods to `_handleClickDown` and `_handleClickUp`?

src/interactions/clickInteraction.ts
bluong: I remember a statement a while back that we should be using getters wherever possible.  Should we stick to that or?

src/interactions/clickInteraction.ts
bluong: I think I like those names.  Let's go with that.

jtlan:
@bluong wrote:
> I remember a statement a while back that we should be using getters wherever possible. Should we stick to that or?

I think within a class it's fine to use the actual reference instead of the getter. *Setting* should be done with the setter though.

bluong:
> I think within a class it's fine to use the actual reference instead of the getter. Setting should be done with the setter though.

The difference comes down to the implementation of the getter.  Note that this is something that I'm unsure of myself, as I can imagine some scenarios when you want to bypass the getter logic to get the raw variable value.  While it does look cleaner to not use the getter here, I'm not sure on what's the best move to make for consistency moving forward

crmorford:
http://jsfiddle.net/cpyvjkr5/10/ Points are being placed to the right of where I expect. Anyone know why?

src/interactions/clickInteraction.ts
```diff
@@ -3,41 +3,58 @@
...
+
+      this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> component.content().node());
+      this._touchDispatcher.onTouchStart("Interaction.Click" + this.getID(), (p: Point) => this._handleClickDown(p));
+      this._touchDispatcher.onTouchEnd("Interaction.Click" + this.getID(), (p: Point) => this._handleClickUp(p));
     }
 
-    public _requiresHitbox() {
-      return true;
+    private _handleClickDown(p: Point) {
+      if (this._isInsideComponent(p)) {
```
jtlan: Have to translate the point `p`:
```Typescript
var translatedP = this._translateToComponentSpace(p);
if (this._isInsideComponent(translatedP)) {
  ...
```

src/interactions/clickInteraction.ts
```diff
@@ -3,41 +3,58 @@
...
+        this._clickedDown = true;
+      }
     }
 
-    protected _listenTo(): string {
-      return "click";
+    private _handleClickUp(p: Point) {
+      if (this._clickedDown && this._isInsideComponent(p) && (this._clickCallback != null)) {
+        this._clickCallback(p);
```
jtlan: Same translation problem here.

PR #1769, opened 2015-03-26
fixing overlay tests

bluong:
The API-Break in the conversion of the `merge` API resulted in a bunch of API breaks in our quicktests


jtlan:
I think this can be NOQE?

bluong:
Probably.  Merging

PR #1770, opened 2015-03-26
[MouseDispatcher] - Adding onDblClick

bluong:
In preparation for DoubleClickInteraction / DblClickInteraction


jtlan:
I think that QE should only test `Interaction`s, and `Dispatcher`s should be handled via unit tests.

bluong:
Note that I did find an error in implementation because my unit test was incorrect and it was only in using the dispatcher that I realized that something was wrong

PR #1772, opened 2015-03-26
[SymbolDrawer] - Extends from ElementDrawer

### LINKED ISSUE #1771: Plot.Scatter shows invalid points
Malformed data is still drawn on `Plot.Scatter`:
screen shot 2015-03-26 at 3 22 04 pm
**Actual**: Big point drawn in upper-left corner, console shows "`Error: Invalid value for <path> attribute transform="translate(199.99999999999994,NaN) scale(0.06)"`"
**Expected**: Malformed data point is not drawn.

bluong:
`Drawer.Element` has functionality that should have been picked up in `Drawer.Symbol` but due to the lack of extension, that was unfortunately not picked up.  Now the hierarchy is made.

Fixes #1771

### DIFF
--- a/src/drawers/symbolDrawer.ts
+++ b/src/drawers/symbolDrawer.ts
@@ -2,23 +2,16 @@
 
 module Plottable {
 export module _Drawer {
-  export class Symbol extends AbstractDrawer {
+  export class Symbol extends Element {
 
-    protected _enterData(data: any[]) {
-      super._enterData(data);
-      var dataElements = this._getDrawSelection().data(data);
-      dataElements.enter().append("path");
-      dataElements.exit().remove();
-      dataElements.classed("symbol", true);
-    }
-
-    private _getDrawSelection() {
-      return this._getRenderArea().selectAll("path");
+    constructor(key: string) {
+      super(key);
+      this._svgElement = "path";
+      this._className = "symbol";
     }
 
     protected _drawStep(step: AppliedDrawStep) {
-      super._drawStep(step);
-      var attrToProjector = <AttributeToAppliedProjector>_Util.Methods.copyMap(step.attrToProjector);
+      var attrToProjector = step.attrToProjector;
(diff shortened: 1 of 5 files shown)


test/components/plots/scatterPlotTests.ts
```diff
@@ -111,6 +111,40 @@ describe("Plots", () => {
...
+      var plot = new Plottable.Plot.Scatter(xScale, yScale);
+      plot.addDataset(dataset)
+          .project("x", "foo", xScale)
+          .project("y", "bar", yScale);
+      plot.renderTo(svg);
+
+      var dataWithNaN = data.slice();
+      dataWithNaN[2] = { foo: 0.4, bar: NaN };
+      dataset.data(dataWithNaN);
+      assert.strictEqual(plot.getAllSelections().size(), 4, "does not retrieve NaN point");
```
jtlan: More accurately, "does not draw NaN point". Can we do this using direct D3 selection on `content()`, rather than using `getAllSelections()`, to make this independent?

src/drawers/symbolDrawer.ts
```diff
@@ -2,23 +2,16 @@
...
-    private _getDrawSelection() {
-      return this._getRenderArea().selectAll("path");
+    constructor(key: string) {
+      super(key);
+      this._svgElement = "path";
+      this._className = "symbol";
     }
 
     protected _drawStep(step: AppliedDrawStep) {
-      super._drawStep(step);
```
jtlan: No `super`?

bluong: The `super` is moved to the end

test/components/plots/scatterPlotTests.ts
bluong: The most independent would technically be using raw D3/Javascript on the svg and then searching using svg.selectAll("path")...  Should we use that instead?

jtlan: Maybe? I'm trying to flag that we're trying to assert that the point doesn't get drawn, which is subtly different from "doesn't get returned".

PR #1773, opened 2015-03-27
Non scaling stroke ie

bluong:


src/components/plots/scatterPlot.ts
```diff
@@ -35,6 +35,16 @@ export module Plot {
       attrToProjector["fill"] = attrToProjector["fill"] || d3.functor(this._defaultFillColor);
       attrToProjector["symbol"] = attrToProjector["symbol"] || Plottable.SymbolGenerators.d3Symbol("circle");
       attrToProjector["vector-effect"] = attrToProjector["vector-effect"] || d3.functor("non-scaling-stroke");
+
+      if (_Util.Methods.isIE()) {
+        var strokeWidthProjector = attrToProjector["stroke-width"];
+        attrToProjector["stroke-width"] = (d, i, u, m) => {
+          if (attrToProjector["vector-effect"](d, i, u, m) === "non-scaling-stroke") {
+            return strokeWidthProjector(d, i, u, m) * SymbolGenerators.SYMBOL_GENERATOR_RADIUS / attrToProjector["r"](d, i, u, m);
+          }
```
jtlan: This doesn't return anything in the case that `vector-effect` is not `non-scaling-stroke`. Also, wouldn't you want to perform the division if `vector-effect` is *not* `non-scaling-stroke`?

jtlan:
We should also label this with HACKHACK and the relevant ticket (https://connect.microsoft.com/IE/feedback/details/788819/svg-non-scaling-stroke)

crmorford:
http://jsfiddle.net/cg4zv553/2/

PR #1777, opened 2015-03-30
[BarChart] Hide labels that are cut off

### LINKED ISSUE #1728: Bar Chart Labels are only partially rendered if cut off by min/max
When enabling labels on a bar chart, where one bar is partially cut off by a min/max, the label is still rendered, but cut off. If the render is too small to show the label fully, the label should be removed.

Pic attached with a cut off label.

rcchen:
Fixes #1728 

JSFiddle: http://jsfiddle.net/rcchen/t648zgq2/

### DIFF
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -66,7 +66,10 @@ export module Plot {
       super._setup();
       this._renderArea = this._content.append("g").classed("render-area", true);
       // HACKHACK on 591
-      this._getDrawersInOrder().forEach((d) => d.setup(this._renderArea.append("g")));
+      this._getDrawersInOrder().forEach((d) => {
+        d.setup(this._renderArea.append("g"));
+        d.boundingBox(this._boundingBox);
+      });
     }
 
     public remove() {
--- a/src/components/plots/abstractPlot.ts
+++ b/src/components/plots/abstractPlot.ts
@@ -66,10 +66,7 @@ export module Plot {
       super._setup();
       this._renderArea = this._content.append("g").classed("render-area", true);
       // HACKHACK on 591
-      this._getDrawersInOrder().forEach((d) => {
-        d.setup(this._renderArea.append("g"));
-        d.boundingBox(this._boundingBox);
-      });
+      this._getDrawersInOrder().forEach((d) => d.setup(this._renderArea.append("g"), this._boundingBox));
     }
 
     public remove() {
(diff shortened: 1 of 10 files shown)


rcchen:
@bluong let me know if this should be approached differently; I'm using the bounding box on the parent plot to see if the label should be shown or not, similar to the approach that I took on Axis a while ago

bluong:
Preface to all of this:  Labels on the barPlot are not well-architected and thus should honestly be overall looked at.  Somewhat due to this, I'm slightly not too much in care for the architecture of this endpoint, but I do care in finding a good solution to this problem.  Thus, feel free to experiment or if that takes too much time, I'm going to care about the structure a bit less.

src/drawers/rectDrawer.ts
```diff
@@ -33,7 +33,8 @@ export module _Drawer {
       return this._labelsTooWide;
     }
 
-    public drawText(data: any[], attrToProjector: AttributeToProjector, userMetadata: any, plotMetadata: Plot.PlotMetadata) {
+    public drawText(data: any[], attrToProjector: AttributeToProjector, boundingBox: D3.Selection, userMetadata: any,
+        plotMetadata: Plot.PlotMetadata) {
```
bluong: Instead of needing to use the boundingBox here, I believe drawers have a renderArea selection that knows about the dimensions in which the drawer can render.  Can we use that instead?

rcchen: I tried it out and apparently we can't because the renderArea encapsulates the entire "bar render area" and does not actually match the visible area.

image

bluong: ?
Note that the renderArea from the drawer is gotten from the renderArea from the plot, which I believe was the reliable source of information for visible data

src/components/plots/abstractPlot.ts
```diff
@@ -66,7 +66,10 @@ export module Plot {
       super._setup();
       this._renderArea = this._content.append("g").classed("render-area", true);
       // HACKHACK on 591
-      this._getDrawersInOrder().forEach((d) => d.setup(this._renderArea.append("g")));
+      this._getDrawersInOrder().forEach((d) => {
+        d.setup(this._renderArea.append("g"));
+        d.boundingBox(this._boundingBox);
```
bluong: Should the _boundingBox be inputted into setup or do you think this is fine?

src/drawers/abstractDrawer.ts
```diff
@@ -19,6 +19,7 @@ export module _Drawer {
 
   export class AbstractDrawer {
     private _renderArea: D3.Selection;
+    protected _boundingBox: D3.Selection;
```
bluong: you can make this var private and use the getter

crmorford:
Because I'm backlogged, help me by making a jsfiddle repro?

rcchen:

PR #1778, opened 2015-03-31
[SymbolGenerator] - Rename to SymbolFactory, Only incorporates "r" arg, Legends take in accessor

bluong:
First off, we found that `SymbolGenerator` is quite the mouthful, so we are renaming the idea to the bit more concise and also recognizable `SymbolFactory`

The idea surrounding `SymbolFactories` is that they would work within a 100-by-100 pixel space and scale accordingly.  We found that this idea is not workable, specifically in Internet Explorer.

Thus, we are enforcing that `SymbolFactories` must take in an argument defining the radius of the symbol, and that no scaling will be done by Plottable to have the symbol render to the correct size.  Plottable will input a radius amount into the `SymbolFactory` and retrieve the symbol that way.

In consequence, `Legend`s no longer can take in a `SymbolFactory` and can only take in an accessor function that returns a `SymbolFactory` since this will cover many usecases and we have no feasible way to differentiate between the two.

API-Breaks:
* The `Plottable.SymbolGenerators` module has been renamed to `Plottable.SymbolFactories`

* `Plot.Scatter` no longer uses an "r" projection but instead a "size" projection which denotes the side length of a square that contains the symbol.

* A `SymbolFactory` is now defined as
```(size: number) => string```
where size is the side length of a square that contains the symbol.

* The d3Symbol `SymbolFactory` has been expanded into each of the symbol types as below
```typescript
Plottable.SymbolFactories.square();
Plottable.SymbolFactories.circle();
Plottable.SymbolFactories.diamond();
Plottable.SymbolFactories.triangleUp();
Plottable.SymbolFactories.triangleDown();
Plottable.SymbolFactories.cross();
```


jtlan:
Should **symbolGenerators.ts** be renamed to **symbolGeneratorAccessors.ts** ?

jtlan:
Also, I'm surprised that there are not tests that had to be drastically changed.

src/drawers/symbolDrawer.ts
```diff
@@ -23,13 +23,13 @@ export module _Drawer {
...
       attrToProjector["transform"] = (datum: any, index: number) =>
-        "translate(" + xProjector(datum, index) + "," + yProjector(datum, index) + ") " +
-        "scale(" + rProjector(datum, index) / 50 + ")";
+        "translate(" + xProjector(datum, index) + "," + yProjector(datum, index) + ")";
 
       var symbolProjector = attrToProjector["symbol"];
       delete attrToProjector["symbol"];
 
-      attrToProjector["d"] = symbolProjector;
+      attrToProjector["d"] = (datum: any, index: number) =>
```
jtlan: Shouldn't this occur in `generateAttrToProjector()` on `Plot.Scatter`?

For that matter, should we add ` = attrToProjector["d"] || ..." on the off-chance that some expert has decided to write a projector that directly returns path-strings?

src/components/legend.ts
```diff
@@ -300,23 +301,23 @@ export module Component {
...
+     * Sets the SymbolGeneratorAccessor of the legend
      *
      * @returns {Legend} The calling Legend
      */
-    public symbolGenerator(symbolGenerator: SymbolGenerator): Legend;
-    public symbolGenerator(symbolGenerator?: SymbolGenerator): any {
-      if (symbolGenerator == null) {
-        return this._symbolGenerator;
+    public symbolGeneratorAccessor(symbolGeneratorAccessor: SymbolGeneratorAccessor): Legend;
+    public symbolGeneratorAccessor(symbolGeneratorAccessor?: SymbolGeneratorAccessor): any {
```
jtlan: Why use this type rather than just `(d: any, i: number) => SymbolGenerator ?

typings/d3/d3.d.ts
```diff
@@ -1674,7 +1674,7 @@ declare module D3 {
         export interface Symbol {
             type: (symbolType: string | ((datum: any, index: number) => string)) => Symbol;
             size: (size: number | ((datum: any, index: number) => number)) => Symbol;
-            (datum:any, index:number): string;
+            (datum?: any, index?: number): string;
```
jtlan: Is there a PR going to DefinitelyTyped with this, or did we just upgrade the typing file?

src/components/legend.ts
bluong: Thought that `(d: any, i: number) => SymbolGenerator` was too much of a mouthful.  I could switch back, but `SymbolGeneratorAccessor` just seemed the easiest to use

typings/d3/d3.d.ts
bluong: I'll make the PR

src/components/legend.ts
jtlan: You still have to look up what `SymbolGeneratorAccessor` is though, so that's not really helpful.

typings/d3/d3.d.ts
bluong: https://github.com/borisyankov/DefinitelyTyped/pull/4017

src/components/legend.ts
bluong: I'll revert it and see how it looks.

src/components/legend.ts
```diff
@@ -300,23 +298,23 @@ export module Component {
     }
 
     /**
-     * Gets the SymbolGenerator of the legend, which dictates how
+     * Gets the SymbolGeneratorAccessor of the legend, which dictates how
      * the symbol in each entry is drawn.
      *
-     * @returns {SymbolGenerator} The SymbolGenerator of the legend
+     * @returns {symbolGeneratorAccessor} The SymbolGeneratorAccessor of the legend
```
jtlan: Docs should reflect new typing?

src/components/legend.ts
```diff
@@ -300,23 +298,23 @@ export module Component {
...
      */
-    public symbolGenerator(): SymbolGenerator;
+    public symbolGeneratorAccessor(): (datum: any, index: number) => SymbolGenerator;
     /**
-     * Sets the SymbolGenerator of the legend
+     * Sets the SymbolGeneratorAccessor of the legend
```
jtlan: missing @param in these docs.

bluong: Param is now set but the github diff doesnt show it...

jtlan:
Suggestions after talking to @derekcicerone:

Consider a shorter name than "SymbolGenerators", using a more common word than "generator". Maybe "SymbolFactory"?

We should probably hold off on the convenience method, to avoid giving users too many ways to do things that might confuse them.

bluong:
I'll go with "SymbolFactory" in this next commit but what about "SymbolMaker" ?

derekcicerone:
I think "maker" is too unfamiliar to the dev world.

Derek
Sent from a tablet

On Apr 2, 2015, at 3:31 AM, bluong <notifications@github.com<mailto:notifications@github.com>> wrote:


I'll go with "SymbolFactory" in this next commit but what about "SymbolMaker" ?

jtlan:
"`ShapeFactory`" ?

derekcicerone:
"Symbol" is fine if it matches the D3 community.

Derek
Sent from my iPhone

On Apr 2, 2015, at 1:27 PM, jtlan <notifications@github.com<mailto:notifications@github.com>> wrote:


"ShapeFactory" ?

PR #1779, opened 2015-03-31
[BarPlot] - Using different algorithm for finding closestPlotData

bluong:
Looks by Position then by Value

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -398,6 +398,7 @@ export module Plot {
             var valueDist = this._isVertical ? vertDist : horizDist;
 
             if (positionDist < closestPositionDistance && valueDist < withinValue) {
+              closestPositionDistance = positionDist;
               closestValueDistance = withinValue;
             }
 
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -379,6 +379,56 @@ export module Plot {
       return barPixelWidth;
     }
 
+    protected _getClosestPlotData(queryPoint: Point, datasetKeys: string[], withinValue = Infinity) {
+      var closestPositionDistance = withinValue;
+      var closestValueDistance = withinValue;
+
+      var closestDatum: any;
+      var closestPixelPoint: Point;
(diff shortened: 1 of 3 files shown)


src/components/plots/barPlot.ts
```diff
@@ -379,6 +379,57 @@ export module Plot {
...
+      return {data: [closestDatum],
+              pixelPoints: [closestPixelPoint],
+              selection: closestSelection};
+    }
+
+    private static _distanceFromRange(point: number, start: number, end: number) {
+      if (point < start) {
+        return start - point;
+      } else if (point > end) {
+        return end - point;
```
jtlan: This value will be negative, potentially resulting in some odd fail-out if used as a closest-point measurement.

src/components/plots/barPlot.ts
```diff
@@ -379,6 +379,57 @@ export module Plot {
...
+          var positionDist = this._isVertical ? horizDist : vertDist;
+          if (positionDist <= closestPositionDistance) {
+            var valueDist = this._isVertical ? vertDist : horizDist;
+
+            if (positionDist < closestPositionDistance && valueDist < withinValue) {
+              closestPositionDistance = positionDist;
+              closestValueDistance = withinValue;
+            }
+
+            if (valueDist < closestValueDistance) {
```
jtlan: This looks like it will override the closest information if `valueDist` is closer, even if `positionDist` is farther. That seems incorrect...

jtlan:
What is the expected behavior here? Can we get some tests for it?

src/components/plots/barPlot.ts
bluong: Note that this if is inside 
`positionDist <= closestPositionDistance`

bluong:
The expected behavior in the end should be to check the position distances of the bar (on a verticalBarPlot, check the x coordinate of the point against the x coordinate of the bars).  Of those distances that are closest, the value distances (y coordinate in the verticalBarPlot case) are checked to essentially act as a tiebreaker.

There is also a bias towards if the point is inside a bar

bluong:
Closing this as we're going with a different approach

PR #1780, opened 2015-03-31
[LinePlot] - getClosestPlotData algorithm change

bluong:
Uses X then Y.

### DIFF
--- a/src/components/plots/linePlot.ts
+++ b/src/components/plots/linePlot.ts
@@ -96,7 +96,8 @@ export module Plot {
     }
 
     protected _getClosestPlotData(queryPoint: Point, datasetKeys: string[], withinValue = Infinity) {
-      var closestDistanceSquared = withinValue;
+      var closestXDistance = withinValue;
+      var closestYDistance = withinValue;
       var closestDatum: any;
       var closestSelection: D3.Selection;
       var closestPoint: Point;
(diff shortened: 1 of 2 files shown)


src/components/plots/linePlot.ts
```diff
@@ -104,12 +105,20 @@ export module Plot {
...
-            closestSelection = plotData.selection;
+          var xDist = Math.abs(queryPoint.x - pixelPoint.x);
+          if (xDist <= closestXDistance) {
+            var yDist = Math.abs(queryPoint.y - pixelPoint.y);
+            if (xDist < closestXDistance && yDist < withinValue) {
+              closestXDistance = xDist;
+              closestYDistance = withinValue;
+            }
+
+            if (yDist < closestYDistance) {
```
jtlan: This seems incorrect. If `yDist` is less than the current closest, then the current closest gets overridden even though it might be closer by x.

jtlan:
I don't think "x then y" is a good default for `Plot.Line`, since we assume nothing about the ordering of the data values or the orientation of the plot. Either this algorithm should be a different function, or we should come up with some way for people to specify the distance function.

bluong:
Was wondering if this would be a more sensible default.  Your reasoning is sound, though.  Closing pull request.

PR #1783, opened 2015-04-02
XDragBoxLayer and YDragBoxLayer

jtlan:
Implemented `XDragBoxLayer` and `YDragBoxLayer`.

### DIFF
--- a/src/components/interactive/dragBoxLayer.ts
+++ b/src/components/interactive/dragBoxLayer.ts
@@ -22,8 +22,7 @@ export module Component {
       private _detectionCornerBR: D3.Selection;
 
       private _detectionRadius = 3;
-      protected _xResizable = false;
-      protected _yResizable = false;
+      private _resizable = false;
 
       private _dragStartCallback: (b: Bounds) => any;
       private _dragCallback: (b: Bounds) => any;
@@ -45,15 +44,7 @@ export module Component {
         var startedNewBox: boolean;
 
         this._dragInteraction.onDragStart((s: Point) => {
-          resizingEdges = this._getEdges(s);
-          if (!this._yResizable) {
-            resizingEdges.top = false;
-            resizingEdges.bottom = false;
-          }
-          if (!this._xResizable) {
-            resizingEdges.left = false;
-            resizingEdges.right = false;
-          }
+          resizingEdges = this._getResizingEdges(s);
 
           if (!this.boxVisible() ||
               (!resizingEdges.top && !resizingEdges.bottom &&
@@ -140,7 +131,7 @@ export module Component {
         this._detectionCornerBR = createCorner().classed("drag-corner-br", true);
       }
 
-      private _getEdges(p: Point) {
+      private _getResizingEdges(p: Point) {
         var edges = {
           top: false,
           bottom: false,
@@ -155,14 +146,15 @@ export module Component {
         var r = bounds.bottomRight.x;
         var rad = this._detectionRadius;
 
+        var resizable = this.resizable();
         if (l - rad <= p.x && p.x <= r + rad) {
-          edges.top = (t - rad <= p.y && p.y <= t + rad);
-          edges.bottom = (b - rad <= p.y && p.y <= b + rad);
+          edges.top = resizable && (t - rad <= p.y && p.y <= t + rad);
+          edges.bottom = resizable && (b - rad <= p.y && p.y <= b + rad);
         }
 
         if (t - rad <= p.y && p.y <= b + rad) {
-          edges.left = (l - rad <= p.x && p.x <= l + rad);
-          edges.right = (r - rad <= p.x && p.x <= r + rad);
+          edges.left = resizable && (l - rad <= p.x && p.x <= l + rad);
+          edges.right = resizable && (r - rad <= p.x && p.x <= r + rad);
         }
 
         return edges;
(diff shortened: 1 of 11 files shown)


src/components/interactive/yDragBoxLayer.ts
```diff
@@ -0,0 +1,20 @@
...
+    export class YDragBoxLayer extends DragBoxLayer {
+      protected _setBounds(newBounds: Bounds) {
+        super._setBounds({
+          topLeft: { x: 0, y: newBounds.topLeft.y },
+          bottomRight: { x: this.width(), y: newBounds.bottomRight.y }
+        });
+      }
+
+      protected _setResizableClasses(canResize: boolean) {
+        this.classed("y-resizable", canResize);
```
bluong: should "x-resizable" class be unset here or not required?

src/components/interactive/xDragBoxLayer.ts
```diff
@@ -0,0 +1,20 @@
...
+    export class XDragBoxLayer extends DragBoxLayer {
+      protected _setBounds(newBounds: Bounds) {
+        super._setBounds({
+          topLeft: { x: newBounds.topLeft.x, y: 0 },
+          bottomRight: { x: newBounds.bottomRight.x, y: this.height() }
+        });
+      }
+
+      protected _setResizableClasses(canResize: boolean) {
+        this.classed("x-resizable", canResize);
```
bluong: Same question as for y

src/components/interactive/yDragBoxLayer.ts
jtlan: Not required, but I think it might be good safety.

bluong: I personally like using `super`'s where applicable.  Not sure if applicable here.

Whatever you think here is good.  I don't think I currently have a strong opinion on how to do this well.

src/components/interactive/dragBoxLayer.ts
```diff
@@ -155,14 +146,15 @@ export module Component {
         var r = bounds.bottomRight.x;
         var rad = this._detectionRadius;
 
+        var resizable = this.resizable();
```
bluong: It seems like
This seems like
```
if (!this.resizable()) {
  return edges;
}
```

would work here and would skip all the complexity below.  Is this true?

English-wise, this would literally read, if the box isn't resizable then there are no resizing edges gotten

src/components/interactive/dragBoxLayer.ts
```diff
@@ -242,18 +234,17 @@ export module Component {
...
       public resizable(canResize?: boolean): any {
         if (canResize == null) {
-          return this._xResizable || this._yResizable;
+          return this._resizable;
         }
-        this._setResizable(canResize);
-        this.classed("x-resizable", this._xResizable);
-        this.classed("y-resizable", this._yResizable);
+        this._resizable = canResize;
+        this._setResizableClasses(canResize);
```
bluong: This seems a ton easier to read.  Nice cleanup

test/components/interactive/xDragBoxLayerTests.ts
```diff
@@ -0,0 +1,73 @@
...
+      var actualBounds = dbl.bounds();
+      var dragTo = {
+        x: SVG_WIDTH * 3 / 4,
+        y: SVG_HEIGHT / 2
+      };
+      var target = dbl.background();
+      triggerFakeDragSequence(target, actualBounds.bottomRight, dragTo);
+      actualBounds = dbl.bounds();
+      assert.strictEqual(actualBounds.bottomRight.x, dragTo.x, "resized in x");
+      assert.strictEqual(actualBounds.bottomRight.y, dbl.height(), "height did not change");
```
bluong: the test message does not actually suggest what is being tested.

The test currently reads, "assert that the bottom of the bounds is at the bottom of the layer after this dragTo".  We will probably want this test but this is the wrong message for it.

We do however want the test with the message "height did not change", in which case, you should find the oldHeight before the dragSequence (using bounds.bottom.y - bounds.top.y) and compare it to the height after the dragSequence and then assert these 2 are the same.

test/components/interactive/yDragBoxLayerTests.ts
```diff
@@ -0,0 +1,73 @@
...
+        x: SVG_WIDTH / 2,
+        y: SVG_HEIGHT * 3 / 4
+      };
+      var target = dbl.background();
+      triggerFakeDragSequence(target, actualBounds.bottomRight, dragTo);
+      actualBounds = dbl.bounds();
+      assert.strictEqual(actualBounds.bottomRight.x, dbl.width(), "width did not change");
+      assert.strictEqual(actualBounds.bottomRight.y, dragTo.y, "resized in y");
+      svg.remove();
+    });
```
bluong: Same as above.

src/components/interactive/dragBoxLayer.ts
jtlan: Nice catch.

src/components/interactive/yDragBoxLayer.ts
jtlan: It's odd in this case because the subclass has *less* functionality than its parent, so we don't really want to call `super`.

bluong:
What about an X/YSelectionBoxLayer

src/components/interactive/xDragBoxLayer.ts
```diff
@@ -0,0 +1,27 @@
+///<reference path="../../reference.ts" />
+
+module Plottable {
+export module Component {
+  export module Interactive {
+    export class XDragBoxLayer extends DragBoxLayer {
+      protected _hasCorners = false;
```
bluong: Set this to false in the constructor

src/components/interactive/yDragBoxLayer.ts
```diff
@@ -0,0 +1,27 @@
+///<reference path="../../reference.ts" />
+
+module Plottable {
+export module Component {
+  export module Interactive {
+    export class YDragBoxLayer extends DragBoxLayer {
+      protected _hasCorners = false;
```
bluong: Same as above.

test/components/interactive/xDragBoxLayerTests.ts
```diff
@@ -0,0 +1,74 @@
...
+      var dragTo = {
+        x: SVG_WIDTH * 3 / 4,
+        y: SVG_HEIGHT / 2
+      };
+      var target = dbl.background();
+      triggerFakeDragSequence(target, actualBounds.bottomRight, dragTo);
+      actualBounds = dbl.bounds();
+      assert.strictEqual(actualBounds.bottomRight.x, dragTo.x, "resized in x");
+      assert.strictEqual(actualBounds.topLeft.y, 0, "box still starts at top");
+      assert.strictEqual(actualBounds.bottomRight.y, dbl.height(), "box still ends at bottom");
```
bluong: Do we want the previous test about the height staying the same to exist still?

jtlan: That test was trying to test what was tested here.

PR #1784, opened 2015-04-02
[Interaction.DoubleClick] - Move to new architecture

bluong:
`Interaction.DoubleClick` has been remade to conform to the new architecture.  In accordance, there is no longer a `callback` endpoint but instead a `onDblClick` endpoint that takes in a callback as the below signature:

```typescript
onDblClick(callback: (p: Point) => any)
```

The double click callback is triggered when Plottable knows that a user has clicked down, up, down then up at the same location within a small amount of time.  This small amount of time is system dependent.

The `onDblClick` endpoint can also be used as a getter

Example JSFiddle: http://jsfiddle.net/te4kzpp4/1/


rcchen:
Thought that interactions might not be testable but it looks like we have unit tests for existing interactions, so add them for DoubleClick?

jtlan:
Second fiddle, for testing what happens when the first and second clicks don't occur in the same `Component`:
http://jsfiddle.net/w2mzg137/

jtlan:
Also, I think we should use `onDoubleClick()` instead of `onDblClick()`. Full names when possible.

crmorford:
This does not work on touch. We have to decide if that's okay or not.

PR #1786, opened 2015-04-03
[Quicktest] - clickInteraction fixes

bluong:


bluong:
This seems like NOQE. I'll verify the quicktests on my end and see if this can merge.

PR #1787, opened 2015-04-03
[AbstractPlot] - getPlotData()

bluong:
Similar to getClosestPlotData / getBars, getPlotData takes in 2 values (x, y) or extents (xExtent, yExtent) to determine all of the plot data that is contained at or in the 2 arguments.

Behavior is different for each plot.

AbstractPlot - Draw a BBox over the element and check if there is an intersection

ScatterPlot - Check if the point is at the same query point or within the extents

LinePlot - Check if the individual points are at the same query point or within the extents

PiePlot - Check if the sectors are in intersection with the query point or the center of the extents


src/components/plots/abstractPlot.ts
```diff
@@ -556,6 +556,54 @@ export module Plot {
...
+      return (typeof input === "number") ? {min: <number> input, max: <number> input} : <Extent> input;
+    }
+
+    protected _getPlotData(xExtent: Extent, yExtent: Extent, datasetKeys: string[]): PlotData {
+      var data: any[] = [];
+      var pixelPoints: Point[] = [];
+      var selections: EventTarget[] = [];
+      var plotData = this.getAllPlotData(datasetKeys);
+      plotData.selection.each(function(datum, index) {
+        var bbox = this.getBBox();
```
jtlan: This parsing may not be safe for some types of Plots that nest their selections...

bluong: I could move this down to the RectanglePlot level if that's preferable as this implementation makes the most sense there.  I am unsure on what could be a valid generic implementation (maybe not have one?)

jtlan:
Probably better to break this into two methods:
```Typescript
plotDataAtPoint(p: Point): PlotData;
plotDataInRange(xRange: number[], number[]): PlotData;
```
You can use the `type point = number` trick on the ranges if you want to make it clearer.

src/components/plots/abstractPlot.ts
jtlan: I think we should probably just use direct pixelPoint comparison on the abstract level. This does mean that trying to get the plot data with a specific point instead of ranges will fail ... but that makes sense.

src/components/plots/linePlot.ts
```diff
@@ -123,6 +123,31 @@ export module Plot {
...
     }
 
+    protected _getPlotData(xExtent: Extent, yExtent: Extent, datasetKeys: string[]): PlotData {
+      var data: any[] = [];
+      var pixelPoints: Point[] = [];
+      var selections: EventTarget[] = [];
+      datasetKeys.forEach((datasetKey) => {
+        var plotData = this.getAllPlotData(datasetKey);
+        var lineSelected = false;
+        plotData.pixelPoints.forEach((pixelPoint, index) => {
```
jtlan: [`Array.filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) just seems like a better way to do this. You can check if the line is selected afterwards by seeing if the filtered array is non-empty.

rcchen:

PR #1792, opened 2015-04-03
Added sinusoidal data generator for exampleUtils

rcchen:


bluong:
Note that while this data generator seems useful... I don't personally use the exampleUtils functions at all.  I usually just make the data myself.  If this is useful to people though, I don't see much reason not to let this through.

PR #1794, opened 2015-04-03
[StackedBar/ClusteredBar] - Adding missing values on horizontal version unit tests

bluong:


test/components/plots/stackedBarPlotTests.ts
```diff
@@ -301,4 +301,66 @@ describe("Plots", () => {
       svg.remove();
     });
   });
+
+  describe("Horizontal Stacked Bar Plot Weird Values", () => {
```
jtlan: "with missing values"? instead of "weird values"?

Issue #1255, opened 2014-10-27
Setting alignment of Components in a Group doesn't work correctly


jtlan:
![screen shot 2014-10-27 at 11 23 18 am](https://cloud.githubusercontent.com/assets/1440449/4796817/5b9fe5ac-5e06-11e4-964b-e6a36c6e2ebf.png)

http://jsfiddle.net/c0kskauq/

Actual: Labels bunch up together
Expected: Each label is aligned the way they would be without a `Component.Group`.

PR #1799, opened 2015-04-05
Make Component.Group expand only if necessary

### LINKED ISSUE #1791: Grouping fixed-size Components produces unexpected behavior
Chart with an `Axis` at the bottom: http://jsfiddle.net/yzfauLe7/
screen shot 2015-04-03 at 1 29 49 pm

Chart with a `Label` at the bottom: http://jsfiddle.net/yzfauLe7/1/
screen shot 2015-04-03 at 1 30 25 pm

Chart with `Group` containing an `Axis` and a `Label`: http://jsfiddle.net/yzfauLe7/2/
screen shot 2015-04-03 at 1 31 14 pm

????

jtlan:
Now, if a Group contains only fixed-width Components, its width will
expand to fill the available space only if those Components do not
share the same x alignment. Same for height, based on y alignment.

Added "Mocks" module which contains classes for use in testing.

Wrote better tests for checking Group behavior based on contents.
Moved a misplaced test.

Close #1791.

### DIFF
--- a/src/components/label.ts
+++ b/src/components/label.ts
@@ -9,8 +9,6 @@ export module Component {
     private _measurer: SVGTypewriter.Measurers.Measurer;
     private _wrapper: SVGTypewriter.Wrappers.Wrapper;
     private _writer: SVGTypewriter.Writers.Writer;
-    private _xAlignment: string;
-    private _yAlignment: string;
     private _padding: number;
 
     /**
@@ -32,34 +30,6 @@ export module Component {
       this._fixedHeightFlag = true;
       this._fixedWidthFlag = true;
       this._padding = 0;
-    }
-
-    /**
-     * Sets the horizontal side the label will go to given the label is given more space that it needs
-     *
-     * @param {string} alignment The new setting, one of `["left", "center",
-     * "right"]`. Defaults to `"center"`.
-     * @returns {Label} The calling Label.
-     */
-    public xAlign(alignment: string): Label {
-      var alignmentLC = alignment.toLowerCase();
-      super.xAlign(alignmentLC);
-      this._xAlignment = alignmentLC;
-      return this;
-    }
-
-    /**
-     * Sets the vertical side the label will go to given the label is given more space that it needs
-     *
-     * @param {string} alignment The new setting, one of `["top", "center",
-     * "bottom"]`. Defaults to `"center"`.
-     * @returns {Label} The calling Label.
-     */
-    public yAlign(alignment: string): Label {
-      var alignmentLC = alignment.toLowerCase();
-      super.yAlign(alignmentLC);
-      this._yAlignment = alignmentLC;
-      return this;
     }
 
     public _requestedSpace(offeredWidth: number, offeredHeight: number): _SpaceRequest {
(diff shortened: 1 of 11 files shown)


jtlan:
Marking this as [BLOCKER] because it interferes with desired behavior when merging `SelectionBoxLayer` with fixed-width and fixed-height `Component`s (like an `Axis` of any type).

jtlan:
The intended behavior should produce sensible results in general, and also not break #1255...

src/components/label.ts
```diff
@@ -34,34 +32,6 @@ export module Component {
...
-     * @param {string} alignment The new setting, one of `["top", "center",
-     * "bottom"]`. Defaults to `"center"`.
-     * @returns {Label} The calling Label.
-     */
-    public yAlign(alignment: string): Label {
-      var alignmentLC = alignment.toLowerCase();
-      super.yAlign(alignmentLC);
-      this._yAlignment = alignmentLC;
-      return this;
-    }
```
bluong: Yay for code removal

test/core/componentGroupTests.ts
```diff
@@ -164,35 +126,80 @@ describe("ComponentGroups", () => {
...
+
+    it("fixed-size Components, same x and same y alignments", () => {
+      var svg = generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var tall = new Mocks.FixedSizeComponent(SVG_WIDTH / 4, SVG_HEIGHT / 2);
+      var wide = new Mocks.FixedSizeComponent(SVG_WIDTH / 2, SVG_HEIGHT / 4);
+      var cg = new Plottable.Component.Group([ tall, wide ]);
+      cg.renderTo(svg);
+      var largestWidth = cg.components().map((c) => c.width()).sort((a, b) => b - a)[0];
+      assert.strictEqual(cg.width(), largestWidth, "takes on the width of widest fixed-size component");
+      var largestHeight = cg.components().map((c) => c.height()).sort((a, b) => b - a)[0];
```
bluong: Instead use the _Util.max

test/mocks.ts
```diff
@@ -0,0 +1,25 @@
+///<reference path="testReference.ts" />
+
+module Mocks {
```
bluong: Consider maybe incorporating it in the Plottable module?

test/mocks.ts
```diff
@@ -0,0 +1,25 @@
...
+  export class FixedSizeComponent extends Plottable.Component.AbstractComponent {
```
bluong: Is there not some Javscript mocking infrastructure in place? sinon / etc.?

bluong:
We should see if we can improve on this testing infrastructure.  There is the SinonJS library that I believe is used sparingly in our codebase that we can possibly use more to our advantage.  Would using that be a good idea?

jtlan:
Sinon only creates fake functions, not objects. However, I think it might be better if we change the module name to "`Stubs`".

test/mocks.ts
jtlan: It doesn't belong there since it's test code, rather than part of the main library.

jtlan:
I think I came up with a better way to do this.

PR #1803, opened 2015-04-07
[Table] - addComponent to already occupied cell appends component

### LINKED ISSUE #1714: Adding Component to filled cell should automatically create Group
http://jsfiddle.net/kykxLta8/1/

Trying to add a `Component` to a `Table` cell that is already occupied yields the message:
```
Table.addComponent cannot be called on a cell where a component already exists (for the moment)
```

Should we add in this functionality?

aicioara:
Closes #1714 
As per title. addComponent used to throw an error if the cell was already containing a component. Now it just appends the component to a group (and creates the group if necessary). Unit test included.

### DIFF
--- a/test/core/tableTests.ts
+++ b/test/core/tableTests.ts
@@ -78,7 +78,7 @@ describe("Tables", () => {
     t.addComponent(0, 0, c2);
     t.addComponent(0, 2, c3);
 
-    assert.isTrue(Plottable.Component.Group.prototype.isPrototypeOf((<any> t)._rows[0][2]));
+    assert.isTrue(Plottable.Component.Group.prototype.isPrototypeOf((<any> t)._rows[0][2]), "A group was created");
 
     var components: Plottable.Component.AbstractComponent[] = (<any> t)._rows[0][2].components();
     assert.lengthOf(components, 2, "The group created should have 2 components");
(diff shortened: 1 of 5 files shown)


test/core/tableTests.ts
```diff
@@ -68,14 +68,41 @@ describe("Tables", () => {
...
     var c2 = new Plottable.Component.AbstractComponent();
     var c3 = new Plottable.Component.AbstractComponent();
     var t = new Plottable.Component.Table();
+
     t.addComponent(0, 2, c1);
     t.addComponent(0, 0, c2);
-    assert.throws(() => t.addComponent(0, 2, c3), Error, "component already exists");
+    t.addComponent(0, 2, c3);
+
+    assert.isTrue(Plottable.Component.Group.prototype.isPrototypeOf((<any> t)._rows[0][2]));
```
jtlan: Probably add a message to this `assert` statement, something like "A group was created". Consider that if this test fails all that the developer will see is "Expected false to be true".

test/core/tableTests.ts
```diff
@@ -68,14 +68,41 @@ describe("Tables", () => {
...
+
+  it("Add a component where a group already exists adds the component to the group", () => {
+    var c1 = new Plottable.Component.AbstractComponent();
+    var c2 = new Plottable.Component.AbstractComponent();
+    var c3 = new Plottable.Component.AbstractComponent();
+    var t = new Plottable.Component.Table();
+
+    t.addComponent(0, 2, c1);
+    t.addComponent(0, 2, c2);
+    t.addComponent(0, 2, c3);
```
jtlan: I'd prefer that instead of doing this, you create a `Group` and add it to the cell. This makes this test independent of the previous one (which checks for `Group` creation on collision).

test/core/tableTests.ts
```diff
@@ -68,14 +68,41 @@ describe("Tables", () => {
...
+
+    assert.isTrue(Plottable.Component.Group.prototype.isPrototypeOf((<any> t)._rows[0][2]));
```
jtlan: Same comment as above regarding message.

test/core/tableTests.ts
```diff
@@ -68,14 +68,41 @@ describe("Tables", () => {
...
+
+    var components: Plottable.Component.AbstractComponent[] = (<any> t)._rows[0][2].components();
+    assert.lengthOf(components, 3, "The group created should have 2 components");
```
jtlan: 3 `Component`s?

test/core/tableTests.ts
```diff
@@ -68,14 +68,41 @@ describe("Tables", () => {
...
+    assert.equal(components[0], c1, "First element in the group at (0, 2) should be c1");
+    assert.equal(components[1], c2, "Second element in the group at (0, 2) should be c2");
+    assert.equal(components[2], c3, "Third element in the group at (0, 2) should be c3");
```
jtlan: If you change this to use a pre-created `Group`, the `assert` statement could probably be changed to check that "The `Component` was added to the existing `Group`".

bluong:
In the future @aicioara , the title of the pull request should be a small descriptor of the pull request contents, with the description being a bit more elaborate.  I personally like having the title pinpoint exactly which piece of the codebase has been affected, even though this is a personal style.

For example, I would maybe name this pr title to
"[Table] - addComponent to already occupied cell appends component"

possibly

PR #1804, opened 2015-04-07
[DONOTMERGE] [Behavior.DragPan] - Initial Implementation

bluong:
Mainly leaving this pull request up here for comments on how this idea is implemented.

The file to look at is dragPanBehavior.ts

I'm currently thinking of 3 API namespaces on where to put this behavior:
 * Plottable.Behavior.DragPan - (current) Recognizes the fact that it is a DragInteraction that initiates the behavior.
 * Plottable.Behavior.Pan.Drag - Recognizes that DragInteraction will initiate the behavior and also shelves it into a nice Pan module.
 * Plottable.Behavior.Pan - The way to get at the DragInteraction will be possibly through a getDragInteraction call, which means that if we want to develop more interactions dealing with Pan, we build another getXInteraction API endpoint rather than make a new class


src/behaviors/dragPanBehavior.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Creates a DragPan Behavior.
+     *
+     * The allows you to move around and zoom in on a plot, interactively. It
+     * does so by changing the xScale and yScales' domains repeatedly.
+     *
+     * @constructor
+     * @param {QuantitativeScale} [xScale] The X scale to update on panning/zooming.
+     * @param {QuantitativeScale} [yScale] The Y scale to update on panning/zooming.
```
jtlan: wrong jsdoc?

src/behaviors/dragPanBehavior.ts
```diff
@@ -0,0 +1,89 @@
...
+      this._setupInteraction(this._dragInteraction);
+      this._verticalPan = isVertical;
+    }
+
+    public getInteraction(): Interaction.Drag {
+      return this._dragInteraction;
+    }
+
+    public leftBounds(): number[];
+    public leftBounds(newBounds: number[]): Behavior.DragPan;
```
jtlan: Why does this take an array and not a number?

src/behaviors/dragPanBehavior.ts
```diff
@@ -0,0 +1,89 @@
...
+        var leftRangeValues = this.leftBounds().map((leftBound) => safeScale(leftBound) - this._scale.range()[0]);
+        var rightRangeValues = this.rightBounds().map((rightBound) => safeScale(rightBound) - this._scale.range()[1]);
+
+        if (dragAmount > 0) {
+          dragAmount = Math.min(dragAmount, -Math.max(leftRangeValues[0], rightRangeValues[0]));
+        } else {
+          dragAmount = Math.max(dragAmount, -Math.min(leftRangeValues[1], rightRangeValues[1]));
+        }
+
+        ScaleDomainTransformers.translate(this._scale, dragAmount);
```
jtlan: If the API was non-destructive here, it might be easier to implement the transform, since you could just no-op if the transformed domain was too far away.

bluong: I don't think that is the behavior that we would prefer.  Imagine if a consumer were to pan to one of the extreme edges.  With the no-op behavior, the panning could end at some random point away from the edge.  What this would do is actually set the panning to the extreme edge, which I believe is better behavior.

src/behaviors/dragPanBehavior.ts
bluong: min/max

src/behaviors/dragPanBehavior.ts
bluong: The doc is incorrect.  I considered having an xScale/yScale but then all of the below API endpoints would have to incorporate the addition of other scales, which is why I went for the simpler single scale constructor appraoch.

src/behaviors/dragPanBehavior.ts
bluong: To be more explicit, this is so users can enact the min/max of the left edge of the domain.

bluong:
Behaviors are not too distinct from Interactions to the point where they need to reside in a separate module.  Closing this.

PR #1805, opened 2015-04-07
Enable clipPath on DragBoxLayer.

jtlan:
Previously, a `DragBoxLayer` whose box was dragged all the way to the edge would show a "resizing" cursor, even though the user's cursor might be outside the `DragBoxLayer` and hence unable to click on it. This is because the hover-sensitive area is larger than the box itself.

Resolved by using a clipPath to hide the parts of the drag-box that are outside the `DragBoxLayer`.

### DIFF
--- a/test/components/interactive/dragBoxLayerTests.ts
+++ b/test/components/interactive/dragBoxLayerTests.ts
@@ -50,6 +50,15 @@ describe("Interactive Components", () => {
       svg.remove();
     });
 
+    it("clipPath enabled", () => {
+      /*
+       * clipPath has to be enabled so the resize handle doesn't show on hovering
+       * over the edge of the drag box
+       */
+      var dbl = new Plottable.Component.Interactive.DragBoxLayer();
+      assert.isTrue(dbl.clipPathEnabled);
+    });
+
     it("detectionRadius()", () => {
       var dbl = new Plottable.Component.Interactive.DragBoxLayer();
 
--- a/test/components/interactive/dragBoxLayerTests.ts
+++ b/test/components/interactive/dragBoxLayerTests.ts
@@ -51,12 +51,8 @@ describe("Interactive Components", () => {
     });
 
     it("clipPath enabled", () => {
-      /*
-       * clipPath has to be enabled so the resize handle doesn't show on hovering
-       * over the edge of the drag box
-       */
       var dbl = new Plottable.Component.Interactive.DragBoxLayer();
-      assert.isTrue(dbl.clipPathEnabled);
+      assert.isTrue(dbl.clipPathEnabled, "uses clipPath (to hide detection edges)");
     });
 
     it("detectionRadius()", () => {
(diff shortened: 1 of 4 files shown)


test/components/interactive/dragBoxLayerTests.ts
```diff
@@ -50,6 +50,15 @@ describe("Interactive Components", () => {
       svg.remove();
     });
 
+    it("clipPath enabled", () => {
+      /*
+       * clipPath has to be enabled so the resize handle doesn't show on hovering
+       * over the edge of the drag box
+       */
```
bluong: Let's put this comment either in the pull request description or in the source code as a comment.  Putting it here is not visible to anyone really.

bluong: In regards to this comment, is this mainly to avoid having the drag box spill over the edge of the cell?  What's the context of this change?

Can you provide context in the pull request description?

bluong:
How does this idea interact with the clamping idea on `Interaction.Drag`?

bluong:
The implementation looks simple enough.  I just have questions regarding consequences of this fix

bluong:
Ok now that the comment is there, this issue makes a lot more sense.  Thus, a few questions...

* Would this happen mainly if the DragInteraction did not clamp inside the component space? (Interactivity can happen outside)

* Unless I'm mistaken the corners are set through the setBounds call correct?  Does that mean that the box also runs outside the boundaries of the component space?  I guess a better question to ask is why doesn't bounds clamp to the component space, meaning that this issue would be inherently solved that way?

jtlan:
@bluong wrote:
> Would this happen mainly if the DragInteraction did not clamp inside the component space? (Interactivity can happen outside)

In that case, the user could drag the edge of the box outside the visible area, which isn't desirable.

>Unless I'm mistaken the corners are set through the setBounds call correct? Does that mean that the box also runs outside the boundaries of the component space? I guess a better question to ask is why doesn't bounds clamp to the component space, meaning that this issue would be inherently solved that way?

There are cases where you would want to set the `bounds()` of a `SelectionBoxLayer` outside the visible area -- when the bounds are driven by an external source for example. In this case, clamping wouldn't solve the entire problem, because the position of a `<line>` (like the detection edges) describes the center, not the outer edge, so half the stroke would be outside even if, as I am doing, the bounds are clamped to the visible area.

> Mainly a question on clip-path, but if you were to enable clip-path hereand merge this component with another component with clip-path disabled, what happens?

clipPath only affects the `Component` it is on (unless somehow you enable it on a `Group` or `Table`).

bluong:
Almost everything makes sense now.  Just curious on one question now

Would setting the `bounds()` of a `SelectionBoxLayer` to outside the visible area lead to bugs?  Imagine a `getPlotData` call that uses the `bounds()` of the `SelectionBoxLayer` to determine in the extents to search in.

jtlan:
@bluong wrote:
>Would setting the `bounds()` of a `SelectionBoxLayer` to outside the visible area lead to bugs?  Imagine a `getPlotData` call that uses the `bounds()` of the `SelectionBoxLayer` to determine in the extents to search in.

No, I think this is intended. Consider the case where someone saves off the selection, then pans, then re-applies the saved selection. The box might be off the screen, but it should still be drawn at its actual position.

Issue #1800, opened 2015-04-06
One tier time axis configurations leaves white space in the unused second tier


vivianliang:
One tier time axis configurations, for example:
```
[
    { interval: d3.time.month, step: 1, formatter: Plottable.Formatters.time("%b d") }
]
```
 leaves white space in the unused second tier instead of eliminating the second tier altogether.

jtlan:
Thanks for flagging this!

In the case where the `Axis.Time` has some configurations that use two tiers, this is intentional. However, in your example, the `Axis.Time` should only occupy one tier because it never has more than one tier.

Repro: http://jsfiddle.net/9kk194m6/1/

aicioara:
@jtlan would we ever want to support more than 2 tiers?

jtlan:
@aicioara We thought about doing that. I'm not sure whether that's best done by adding the ability to specify more configurations, or by coming up with a good way to stack multiple axes (using a `Table).

jtlan:
More tiers tracked at #1820.

PR #1806, opened 2015-04-07
[timeAxis] Height computed from maximum number of tiers

aicioara:
Before (see URL) we were always setting the number of tiers to the maximum value (2). That meant that if we were only ever using 1 tier, there would be whitespace reserved for the second tier, which will never be used.

We now look through configurations and derive the maximum number of tiers.

Catch: If there is a chance to ever display 2 tiers, we will leave space for 2 tiers even when only 1 tier is displayed, because we don't want the graph to jump on resize of the axis

Also included small refactoring:
- added a private static default configurations _DEFAULT_TIME_AXIS_CONFIGURATIONS
All should help readability 

Before: http://jsfiddle.net/9kk194m6/2/
After: http://jsfiddle.net/6yvtwzvm/

### DIFF
--- a/src/components/axes/timeAxis.ts
+++ b/src/components/axes/timeAxis.ts
@@ -27,7 +27,7 @@ export module Axis {
     /*
      * Default possible axis configurations.
      */
-    private _possibleTimeAxisConfigurations: TimeAxisConfiguration[] = [
+    private _DEFAULT_TIME_AXIS_CONFIGURATIONS: TimeAxisConfiguration[] = [
       [
         {interval: d3.time.second, step: 1, formatter: Formatters.time("%I:%M:%S %p")},
         {interval: d3.time.day, step: 1, formatter: Formatters.time("%B %e, %Y")}
@@ -141,7 +141,8 @@ export module Axis {
     private _tierMarkContainers: D3.Selection[];
     private _tierBaselines: D3.Selection[];
     private _tierHeights: number[];
-    private _maximumTires: number = 2;
+    private _currentTimeAxisConfigurations: TimeAxisConfiguration[];
+    private _maximumTires: number;
     private _measurer: SVGTypewriter.Measurers.Measurer;
 
     private _mostPreciseConfigIndex: number;
@@ -169,6 +170,7 @@ export module Axis {
       this.classed("time-axis", true);
       this.tickLabelPadding(5);
       this.tierLabelPositions(["between", "between"]);
+      this.axisConfigurations(this._DEFAULT_TIME_AXIS_CONFIGURATIONS);
     }
 
     public tierLabelPositions(): string[];
@@ -203,10 +205,10 @@ export module Axis {
     public axisConfigurations(configurations: TimeAxisConfiguration[]): Time;
     public axisConfigurations(configurations?: any): any {
       if(configurations == null){
-        return this._possibleTimeAxisConfigurations;
+        return this._currentTimeAxisConfigurations;
       }
-      this._possibleTimeAxisConfigurations = configurations;
-      this._maximumTires = Math.max.apply(null, configurations.map(function(e) {return e.length}));
+      this._currentTimeAxisConfigurations = configurations;
+      this._maximumTires = Math.max.apply(null, configurations.map((config: TimeAxisConfiguration) => config.length));
       this._invalidateLayout();
       return this;
     }
@@ -215,15 +217,15 @@ export module Axis {
      * Gets the index of the most precise TimeAxisConfiguration that will fit in the current width.
      */
     private _getMostPreciseConfigurationIndex(): number {
-      var mostPreciseIndex = this._possibleTimeAxisConfigurations.length;
-      this._possibleTimeAxisConfigurations.forEach((interval: TimeAxisConfiguration, index: number) => {
+      var mostPreciseIndex = this._currentTimeAxisConfigurations.length;
+      this._currentTimeAxisConfigurations.forEach((interval: TimeAxisConfiguration, index: number) => {
         if (index < mostPreciseIndex && interval.every((tier: TimeAxisTierConfiguration) =>
           this._checkTimeAxisTierConfigurationWidth(tier))) {
           mostPreciseIndex = index;
         }
       });
 
-      if (mostPreciseIndex === this._possibleTimeAxisConfigurations.length) {
+      if (mostPreciseIndex === this._currentTimeAxisConfigurations.length) {
         _Util.Methods.warn("zoomed out too far: could not find suitable interval to display labels");
         --mostPreciseIndex;
       }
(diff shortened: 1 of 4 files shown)


aicioara:
Closes #1800

aicioara:
Fails the build. Hang on...

src/components/axes/timeAxis.ts
```diff
@@ -141,6 +141,8 @@ export module Axis {
     private _tierMarkContainers: D3.Selection[];
     private _tierBaselines: D3.Selection[];
     private _tierHeights: number[];
+    private _currentTimeAxisConfigurations: TimeAxisConfiguration[];
+    private _maximumTires: number;
```
bluong: _maximumTiers?

src/components/axes/timeAxis.ts
```diff
@@ -202,9 +205,10 @@ export module Axis {
     public axisConfigurations(configurations: TimeAxisConfiguration[]): Time;
     public axisConfigurations(configurations?: any): any {
       if(configurations == null){
-        return this._possibleTimeAxisConfigurations;
+        return this._currentTimeAxisConfigurations;
       }
-      this._possibleTimeAxisConfigurations = configurations;
+      this._currentTimeAxisConfigurations = configurations;
+      this._maximumTires = Math.max.apply(null, configurations.map((config: TimeAxisConfiguration) => config.length));
```
bluong: Use _Util.max

src/components/axes/timeAxis.ts
```diff
@@ -202,9 +205,10 @@ export module Axis {
     public axisConfigurations(configurations: TimeAxisConfiguration[]): Time;
     public axisConfigurations(configurations?: any): any {
       if(configurations == null){
-        return this._possibleTimeAxisConfigurations;
+        return this._currentTimeAxisConfigurations;
       }
-      this._possibleTimeAxisConfigurations = configurations;
+      this._currentTimeAxisConfigurations = configurations;
+      this._maximumTiers = _Util.Methods.max(configurations.map((config: TimeAxisConfiguration) => config.length), 0);
```
bluong: Is this variable needed as an instance variable?

src/components/axes/timeAxis.ts
```diff
@@ -240,8 +244,12 @@ export module Axis {
 
     public _computeHeight() {
       var textHeight = this._measurer.measure().height;
-      this._tierHeights = this._tierLabelPositions.map((pos: string) =>
-        textHeight + this.tickLabelPadding() + ((pos === "between") ? 0 : this._maxLabelTickLength()));
+      this._tierHeights = this._tierLabelPositions.map((pos: string, index: number) => {
+        if (index >= this._maximumTiers) {
+          return 0;
+        }
```
bluong: Why is this._tierLabelPositions.length > this._maximumTiers to begin with?

src/components/axes/timeAxis.ts
jtlan: Yes, since it depends on the configuration.

src/components/axes/timeAxis.ts
```diff
@@ -141,6 +141,8 @@ export module Axis {
     private _tierMarkContainers: D3.Selection[];
     private _tierBaselines: D3.Selection[];
     private _tierHeights: number[];
+    private _currentTimeAxisConfigurations: TimeAxisConfiguration[];
```
jtlan: Why was the name of this variable changed? "current" is probably not the right description here...

src/components/axes/timeAxis.ts
jtlan: Say someone sets the tier label positions, then sets the number of tiers in the configuration to a smaller number. Now they're out of sync...

jtlan:
Some comments. Also, please write a unit test that checks for the bug you fixed.

src/components/axes/timeAxis.ts
aicioara: Should I just leave it as it was before? I thought that `_possibleTimeAxisConfigurations` is an inappropriate name

src/components/axes/timeAxis.ts
aicioara: it depends how often the `_computeHeight()` is called. I will double check tomorrow. I personally prefer it not to be an instance variable, but it might be an overhead to calculate the max number of Tiers every time we compute the height

src/components/axes/timeAxis.ts
```diff
@@ -205,6 +208,7 @@ export module Axis {
         return this._possibleTimeAxisConfigurations;
       }
       this._possibleTimeAxisConfigurations = configurations;
+      this._maximumTiers = _Util.Methods.max(configurations.map((config: TimeAxisConfiguration) => config.length), 0);
```
bluong: I guess I'm still a bit confused why this instance variable must be set here instead of calculating this value at the usage down below in _computeHeight

src/components/axes/timeAxis.ts
```diff
@@ -240,8 +244,13 @@ export module Axis {
 
     public _computeHeight() {
       var textHeight = this._measurer.measure().height;
-      this._tierHeights = this._tierLabelPositions.map((pos: string) =>
-        textHeight + this.tickLabelPadding() + ((pos === "between") ? 0 : this._maxLabelTickLength()));
+
+      this._tierHeights = [];
```
bluong: Why can't we instead use...
var maximumTiers = _Util.Methods.max(this._possibleTimeAxisConfigurations.map((config: TimeAxisConfiguration) => config.length), 0);

bluong:
I'm not a fan of having unneeded instance variables if we can avoid it.

src/components/axes/timeAxis.ts
jtlan: Hmm, I guess we don't use the number of tiers anywhere else. This might be a good change to make.

aicioara:
I am not a big fan of unneeded instance variables either, so I will implement the suggested changes

src/components/axes/timeAxis.ts
```diff
@@ -240,8 +242,14 @@ export module Axis {
 
     public _computeHeight() {
       var textHeight = this._measurer.measure().height;
-      this._tierHeights = this._tierLabelPositions.map((pos: string) =>
-        textHeight + this.tickLabelPadding() + ((pos === "between") ? 0 : this._maxLabelTickLength()));
+      var maximumTiers = _Util.Methods.max(this._possibleTimeAxisConfigurations.map((config: TimeAxisConfiguration) => config.length), 0);
```
bluong: Is it possible to actually know the number of tiers that we will have instead of the maximum?  If maximum > actual, then we will be left with extra space right?

bluong:
Please clean up the PR title / description so to facilitate QE

crmorford:
How does it know the maximum number of tiers? I think I've only really looked at 2-tier cases before.
Is it number of objects when you set the time axis configurations?

aicioara:
@crmorford A configuration like this should have 1 tier
```typescript
[[
    { interval: d3.time.month, step: 1, formatter: Plottable.Formatters.time("%b d") }
]]
```

A configuration like this should also have 1 tier
```typescript
[[
    { interval: d3.time.month, step: 1, formatter: Plottable.Formatters.time("%b d") }
], 
[
        {interval: d3.time.day, step: 1, formatter: Formatters.time("%B %e, %Y")}
]]
```
I hope it makes sense

src/components/axes/timeAxis.ts
aicioara: For completeness, I am adding the offline conversation. Yes, we will have extra space, but if we zoom in and out, at some point we will hit the 2 tier configuration and we don't want our graph to jump at this point

Issue #1801, opened 2015-04-07
Tests fail if "#" is present in the URL


jtlan:
![screen shot 2015-04-04 at 7 34 57 pm](https://cloud.githubusercontent.com/assets/1440449/7015267/f6badaf8-dc85-11e4-9065-c82068a3624f.png)

If the "#" symbol is present in the URL, the test shown fails. This may also be a sign that `clip-path` is not being used correctly...

PR #1807, opened 2015-04-07
2 tests were failing if the # symbol was presented in the URL

aicioara:
To reproduce 
http://localhost:9999/test/tests.html#

Fix: regex strip down of suffix.


aicioara:
Closes #1801

bluong:
In the future, it'd be great to put the description of the current failure mode in the comment/description rather than in the pull request title.  Think of the first comment to be a description field that could be as elaborate as you want for reviewers to get context before they review.  The title should be a small snippet of describing a fix or some introduction of a new feature.

All in all, the title should be as concise as possible, whereas the comments can be a place for elaborate discussion.

aicioara:
Sure

PR #1809, opened 2015-04-08
Component.Group now occupies all offered space

### LINKED ISSUE #1791: Grouping fixed-size Components produces unexpected behavior

jtlan:
This change causes `Component.Group` to behave as though it's not there by occupying the entirety of any space in the final offer, although it will still request no more space than is necessary to fit its `Components`.

Close #1791.
Also should not break #1255.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -3459,6 +3459,7 @@ var Plottable;
              * @param {number} availableHeight available height for the Component to render in
              */
             AbstractComponent.prototype._computeLayout = function (offeredXOrigin, offeredYOrigin, availableWidth, availableHeight) {
+                var _this = this;
                 if (offeredXOrigin == null || offeredYOrigin == null || availableWidth == null || availableHeight == null) {
                     if (this._element == null) {
                         throw new Error("anchor must be called before computeLayout");
@@ -3489,13 +3490,12 @@ var Plottable;
                 this._yOrigin = offeredYOrigin + this._yOffset + (availableHeight - this.height()) * this._yAlignProportion;
                 ;
                 this._element.attr("transform", "translate(" + this._xOrigin + "," + this._yOrigin + ")");
+                this._boxes.forEach(function (b) { return b.attr("width", _this.width()).attr("height", _this.height()); });
             };
             AbstractComponent.prototype._setSize = function (availableWidth, availableHeight) {
-                var _this = this;
                 var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
                 this._width = this._isFixedWidth() ? Math.min(availableWidth, requestedSpace.width) : availableWidth;
                 this._height = this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.height) : availableHeight;
-                this._boxes.forEach(function (b) { return b.attr("width", _this.width()).attr("height", _this.height()); });
             };
             AbstractComponent.prototype._render = function () {
                 if (this._isAnchored && this._isSetup && this.width() >= 0 && this.height() >= 0) {
--- a/plottable.js
+++ b/plottable.js
@@ -3485,17 +3485,20 @@ var Plottable;
                         throw new Error("null arguments cannot be passed to _computeLayout() on a non-root node");
                     }
                 }
-                this._setSize(availableWidth, availableHeight);
+                var size = this._getSize(availableWidth, availableHeight);
+                this._width = size.width;
+                this._height = size.height;
                 this._xOrigin = offeredXOrigin + this._xOffset + (availableWidth - this.width()) * this._xAlignProportion;
                 this._yOrigin = offeredYOrigin + this._yOffset + (availableHeight - this.height()) * this._yAlignProportion;
-                ;
                 this._element.attr("transform", "translate(" + this._xOrigin + "," + this._yOrigin + ")");
                 this._boxes.forEach(function (b) { return b.attr("width", _this.width()).attr("height", _this.height()); });
             };
-            AbstractComponent.prototype._setSize = function (availableWidth, availableHeight) {
+            AbstractComponent.prototype._getSize = function (availableWidth, availableHeight) {
                 var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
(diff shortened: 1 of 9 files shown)


src/components/componentGroup.ts
```diff
@@ -49,12 +49,17 @@ export module Component {
...
+      this._height = availableHeight;
+    }
+
     public _isFixedWidth(): boolean {
-      return false;
+      return this.components().every((c) => c._isFixedWidth());
     }
 
     public _isFixedHeight(): boolean {
-      return false;
```
bluong: What happens in the case of alignments?

test/core/componentGroupTests.ts
```diff
@@ -164,32 +126,54 @@ describe("ComponentGroups", () => {
...
+      ]);
+
+      var request = cg._requestedSpace(SVG_WIDTH, SVG_HEIGHT);
+      verifySpaceRequest(request, 0, 0, false, false, "empty Group doesn't request any space");
+
+      cg.renderTo(svg);
+      assert.strictEqual(cg.width(), SVG_WIDTH, "occupies all offered width");
+      assert.strictEqual(cg.height(), SVG_HEIGHT, "occupies all offered height");
+      svg.remove();
     });
```
bluong: How does this work with a proportional idea?

test/core/componentGroupTests.ts
```diff
@@ -164,32 +126,54 @@ describe("ComponentGroups", () => {
...
+
+      var request = cg._requestedSpace(SVG_WIDTH, SVG_HEIGHT);
+      assert.strictEqual(request.width, SVG_WIDTH/2, "requested enough space for widest Component");
+      assert.strictEqual(request.height, SVG_HEIGHT/2, "requested enough space for tallest Component");
+
+      cg.renderTo(svg);
+      assert.strictEqual(cg.width(), SVG_WIDTH, "occupies all offered width");
+      assert.strictEqual(cg.height(), SVG_HEIGHT, "occupies all offered height");
+      svg.remove();
     });
```
bluong: Can you write a test that describes the dragbox case?

bluong:
Implementation looks fine.  Just some testing fixes

src/components/componentGroup.ts
jtlan: `Group` no longer accounts for the alignments of sub`Component`s. If it did, the resultant layout behavior would be dependent on the alignments, which is contrary to the goal of having `Group` behave as though it isn't there during layout.

Since the `Group` occupies the whole cell, its own alignment has no effect.

test/core/componentGroupTests.ts
jtlan: I should probably assert something about `_isFixedWidth()`. It doesn't request space, but it's non-fixed so it will accept more.

The problem is that this behavior really manifests when inside a `Table`, but that's more of an integration test.

test/core/componentGroupTests.ts
jtlan: Not yet, since `DragBoxLayer` has yet to merge. There aren't many `Component`s that don't participate in layout but occupy all space in the final offer ... other than `Group` itself.

test/core/componentGroupTests.ts
bluong: Oh I guess my main confusion was why the test was named "with proportional Components" when I see nothing really proportional about this test...?

test/core/componentGroupTests.ts
bluong: Well when I say the dragbox case, I mean create the scenario surrounding it (A group with something of fixed height and something that doesn't care about the space it's in) and see how that reacts

bluong:
Just fix up the tests and this should be good to go

test/core/componentGroupTests.ts
jtlan: The "liquid" `Component`s are essentially fixed-size with 0 width and 0 height, but if they happened to be offered more in the final offer they will take it. Based on that, the test case `"with fixed-size Components"` should cover it.

plottable.js
```diff
@@ -3485,15 +3485,18 @@ var Plottable;
                         throw new Error("null arguments cannot be passed to _computeLayout() on a non-root node");
                     }
                 }
-                var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
-                this._width = this._isFixedWidth() ? Math.min(availableWidth, requestedSpace.width) : availableWidth;
-                this._height = this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.height) : availableHeight;
+                this._setSize(availableWidth, availableHeight);
                 this._xOrigin = offeredXOrigin + this._xOffset + (availableWidth - this.width()) * this._xAlignProportion;
                 this._yOrigin = offeredYOrigin + this._yOffset + (availableHeight - this.height()) * this._yAlignProportion;
                 ;
```
bluong: Might as well remove this semicolon

src/components/abstractComponent.ts
```diff
@@ -134,17 +134,19 @@ export module Component {
           throw new Error("null arguments cannot be passed to _computeLayout() on a non-root node");
         }
       }
-      var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
-      this._width  = this._isFixedWidth()  ? Math.min(availableWidth , requestedSpace.width)  : availableWidth ;
-      this._height = this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.height) : availableHeight;
-
+      this._setSize(availableWidth, availableHeight);
       this._xOrigin = offeredXOrigin + this._xOffset + (availableWidth - this.width()) * this._xAlignProportion;
       this._yOrigin = offeredYOrigin + this._yOffset + (availableHeight - this.height()) * this._yAlignProportion;;
```
bluong: There's an extra semicolon if you wanna take that out

PR #1810, opened 2015-04-08
[Interaction.Scroll] - Initial Implementation

bluong:
Wraps the dispatchers return values to ensure that the deltaY is returned in pixel space.

NOTE: I cannot really think of much capability of this since we are currently using it for the usecase of zooming.  At the very least, I do want consumers to be able to interact with their components with their mouse wheel.

NOTE: This is also prefacing the idea of zooming which will go down a similar path as pan.


src/interactions/scrollInteraction.ts
```diff
@@ -0,0 +1,48 @@
...
+      super._anchor(component, hitBox);
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._mouseDispatcher.onWheel("Interaction.Scroll" + this.getID(), (p: Point, e: WheelEvent) => this._handleScrollEvent(p, e));
+    }
+
+    private _handleScrollEvent(p: Point, e: WheelEvent) {
+      var translatedP = this._translateToComponentSpace(p);
+      if (this._isInsideComponent(translatedP)) {
+        if (this._scrollCallback) {
+          var deltaPixelAmount = e.deltaY * (e.deltaMode ? 120 : 1);
```
jtlan: Is there testing for this functionality?

Also it'd be good if we could find a reference for this 120...

jtlan:
Open discussion:

Should we type-define callback functions?
**arguments for**
* Distinguishes callbacks from functions with the same signature
* Allows documentation of callback parameters and return types

**arguments against**
* Methods passed or returning the callback type are hard to use without cross-referencing the callback type definition

src/interactions/scrollInteraction.ts
bluong: I looked through the d3 js source and found the 120 there with pretty much this exact line of code
`e.deltaY * (e.deltaMode ? 120 : 1);`

src/interactions/scrollInteraction.ts
```diff
@@ -0,0 +1,58 @@
...
+export module Interaction {
+
+  export type ScrollCallback = (p: Point, deltaAmount: number) => any;
+
+  export class Scroll extends Interaction.AbstractInteraction {
+
+    /**
+     * The number of pixels occupied in a line.
+     */
+    public static PIXELS_PER_LINE = 120;
```
jtlan: Can we find a source for this value?

jtlan:
Since it's called `Interaction.Scroll`, should we also capture non-wheel scroll events, such as a two-finger scroll gesture?

plottable.js
```diff
@@ -10261,6 +10261,55 @@ var Plottable;
...
+                _super.prototype._anchor.call(this, component, hitBox);
+                this._mouseDispatcher = Plottable.Dispatcher.Mouse.getDispatcher(this._componentToListenTo.content().node());
+                this._mouseDispatcher.onWheel("Interaction.Scroll" + this.getID(), function (p, e) { return _this._handleScrollEvent(p, e); });
+            };
+            Scroll.prototype._handleScrollEvent = function (p, e) {
+                var translatedP = this._translateToComponentSpace(p);
+                if (this._isInsideComponent(translatedP)) {
+                    if (this._scrollCallback) {
+                        e.preventDefault();
+                        var deltaPixelAmount = e.deltaY * (e.deltaMode ? Scroll.PIXELS_PER_LINE : 1);
```
jtlan: Aren't there three delta modes? https://developer.mozilla.org/en-US/docs/Web/Events/wheel

src/interactions/scrollInteraction.ts
```diff
@@ -0,0 +1,58 @@
...
+      this._mouseDispatcher = Dispatcher.Mouse.getDispatcher(<SVGElement> this._componentToListenTo.content().node());
+      this._mouseDispatcher.onWheel("Interaction.Scroll" + this.getID(), (p: Point, e: WheelEvent) => this._handleScrollEvent(p, e));
+    }
+
+    private _handleScrollEvent(p: Point, e: WheelEvent) {
+      var translatedP = this._translateToComponentSpace(p);
+      if (this._isInsideComponent(translatedP)) {
+        if (this._scrollCallback) {
+          e.preventDefault();
+          var deltaPixelAmount = e.deltaY * (e.deltaMode ? Scroll.PIXELS_PER_LINE : 1);
```
jtlan: Aren't there three delta modes? https://developer.mozilla.org/en-US/docs/Web/Events/wheel

bluong: While true, I've looked around implementations and only saw the 2 modes being cased out... I would need the PIXELS_PER_PAGE to switch off

src/interactions/scrollInteraction.ts
bluong: The best I've got is this code from D3
`-d3.event.deltaY * (d3.event.deltaMode ? 120 : 1);`

I'll go look around for a more legitimate source

bluong: Based on this post http://stackoverflow.com/questions/20110224/what-is-the-height-of-a-line-in-a-wheel-event-deltamode-dom-delta-line

it's not consistent.  Thus, D3 is basically using a guess estimate.  We can stick with the estimate to get behavior similar to D3 or we can come up with our own logic.

bluong:
This does not seem relevant anymore since it will be rolled to be an internal thing inside PanZoom

PR #1811, opened 2015-04-09
Change order of arguments to addComponent()

jtlan:
On `Table`, `addComponent` has been changed so that the `Component` to add is the first argument:
```Typescript
public addComponent(componentToAdd: AbstractComponent, row: number, col: number): Table;
```

This signature is much more readable, as `addComponent(comp, r, c)` can be read as "add `Component` *comp* to row *r* and column *c*". The original argument ordering was less memorable and was thus more likely to read to mistakes. In particular, it is reasonable to expect that a method `addThing()` takes a `Thing` as the first argument.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2534,28 +2534,25 @@ declare module Plottable {
              */
             constructor(rows?: AbstractComponent[][]);
             /**
-             * Adds a Component in the specified cell.
+             * Adds a Component at the specified row and column position.
              *
-             * If the cell is already occupied, there are 3 cases
-             *  - Component + Component => Group containing both components
-             *  - Component + Group => Component is added to the group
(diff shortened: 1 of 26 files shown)


jtlan:
@jtlan wrote:
>In particular, it is reasonable to expect that a method addThing() takes a Thing as the first argument.

This might be a good argument for having the *key* be the second argument in `addDataset()` ...

bluong:
LGTM  was there some specific reason for doing this other than for API neatness?

jtlan:
Mostly wanted to kick-start the process of API cleanup. I think there are a few other cases where the method names and/or signatures are weird.

bluong:
Please note that this is an API-break.  Otherwise, nice!

jtlan:
Going to group this with other breaks.

PR #1812, opened 2015-04-09
[numericAxis] Constrained tick labels look good now

### LINKED ISSUE #1118: Tick labels can be positioned badly under constrained circumstances
screen shot 2014-09-29 at 1 54 36 pm

http://jsfiddle.net/dFf3s/7/

aicioara:
Closes #1118 

In this patch: 
- For Bottom / Top / Left / Right `tickLabelPosition`, we are only showing the ticks where we have labels
- Since we are doing the above, we are now spacing the labels a bit more, such that they are `padding` distance from their tickMark and 2*`padding` distance or more from the next tickMark, to avoid confusion
- Added comments on existing methods touched
- the `_showAllTickMarks()` method is needed because we are hiding the tick marks in the view. If we decide to change the padding dynamically, the hidden marks have to be shown again before deciding which marks to hide. Feel free to comment on this

Issues:
- I don't really like that the logic of hiding some tick marks is in the view, while tick mark generation is in the model. 
- Because of the above the `_showAllTickMarks()` method was needed

before http://jsfiddle.net/dywyfpj2/4/
after http://jsfiddle.net/dywyfpj2/5/


bluong:
Fix the build please

aicioara:
Closes #1118 
Writing a unit test for this now.

src/components/axes/abstractAxis.ts
```diff
@@ -25,7 +25,7 @@ export module Axis {
     protected _computedHeight: number;
     private _endTickLength = 5;
     private _tickLength = 5;
-    private _tickLabelPadding = 10;
+    protected _tickLabelPadding = 10;
```
jtlan: Use the public `tickLabelPadding()` method to get and set this value, rather than exposing it as `protected`.

jtlan:
@aicioara wrote:
>I don't really like that the logic of hiding some tick marks is in the view, while tick mark generation is in the model.

This sort of makes sense, if you consider [the list of available ticks] to be a model feature. The list might be suitable as part of the model to coordinate multiple views that depend on it -- for example, gridlines and axes backed by the same `Scale` should agree as to where the ticks are.

src/components/axes/numericAxis.ts
```diff
@@ -194,12 +194,51 @@ export module Axis {
       var labelGroupTransform = "translate(" + labelGroupTransformX + ", " + labelGroupTransformY + ")";
       this._tickLabelContainer.attr("transform", labelGroupTransform);
 
+      this._showAllTickMarks();
```
jtlan: If this method is only invoked once, why not just inline it?

src/components/axes/abstractAxis.ts
aicioara: Thank you.

test/components/numericAxisTests.ts
```diff
@@ -427,4 +427,48 @@ describe("NumericAxis", () => {
...
+          return (visibility === "visible") || (visibility === "inherit");
+        });
+
+
+    tickLabels.each(function() {
+      var tickLabelBox = this.getBoundingClientRect();
+
+      tickMarks.each(function() {
+        var tickMarkBox = this.getBoundingClientRect();
+          assert.isFalse(Plottable._Util.DOM.boxesOverlap(tickLabelBox, tickMarkBox));
```
jtlan: The last argument in any `assert` method is a message. Usually we try to write something detailed there; consider another developer who is working on code that causes this test to fail. They see this:

`Expected true to be false`

with no explanation as to what went wrong. A detailed message provides context when the test fails:

`Expected true to be false: tick label does not overlap a tick mark`

jtlan:
Mostly stylistic things.

src/components/axes/numericAxis.ts
aicioara: It's neater as it encapsulates the semantics in the name. I prefer this when reading code.

aicioara:
Addressed all the issues (or commented on them)

src/components/axes/numericAxis.ts
jtlan: I would agree in cases where the method's functionality is somewhat complex, but in this case the code is pretty self-explanatory. Naming the method is nice, but now you have to go look in the file for where it's defined...

crmorford:
This is odd, if i pick numTicks(4), i get some mini ticks at the bottom and top
screen shot 2015-04-22 at 12 59 23 pm
http://jsfiddle.net/dywyfpj2/6/

aicioara:
Hmm, this is because the tick length is inconsistently setup in the fiddle.

Fixed: http://jsfiddle.net/dywyfpj2/7/

crmorford:
oh oops, my bad. thank you :)

aicioara:
Merged and waiting for build to pass

aicioara:
Restarted build, as it timed out

PR #1814, opened 2015-04-09
SelectionBoxLayer no longer requests extra space.

jtlan:
Has an effective fixed size of zero, but will occupy all offered space.

### DIFF
--- a/src/components/selectionBoxLayer.ts
+++ b/src/components/selectionBoxLayer.ts
@@ -14,6 +14,8 @@ export module Component {
     constructor() {
       super();
       this.classed("selection-box-layer", true);
+      this._fixedWidthFlag = true;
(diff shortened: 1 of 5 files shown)


src/components/selectionBoxLayer.ts
```diff
@@ -14,6 +14,8 @@ export module Component {
     constructor() {
       super();
       this.classed("selection-box-layer", true);
+      this._fixedWidthFlag = true;
+      this._fixedHeightFlag = true;
```
bluong: If we set these to true, what's the fixedWidth/Height of this component?

jtlan: Zero. Really `_fixedWidthFlag` should be named "Would use extra offered width" or something.

PR #1817, opened 2015-04-10
Refactoring getClosestPlotData()'s API to the new, simpler version.

istarnx:


istarnx:
Oh, by the way, I added getClosestPlotData(), a public method, to the bottom of linePlot.ts, before the public Hover api stuff. This seemed weird. We're usually placing public stuff at the top if files, right?

src/components/plots/abstractPlot.ts
```diff
@@ -516,30 +516,17 @@ export module Plot {
...
+     * Retrieves the closest PlotData across all datasets, where distance is defined to be
+     * the Euclidiean norm.
      *
-     * @param {Point} queryPoint The point to query from
-     * @param {number} withinValue Will only return plot data that is of a distance below withinValue
-     *                             (default = Infinity)
-     * @param {string | string[]} datasetKeys The dataset(s) to retrieve the plot data from.
-     *                                        (default = this.datasetOrder())
-     * @returns {PlotData} The retrieved PlotData.
+     * @param {Point} queryPoint The point to which dataset points should be compared
```
bluong: Let's change the doc possibly to 

Retrieves PlotData with the lowest distance, where distance is defined to be the Euclidiean norm.

src/components/plots/linePlot.ts
```diff
@@ -182,6 +154,37 @@ export module Plot {
...
     }
 
+    public getClosestPlotData(queryPoint: Point): PlotData {
+      var closestDistanceSquared = Infinity;
+      var closestDatum: any;
+      var closestSelection: D3.Selection;
+      var closestPoint: Point;
+      var datasetKeys = this.datasetOrder();
+
+      datasetKeys.forEach((datasetKey: string) => {
```
bluong: Minor, datasetKeys doesn't seem to be used anywhere.  Inline this.datasetOrder with the forEach?

bluong:
When this is finished, we are going to want to document this behavior

bluong:
This seems to be a simple removal of behavior.  Just fix up the docs

src/components/plots/linePlot.ts
istarnx: Good catch... this function is going to be rewritten in the next checkin anyway, so let's not worry about it. :)

src/components/plots/abstractPlot.ts
istarnx: Both of these make sense to me. I'll do 'em.

PR #1824, opened 2015-04-10
[BarPlot] - getAllPlotData pixelPoints corrected for negative-valued bars and barAlignment

bluong:
Bar plots would normally place their pixelPoints for `getAllPlotData` at the top of the bar under the assumption that the bar was positive valued.  This is incorrect if the bar is negative-valued or has a different barAlignment

Bars should now place their pixelPoints at the bottom of the bar if they are negative-valued (left of the bar for horizontal bar plots)

Accompanying JSFiddle - http://jsfiddle.net/8w5v19ry/1/

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -501,7 +501,7 @@ export module Plot {
       plotData.selection.each(function (datum, index) {
         var bar = d3.select(this);
 
-        if (isVertical && +bar.attr("y") + +bar.attr("height") > scaledBaseline) {
+        if (isVertical && +bar.attr("y") >= scaledBaseline) {
           plotData.pixelPoints[index].y += +bar.attr("height");
         } else if (!isVertical && +bar.attr("x") < scaledBaseline) {
           plotData.pixelPoints[index].x -= +bar.attr("width");
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -489,6 +489,26 @@ export module Plot {
       };
     }
     //===== /Hover logic =====
+
+    protected _getAllPlotData(datasetKeys: string[]): PlotData {
+      var plotData = super._getAllPlotData(datasetKeys);
+
+      var valueScale = this._isVertical ? this._yScale : this._xScale;
+      var scaledBaseline = (<Scale.AbstractScale<any, any>> (this._isVertical ? this._yScale : this._xScale)).scale(this.baseline());
+      var isVertical = this._isVertical;
+
+      plotData.selection.each(function (datum, index) {
+        var bar = d3.select(this);
+
+        if (isVertical && +bar.attr("y") + +bar.attr("height") > scaledBaseline) {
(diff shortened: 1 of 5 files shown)


jtlan:
Shouldn't the pixel position also depend on the `barAlignment()`?

src/components/plots/barPlot.ts
```diff
@@ -489,6 +489,33 @@ export module Plot {
       };
     }
     //===== /Hover logic =====
+
+    protected _getAllPlotData(datasetKeys: string[]): PlotData {
+      var plotData = super._getAllPlotData(datasetKeys);
+
+      var valueScale = this._isVertical ? this._yScale : this._xScale;
```
rcchen: We should really refactor the code base to use keyScale/valueScale semantics or something like that instead of xScale/yScale sometime...

test/components/plots/barPlotTests.ts
```diff
@@ -355,6 +391,42 @@ describe("Plots", () => {
...
+            assert.strictEqual(pixelPointX, +barSelection.attr("x"), "negative on left");
+          } else {
+            assert.strictEqual(pixelPointX, +barSelection.attr("x") + +barSelection.attr("width"), "positive on right");
+          }
+        });
+        svg.remove();
+      });
+
+      it("getAllPlotData() pixel points corrected for barAlignment left", () => {
+        barPlot.barAlignment("left");
```
rcchen: Why is it "left" for a horizontalBarPlot? I thought only "top", "center", and "bottom" were supported

test/components/plots/barPlotTests.ts
```diff
@@ -355,6 +391,42 @@ describe("Plots", () => {
...
+        plotData.data.forEach((datum, i) => {
+          var barSelection = d3.select(plotData.selection[0][i]);
+          var pixelPointY = plotData.pixelPoints[i].y;
+          assert.strictEqual(pixelPointY, +barSelection.attr("y"), "barAlignment left y correct");
+        });
+        svg.remove();
+      });
+
+      it("getAllPlotData() pixel points corrected for barAlignment right", () => {
+        barPlot.barAlignment("right");
```
rcchen: Same as 410

rcchen:
Some concerns about the validity of tests for horizontalBarCharts

test/components/plots/barPlotTests.ts
bluong: We used to have "top" "center" "bottom" semantics but to keep semantics consistent, we switched to the same "left" "right" "center"

jtlan:
@bluong , can we write the test in a way that checks all combinations of Vertical/Horizontal `Plot`, positive/negative bars, and all 3 alignments?

jtlan:
`NaN` values screw up: http://jsfiddle.net/8w5v19ry/3/

jtlan:
`NaN` and other values tracked here: #1845.

vchu123:
Leftmost data point doesn't work in IE ONLY (and only that one data point)

http://jsfiddle.net/8w5v19ry/4/

PR #1826, opened 2015-04-12
Plot.Bar.getClosestPlotData()

istarnx:
Here's the Bar case and its tests.


istarnx:
Closing because I'm pulling into the wrong branch. Will re-open.

PR #1827, opened 2015-04-13
[Time Axis] We allow for more than 2 tiers

### LINKED ISSUE #1818: Axis.Time displays strangely when more than two tiers are specified
text is going off the right side of the axis: 
http://jsfiddle.net/pLapukjh/

aicioara:
Closes #1818 

Changes
- The number of tiers is now an instance variable
- Deleting old axis tiers when creating new ones
- Added unit test for this case

Before: http://jsfiddle.net/pLapukjh/
After: http://jsfiddle.net/pLapukjh/2/

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -4524,6 +4524,7 @@ var Plottable;
                     return this._possibleTimeAxisConfigurations;
                 }
                 this._possibleTimeAxisConfigurations = configurations;
+                this._numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
                 this._invalidateLayout();
                 return this;
             };
@@ -4552,9 +4553,9 @@ var Plottable;
             };
             Time.prototype._computeHeight = function () {
                 var textHeight = this._measurer.measure().height;
-                var maximumTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
+                console.log('a');
                 this._tierHeights = [];
-                for (var i = 0; i < maximumTiers; i++) {
+                for (var i = 0; i < this._numTiers; i++) {
                     this._tierHeights.push(textHeight + this.tickLabelPadding() + ((this._tierLabelPositions[i]) === "between" ? 0 : this._maxLabelTickLength()));
                 }
                 this._computedHeight = d3.sum(this._tierHeights);
@@ -4588,8 +4589,7 @@ var Plottable;
                 this._tierBaselines = [];
                 this._tickLabelContainer.remove();
                 this._baseline.remove();
-                var numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
-                for (var i = 0; i < numTiers; ++i) {
+                for (var i = 0; i < this._numTiers; ++i) {
                     var tierContainer = this._content.append("g").classed("time-axis-tier", true);
                     this._tierLabelContainers.push(tierContainer.append("g").classed(Axis.AbstractAxis.TICK_LABEL_CLASS + "-container", true));
                     this._tierMarkContainers.push(tierContainer.append("g").classed(Axis.AbstractAxis.TICK_MARK_CLASS + "-container", true));
@@ -4697,8 +4697,7 @@ var Plottable;
                 var _this = this;
                 this._mostPreciseConfigIndex = this._getMostPreciseConfigurationIndex();
                 var tierConfigs = this._possibleTimeAxisConfigurations[this._mostPreciseConfigIndex];
-                var numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
-                for (var i = 0; i < numTiers; ++i) {
+                for (var i = 0; i < this._numTiers; ++i) {
                     this._cleanTier(i);
                 }
                 tierConfigs.forEach(function (config, i) { return _this._renderTierLabels(_this._tierLabelContainers[i], config, i); });
@@ -4866,10 +4865,6 @@ var Plottable;
                 ]
             ];
             Time._LONG_DATE = new Date(9999, 8, 29, 12, 59, 9999);
-            /**
-             * Number of possible tiers.
-             */
-            Time._NUM_TIERS = 2;
             return Time;
         })(Axis.AbstractAxis);
         Axis.Time = Time;
--- a/plottable.js
+++ b/plottable.js
@@ -4501,9 +4501,9 @@ var Plottable;
              */
             function Time(scale, orientation) {
                 _super.call(this, scale, orientation);
+                this._tierLabelPositions = [];
                 this.classed("time-axis", true);
                 this.tickLabelPadding(5);
-                this.tierLabelPositions(["between", "between"]);
                 this.axisConfigurations(Time._DEFAULT_TIME_AXIS_CONFIGURATIONS);
             }
             Time.prototype.tierLabelPositions = function (newPositions) {
@@ -4525,6 +4525,12 @@ var Plottable;
                 }
                 this._possibleTimeAxisConfigurations = configurations;
                 this._numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
+                var oldLabelPositions = this.tierLabelPositions();
+                var newLabelPositions = [];
+                for (var i = 0; i < this._numTiers; i++) {
+                    newLabelPositions.push(oldLabelPositions[i] ? oldLabelPositions[i] : "between");
+                }
+                this.tierLabelPositions(newLabelPositions);
                 this._invalidateLayout();
                 return this;
             };
@@ -4553,7 +4559,6 @@ var Plottable;
             };
             Time.prototype._computeHeight = function () {
                 var textHeight = this._measurer.measure().height;
-                console.log('a');
                 this._tierHeights = [];
                 for (var i = 0; i < this._numTiers; i++) {
                     this._tierHeights.push(textHeight + this.tickLabelPadding() + ((this._tierLabelPositions[i]) === "between" ? 0 : this._maxLabelTickLength()));
--- a/plottable.js
+++ b/plottable.js
@@ -4588,7 +4588,8 @@ var Plottable;
                 this._tierBaselines = [];
                 this._tickLabelContainer.remove();
                 this._baseline.remove();
-                for (var i = 0; i < Time._NUM_TIERS; ++i) {
+                var numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
+                for (var i = 0; i < numTiers; ++i) {
                     var tierContainer = this._content.append("g").classed("time-axis-tier", true);
                     this._tierLabelContainers.push(tierContainer.append("g").classed(Axis.AbstractAxis.TICK_LABEL_CLASS + "-container", true));
                     this._tierMarkContainers.push(tierContainer.append("g").classed(Axis.AbstractAxis.TICK_MARK_CLASS + "-container", true));
@@ -4696,7 +4697,8 @@ var Plottable;
                 var _this = this;
                 this._mostPreciseConfigIndex = this._getMostPreciseConfigurationIndex();
                 var tierConfigs = this._possibleTimeAxisConfigurations[this._mostPreciseConfigIndex];
-                for (var i = 0; i < Time._NUM_TIERS; ++i) {
+                var numTiers = Plottable._Util.Methods.max(this._possibleTimeAxisConfigurations.map(function (config) { return config.length; }), 0);
+                for (var i = 0; i < numTiers; ++i) {
                     this._cleanTier(i);
                 }
                 tierConfigs.forEach(function (config, i) { return _this._renderTierLabels(_this._tierLabelContainers[i], config, i); });
--- a/plottable.js
+++ b/plottable.js
@@ -4594,14 +4594,9 @@ var Plottable;
                 // Makes sure that the size it requires is a multiple of tier sizes, such that
                 // we have no leftover tiers
                 var size = _super.prototype._getSize.call(this, availableWidth, availableHeight);
-                var adjustedHeight = 0;
-                for (var i = 0; i < this._tierHeights.length; i++) {
-                    if (adjustedHeight + this._tierHeights[i] > size.height) {
-                        break;
-                    }
-                    adjustedHeight += this._tierHeights[i];
-                }
-                size.height = adjustedHeight;
+                size.height = this._tierHeights.reduce(function (prevValue, currValue, index, arr) {
+                    return (prevValue + currValue > size.height) ? prevValue : (prevValue + currValue);
+                });
                 return size;
             };
             Time.prototype._setup = function () {
@@ -4756,10 +4751,10 @@ var Plottable;
                 var usedHeight = 0;
                 var visibleAxisTiers = this._element.selectAll("." + Time.TIME_AXIS_TIER_CLASS).filter(function (d, i) {
                     var visibility = d3.select(this).style("visibility");
-                    return visibility === "visible" || visibility === "inherit";
+                    return visibility === "visible";
                 }).attr("visibility", function (d, i) {
                     usedHeight += _this._tierHeights[i];
-                    return usedHeight <= availableHeight ? "visible" : "hidden";
+                    return usedHeight <= availableHeight ? "inherit" : "hidden";
                 });
             };
             Time.prototype._hideOverlappingAndCutOffLabels = function (index) {
--- a/plottable.js
+++ b/plottable.js
@@ -4754,12 +4754,11 @@ var Plottable;
             };
             Time.prototype._hideOverflowingTiers = function () {
                 var _this = this;
-                var visibleAxisTiers = this._element.selectAll(".time-axis-tier").filter(function (d, i) {
-                    return d3.select(this).style("visibility") === "visible";
-                });
                 var availableHeight = this.height();
                 var usedHeight = 0;
-                visibleAxisTiers.attr("visibility", function (d, i) {
+                var visibleAxisTiers = this._element.selectAll(".time-axis-tier").filter(function (d, i) {
+                    return d3.select(this).style("visibility") === "visible";
+                }).attr("visibility", function (d, i) {
                     usedHeight += _this._tierHeights[i];
                     return usedHeight <= availableHeight ? "visible" : "hidden";
                 });
--- a/plottable.js
+++ b/plottable.js
@@ -4590,6 +4590,22 @@ var Plottable;
                 var worstWidth = this._maxWidthForInterval(config) + 2 * this.tickLabelPadding();
                 return Math.min(this._getIntervalLength(config), this.width()) >= worstWidth;
             };
+            /**
+             * Makes sure that the size it requires is a multiple of tier sizes, such that
+             * we have no leftover tiers
+             */
+            Time.prototype._getSize = function (availableWidth, availableHeight) {
+                var size = _super.prototype._getSize.call(this, availableWidth, availableHeight);
+                var adjustedHeight = 0;
+                for (var i = 0; i < this._tierHeights.length; i++) {
+                    if (adjustedHeight + this._tierHeights[i] > size.height) {
+                        break;
+                    }
+                    adjustedHeight += this._tierHeights[i];
+                }
+                size.height = adjustedHeight;
+                return size;
+            };
             Time.prototype._setup = function () {
                 _super.prototype._setup.call(this);
                 this._setupDomElements();
--- a/plottable.js
+++ b/plottable.js
@@ -4590,11 +4590,9 @@ var Plottable;
                 var worstWidth = this._maxWidthForInterval(config) + 2 * this.tickLabelPadding();
                 return Math.min(this._getIntervalLength(config), this.width()) >= worstWidth;
             };
-            /**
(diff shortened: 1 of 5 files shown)


aicioara:
Creating JSFiddle now

aicioara:
.

src/components/axes/timeAxis.ts
```diff
@@ -207,6 +202,19 @@ export module Axis {
...
+      if (this._isAnchored) {
+        this._createDomElements();
+      }
+
+      var oldLabelPositions: string[] = this.tierLabelPositions();
+      var newLabelPositions: string[] = [];
+      for (var i = 0; i < this._numTiers; i++) {
+        newLabelPositions.push(oldLabelPositions[i] || "between");
+      }
+      this.tierLabelPositions(newLabelPositions);
```
bluong: Do we want the tierLabelPositions instance variable to always contain complete valid information (inputting "between" where needed)?

Or do we want to instead fill in the gaps in the instance variable when we are using tierLabelPositions?

The good thing that you've done here is that you've made the positions valid at every step.

The bad thing is that now this setter is really heavy, doing more than possibly the user expects.

src/components/axes/timeAxis.ts
```diff
@@ -207,6 +202,19 @@ export module Axis {
         return this._possibleTimeAxisConfigurations;
       }
       this._possibleTimeAxisConfigurations = configurations;
+      this._numTiers = _Util.Methods.max(this._possibleTimeAxisConfigurations.map((config: TimeAxisConfiguration) => config.length), 0);
```
bluong: Why store this value rather than calculate it on the fly?

src/components/axes/timeAxis.ts
```diff
@@ -207,6 +202,19 @@ export module Axis {
...
+
+      if (this._isAnchored) {
+        this._createDomElements();
+      }
```
bluong: Instead of this, is there a way we can just route the setter to go back to the _setup function rather than creating them on the fly?

It's just more conventional for us to do the minimum to affect state in our setters and then calling the right function that will handle the rest in terms of DOM manipulation

src/components/axes/timeAxis.ts
```diff
@@ -242,10 +250,9 @@ export module Axis {
 
     public _computeHeight() {
       var textHeight = this._measurer.measure().height;
-      var maximumTiers = _Util.Methods.max(this._possibleTimeAxisConfigurations.map((config: TimeAxisConfiguration) => config.length), 0);
 
       this._tierHeights = [];
-      for (var i = 0; i < maximumTiers; i++) {
+      for (var i = 0; i < this._numTiers; i++) {
```
bluong: Should be do the computation here instead?

src/components/axes/timeAxis.ts
```diff
@@ -306,10 +320,20 @@ export module Axis {
...
-      this._tierLabelContainers[index].selectAll("." + AbstractAxis.TICK_LABEL_CLASS).remove();
-      this._tierMarkContainers[index].selectAll("." + AbstractAxis.TICK_MARK_CLASS).remove();
-      this._tierBaselines[index].style("visibility", "hidden");
+    private _cleanTiers() {
+      var oldTiersNumber = this._tierLabelContainers.length;
+
+      if (! (this._tierLabelContainers.length === this._tierMarkContainers.length &&
+             this._tierLabelContainers.length === this._tierBaselines.length)) {
+        throw Error("Tiers could not be cleaned properly");
+      }
```
bluong: Explain where this situation can occur?

src/components/axes/timeAxis.ts
```diff
@@ -280,12 +287,19 @@ export module Axis {
 
     protected _setup() {
       super._setup();
+      this._createDomElements();
+    }
+
+    private _createDomElements() {
+      d3.selectAll(".time-axis-tier").remove();
```
bluong: the function is somewhat misnamed.  You're creating Dom elements by first removing the existing ones.  Is there a way that we can keep this logic being _setup and then having the setter for axis configuration logic route to here?

src/components/axes/timeAxis.ts
```diff
@@ -280,12 +287,19 @@ export module Axis {
...
       this._tierMarkContainers = [];
       this._tierBaselines = [];
       this._tickLabelContainer.remove();
       this._baseline.remove();
-      for (var i = 0; i < Time._NUM_TIERS; ++i) {
+
+      for (var i = 0; i < this._numTiers; ++i) {
         var tierContainer = this._content.append("g").classed("time-axis-tier", true);
         this._tierLabelContainers.push(tierContainer.append("g").classed(AbstractAxis.TICK_LABEL_CLASS + "-container", true));
         this._tierMarkContainers.push(tierContainer.append("g").classed(AbstractAxis.TICK_MARK_CLASS + "-container", true));
```
bluong: As an aside, why not use D3 to append the "g"'s rather than us manually doing it?

src/components/axes/timeAxis.ts
aicioara: I would probably expect the latter. But then every time we use the tierLabelPositions we have to 1) Check for the size of the array 2) Check for the attribute, eventually resize the array. Also, the number of tiers is always 1 or 2, this method is not called that often, so I think it is acceptable to recreate an array of 2 elements all the time

src/components/axes/timeAxis.ts
aicioara: Because it makes sense now. The number of tiers is a property of the axis, which we may want to query in the future (or test).

src/components/axes/timeAxis.ts
aicioara: I was thinking of that, but the problem is the following: In order to compute the layout, we need the size of the tiers. But in order to know the size of the tiers, we need to first create the DOM elements, put them at the right position, and measure them (then recycle them). This is what I understood from a discussion with @jtlan . Please correct me if I am wrong

src/components/axes/timeAxis.ts
aicioara: Is this related to your previous comment (of computing the value on the fly)?

src/components/axes/timeAxis.ts
aicioara: It should never occur. It is just defensive code (which I believe we are missing a lot in the codebase). I can remove it if you want. This is just because the method (and what was here before) assumes that the three arrays are having the same length. 

What is our policy about defensive code like this. It would be great to have some Java-like assert system

src/components/axes/timeAxis.ts
aicioara: This entire logic is a bit weird actually, because I was in doubts of how to do it properly. Can we maybe discuss this, so I see your thoughts?

src/components/axes/timeAxis.ts
bluong: Ok I see what you are getting at.

Can we remove the newLabelPositions variable?  I don't think that variable is needed.  Also, if you remove that variable, can you remove the `old` from `oldLabelPositions`

src/components/axes/timeAxis.ts
bluong: We currently don't have a clear idea of error catching in Plottable.  I assume that this is similar to an IllegalStateException in Java land, but if this error were to be reported to the user... I don't see too much of how this would help them debug.

I would say remove it for now for consistency and since I'm unsure of how helpful this would be.  If you can think of a good error infrastructure though, we should definitely discuss it in the near future.

src/components/axes/timeAxis.ts
bluong: Maybe change function name to _setupDomElements?

aicioara:
All issues have either been solved or discussed. This should be fine now

src/components/axes/timeAxis.ts
```diff
@@ -306,10 +320,15 @@ export module Axis {
         );
     }
 
-    private _cleanTier(index: number) {
-      this._tierLabelContainers[index].selectAll("." + AbstractAxis.TICK_LABEL_CLASS).remove();
-      this._tierMarkContainers[index].selectAll("." + AbstractAxis.TICK_MARK_CLASS).remove();
-      this._tierBaselines[index].style("visibility", "hidden");
+    private _cleanTiers() {
+      var oldTiersNumber = this._tierLabelContainers.length;
```
bluong: This variable can be inlined

aicioara:
Edge case from @vchu123 http://jsfiddle.net/pLapukjh/13/

aicioara:
JSFiddle here http://jsfiddle.net/pLapukjh/13/ . Not sure why the whitespace at the bottom. Basically the `timeAxis` said that it only needs less size that it was given, but the `computeLayout` does not distribute that value back to the plot

src/components/axes/timeAxis.ts
jtlan: @bluong Are you saying it would be clearer like this?
```Typescript
this.tierLabelPositions(this.tierLabelPositions.map((pos) => pos || between));
```

src/components/axes/timeAxis.ts
```diff
@@ -278,14 +285,40 @@ export module Axis {
       return Math.min(this._getIntervalLength(config), this.width()) >= worstWidth;
     }
 
+    /**
+     * Makes sure that the size it requires is a multiple of tier sizes, such that
+     * we have no leftover tiers
+     */
```
jtlan: This comment is leaking into the plottable.d.ts; can you move it inside the implementation?

src/components/axes/timeAxis.ts
```diff
@@ -278,14 +285,40 @@ export module Axis {
...
+      return size;
+    }
+
     protected _setup() {
       super._setup();
+      this._setupDomElements();
+    }
+
+    private _setupDomElements() {
+      d3.selectAll(".time-axis-tier").remove();
```
jtlan: We should save this class off as a static string.

src/components/axes/timeAxis.ts
```diff
@@ -445,6 +480,20 @@ export module Axis {
...
     }
 
+    private _hideOverflowingTiers() {
+      var availableHeight = this.height();
+      var usedHeight = 0;
+
+      var visibleAxisTiers = this._element
+        .selectAll(".time-axis-tier")
+        .filter(function(d: Element, i: number) {
+          return d3.select(this).style("visibility") === "visible";
```
jtlan: You use `"visible"` here, but `"inherit"` elsewhere. We probably should use one of them everywhere (probably `"inherit"`).

test/components/timeAxisTests.ts
```diff
@@ -179,7 +179,79 @@ describe("TimeAxis", () => {
...
+    xAxis.axisConfigurations([
+        [
+           {interval: d3.time.day, step: 2, formatter: Plottable.Formatters.time("%a %e")}
+        ],
+    ]);
+
+    var initialTierSize: number = xAxis.height();
+
+    assert.strictEqual(initialTierSize, oneTierSize,
+                      "2-tier time axis should shrink when prezented new configuration with 1 tier");
```
jtlan: "presented"

test/components/timeAxisTests.ts
```diff
@@ -179,7 +179,79 @@ describe("TimeAxis", () => {
...
+        [
+           {interval: d3.time.day, step: 2, formatter: Plottable.Formatters.time("%a %e")},
+           {interval: d3.time.day, step: 2, formatter: Plottable.Formatters.time("%a %e")},
+           {interval: d3.time.day, step: 2, formatter: Plottable.Formatters.time("%a %e")},
+        ],
+    ]);
+
+    var threeTierAxisHeight: number = xAxis.height();
+
+    assert.strictEqual(threeTierAxisHeight * 2, twoTierAxisHeight * 3, "two-tier axis is twice as tall as one-tier axis");
```
jtlan: Wrong assert message for the thing being checked here. Probably better to assert it like this:
```Typescript
assert.strictEqual(threeTierAxisHeight, twoTierAxisHeight * 3 / 2, "three-tier-axis is 50% taller than two-tier axis");
```

test/components/timeAxisTests.ts
```diff
@@ -179,7 +179,79 @@ describe("TimeAxis", () => {
...
+    xAxis.axisConfigurations([configuration]);
+
+    xAxis.renderTo(svg);
+
+    var numberOfVisibleTiers = (<any> xAxis)._element.selectAll(".time-axis-tier").filter(function() {
+      return d3.select(this).style("visibility") === "visible";
+    })[0].length;
+
+    assert.notStrictEqual(numberOfVisibleTiers, tiersToCreate,
+      "there should be less than " + tiersToCreate + " visible tiers");
```
jtlan: This is probably the wrong assert for the test.
This test is trying to check that "none of the visible tiers are drawn outside the `Axis`' bounding box", so we should probably write a check that tests that more directly. Maybe something like filtering the visible tiers, computing their bounding box, and making sure it's inside the `Axis`' bounding box?

src/components/axes/timeAxis.ts
bluong: This was discussed in person (I should've stated it in this thread), but @aicioara 's solution is probably the cleanest, most correct way to get the wanted behavior.

I am unsure how your solution would work @jtlan since this.tierLabelPositions() should never have a null position in the array

jtlan: Right, forgot about the number of tiers changing.

src/components/axes/timeAxis.ts
```diff
@@ -278,15 +289,40 @@ export module Axis {
...
+
+      var size = super._getSize(availableWidth, availableHeight);
+
+      var adjustedHeight = 0;
+      for (var i = 0; i < this._tierHeights.length; i++) {
+        if (adjustedHeight + this._tierHeights[i] > size.height) {
+          break;
+        }
+        adjustedHeight += this._tierHeights[i];
+      }
```
bluong: I think this version is the most performant, but what would happen if we were to encapsulate this into a reduce statement?

plottable.js
```diff
@@ -4716,12 +4743,25 @@ var Plottable;
...
+            Time.prototype._hideOverflowingTiers = function () {
+                var _this = this;
+                var availableHeight = this.height();
+                var usedHeight = 0;
+                var visibleAxisTiers = this._element.selectAll("." + Time.TIME_AXIS_TIER_CLASS).filter(function (d, i) {
+                    var visibility = d3.select(this).style("visibility");
+                    return visibility === "visible" || visibility === "inherit";
+                }).attr("visibility", function (d, i) {
+                    usedHeight += _this._tierHeights[i];
+                    return usedHeight <= availableHeight ? "visible" : "hidden";
```
jtlan: Use `"inherit"` instead of `"visible"`?

aicioara:
Test fails because IE9 considers that `window.getComputedStyle(element)` can have value `inherit`, while other browsers don't

src/components/axes/timeAxis.ts
```diff
@@ -420,11 +450,11 @@ export module Axis {
       );
 
       var baselineOffset = 0;
-      for (i = 0; i < Math.max(tierConfigs.length, 1); ++i) {
+      for (var i = 0; i < Math.max(tierConfigs.length, 1); ++i) {
         var attr = this._generateBaselineAttrHash();
         attr["y1"] += (this.orient() === "bottom") ? baselineOffset : -baselineOffset;
         attr["y2"] = attr["y1"];
-        this._tierBaselines[i].attr(attr).style("visibility", "visible");
+        this._tierBaselines[i].attr(attr).style("visibility", "inherit");
```
bluong: Why this change?

src/components/axes/timeAxis.ts
```diff
@@ -445,6 +476,21 @@ export module Axis {
       return this;
     }
 
+    private _hideOverflowingTiers() {
+      var availableHeight = this.height();
+      var usedHeight = 0;
+
+      var visibleAxisTiers = this._element
+        .selectAll("." + Time.TIME_AXIS_TIER_CLASS)
+        .filter(function(d: Element, i: number) {
```
bluong: d,i unneccessary

src/components/axes/timeAxis.ts
```diff
@@ -445,6 +476,21 @@ export module Axis {
...
+          var visibility = d3.select(this).style("visibility");
+          return visibility === "visible";
+        }).attr("visibility", (d: any, i: number) => {
```
bluong: Does this only set the visibility attribute on tiers that are "visible"?

src/components/axes/timeAxis.ts
aicioara: Suggestion from @jtlan . It makes more sense to enforce the visibility to inherit as that is the default behavior. It is also better practice because in the future the parent element might be set to `hidden` and this will break that change by enforcing visibility

src/components/axes/timeAxis.ts
aicioara: Agree. Btw, do we have defined coding standards anywhere so I can read what is the preferred way for this codebase. Some schools prefer to define all parameters for a function because:
- Changing the function later and making use of current parameters is easier (no need to look it up)

However, I want to be consistent with the rest of the codebase. Do you think I have to remove?

src/components/axes/timeAxis.ts
aicioara: You're right, that was wrong

aicioara:
There are lots of issues with IE in general. Label overlapping is not detected correctly
screen shot 2015-04-20 at 4 10 34 pm

Setting dates which work in Chrome 
`xScale.domain([new Date("2013-03-23 10:30"), new Date("2013-04-03 10:30")]);` Causes the entire axis to show NaN values.
screen shot 2015-04-20 at 4 13 28 pm
(Will fix in #1849 )

However, the feature itself works properly

src/components/axes/timeAxis.ts
aicioara: Can I make this show somehow?

aicioara:
@bluong The CR line got outdated, but I said at some point 

```
@bluong
d,i unneccessary
```

```
@aicioara
Agree. Btw, do we have defined coding standards anywhere so I can read what is the preferred way for this codebase. Some schools prefer to define all parameters for a function because:
- Changing the function later and making use of current parameters is easier (no need to look it up)

However, I want to be consistent with the rest of the codebase. Do you think I have to remove?
```

bluong:
I personally like removing arguments and then adding them in when needed.  Less complexity when reading in my opinion

src/components/axes/timeAxis.ts
```diff
@@ -278,15 +289,33 @@ export module Axis {
...
+    protected _getSize(availableWidth: number, availableHeight: number) {
+      // Makes sure that the size it requires is a multiple of tier sizes, such that
+      // we have no leftover tiers
+
+      var size = super._getSize(availableWidth, availableHeight);
+      size.height = this._tierHeights.reduce(function(prevValue, currValue, index, arr) {
+        return (prevValue + currValue > size.height) ? prevValue : (prevValue + currValue);
+      });
+      return size;
+    }
```
bluong: @jtlan Can you verify if this is the correct usage of _getSize? I know that you added this endpoint recently so you should have the best context here

src/components/axes/timeAxis.ts
```diff
@@ -469,7 +513,8 @@ export module Axis {
       var visibleTickLabels = this._tierLabelContainers[index]
                                     .selectAll("." + AbstractAxis.TICK_LABEL_CLASS)
                                     .filter(function(d: Element, i: number) {
-                                      return d3.select(this).style("visibility") === "visible";
+                                      var visibility = d3.select(this).style("visibility");
+                                      return visibility === "visible" || visibility === "inherit";
```
bluong: What happens if the parent's visbility is not visible?

aicioara: If the parent visibility is not visible 
```typescript
d3.select(this).style("visibility");
```
Returns "hidden" in everything except IE9. For IE9, I could not figure out what is the rule, nor I could find resources to tell me. It seems like sometimes it correctly computes the visibility, other times it does not. This filter is an optimization anyway, so it will work fine, even if more elements are returned.

aicioara:
@jtlan There is a question from @bluong regarding the `_getSize()` method

src/components/axes/timeAxis.ts
jtlan: This is exactly correct. `_getSize()` is used to retrieve the true size given a particular value of available space.

src/components/axes/timeAxis.ts
jtlan: Specifically #1033 was caused by not doing this.

src/components/axes/timeAxis.ts
```diff
@@ -278,15 +289,33 @@ export module Axis {
       return Math.min(this._getIntervalLength(config), this.width()) >= worstWidth;
     }
 
+    protected _getSize(availableWidth: number, availableHeight: number) {
+      // Makes sure that the size it requires is a multiple of tier sizes, such that
+      // we have no leftover tiers
+
+      var size = super._getSize(availableWidth, availableHeight);
+      size.height = this._tierHeights.reduce(function(prevValue, currValue, index, arr) {
```
jtlan: No fat-arrow notation? Generally our preferred style is to use `=>` except in cases where the existing `this` context is important.

aicioara:
@jtlan addressed. I left the `{}` in though because I had to split it on multiple lines anyway

crmorford:
it looks like it can handle up to 8 tiers, and at 10, you get a little extra white space at the bottom.
http://jsfiddle.net/pLapukjh/14/

aicioara:
If you increase the height, problem solved http://jsfiddle.net/pLapukjh/15/

crmorford:
is there a way to get it to fail without white space?

crmorford:
QE+1 as is, unless you try to fix white space bug, in which case i'll look at it again.

jtlan:
I think fixing the bug would require digging into the layout engine, so let's leave it for now. It's fairly minor.

aicioara:
Alright, I'll merge then.

PR #1830, opened 2015-04-13
Bar plot closest plot data

istarnx:


src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+     *
+     * @returns {PlotData} The PlotData closest to queryPoint
+     */
+    public getClosestPlotData(queryPoint: Point): PlotData {
+      var keys = this.datasetOrder();
+      var chartXExtent = { min: 0, max: this.width() };
+      var chartYExtent = { min: 0, max: this.height() };
+
+      var minDist = Infinity;
+      var closest: any;
```
bluong: Can we get a proper typing here?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
     }
 
     /**
+     * Retrieves the closest PlotData to queryPoint.
+     *
+     * Bars containing the queryPoint are considered closest. If queryPoint lies outside
+     * of all bars, we return the closest in the dominant direction (x for horizontal
+     * charts, y for vertical) and break ties
```
bluong: How are ties broken?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+     *
+     * @param {Point} queryPoint The point to which dataset points should be compared
```
bluong: The point being queried or similar.  I want to avoid the idea of dataset points as that's extraneous

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+     *
+     * @returns {PlotData} The PlotData closest to queryPoint
+     */
+    public getClosestPlotData(queryPoint: Point): PlotData {
+      var keys = this.datasetOrder();
```
bluong: Let's move this keys idea down to inline.  It reduces the number of variables we have to keep to keep track of.

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+      var minDist = Infinity;
+      var closest: any;
+
+      var accessor: { [axis: string]: (p: Point) => number } = {
+        "x": (p) => p.x,
+        "y": (p) => p.y
+      };
+
+      var dominantAxis: string = this._isVertical ? "x" : "y";
+      var secondaryAxis: string = this._isVertical ? "y" : "x";
```
bluong: We've been consistent with a primaryAttr / secondaryAttr throughout the codebase.  Let's rename to such

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+
+      keys.forEach((key) => {
```
bluong: What's happening in this huge block.  Comment or separate into a function perhaps.

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+        var plotData = this.getAllPlotData(key);
+        if (plotData != null && plotData.pixelPoints != null) {
```
bluong: When would these values return null?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+     * charts, y for vertical) and break ties
+     *
+     * @param {Point} queryPoint The point to which dataset points should be compared
+     *
+     * @returns {PlotData} The PlotData closest to queryPoint
+     */
+    public getClosestPlotData(queryPoint: Point): PlotData {
+      var keys = this.datasetOrder();
+      var chartXExtent = { min: 0, max: this.width() };
+      var chartYExtent = { min: 0, max: this.height() };
```
bluong: plots have an idea of a BBox, why can't we just query against the BBox rather than hold these extents here?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+      keys.forEach((key) => {
+        var plotData = this.getAllPlotData(key);
+        if (plotData != null && plotData.pixelPoints != null) {
+          plotData.pixelPoints.forEach((plotPt, i) => {
+            var bar = plotData.selection[0][i];
+
+            if (!Bar._intersectsBBox(chartXExtent, chartYExtent, bar.getBBox())) {
+              // bar isn't visible on plot; ignore it
+              return;
+            }
```
bluong: We don't currently have a function for bar visibility.  I think this function should be built out rather than using _intersectsBBox to proxy for it.

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+
+            var dist = Infinity;
+
+            if (Bar._intersectsBBox(queryPoint.x, queryPoint.y, bar.getBBox())) {
+              // queryPoint is inside of this bar
+              dist = -Infinity;
```
bluong: Why is dist not 0?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+        "y": (p) => p.y
+      };
+
+      var dominantAxis: string = this._isVertical ? "x" : "y";
+      var secondaryAxis: string = this._isVertical ? "y" : "x";
+
+      keys.forEach((key) => {
+        var plotData = this.getAllPlotData(key);
+        if (plotData != null && plotData.pixelPoints != null) {
+          plotData.pixelPoints.forEach((plotPt, i) => {
```
bluong: At this statement, consider a
var primaryAttr = this._isVertical ? plotPt.x : plotPt.y;
var secondaryAttr = this._isVertical ? plotPt.y : plotPt.x;

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+            }
+
+            var dist = Infinity;
+
+            if (Bar._intersectsBBox(queryPoint.x, queryPoint.y, bar.getBBox())) {
+              // queryPoint is inside of this bar
+              dist = -Infinity;
+            } else {
+              dist = Math.abs(accessor[dominantAxis](queryPoint) - accessor[dominantAxis](plotPt));
+            }
```
bluong: The above logic can be summed up into a ternary statement

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+            minDist = dist;
+          }
+
+          if (dist === minDist) {
+            secondaryClosest.push(candidate);
+          }
+        });
+
+        closest = secondaryClosest;
+      }
```
bluong: The above logic looks extremely similar to the logic dealing with being inside bars.  Can these two pieces of logic be collapsed?

bluong: We should be able to since we just need to keep careful track of the value of the minDist var

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,114 @@ export module Plot {
...
+            }
+          });
+        }
+      });
+
+      if (closest == null || closest.length === 0) {
+        return {data: [], pixelPoints: [], selection: d3.select()};
+      }
+
+      if (minDist >= 0) {
```
bluong: Why >= instead of >?

src/components/plots/barPlot.ts
```diff
@@ -165,30 +288,20 @@ export module Plot {
...
-      var tolerance: number = 0.5;
+    private _getBarsFromDataset(key: string, xValOrExtent: number | Extent, yValOrExtent: number | Extent): any[] {
       var bars: any[] = [];
 
       var drawer = <_Drawer.Element>this._key2PlotDatasetKey.get(key).drawer;
       drawer._getRenderArea().selectAll("rect").each(function(d) {
-        var bbox = this.getBBox();
-        if (bbox.x + bbox.width >= xExtent.min - tolerance && bbox.x <= xExtent.max + tolerance &&
-            bbox.y + bbox.height >= yExtent.min - tolerance && bbox.y <= yExtent.max + tolerance) {
+        if (Bar._intersectsBBox(xValOrExtent, yValOrExtent, this.getBBox())) {
```
bluong: I would very much like to keep the idea of only using extents in development code

bluong:
We can collapse this code by a significant amount

src/components/plots/barPlot.ts
```diff
@@ -46,6 +46,32 @@ export module Plot {
...
+     */
+    private static _intersectsBBox(xValOrExtent: number | Extent, yValOrExtent: number | Extent,
+      bbox: SVGRect, tolerance = 0.5) {
+      var xExtent: Extent = Bar._parseExtent(xValOrExtent);
+      var yExtent: Extent = Bar._parseExtent(yValOrExtent);
+
+      // SVGRects are positioned with sub-pixel accuracy (the default unit
+      // for the x, y, height & width attributes), but user selections (e.g. via
+      // mouse events) usually have pixel accuracy. A tolerance of half-a-pixel
+      // seems appropriate.
```
bluong: Looking back at this comment.  It doesn't seem all that useful to me.  Do you think it should be removed?

bluong: Consider moving this to a Util?

src/components/plots/barPlot.ts
```diff
@@ -46,6 +46,32 @@ export module Plot {
...
+
+    private static _parseExtent(input: any): Extent {
+      if (typeof (input) === "number") {
+        return { min: input, max: input };
+      } else if (input instanceof Object && "min" in input && "max" in input) {
+        return <Extent> input;
+      } else {
+        throw new Error("input '" + input + "' can't be parsed as an Extent");
+      }
+    }
```
bluong: Consider moving this to a Util?

src/components/plots/barPlot.ts
```diff
@@ -151,6 +166,81 @@ export module Plot {
...
+          if (!Bar._intersectsBBox(chartXExtent, chartYExtent, bar.getBBox())) {
+            // bar isn't visible on plot; ignore it
+            return;
+          }
+
+          var primaryDist: number, secondaryDist: number;
+          if (Bar._intersectsBBox(queryPoint.x, queryPoint.y, bar.getBBox())) {
+            // queryPoint is inside of this bar; this is as close as it can be to the bar
+            primaryDist = -Infinity;
+            secondaryDist = -Infinity;
```
bluong: What's the difference between -Infinity and 0?

bluong:
Just a question or 2 at this point.  Code mostly looks solid

src/components/plots/barPlot.ts
istarnx: This is a remnant of my previous hack at the problem. My intent was to differentiate between being *inside* of a bar and being on top of (zero-distance-from) a pixel point. Thinking on it again, however, it looks like the "on top of" case will trigger _intersectsBBox() anyway. Zero it is.

src/components/plots/barPlot.ts
istarnx: Y'know, I was thinking that... ;)

src/components/plots/barPlot.ts
istarnx: I think the comment might be helpful here, if only because we're overloading the term "intersects"? In general, is being inside of a SVGRect typically considered intersection? If you think it's clear without the comment, I'll remove it.

istarnx:
Ugh. I keep forgetting to "grunt test" locally. I've got some issues to fix up here.

src/components/plots/barPlot.ts
```diff
@@ -151,6 +140,80 @@ export module Plot {
...
+          if (!_Util.Methods.intersectsBBox(chartXExtent, chartYExtent, bar.getBBox())) {
+            // bar isn't visible on plot; ignore it
+            return;
+          }
+
+          var primaryDist = 0;
+          var secondaryDist = 0;
+
+          // if we're inside a bar, distance in both directions should stay 0
+          if (!_Util.Methods.intersectsBBox(queryPoint.x, queryPoint.y, bar.getBBox())) {
```
bluong: Let's store off the BBox since I think above? that is an expensive call

bluong:
I think it's just the one thing

PR #1831, opened 2015-04-14
[abstractComponentContainer] Can now merge groups that are already anchored

### LINKED ISSUE #1823: Can't merge Component that's already anchored
Repro: http://jsfiddle.net/0r93afbf/
Trying to add the label again produces the following message:
```
Uncaught Error: Can't presently merge a component that's already been anchored
```

Expected: adding a `Component` to a new `Group` should work even if the Component is already rendered.

aicioara:
Changelog:
- No longer setting the parent field, but rather have a method for doing that, which also detaches the component from its old parent (if there's the case)
- No longer throwing an error if the component was already setup or anchored, as we can now do it
- Throwing an error if we are trying to add `null` to a table cell

Closes #1823 

Before: http://jsfiddle.net/0r93afbf/
After: http://jsfiddle.net/0r93afbf/5/

Fiddle explanation: Click on the button. If it works correctly, the label "AWESOME" should be taken from the x-Axis and put inside the graph. The default position is center (vertically and horizontally)

### DIFF
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -194,6 +194,7 @@ export module Component {
      * @returns {Component} The calling component.
      */
     public renderTo(element: String | D3.Selection): AbstractComponent {
+      this.detach();
       if (element != null) {
         var selection: D3.Selection;
         if (typeof(element) === "string") {
@@ -210,6 +211,7 @@ export module Component {
         throw new Error("If a component has never been rendered before, then renderTo must be given a node to render to, \
           or a D3.Selection, or a selector string");
       }
+
       this._computeLayout();
       this._render();
       // flush so that consumers can immediately attach to stuff we create in the DOM
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -423,9 +423,6 @@ export module Component {
 
     public _merge(c: AbstractComponent, below: boolean): Component.Group {
       var cg: Component.Group;
-      if (this._isSetup || this._isAnchored) {
-        throw new Error("Can't presently merge a component that's already been anchored");
-      }
       if (Plottable.Component.Group.prototype.isPrototypeOf(c)) {
         cg = (<Plottable.Component.Group> c);
         cg._addComponent(this, below);
@@ -489,6 +486,11 @@ export module Component {
       this._isAnchored = false;
       this._parent = null;
       return this;
+    }
+
+    public _setParent(parent: AbstractComponent) {
+      this.detach();
+      this._parent = parent;
     }
 
     /**
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -488,7 +488,7 @@ export module Component {
       return this;
     }
 
-    public _setParent(parent: AbstractComponent) {
+    public _setParent(parent: AbstractComponentContainer) {
       this.detach();
       this._parent = parent;
     }
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -211,7 +211,6 @@ export module Component {
         throw new Error("If a component has never been rendered before, then renderTo must be given a node to render to, \
           or a D3.Selection, or a selector string");
       }
-
       this._computeLayout();
       this._render();
       // flush so that consumers can immediately attach to stuff we create in the DOM
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -12,7 +12,7 @@ export module Component {
     private _xOrigin: number; // Origin of the coordinate space for the component. Passed down from parent
     private _yOrigin: number;
 
-    public _parent: AbstractComponentContainer;
+    private _parentElement: AbstractComponentContainer;
     private _xAlignProportion = 0; // What % along the free space do we want to position (0 = left, .5 = center, 1 = right)
     private _yAlignProportion = 0;
     protected _fixedHeightFlag = false;
@@ -182,7 +182,7 @@ export module Component {
         if (this._isTopLevelComponent) {
           this._scheduleComputeLayout();
         } else {
-          this._parent._invalidateLayout();
+          this._parent()._invalidateLayout();
         }
       }
     }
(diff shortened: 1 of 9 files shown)


aicioara:
.

src/components/abstractComponent.ts
```diff
@@ -491,6 +488,11 @@ export module Component {
       return this;
     }
 
+    public _setParent(parent: AbstractComponentContainer) {
```
jtlan: Since we have this method now, please make `_parent` private or protected.

test/core/componentGroupTests.ts
```diff
@@ -185,6 +185,48 @@ describe("ComponentGroups", () => {
...
+
+    it("can move components to other groups after anchoring", () => {
+      var svg = generateSVG();
+
+      var xScale = new Plottable.Scale.Category();
+      xScale.domain(["A", "B", "C"]);
+      var xAxis = new Plottable.Axis.Category(xScale, "bottom");
+      var yScale = new Plottable.Scale.Linear();
+      var yAxis = new Plottable.Axis.Numeric(yScale, "left");
+      var chart = new Plottable.Component.Table([
```
jtlan: You shouldn't have to use a `Table` in this test; simply create two `Group`s in different `<svg>`s and move a `Component` between them after rendering.

src/components/table.ts
```diff
@@ -51,14 +51,16 @@ export module Component {
       this.classed("table", true);
       rows.forEach((row, rowIndex) => {
         row.forEach((component, colIndex) => {
-          this.addComponent(rowIndex, colIndex, component);
+          if (component != null) {
```
jtlan: Is there a test associated with this change?

jtlan:
In general, when writing tests, try to use as few parts of the code not under test as possible. In this case, for example, your test includes a `Table` and two `Axes` -- but the bug in question can manifest using only a pair of `Group`s (the `Component` being tested) and an `AbstractComponent` (the parent class).

You can think of a test as though it were the proof of a theorem -- you want to base them either on axioms (in this analogy, code assumed to work, like fundamental parts of Javascript) or other theorems you've already proven, preferably those the shortest distance from an axiom (in this analogy, other tested parts of the code, higher up the inheritance chain if possible).

src/components/table.ts
aicioara: This is just an optimization. The case is caught later in the abstractComponentContainer._addComponent(). Not sure what you mean with testing

aicioara:
Fixed the test.

src/components/table.ts
jtlan: Maybe we should test the behavior of `this.addComponent(row, col, null)` if the cell is already occupied.

test/core/componentGroupTests.ts
```diff
@@ -185,6 +185,46 @@ describe("ComponentGroups", () => {
...
+      );
+
+      assert.strictEqual(cg2.components().length, 1,
+        "Second group should have 1 component after movement");
+
+      assert.strictEqual(cg1.components().length, 0,
+        "First group should have no components after movement");
+
+      assert.strictEqual(c._parent(), cg2,
+        "component's parent after movement should be the group 1"
```
jtlan: Should be group 2?

jtlan:
Minor test error, and we probably should have a test for `Table`'s `addComponent()` method.

src/components/table.ts
aicioara: Done

aicioara:
All suggested changes added. Waiting for compilation.

src/components/table.ts
```diff
@@ -83,8 +85,7 @@ export module Component {
       var currentComponent = this._rows[row] && this._rows[row][col];
 
       if (currentComponent) {
-        currentComponent.detach();
-        component = component.above(currentComponent);
+        component = currentComponent.below(component);
```
jtlan: Why this change?

aicioara:
Discussion with @jtlan and @bluong concluded that there are 3 options when adding null to a table cell where there is something
- Throw an Error - We cannot add null to a table cell. Pro for this was that the user is notified straight away that something wrong has happened. Code does not fail later on unexpectedly
- Do not do anything - fail silently. Pro for this would be that user might have a list of components, some of which are null and wants to add them to the cell
- Create a group containing only 1 element (the element which was originally in the cell). Pro for this was that user expects a group, even though he added a null

Went for option 1

Solved that. Adding `null` to a table cell now throws an exception. Tests changed accordingly.

test/core/tableTests.ts
```diff
@@ -106,6 +106,13 @@ describe("Tables", () => {
     assert.equal(components[2], c3, "The Component was added to the existing Group");
   });
 
+  it("add null to a table cell where there was a group should throw an error", () => {
```
jtlan: "where there was a group" is no longer necessary in this description.

jtlan:
One minor thing.

crmorford:
should i be able to steal back an anchored plot and render it to its own svg? http://jsfiddle.net/ck1psjL3/22/, lines 26&28

crmorford:
Update on last comment. Some sort of bug. 
http://jsfiddle.net/ck1psjL3/25/ (i can't un-anchor a plot and render it to a different size svg correctly)
vs http://jsfiddle.net/ck1psjL3/24/ (i can render a plot back and forth between svgs no problem)

aicioara:
Working on that

aicioara:
Fixed:
Note the detach http://jsfiddle.net/ck1psjL3/27/
@jtlan do you want this in code as well? See my commit.

jtlan:
Yes, a `Component` that gets sent anywhere should disconnect itself from its parent.

aicioara:
Fixed

test/core/componentTests.ts
```diff
@@ -403,6 +403,33 @@ describe("Component behavior", () => {
     svg.remove();
   });
 
+  it("rendering to a new svg detaches the component", () => {
```
jtlan: I think properly this test belongs somewhere else, but I think we'll leave it for now.

aicioara: I actually gave it a thought. I put it in the same category with the other detach() methods on the component.

PR #1832, opened 2015-04-14
Fix attr doc

### LINKED ISSUE #1829: attr() documentation refers to "r" instead of "size"
http://plottablejs.org/docs/classes/plottable.plot.scatter.html#attr

bluong:
Closes #1829


plottable.d.ts
```diff
@@ -2682,12 +2682,12 @@ declare module Plottable {
              *
              * Here's a common use case:
              * ```typescript
-             * plot.attr("r", function(d) { return d.foo; });
+             * plot.attr("x", function(d) { return d.foo; });
              * ```
              * This will set the radius of each datum `d` to be `d.foo`.
```
jtlan: Description still refers to radius.

jtlan:
When fixing doc make sure you also fix the english parts, not just the source code parts...

src/components/plots/abstractPlot.ts
```diff
@@ -149,12 +149,13 @@ export module Plot {
...
+     * plot.attr("x", function(d) { return d.foo; }, xScale);
      * ```
-     * This will set the radius of each datum `d` to be `d.foo`.
+     * This will set the x accessor of each datum `d` to be `d.foo`,
+     * scaled in accordance with `xScale`
      *
      * @param {string} attrToSet The attribute to set across each data
-     * point. Popular examples include "x", "y", "r". Scales that inherit from
+     * point. Popular examples include "x", "y". Scales that inherit from
      * Plot define their meaning.
```
jtlan: "Scales that inherit from Plot define their meaning"

... This doesn't make any sense.

jtlan:
More weirdness in the doc we should probably correct now, since we're here.

PR #1833, opened 2015-04-14
fixing build files

bluong:
The build files were not updated when the typings files were not updated...  New branches off of develop will include this diff...


jtlan:
? I don't see this same diff when I pull new **develop**...

PR #1835, opened 2015-04-14
First pass Line.getClosestPlotData() impl.

istarnx:
Brandon, a couple of questions: 

1. I've repeated code in linePlotTests.ts that could be shared once this merges with the Bar closest plot data work. See assertPlotDataEqual(). What are your thoughts on sharing this code? And where would you prefer to put it if we were to share?

2. There's also a block of code at the end of getClosestPlotData() that could be shared between the bar and line case. We can talk about that as well.

### DIFF
--- a/src/components/plots/linePlot.ts
+++ b/src/components/plots/linePlot.ts
@@ -168,11 +168,9 @@ export module Plot {
       var minXDist = Infinity;
       var minYDist = Infinity;
 
-      var closest: {
-        datum: any;
-        pixelPoint: Point;
-        node: Node
-      }[];
+      var closestData: any[] = [];
+      var closestPixelPoints: Point[] = [];
+      var closestElements: Element[] = [];
 
       this.datasetOrder().forEach((key: string) => {
         var plotData = this.getAllPlotData(key);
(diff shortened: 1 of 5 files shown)


src/components/plots/linePlot.ts
```diff
@@ -154,33 +154,70 @@ export module Plot {
...
-      var closestDatum: any;
-      var closestSelection: D3.Selection;
-      var closestPoint: Point;
+      var minXDist = Infinity;
+      var minYDist = Infinity;
+
+      var closest: {
+        datum: any;
+        pixelPoint: Point;
+        node: Node
```
bluong: For this I've been using EventTarget consistently, but I'm not sure if I should've used Node instead.  Which one do you think we should use?

src/components/plots/linePlot.ts
```diff
@@ -154,33 +154,70 @@ export module Plot {
...
+              node: plotData.selection[0][0]
+            });
           }
         });
       });
 
-      if (closestDatum == null) {
+      if (minXDist === Infinity) {
         return { data: [], pixelPoints: [], selection: d3.select() };
       }
```
bluong: Why do you need this ifcase if closest is empty?

src/components/plots/linePlot.ts
```diff
@@ -154,33 +154,70 @@ export module Plot {
...
 
+      var data: any[] = [];
+      var pixelPoints: Point[] = [];
+      var nodes: Node[] = [];
+
+      closest.forEach((c: any) => {
+        data.push(c.datum);
+        pixelPoints.push(c.pixelPoint);
+        nodes.push(c.node);
+      });
```
bluong: Instead of playing with these closest objects, why not deal directly with data, pixelPoints, and nodes.  It should cut down quite a bit on number of lines.

bluong:
A lot of this looks fine.  Let's just package the code a bit more neatly

src/components/plots/linePlot.ts
istarnx: Brandon and I discussed this and we think we should be as specific as possible. In this case, I'll go with Element.

test/components/plots/linePlotTests.ts
```diff
@@ -242,17 +242,77 @@ describe("Plots", () => {
...
+        expected = {
+          data: [],
+          pixelPoints: [],
+          selection: d3.selection()
+        };
+
+        closest = linePlot.getClosestPlotData({x: d0Px.x, y: d0Px.y});
+        assert.lengthOf(closest.data, 0, "empty plots return empty data");
+        assert.lengthOf(closest.pixelPoints, 0, "empty plots return empty pixelPoints");
+        assert.isTrue(closest.selection.empty(), "empty plots return empty selection");
```
bluong: Can you break up this test?  It seems to be testing way too much.

For example "only in-view points are considered" I think should be a test

assert.lengthOf(closest.data, 0, "empty plots return empty data");
+        assert.lengthOf(closest.pixelPoints, 0, "empty plots return empty pixelPoints");
+ assert.isTrue(closest.selection.empty(), "empty plots return empty selection"); should be encapulsated into a single test

bluong:
Testing cleanup

PR #1836, opened 2015-04-14
[colorScale] Quickfix for infinite recursion bug on CSS colors

### LINKED ISSUE #1834: Reading colors from CSS can lock up browser
`Scale.Color` reads the colors from **plottable.css** by applying the `plottable-colors-N` css classes to a test `<div>`. If you happen to have this in your styles:
```CSS
div {
  background-color: yellow;
}
```
it will loop forever trying to load colors and lock up the browser.

No JSFiddle repro because, well, it'd lock up your browser.

aicioara:
BLUF Creating a `<plottable_color_testing>` element is less likely to be background-colored by an external stylesheet

Added unit test. Testing the exact bug fix is not a good idea, since the bug itself freezes the entire browser. Instead UnitTest is testing if the CSS specified color is applied. By doing this we can test cross-browser solution of the fix.

Introduction: We are having an internal representation of all Components. The ColorScale knows the possible colors it can have. It does this by creating `<div>` elements and getting their color (such that CSS is also taken into account). 
Problem: Since the user can specify more than the `20` colors we use by default, the JS keeps creating `<div>`s until the divs don't have a color anymore. Also, the JS cannot read the CSS directly. Now, if the `<div>`s themselves have color set, the loop that gets the colors is infinite (and crashes browser)
```css
div { background-color: blue}
``` 
Solution (hack): Create a `<plottable-color-testing>` element, of which we get the color. There is little to no chance this will be specified in the CSS

Also added extra layers of security
- We cap the maximum number of colors we read from the CSS to 256 (such that we never ever crash the browser with an infinite loop)
- We try to recover if someone uses the * selector and we detect such thing. The way this works is: we check what is the default `background-color` of the `<plottable-color-testing>` element, call it x. We read colors until we find color x occurring twice in a row, at which point we stop, adding it only once to the final color array.

Better solutions might exist

Closes #1834

### DIFF
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -60,7 +60,7 @@ export module Scale {
 
     private static _getPlottableColors(): string[] {
       var plottableDefaultColors: string[] = [];
-      var colorTester = d3.select("body").append("plottable_color_testing");
+      var colorTester = d3.select("body").append("plottable-color-tester");
       var i = 0;
       var colorHex: string;
       while ((colorHex = _Util.Methods.colorTest(colorTester, "plottable-colors-" + i)) !== null) {
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -68,7 +68,6 @@ export module Scale {
         i++;
       }
       colorTester.remove();
-      console.log("AAA", plottableDefaultColors);
       return plottableDefaultColors;
     }
 
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -6,6 +6,7 @@ export module Scale {
 
     private static HEX_SCALE_FACTOR = 20;
     private static LOOP_LIGHTEN_FACTOR = 1.6;
+    private static MAXIMUM_COLORS = 256;
 
     /**
      * Constructs a ColorScale.
@@ -63,7 +64,8 @@ export module Scale {
       var colorTester = d3.select("body").append("plottable-color-tester");
       var i = 0;
       var colorHex: string;
-      while ((colorHex = _Util.Methods.colorTest(colorTester, "plottable-colors-" + i)) !== null) {
+      while ((colorHex = _Util.Methods.colorTest(colorTester, "plottable-colors-" + i)) !== null &&
+              i < this.MAXIMUM_COLORS) {
         plottableDefaultColors.push(colorHex);
         i++;
       }
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -60,7 +60,7 @@ export module Scale {
 
     private static _getPlottableColors(): string[] {
       var plottableDefaultColors: string[] = [];
-      var colorTester = d3.select("body").append("div");
+      var colorTester = d3.select("body").append("plottable");
       var i = 0;
       var colorHex: string;
       while ((colorHex = _Util.Methods.colorTest(colorTester, "plottable-colors-" + i)) !== null) {
@@ -68,6 +68,7 @@ export module Scale {
         i++;
       }
       colorTester.remove();
+      console.log("AAA", plottableDefaultColors);
       return plottableDefaultColors;
     }
 
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -6,6 +6,8 @@ export module Scale {
 
     private static HEX_SCALE_FACTOR = 20;
     private static LOOP_LIGHTEN_FACTOR = 1.6;
+    //The maximum number of colors we are getting from CSS stylesheets
+    private static MAXIMUM_COLORS_FROM_CSS = 256;
 
     /**
      * Constructs a ColorScale.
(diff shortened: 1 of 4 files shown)


aicioara:
Checking if it works for all browsers

src/scales/colorScale.ts
```diff
@@ -68,6 +68,7 @@ export module Scale {
         i++;
       }
       colorTester.remove();
+      console.log("AAA", plottableDefaultColors);
```
bluong: Remove console.log

src/scales/colorScale.ts
```diff
@@ -60,7 +60,7 @@ export module Scale {
 
     private static _getPlottableColors(): string[] {
       var plottableDefaultColors: string[] = [];
-      var colorTester = d3.select("body").append("div");
+      var colorTester = d3.select("body").append("plottable");
```
bluong: How would this work?  There's no such thing as a plottable element?

bluong:
Can you explain the fix and possibly the underlying issue of the buggy behavior?

src/scales/colorScale.ts
aicioara: What we were doing before was a hack anyway. Elements don't have to exist in order to have CSS classes attached to them

aicioara:
This is a quickfix. After talking to @jtlan we decided quickfix is better now, then create a new ticket if needed.

aicioara:
@bluong Fix explained in the description now

aicioara:
Tested in Safari, IE9, Chrome, PhantomJS and Firefox. Creating fake (unsupported) `<plottable>` elements is actually fine in all those browsers

bluong:
Just need to change the element name that we are appending to be like <plottable-color-tester> or so and we should be good

aicioara:
The DOM Element we create is called `<plottable_color_testing>`

bluong:
Rename to plottable-color-tester please

jtlan:
Technically this isn't immune to the problem, since someone could write:
```CSS
plottable-color-tester {
  fill: blue;
}
```
to crash it; this requires malicious intent though.

aicioara:
The idea is to lower the chance. We need a proper way to implement this anyway, since I believe dynamic CSS adding should also trigger graph color change

src/scales/colorScale.ts
```diff
@@ -6,6 +6,7 @@ export module Scale {
 
     private static HEX_SCALE_FACTOR = 20;
     private static LOOP_LIGHTEN_FACTOR = 1.6;
+    private static MAXIMUM_COLORS = 256;
```
jtlan: Probably should give this a name that indicates it's the maximum number of colors when loading them via CSS; the user can still set more colors using `.range()`.

src/scales/colorScale.ts
```diff
@@ -60,11 +60,16 @@ export module Scale {
...
       var plottableDefaultColors: string[] = [];
-      var colorTester = d3.select("body").append("div");
+      var colorTester = d3.select("body").append("plottable-color-tester");
+
+      var defaultColorHex: string = _Util.Methods.colorTest(colorTester, "");
       var i = 0;
       var colorHex: string;
       while ((colorHex = _Util.Methods.colorTest(colorTester, "plottable-colors-" + i)) !== null) {
         plottableDefaultColors.push(colorHex);
+        if (colorHex === defaultColorHex && colorHex === plottableDefaultColors[plottableDefaultColors.length - 1]) {
```
jtlan: This should probably be one line higher, so we break before the color is added a second time, not after.

jtlan:
We should still break out after too many iterations. This selector can break what you've written:
```CSS
[class^="plottable-"] {
  background-color: pink;
}
```

aicioara:
Also added extra layers of security
- We cap the maximum number of colors we read from the CSS to 256 (such that we never ever crash the browser with an infinite loop)
- We try to recover if someone uses the * selector and we detect such thing. The way this works is: we check what is the default `background-color` of the `<plottable-color-testing>` element, call it x. We read colors until we find color x occurring twice in a row, at which point we stop, adding it only once to the final color array.

Added unit tests to check for this

test/scales/scaleTests.ts
```diff
@@ -291,6 +291,65 @@ describe("Scales", () => {
...
+    it("accepts CSS specified colors", () => {
+
+      var style = d3.select("body").append("style");
+      style.html(".plottable-colors-0 {background-color: #ff0000 !important; }");
+
+      var scale = new Plottable.Scale.Color();
+      assert.strictEqual(scale.range()[0], "#ff0000", "User has specified red color for first color scale color");
+      assert.strictEqual(scale.range()[1], "#fd373e", "The second color of the color scale should be the same");
+
+      style.remove();
```
jtlan: Minor thing; we should move this line (and subsequent `maliciousStyle.remove()` lines) to immediately after the test scale is created. That way, if any `assert` statements fail, the `<style>` element doesn't remain stranded on the page, affecting all other tests.

jtlan:
+1 from (substitute) QE.

PR #1840, opened 2015-04-15
Interaction.Drag no longer triggers on right-click.

jtlan:


test/testUtils.ts
```diff
@@ -32,25 +32,6 @@ function makeFakeEvent(x: number, y: number): D3.D3Event {
     };
 }
 
-function fakeDragSequence(anyedInteraction: any, startX: number, startY: number, endX: number, endY: number) {
```
jtlan: This code was added back in by mistake during a merge...

src/interactions/dragInteraction.ts
```diff
@@ -44,6 +44,10 @@ export module Interaction {
     }
 
     private _startDrag(p: Point, e: UIEvent) {
+      var button = (<MouseEvent> e).button;
+      if (button != null && button !== 0) {
+        return;
+      }
```
bluong: Wouldn't a more semantical check be...

if (e instanceof MouseEvent && button !== 0) {
  return;
}

?

jtlan: Maybe; Checking for the button property is a sufficient proxy though, since that's what we're actually interested in.

PR #1842, opened 2015-04-16
1. Tweaked the closest bar algo such that being within a bar's secondary...

istarnx:
... axis span sets its distance to 0.

2. Now test bbox intersection with a tolerance of zero: gives us stable behaviour when moving along the dominant axis at the same secondary axis value.
3. Tests. :)

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -164,13 +164,18 @@ export module Plot {
       var queryPtPrimary = this._isVertical ? queryPoint.x : queryPoint.y;
       var queryPtSecondary = this._isVertical ? queryPoint.y : queryPoint.x;
 
+      // SVGRects are positioned with sub-pixel accuracy (the default unit
+      // for the x, y, height & width attributes), but user selections (e.g. via
+      // mouse events) usually have pixel accuracy. We add a tolerance of 0.5 pixels.
+      var tolerance = 0.5;
+
       this.datasetOrder().forEach((key) => {
         var plotData = this.getAllPlotData(key);
         plotData.pixelPoints.forEach((plotPt, i) => {
           var bar = plotData.selection[0][i];
           var barBBox = bar.getBBox();
 
-          if (!_Util.Methods.intersectsBBox(chartXExtent, chartYExtent, barBBox, 0)) {
+          if (!_Util.Methods.intersectsBBox(chartXExtent, chartYExtent, barBBox, tolerance)) {
             // bar isn't visible on plot; ignore it
             return;
           }
@@ -179,7 +184,7 @@ export module Plot {
           var secondaryDist = 0;
 
           // if we're inside a bar, distance in both directions should stay 0
-          if (!_Util.Methods.intersectsBBox(queryPoint.x, queryPoint.y, barBBox, 0)) {
+          if (!_Util.Methods.intersectsBBox(queryPoint.x, queryPoint.y, barBBox, tolerance)) {
             var plotPtPrimary = this._isVertical ? plotPt.x : plotPt.y;
             primaryDist = Math.abs(queryPtPrimary - plotPtPrimary);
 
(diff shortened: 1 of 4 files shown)


test/components/plots/stackedBarPlotTests.ts
```diff
@@ -89,6 +89,50 @@ describe("Plots", () => {
       assert.deepEqual(dataset2.data(), originalData2, "underlying data is not modified");
       svg.remove();
     });
+
+    it("returns correct closest plot data", () => {
```
bluong: test name is not precise enough

src/components/plots/barPlot.ts
```diff
@@ -179,12 +179,21 @@ export module Plot {
           var secondaryDist = 0;
 
           // if we're inside a bar, distance in both directions should stay 0
-          if (!_Util.Methods.intersectsBBox(queryPoint.x, queryPoint.y, barBBox)) {
+          if (!_Util.Methods.intersectsBBox(queryPoint.x, queryPoint.y, barBBox, 0)) {
```
bluong: Why this change?

src/components/plots/barPlot.ts
```diff
@@ -179,12 +179,21 @@ export module Plot {
...
-            var plotPtSecondary = this._isVertical ? plotPt.y : plotPt.x;
-
             primaryDist = Math.abs(queryPtPrimary - plotPtPrimary);
-            secondaryDist = Math.abs(queryPtSecondary - plotPtSecondary);
+
+            // compute this bar's min and max along the secondary axis
+            var barMinSecondary = this._isVertical ? barBBox.y : barBBox.x;
+            var barMaxSecondary = barMinSecondary + (this._isVertical ? barBBox.height : barBBox.width);
+
+            if (queryPtSecondary > barMinSecondary && queryPtSecondary <= barMaxSecondary) {
```
bluong: Why > instead of >=

istarnx: So, my intent was to tie-break using this approach, but after thinking about it again, I'm not sure. _intersectsBBox() will end up picking two bars if you're exactly on their border value (note the >= and <= in its comparison code) so perhaps it's reasonable to do the same thing here and let the user deal with multiple entities exactly the same distance apart.

 so it's reasonable that this logic do the same and leave it up to the user to deal with two

src/components/plots/barPlot.ts
istarnx: Note the commit comment: "Now test bbox intersection with a tolerance of zero: gives us stable behaviour when moving along the dominant axis at the same secondary axis value." 

It's to make sliding the mouse outside of a bar, but on the same secondary axis value, keep that bar as closest. I.e., when the logic switches from the "within bar" to "within bar's y-range" the resultant bar was different for the same y value when I slid the mouse along x.

bluong:
Needs to obey tolerance

src/components/plots/barPlot.ts
```diff
@@ -179,12 +184,21 @@ export module Plot {
...
-            var plotPtSecondary = this._isVertical ? plotPt.y : plotPt.x;
-
             primaryDist = Math.abs(queryPtPrimary - plotPtPrimary);
-            secondaryDist = Math.abs(queryPtSecondary - plotPtSecondary);
+
+            // compute this bar's min and max along the secondary axis
+            var barMinSecondary = this._isVertical ? barBBox.y : barBBox.x;
+            var barMaxSecondary = barMinSecondary + (this._isVertical ? barBBox.height : barBBox.width);
+
+            if (queryPtSecondary >= barMinSecondary - tolerance && queryPtSecondary <= barMaxSecondary + tolerance) {
```
bluong: This makes sense from a code standpoint, but from a semantical standpoint, can we change it to have the same semantics as intersectsBBox?

aka

queryPtSecondary + tolerance >= barMinSecondary && queryPtSecondary - tolerance <= barMaxSecondary

This corresponds with the fact that the queryPt has an invisible rectangle around itself rather than the bbox of the bar having that invisible rectangle

PR #1844, opened 2015-04-16
[abstractStackedPlot] stackedBar fail gracefully when presented with wrong data

### LINKED ISSUE #1838: StackedBar doesn't handle string values correctly
If string values are supplied in the stacking direction of `StackedBar`, the following error is thrown:
```
Error: Invalid value for <rect> attribute y="NaN"
```

Repro: http://jsfiddle.net/d2te4yct/

aicioara:
Closes #1838

Before: http://jsfiddle.net/d2te4yct/
After: http://jsfiddle.net/d2te4yct/2/

Technical description: If the y-value of one of the data sets was not convertible to string for a given x-value, all subsequent bars that would stack on top of the broken bar for that x-value (for example in the fiddle the red bar is in air)
Exception thrown: `Error: Invalid value for <rect> attribute y="NaN"`
Solution: Default y-values to 0, when the string cannot be converted to the number

**Alternative approach**
Throw an exception when user presents such data. The difference is that the exception should be more meaningful. User most probably did not want to specify a non-convertible string for a numeric y-axis. @jtlan @bluong What do you think?
**Decision** : We do not want that behavior

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -248,8 +248,7 @@ export module _Util {
     }
 
     /**
-     * Returns true if and only if x is NaN
-     * Will return false for values like undefined, null, infinity, strings
+     * Returns true **only** if x is NaN
      */
     export function isNaN(n: any) {
       return n !== n;
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -248,6 +248,14 @@ export module _Util {
     }
 
     /**
+     * Returns true if and only if x is NaN
+     * Will return false for values like undefined, null, infinity, strings
+     */
+    export function isNaN(n: any) {
+      return n !== n;
+    }
+
+    /**
      * Creates shallow copy of map.
      * @param {{ [key: string]: any }} oldMap Map to copy
      *
--- a/test/components/plots/stackedPlotTests.ts
+++ b/test/components/plots/stackedPlotTests.ts
@@ -267,52 +267,6 @@ describe("Plots", () => {
     });
   });
 
-  describe("fail safe tests", () => {
-
-    var svg: D3.Selection;
-    var xScale: Plottable.Scale.Category;
-    var yScale: Plottable.Scale.Linear;
-    var data1: Object[];
-    var data2: Object[];
-
-    beforeEach(() => {
-      svg = generateSVG(600, 400);
-
-      data1 = [
-          { x: "A", y: "s", fill: "blue" },
-      ];
-      data2 = [
-          { x: "A", y: 1, fill: "red"},
-      ];
-      xScale = new Plottable.Scale.Category();
-      yScale = new Plottable.Scale.Linear();
-    });
-
-    afterEach(() => {
-      svg.remove();
-    });
-
-    it("conversion fails should be silent in Plot.StackedBar", () => {
-      var plot = new Plottable.Plot.StackedBar(xScale, yScale);
-      plot.addDataset("d1", data1);
-      plot.addDataset("d2", data2);
-      plot.project("fill", "fill");
-      plot.project("x", "x", xScale).project("y", "y", yScale);
-
-      var ds1PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d1").plotMetadata;
-      var ds2PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d2").plotMetadata;
-      var ds1FirstColumnOffset = ds1PlotMetadata.offsets.get("A");
-      var ds2FirstColumnOffset = ds2PlotMetadata.offsets.get("A");
-
-      assert.strictEqual(typeof ds1FirstColumnOffset, "number", "ds1 offset should be a valid number");
-      assert.strictEqual(typeof ds2FirstColumnOffset, "number", "ds2 offset should be a valid number");
-
-      assert.isFalse(Plottable._Util.Methods.isNaN(ds1PlotMetadata.offsets.get("A")), "ds1 offset should be a valid number");
-      assert.isFalse(Plottable._Util.Methods.isNaN(ds2PlotMetadata.offsets.get("A")), "ds2 offset should be a valid number");
-    });
-
-  });
-
   describe("scale extent updates", () => {
     var svg: D3.Selection;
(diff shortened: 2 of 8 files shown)


aicioara:
Writing more documentation on the issue and waiting for the commit to compile.

jtlan:
Probably better to fail gracefully on messy data -- messy data is a reasonably common scenario.

test/components/plots/stackedPlotTests.ts
```diff
@@ -267,6 +267,61 @@ describe("Plots", () => {
...
+      plot.addDataset("d1", data1);
+      plot.addDataset("d2", data2);
+      plot.project("fill", "fill");
+      plot.project("x", "x", xScale).project("y", "y", yScale);
+
+      var ds1PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d1").plotMetadata;
+      var ds2PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d2").plotMetadata;
+
+      assert.isFalse(isNaN(ds1PlotMetadata.offsets.get("A")), "ds1 offset should be a number");
+      assert.isFalse(isNaN(ds2PlotMetadata.offsets.get("A")), "ds2 offset should be a number");
```
jtlan: Don't use `isNaN()`, as it returns `true` for non-`NaN` values:
```Javascript
isNaN(undefined); // returns true;
isNaN({}); // returns true;
```

test/components/plots/stackedPlotTests.ts
```diff
@@ -267,6 +267,61 @@ describe("Plots", () => {
...
+      ];
+      xScale = new Plottable.Scale.Category();
+      yScale = new Plottable.Scale.Linear();
+    });
+
+    afterEach(() => {
+      svg.remove();
+    });
+
+    it("conversion fails should be silent in stackedBarPlot", () => {
```
jtlan: It's probably more consistent to use the reverse-dot naming whenever you refer to a `Plottable` object; so "`Plot.StackedBar`" (the actual name), not "`stackedBarPlot`".

jtlan:
Some minor test stuff.

aicioara:
For the `stackedArea`, this also happens
screen shot 2015-04-16 at 10 59 59 am
Ideally we want this 
screen shot 2015-04-16 at 11 01 08 am
Test data
```typescript
var data1 = [
    { x: "A", y: 0, fill: "blue" },
    { x: "B", y: 2, fill: "blue" },
    { x: "C", y: 3, fill: "blue" },
];
var data2 = [
    { x: "A", y: 1, fill: "red"},
    { x: "B", y: 2, fill: "blue" },
    { x: "C", y: 3, fill: "blue" },
];

var xScale = new Plottable.Scale.Category();
var yScale = new Plottable.Scale.Linear();
var plot = new Plottable.Plot.StackedArea(xScale, yScale);
plot.addDataset(data1);
plot.addDataset(data2);
plot.project("fill", "fill");
plot.project("x", "x", xScale).project("y", "y", yScale);

var chart = new Plottable.Component.Table([[plot]]);

plot.renderTo("#chart");
```

test/components/plots/stackedPlotTests.ts
aicioara: `"string" !== "another string"` also returns `true`. How about checking for undefined then? or checking `typeof x === "number"`

aicioara:
Decided to open a new pull request for the stackedArea itself as it seems more complex and outside the scope of this ticket. See #1848

aicioara:
There is still a type error as the StackedArea cannot receive a `
`Plottable.Scale.Category` scale as its argument

aicioara:
Deleted the Plot.StackedArea test as it will be covered in the newly opened ticket and now it passes type checks

src/utils/utils.ts
```diff
@@ -248,6 +248,13 @@ export module _Util {
     }
 
     /**
+     * Returns true **only** if x is NaN
+     */
+    export function isNaN(n: any) {
+      return n !== n;
```
jtlan: It's this type of stuff that makes me miss `typedef` from C...

test/components/plots/stackedPlotTests.ts
```diff
@@ -267,6 +267,52 @@ describe("Plots", () => {
...
+      ];
+      xScale = new Plottable.Scale.Category();
+      yScale = new Plottable.Scale.Linear();
+    });
+
+    afterEach(() => {
+      svg.remove();
+    });
+
+    it("conversion fails should be silent in Plot.StackedBar", () => {
```
jtlan: Wouldn't it make more sense for this to be in **stackedBarPlotTests.ts** ?

test/components/plots/stackedPlotTests.ts
```diff
@@ -267,6 +267,52 @@ describe("Plots", () => {
...
+
+      var ds1PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d1").plotMetadata;
+      var ds2PlotMetadata = <Plottable.Plot.StackedPlotMetadata>(<any> plot)._key2PlotDatasetKey.get("d2").plotMetadata;
+      var ds1FirstColumnOffset = ds1PlotMetadata.offsets.get("A");
+      var ds2FirstColumnOffset = ds2PlotMetadata.offsets.get("A");
+
+      assert.strictEqual(typeof ds1FirstColumnOffset, "number", "ds1 offset should be a valid number");
+      assert.strictEqual(typeof ds2FirstColumnOffset, "number", "ds2 offset should be a valid number");
+
+      assert.isFalse(Plottable._Util.Methods.isNaN(ds1PlotMetadata.offsets.get("A")), "ds1 offset should be a valid number");
```
jtlan: This `assert` has the same message as the one on Line 307:
* What's the difference between the two `assert`s?
* The message for this one should more accurately reflect what is being asserted. In this case, though, we should probably check that the value in question is a `number` and *also* that it is not `NaN`.

jtlan:
I see that this change fixes the stacking problem by treating non-valid values as `0` for the purposes of stacking; however, there is no test that checks for this particular functionality (the tests only make sure that the offsets are valid numbers). Can we make sure that the intended behavior of the new functionality is actually being checked in the test code?

aicioara:
Blocking #1850 and #1848

PR #1847, opened 2015-04-16
Group get closest plot data

istarnx:
Note that PlotData now includes "plot." I updated the code in a number of places to include this data.


src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
...
+     * @param {Point} queryPoint The point to which plot data should be compared
+     *
+     * @returns {PlotData} The PlotData closest to queryPoint
+     */
+    public getClosestPlotData(queryPoint: Point): Plot.PlotData {
+      var minDistSquared = Infinity;
+      var closestPlotData: Plot.PlotData = {
+        data: [],
+        pixelPoints: [],
+        plot: null,
```
bluong: Is this plot parameter valid?  Imagine if I were to merge a scatterPlot and a barPlot

bluong: Oh so closestPlotData only returns a singular item?  Is this not consistent with what plot.getClosestPlotData does?

src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
       components.forEach((c: AbstractComponent) => this._addComponent(c));
     }
 
+    /**
+     * Retrieves closest PlotData to queryPoint across all plots in this group.
+     * Plots define their own notion of closeness; this function compares results across
```
bluong: How does the function compare results / what are the results you are referring to?

src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
...
+     *
+     * @returns {PlotData} The PlotData closest to queryPoint
+     */
+    public getClosestPlotData(queryPoint: Point): Plot.PlotData {
+      var minDistSquared = Infinity;
+      var closestPlotData: Plot.PlotData = {
+        data: [],
+        pixelPoints: [],
+        plot: null,
+        selection: d3.selectAll([])
```
bluong: Use d3.select() instead.  It'll give you a null selection

src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
...
+      };
+
+      this.components().forEach((c) => {
+        // consider only components implementing this function
+        if ((<any>c).getClosestPlotData !== undefined) {
```
bluong: I personally like to put a space between the cast and the variable like
(<any> c)

bluong: We should not be anycasting this.  You should be able to do a call like...

if (c instanceof Plot.AbstractPlot)

I think

src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
...
+          var cpd = (<any>c).getClosestPlotData(queryPoint);
```
bluong: Cast c to Plot.AbstractPlot

src/components/componentGroup.ts
```diff
@@ -23,6 +23,48 @@ export module Component {
...
+          if (cpd.data.length > 0) {
```
bluong: If all we're caring about is the pixelPoint, let's go with the more semantical appraoch

var plotClosestPixelPoint (or so) = cpd.pixelPoints[0]
if (plotClosestPixelPoint != null) {
do stuff
}

test/core/componentGroupTests.ts
```diff
@@ -284,4 +284,67 @@ describe("ComponentGroups", () => {
...
+        selection: c2Selection
+      };
+
+      var c1 = new Plottable.Component.AbstractComponent();
+      (<any>c1).getClosestPlotData = (p: Plottable.Point) => {
+        return c1PlotData;
+      };
+
+      var c2 = new Plottable.Component.AbstractComponent();
+      (<any>c2).getClosestPlotData = (p: Plottable.Point) => {
```
bluong: Don't anycast if you can help it

istarnx:
Closing to re-target at develop: https://github.com/palantir/plottable/pull/1877

PR #1852, opened 2015-04-17
Closest plot data

istarnx:


vchu123:
QE +1

Known issues to be fixed in follow-up
1.  Scatter Plot closest point can targets points off chart
2.  Bar charts with multiple series (non-stacked) always targets top bar
3.  Possibly expose more options instead of forcing x then y targetting

PR #1855, opened 2015-04-17
[InterpolatedColorLegend] Enabled expansion and gradients

### LINKED ISSUE #1568: Size of InterpolatedColorLegend
screen shot 2015-01-30 at 5 18 05 pm
fiddle: http://jsfiddle.net/df5hjpok/

Does it make sense for `InterpolatedColorLegend` to expand to fill the available space? While it might be reasonable for vertical ones ("right" or "left"), it'd be weird for a horizontal one to expand with a very wide chart...

rcchen:
image

Fixes #1568

JSFiddle: http://jsfiddle.net/owj8pn8u/1/

### DIFF
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -269,9 +269,7 @@ export module Component {
                 textRotation: 0
             };
 
-      // Width of each individual swatch box
       var swatchWidth: number;
-      // Height of each individual swatch box
       var swatchHeight: number;
       var swatchX: (d: any, i: number) => number;
       var swatchY: (d: any, i: number) => number;
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -124,19 +124,29 @@ export module Component {
     }
 
     /**
+     * Gets whether the InterpolatedColorLegend is a gradient or not.
+     *
+     * @returns {boolean} The gradient state of the ICL.
+     */
+    public gradient(): boolean;
+    /**
      * Apply a gradient to the InterpolatedColorLegend.
      *
      * @param {boolean} isGradient Whether there should be a gradient or not.
      *
      * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
      */
-     public gradient(isGradient: boolean): InterpolatedColorLegend {
-       this._isGradient = isGradient;
-       return this;
-     }
+    public gradient(isGradient: boolean): InterpolatedColorLegend;
+    public gradient(isGradient?: boolean): any {
+      if (isGradient == null) {
+        return this._isGradient;
+      }
+      this._isGradient = isGradient;
+      return this;
+    }
 
      /**
-      * Expand the InterpolatedColorLegend to the full length/height of the chart.
+      * Expand the InterpolatedColorLegend to the full length/height of the component space.
       *
       * @param {boolean} isExpanded Whether the legend is expanded or not.
       *
@@ -200,10 +210,10 @@ export module Component {
       if (this._isVertical()) {
         var longestWidth = _Util.Methods.max(labelWidths, 0);
         desiredWidth = this._padding + textHeight + this._padding + longestWidth + this._padding;
-        desiredHeight = offeredHeight;
+        desiredHeight = this._isExpanded ? offeredHeight : numSwatches * textHeight;
       } else {
         desiredHeight = textHeight * 2 + this._padding * 3;
-        desiredWidth = offeredWidth;
+        desiredWidth = this._isExpanded ? offeredWidth : numSwatches * textHeight;
       }
 
       return {
@@ -265,8 +275,8 @@ export module Component {
 
       if (this._isVertical()) {
         var longestTextWidth = Math.max(lowerTextWidth, upperTextWidth);
-        swatchWidth = Math.max(textHeight, 0);
-        swatchHeight = this._isExpanded ? this.height() / numSwatches : swatchWidth;
+        swatchWidth = Math.max(this.width() - 3 * padding - longestTextWidth, 0);
+        swatchHeight = this._isExpanded ? this.height() / numSwatches : Math.max(this.height() / numSwatches, 0);
         swatchY = (d: any, i: number) => (numSwatches - (i + 1)) * swatchHeight;
 
         upperWriteOptions.yAlign = "top";
@@ -289,12 +299,12 @@ export module Component {
         boundingBoxAttr.width = swatchWidth;
         boundingBoxAttr.height = numSwatches * swatchHeight;
       } else { // horizontal
-        swatchHeight = Math.max(textHeight, 0);
-        swatchWidth = this._isExpanded ? this.width() / numSwatches : swatchHeight;
+        swatchHeight = Math.max(this.height() - 3 * padding - textHeight, 0);
+        swatchWidth = this.width() / numSwatches;
         var swatchCenter = this.width() / 2;
         swatchX = (d: any, i: number) => this._isExpanded ? i * swatchWidth :
                                          swatchCenter - (numSwatches / 2 * swatchWidth) + i * swatchWidth;
-        swatchY = (d: any, i: number) => this._orientation === "top" ? this.height() - padding - textHeight : padding;
+        swatchY = (d: any, i: number) => this._orientation === "top" ? this.height() - swatchHeight - padding : padding;
 
         upperWriteOptions.xAlign = "right";
         upperLabelShift.x = -(swatchCenter - numSwatches / 2 * swatchWidth);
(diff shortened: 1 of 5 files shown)


rcchen:
@jtlan tests are currently broken but I'm looking for a CR on the main interpolatedColorLegend.ts source to make sure this implementation looks correct/reasonable

src/components/interpolatedColorLegend.ts
```diff
@@ -109,6 +123,30 @@ export module Component {
...
+     *
+     * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+     */
+     public gradient(isGradient: boolean): InterpolatedColorLegend {
+       this._isGradient = isGradient;
+       return this;
+     }
+
+     /**
+      * Expand the InterpolatedColorLegend to the full length/height of the chart.
```
bluong: of the component space* not of the chart*

src/components/interpolatedColorLegend.ts
```diff
@@ -109,6 +123,30 @@ export module Component {
...
+      *
+      * @param {boolean} isExpanded Whether the legend is expanded or not.
+      *
+      * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+      */
+    public expand(isExpanded: boolean): InterpolatedColorLegend {
+      this._isExpanded = isExpanded;
+      return this;
+    }
```
bluong: Can we have the argument be a number instead of a boolean?  It would be the most configurable answer to set the length of the legend

rcchen: Are we expecting users to know how many pixels their chart is? Seems like asking for a number would be erring on the side of over-configuring. Also doesn't address the question of scaling/resizing of components dynamically (shouldn't it be a ratio of the component area in that case?)

bluong: Ratio is also tricky since right now we essentially have a pixel length I believe of numSwatches * textHeight

I'm mainly considering the case where users would want their legend to be longer than what it currently is but not the full length of the container.

bluong: Maybe a dumb suggestion, but maybe...
```typescript
public length(value: number, isProportion: boolean)
```

jtlan: How do you get the default behavior back after changing this?

src/components/interpolatedColorLegend.ts
```diff
@@ -109,6 +123,30 @@ export module Component {
...
     }
 
+    /**
+     * Apply a gradient to the InterpolatedColorLegend.
+     *
+     * @param {boolean} isGradient Whether there should be a gradient or not.
+     *
+     * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+     */
+     public gradient(isGradient: boolean): InterpolatedColorLegend {
```
jtlan: Should be a getter/setter.

src/components/interpolatedColorLegend.ts
```diff
@@ -109,6 +123,30 @@ export module Component {
...
+     }
+
+     /**
+      * Expand the InterpolatedColorLegend to the full length/height of the chart.
+      *
+      * @param {boolean} isExpanded Whether the legend is expanded or not.
+      *
+      * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+      */
+    public expand(isExpanded: boolean): InterpolatedColorLegend {
```
jtlan: Should also be a getter/setter.

src/components/interpolatedColorLegend.ts
```diff
@@ -109,6 +123,30 @@ export module Component {
...
 
+    /**
+     * Apply a gradient to the InterpolatedColorLegend.
+     *
+     * @param {boolean} isGradient Whether there should be a gradient or not.
+     *
+     * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+     */
+     public gradient(isGradient: boolean): InterpolatedColorLegend {
+       this._isGradient = isGradient;
```
jtlan: Wouldn't "`_usesGradient`" be a more semantically-appropriate descriptor?

src/components/interpolatedColorLegend.ts
```diff
@@ -146,12 +200,10 @@ export module Component {
       if (this._isVertical()) {
         var longestWidth = _Util.Methods.max(labelWidths, 0);
         desiredWidth = this._padding + textHeight + this._padding + longestWidth + this._padding;
-        desiredHeight = this._padding + numSwatches * textHeight + this._padding;
+        desiredHeight = offeredHeight;
```
jtlan: It should only ask for all this space if it's set to expand. Your current implementation breaks `xAlign()`/`yAlign()`:
http://jsfiddle.net/9ortv6ht/

src/components/interpolatedColorLegend.ts
```diff
@@ -197,28 +249,29 @@ export module Component {
                 textRotation: 0
             };
 
+      // Width of each individual swatch box
       var swatchWidth: number;
+      // Height of each individual swatch box
```
jtlan: Why these comments? The variable names are pretty self-explanatory.

src/components/interpolatedColorLegend.ts
```diff
@@ -146,12 +200,10 @@ export module Component {
...
         desiredWidth = this._padding + textHeight + this._padding + longestWidth + this._padding;
-        desiredHeight = this._padding + numSwatches * textHeight + this._padding;
+        desiredHeight = offeredHeight;
       } else {
-        desiredHeight = this._padding + textHeight + this._padding;
-        desiredWidth = this._padding + labelWidths[0] + this._padding
-                        + numSwatches * textHeight
-                        + this._padding + labelWidths[1] + this._padding;
+        desiredHeight = textHeight * 2 + this._padding * 3;
+        desiredWidth = offeredWidth;
```
jtlan: Same here, see comment on line 203.

src/components/interpolatedColorLegend.ts
```diff
@@ -197,28 +249,29 @@ export module Component {
...
         height: 0
       };
 
       if (this._isVertical()) {
-        var longestTextWidth = Math.max(text0Width, text1Width);
-        swatchWidth = Math.max( (this.width() - 3 * padding - longestTextWidth), 0);
-        swatchHeight = Math.max( ((this.height() - 2 * padding) / numSwatches), 0);
-        swatchY = (d: any, i: number) => padding + (numSwatches - (i + 1)) * swatchHeight;
+        var longestTextWidth = Math.max(lowerTextWidth, upperTextWidth);
+        swatchWidth = Math.max(textHeight, 0);
```
jtlan: Shouldn't this always be `textHeight`? `textHeight` can't be < 0 right?

src/components/interpolatedColorLegend.ts
```diff
@@ -236,45 +289,76 @@ export module Component {
         boundingBoxAttr.width = swatchWidth;
         boundingBoxAttr.height = numSwatches * swatchHeight;
       } else { // horizontal
-        swatchWidth = Math.max( ((this.width() - 4 * padding - text0Width - text1Width) / numSwatches), 0);
-        swatchHeight = Math.max( (this.height() - 2 * padding), 0);
-        swatchX = (d: any, i: number) => (padding + text0Width + padding) + i * swatchWidth;
-        swatchY = (d: any, i: number) => padding;
+        swatchHeight = Math.max(textHeight, 0);
```
jtlan: Same comment from line 268...

src/components/interpolatedColorLegend.ts
```diff
@@ -197,28 +249,29 @@ export module Component {
...
       };
 
       if (this._isVertical()) {
-        var longestTextWidth = Math.max(text0Width, text1Width);
-        swatchWidth = Math.max( (this.width() - 3 * padding - longestTextWidth), 0);
-        swatchHeight = Math.max( ((this.height() - 2 * padding) / numSwatches), 0);
-        swatchY = (d: any, i: number) => padding + (numSwatches - (i + 1)) * swatchHeight;
+        var longestTextWidth = Math.max(lowerTextWidth, upperTextWidth);
+        swatchWidth = Math.max(textHeight, 0);
+        swatchHeight = this._isExpanded ? this.height() / numSwatches : swatchWidth;
```
jtlan: The previous logic was really important; it reduced the height of the swatches when the space was constrained. Now you get this:
screen shot 2015-04-21 at 10 50 34 am
http://jsfiddle.net/mantxkxx/

jtlan:
Many layout bugs :(

src/components/interpolatedColorLegend.ts
bluong: Using the call that I described up above.  However, this does involve 2 pieces of information that is not well described to the user (numSwatches and textHeight).  Maybe the approach above is indeed correct... I'm just afraid of building too little and then having another customer asking for a more configurable call.

That said, if we think that this fixes most of the cases and we're ok with an API break more down the line, we should go with this approach I think

src/components/interpolatedColorLegend.ts
bluong: :+1:

src/components/interpolatedColorLegend.ts
rcchen: Eventually we will want to expose the `_numSwatches` private variable. Because swatches are fixed height/width for a given component configuration, a user could keep adding/removing swatches until they reach their desired length. Does this sound like a reasonable way to approach the custom sizing issue?

rcchen:

PR #1856, opened 2015-04-17
More unit tests

bluong:


bluong:
Auto +1 due to side branch workflow.

NOQE since it only affects tests

PR #1860, opened 2015-04-18
[WaterfallPlot] Implementing WaterfallPlot

### LINKED ISSUE #1853: Support for waterfall charts
Example:
image
Seems like the basic data model is one where certain values are base bar values, where as others are just delta values between the base points, e.g. for the above:
```
[
{name: "Net Sales 2010", value: 100, type: "base"},
{name: "Price Effect", value: -10, type: "delta"},
{name: "Price Effect", value: 15, type: "delta"},
...
{name: "Net Sales 2011", value: 110, type: "base"}
...
]
```
Or the API could be even more general, with just some function you can pass to determines if each data point is a delta or a normal bar?

rcchen:
image

Fixes #1853

JSFiddle: http://jsfiddle.net/bL1d5gne/1/

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3361,6 +3361,34 @@ declare module Plottable {
 
 
 declare module Plottable {
+    module Plot {
+        class Waterfall extends Rectangle<any, any> {
+            /**
+             * Constructs a WaterfallPlot.
+             *
+             * A WaterfallPlot is used to display the deltas between certain bar values that are marked as base values.
+             *
+             * @constructor
+             * @param {Scale.AbstractScale} xScale The x scale to use.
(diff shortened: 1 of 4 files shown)


jtlan:
How would you mark a net 0 change? It should be a floating bar with zero height right?

rcchen:
There's an open discussion on that in @1853

jtlan:
We probably want to hold off on this until we have a good idea as to the use case, so that we don't cut it off prematurely. I can see us using a non-`project()`-based API for this:
```Typescript
waterfallPlot.datumX(function(d) { return d.eventName; }, xScale);
waterfallPlot.datumY(function(d) { return d.value; }, yScale);
waterfallplot.datumIsSum(function(d) {
  return d.isSum;
});
```

rcchen:

PR #1864, opened 2015-04-20
[DONOTMERGE] - PanInteraction

bluong:
Second version where Panning and Zooming are separated out for higher configurability.

This is mainly in consideration of what would happen if users wanted to use different types of user actions to trigger a PanInteraction.  In which case, we should consider the namespace in which this default behavior lives in.

NOTE: What we could do to get a neater usage is if Plottable.Interaction.Pan pointed to Plottable.Interaction.Pan.Drag such that we can define a default Pan interaction... but I've attempted at this and no luck so far.


bluong:
Going with the much simpler approach for now.

PR #1865, opened 2015-04-20
[triggerFakeWheelEvent util] - Safari compatible

### LINKED ISSUE #1837: [crossbrowser] Mouse Dispatcher test fails on Safari
To reproduce: run tests on safari.

Test failing is MouseDispatcher onWheel()
screen shot 2015-04-14 at 4 05 05 pm

bluong:
Util function checks if the browser is IE.  If it is, it uses the initWheelEvent workflow that only IE uses.

Fixes #1837


jtlan:
Does this fix #1837?

PR #1867, opened 2015-04-21
[Dispatcher.Touch] - Using multiple touches

bluong:
Adding in the idea of identifiers gotten from the events themselves to correlate which touch correlates with which point.


src/dispatchers/touchDispatcher.ts
```diff
@@ -2,7 +2,7 @@
 
 module Plottable {
 export module Dispatcher {
-  export type TouchCallback = (p: Point, e: TouchEvent) => any;
+  export type TouchCallback = (points: Point[], ids: number[], e: TouchEvent) => any;
```
jtlan: Why are ids needed?

src/dispatchers/touchDispatcher.ts
```diff
@@ -118,12 +118,19 @@ export module Dispatcher {
...
+      var touches = e.changedTouches;
+      this._lastTouchPositions = [];
+      var ids: number[] = [];
+      for (var i = 0; i < touches.length; i++) {
+        var touch = touches[i];
+        var touchID = touch.identifier;
+        var newTouchPosition = this.translator.computePosition(touch.clientX, touch.clientY);
+        if (newTouchPosition != null) {
+          ids.push(touchID);
+          this._lastTouchPositions[touchID] = newTouchPosition;
```
jtlan: Is there a possibility of there being non-consecutive touches in the array?

src/interactions/clickInteraction.ts
```diff
@@ -17,8 +17,11 @@ export module Interaction {
       this._mouseDispatcher.onMouseUp("Interaction.Click" + this.getID(), (p: Point) => this._handleClickUp(p));
 
       this._touchDispatcher = Dispatcher.Touch.getDispatcher(<SVGElement> component.content().node());
-      this._touchDispatcher.onTouchStart("Interaction.Click" + this.getID(), (p: Point) => this._handleClickDown(p));
-      this._touchDispatcher.onTouchEnd("Interaction.Click" + this.getID(), (p: Point) => this._handleClickUp(p));
+      //TODO Deal with no point case
```
jtlan: The below code looks like it deals with it already though?

jtlan:
Questions on the signatures.

src/dispatchers/touchDispatcher.ts
bluong: From the TouchEvent perspective, let's say that you put two fingers down on a mobile device and color a point under one finger red and the other blue.  Now let's move these fingers around.  How does the application know that the red circle is correlated with one finger and the blue is with the other?  It is because a Touch object's ID essentially notes the instance of the Touch

src/dispatchers/touchDispatcher.ts
bluong: the identifiers are essentially unique IDs.  There is no semblance of order

src/interactions/clickInteraction.ts
bluong: It deals with it in a pretty atrocious manner in my opinion

src/dispatchers/touchDispatcher.ts
jtlan: So there will be empty spots in the array? That seems bad.

bluong: I can move the array to be more of an object model if you want.  The idea is similar to a hashmap

jtlan:
I think it'd make more sense to have the signature be something like this:
```Typescript
```

src/dispatchers/touchDispatcher.ts
```diff
@@ -2,7 +2,8 @@
 
 module Plottable {
 export module Dispatcher {
-  export type TouchCallback = (p: Point, e: TouchEvent) => any;
+  export type IdToPoint = { [id: number]: Point; };
```
jtlan: Can we just inline the `IDToPoint` type in the few places it's used?

src/dispatchers/touchDispatcher.ts
```diff
@@ -62,7 +59,7 @@ export module Dispatcher {
     }
 
     protected _getWrappedCallback(callback: Function): Core.BroadcasterCallback<Dispatcher.Touch> {
-      return (td: Dispatcher.Touch, p: Point, e: MouseEvent) => callback(p, e);
+      return (td: Dispatcher.Touch, idToPoint: { [id: number]: Point; }, ids: number[], e: MouseEvent) => callback(ids, idToPoint, e);
```
jtlan: arg order is wrong!

src/dispatchers/touchDispatcher.ts
```diff
@@ -118,22 +115,22 @@ export module Dispatcher {
...
+        var touch = touches[i];
+        var touchID = touch.identifier;
+        var newTouchPosition = this.translator.computePosition(touch.clientX, touch.clientY);
+        if (newTouchPosition != null) {
+          touchPositions[touchID] = newTouchPosition;
+          touchIdentifiers.push(touchID);
+        }
+      };
+      if (touchIdentifiers.length > 0) {
+        b.broadcast(touchPositions, touchIdentifiers, e);
```
jtlan: arg order mismatch

jtlan:
[NOQE] because it's a `Dispatcher`.

PR #1874, opened 2015-04-22
[TSLint Rule] - Adding whitespace check-separator rule

bluong:
Rule will check if there is whitespace after the appropriate separator (like commas) for situations like
Messy: [1,2,3,4,4]
Neat: [1, 2, 3, 4, 4]

### DIFF
--- a/test/dispatchers/touchDispatcherTests.ts
+++ b/test/dispatchers/touchDispatcherTests.ts
@@ -135,10 +135,10 @@ describe("Dispatchers", () => {
       var td = Plottable.Dispatcher.Touch.getDispatcher(<SVGElement> target.node());
 
       var callbackWasCalled = false;
-      var callback = function(p: Plottable.Point, e: TouchEvent) {
+      var callback = function (p: Plottable.Point, e: TouchEvent) {
         callbackWasCalled = true;
-        assert.isNotNull(e, "TouchEvent was passed to the Dispatcher");}
-      ;
+        assert.isNotNull(e, "TouchEvent was passed to the Dispatcher");
+      };
 
       var keyString = "notInDomTest";
       td.onTouchMove(keyString, callback);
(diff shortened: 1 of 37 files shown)


test/dispatchers/touchDispatcherTests.ts
```diff
@@ -135,10 +135,10 @@ describe("Dispatchers", () => {
       var td = Plottable.Dispatcher.Touch.getDispatcher(<SVGElement> target.node());
 
       var callbackWasCalled = false;
-      var callback = function(p: Plottable.Point, e: TouchEvent) {
+      var callback = function (p: Plottable.Point, e: TouchEvent) {
```
aicioara: Why?

aicioara:
Whitespace for functions seems wrong (hence the rule above has not been caught). If you search through `touchDispatcherTests.ts` you will see you have a mixture now (say line 78 compared to 138).

Also, following the feedback, I am aware that there are cases where space is missing before `{` and `:`. I've not seen that caught here. Either rename ticket or handle the case.

bluong:
Note that the pull request is checking specifically for separators and nothing else.  So checking after a comma, semicolon, etc.  Nothing else.

test/dispatchers/touchDispatcherTests.ts
bluong: I think I erred here

aicioara:
Sorry then, I thought `:` is a separator as well. Makes sense

aicioara:
Also, what did you use for the change? Did you do it manually?

bluong:
Manually. If there's a tool that can detect the tslint rules and apply them, I would be glad to use it

jtlan:
Couldn't you just turn the rule on, run tslint, and then just fix all flagged lines?

aicioara:
That's what I mean with manual. There's definitely a formatter. (for example Sublime can do it, but I am not sure how to specify this). Eclipse can also definitely do it (I did it in Java). After you specify the rules in the options, right click on the folder and press format code or similar.

Issue #1866, opened 2015-04-21
[Plots] null and undefined should behave the same


aicioara:
When feeding incorrect data to a plot, we might pass for the primary axis (y-axis in general) values like null or undefined, which should be ignored (by case, every plot might deal with this differently). At the moment, `null` generally behaves like a 0 (which is wrong), while `undefined` behaves differently (most of the cases correctly).

This is probably because of coercion shortcuts in javascript, which we use in many places
```javascript
var x = +null           // x will be 0
var y = +undefined // y will be NaN
```

This might be related to:
#1851 #1850 #1849 #1848 #1838

aicioara:
`NaN` and `undefined` now behave the same. However, do we want `null` to behave the same? In JavaScript `null` is generally considered a 0. For example
```javascript
Math.max(null, -2) === 0
```
I believe JavaScript users would expect the same behavior from our API. @crmorford  What do you think?

crmorford:
null --> undefined please :)

aicioara:
as discussed with @jtlan will do, post api refactor.

Issue #1851, opened 2015-04-16
Plot.Grid fails on bad data values


jtlan:
Repro: http://jsfiddle.net/wd1mmd38/

Actual: Errors are thrown in console. DOM contains a `<rect>` with invalid attributes.
Expected: No errors are thrown. `<rect>`s are only added for valid data points.

aicioara:
Taking this since is related to my other 3 tickets of NaN being set on `<rect>` `svg` objects.

aicioara:
Finished

Issue #1850, opened 2015-04-16
Plot.Pie fails on bad data values


jtlan:
Repro: http://jsfiddle.net/vmzfzx88/3/

Actual: `Plot.Pie` stops rendering, spits errors into console
Expected: More graceful fail; all valid slices should be rendered.

aicioara:
Finished

PR #1876, opened 2015-04-22
[Plot] Graphs fail gracefully when presented with badly formatted data

### LINKED ISSUE #1848: [stackedArea] Graphs should fail gracefully when presented with badly formatted data
Easiest to comment on the JS fiddle
http://jsfiddle.net/d2te4yct/6/

- Graphs 1, 2 and 3 should be the same. "0" should be 0 and "s" should default to 0
- Graph 4 is not rendered until #1838 is merged. For discussion sake, in this jsfiddle http://jsfiddle.net/d2te4yct/7/ you can see that the blue area is not rendered at all (on the right hand side). Ideally we should have that area filled with `blue`

aicioara:
Closes #1848 #1850 #1851 

A couple of graphs, when presented with badly formatted data, were just exploding. We want to handle these cases more gracefully now. The ticket is about `Plot.StackedArea`, `Plot.Pie`, `Plot.Grid`

Plot.StackedArea
===
Before: http://jsfiddle.net/d2te4yct/10/
After: http://jsfiddle.net/d2te4yct/11/

Plot.Pie
===
Before: http://jsfiddle.net/vmzfzx88/3/
After: http://jsfiddle.net/vmzfzx88/4/

Plot.Grid
===
(difference here is more subtle, basically now nothing is thrown in the console)
Before: http://jsfiddle.net/wd1mmd38/
After: http://jsfiddle.net/wd1mmd38/1/

Added unit tests for each

Also, there is a discussion about `null` and `undefined` which should behave the same. That is a discussion for #1866

### DIFF
--- a/src/components/plots/abstractStackedPlot.ts
+++ b/src/components/plots/abstractStackedPlot.ts
@@ -122,7 +122,8 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          if (value === 0) {
+          // We are accepting strings as possible candidates for numbers, hence double equals
+          if (value == 0) {
             offset = isAllNegativeValues ? negativeOffset : positiveOffset;
           } else {
             offset = value > 0 ? positiveOffset : negativeOffset;
--- a/src/components/plots/abstractStackedPlot.ts
+++ b/src/components/plots/abstractStackedPlot.ts
@@ -122,7 +122,7 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          if (!value) {
+          if (!+value) {
             offset = isAllNegativeValues ? negativeOffset : positiveOffset;
           } else {
             offset = value > 0 ? positiveOffset : negativeOffset;
--- a/src/components/plots/abstractStackedPlot.ts
+++ b/src/components/plots/abstractStackedPlot.ts
@@ -122,8 +122,7 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          // We are accepting strings as possible candidates for numbers, hence double equals
-          if (value == 0) {
+          if (value === 0 || value === "0") {
             offset = isAllNegativeValues ? negativeOffset : positiveOffset;
           } else {
             offset = value > 0 ? positiveOffset : negativeOffset;
--- a/src/components/plots/abstractStackedPlot.ts
+++ b/src/components/plots/abstractStackedPlot.ts
@@ -122,7 +122,7 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          if (value === 0 || value === "0") {
+          if (!value) {
             offset = isAllNegativeValues ? negativeOffset : positiveOffset;
           } else {
             offset = value > 0 ? positiveOffset : negativeOffset;
--- a/src/components/plots/rectanglePlot.ts
+++ b/src/components/plots/rectanglePlot.ts
@@ -36,12 +36,12 @@ export module Plot {
       var y2Attr = attrToProjector["y2"];
 
       // Generate width based on difference, then adjust for the correct x origin
-      attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
-      attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m));
+      attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m)) || 0;
+      attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m)) || 0;
 
       // Generate height based on difference, then adjust for the correct y origin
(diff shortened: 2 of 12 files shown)


src/components/plots/abstractStackedPlot.ts
```diff
@@ -122,7 +122,7 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          if (value === 0) {
+          if (value === 0 || value === "0") {
```
jtlan: Should we be accepting stringy values?

src/drawers/arcDrawer.ts
```diff
@@ -39,6 +39,9 @@ export module _Drawer {
     public draw(data: any[], drawSteps: DrawStep[], userMetadata: any, plotMetadata: Plot.PlotMetadata) {
       // HACKHACK Applying metadata should be done in base class
       var valueAccessor = (d: any, i: number) => drawSteps[0].attrToProjector["value"](d, i, userMetadata, plotMetadata);
+
+      data = data.filter(e => !!+valueAccessor(e, null));
```
jtlan: What does `!!` do?

src/components/plots/abstractStackedPlot.ts
aicioara: Well, if we don't this is not the place to begin doing that. Shall I file a new issue about this? Keep in mind that `d3` scales accept strings as arguments, np

src/drawers/arcDrawer.ts
aicioara: 1. It filters numbers other than 0, null, undefined, NaN, unCoercible strings. 
2. If the user wants to draw a slice with size 0, that's not drawn, correct? So filtering is not semantically correct, but achieves the same result

crmorford:
ON stacked area, undefined and NaN make the whole data series not show up:
screen shot 2015-04-23 at 10 25 36 am
 but null causes a gross failure:
screen shot 2015-04-23 at 10 25 27 am
Can someone explain the reasoning why these are different?

aicioara:
@crmorford Agree, the second graph is a bug, will fix that

aicioara:
Fixed an issue and did not address any of the concerns above (however, check my comments)

src/components/plots/abstractStackedPlot.ts
```diff
@@ -122,7 +122,7 @@ export module Plot {
 
           var value = valueAccessor(datum, datumIndex, dataset.metadata(), plotMetadata);
           var offset: number;
-          if (value === 0) {
+          if (!+value) {
             offset = isAllNegativeValues ? negativeOffset : positiveOffset;
           } else {
             offset = value > 0 ? positiveOffset : negativeOffset;
```
jtlan: Shouldn't we use the converted `value` here?

src/drawers/arcDrawer.ts
jtlan: The slice is drawn, but it has size 0. A tooltip would still stick to it for example, but it wouldn't if the data wasn't there at all.

src/components/plots/rectanglePlot.ts
```diff
@@ -36,12 +36,13 @@ export module Plot {
       var y2Attr = attrToProjector["y2"];
 
       // Generate width based on difference, then adjust for the correct x origin
-      attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m));
-      attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m));
+      attrToProjector["width"] = (d, i, u, m) => Math.abs(x2Attr(d, i, u, m) - x1Attr(d, i, u, m)) || 0;
+      attrToProjector["x"] = (d, i, u, m) => Math.min(x1Attr(d, i, u, m), x2Attr(d, i, u, m)) || 0;
```
jtlan: This has the exact opposite problem of the one in `Pie`: Illegal rectangles will be drawn as size 0, instead of being missing.

src/components/plots/abstractStackedPlot.ts
aicioara: you mean 
```typescript 
+offset = value > 0 ? positiveOffset : negativeOffset;
```
I agree

jtlan:
We have some mixups here regarding the effects of 0 and invalid data. Here's what should happen:
* Data that produces objects with `0` size (the angle on a `Pie` slice or the width/height of a `Bar` or `Rectangle`) should still be drawn on screen.
* Data with invalid values (`NaN`, `undefined`) shouldn't be drawn at all: There should be no DOM nodes associated with them.

aicioara:
sounds fair. Looking into it

src/components/plots/rectanglePlot.ts
jtlan: Maybe we do something like this?
```Typescript
var isValid = (d, i, u, m) => {
  return isValidNumber(x1Attr(d, i, u, m)) &&
         isValidNumber(x2Attr(d, i, u, m)) &&
         isValidNumber(y1Attr(d, i, u, m)) &&
         isValidNumber(x2Attr(d, i, u, m));
}
data.filter(isValid);
```

aicioara:
Done all suggested changes

src/utils/utils.ts
```diff
@@ -255,6 +255,14 @@ export module _Util {
     }
 
     /**
+     * Returns true if the argument is a number, which is not NaN
+     * Numbers represented as strings do not pass this function
+     */
+    export function isValidNumber(n: any) {
```
jtlan: I was going to suggest naming this `isFiniteNumber`, but we might want to support infinite values alter, so I think `isValidNumber` is a good name.

PR #1877, opened 2015-04-23
Group get closest plot data

istarnx:
Implementing getClosestPlotData() for Component.Group.


istarnx:
Fiddle: http://jsfiddle.net/z1110v2j/3/

bluong:
Weird all my comments got bashed?

src/components/componentGroup.ts
```diff
@@ -23,6 +23,54 @@ export module Component {
...
+                pixelPoints: [pixelPoint],
+                plot: cpd.plot,
+                selection: d3.select(selection[selectionIndex])
+              };
+            }
+          });
+        }
+      });
+
+      return closestPlotData;
```
bluong: Based on this implementation, it looks like closestPlotData will not be necessarily be obeying all the rules set upon it.  Before it would retrieve all the closestPlotData regardless of how many are closest, whereas this closestPlotData returns a single item.  We should be sure to make these behaviors be near exactly the same if they are to share the same name

istarnx:
Sorry Brandon, your comments were on the PR from this branch into the old feature branch.

PR #1879, opened 2015-04-23
Adding the vertical() accessor to Plot.Bar.

istarnx:


src/components/plots/barPlot.ts
```diff
@@ -74,6 +74,15 @@ export module Plot {
     }
 
     /**
+     * Returns true if the plot is vertical.
```
bluong: We don't need the doc here

test/components/plots/barPlotTests.ts
```diff
@@ -634,6 +639,11 @@ describe("Plots", () => {
         });
 
       });
+
+      it("vertical() returns false", () => {
+        assert.isFalse(barPlot.vertical());
```
bluong: Let's put a small message here like Horizontal Bar plots are always not vertical

test/components/plots/barPlotTests.ts
```diff
@@ -270,6 +270,11 @@ describe("Plots", () => {
           svg.remove();
         });
       });
+
+      it("vertical() returns true", () => {
+        assert.isTrue(barPlot.vertical());
```
bluong: Let's put a small message here like Vertical Bar plots are always vertical

src/components/plots/barPlot.ts
```diff
@@ -74,6 +74,15 @@ export module Plot {
...
+     *
+     * @returns {boolean} true if the plot is vertical
```
bluong: This returns should all that needs to be required.

istarnx:
Sure--I can make the changes if you'd like. I based that doc style off of Java standards, which typically call for a description AND return value.

istarnx:
Also, note that other boolean methods in Bar have both a description and return value comment.

bluong:
Based on our recent API discussion, if we have doc that's verbose/redundant it'd be best to avoid it.  Less is more.  We can discuss more if you want

PR #1880, opened 2015-04-23
[TSLint rule] - adding comment-format check-space rule

bluong:
The rule makes sure that there is a single space between the "/" and the start of a comment in a single-line comment.


bluong:
This is aimed at develop... Closing

PR #1881, opened 2015-04-23
[TSLint Rule] - adding comment-format check-space rule

bluong:
This checks if there is a single space in a single line comment between the / and the comment itself

Neat:
```typescript
// Insert comment here
```

Messy:
```typescript
//Insert Comment here
```

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2171,6 +2171,7 @@ declare module Plottable {
     }
 }
 
+
 declare module Plottable {
     module Axis {
         class Numeric extends AbstractAxis {
--- a/src/components/axes/numericAxis.ts
+++ b/src/components/axes/numericAxis.ts
@@ -1,4 +1,4 @@
-//<reference path="../../reference.ts" />
+///<reference path="../../reference.ts" />
 
 module Plottable {
 export module Axis {
--- a/plottable.js
+++ b/plottable.js
@@ -2399,7 +2399,7 @@ var Plottable;
                 return new Category(this._d3Scale.copy());
             };
             Category.prototype.scale = function (value) {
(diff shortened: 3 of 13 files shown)


plottable.d.ts
```diff
@@ -2171,6 +2171,7 @@ declare module Plottable {
     }
 }
 
+
```
aicioara: This is a bit weird. Could you double check that the compilation actually yielded that?

src/components/axes/numericAxis.ts
```diff
@@ -1,4 +1,4 @@
-//<reference path="../../reference.ts" />
+///<reference path="../../reference.ts" />
```
aicioara: This looked pretty serious. Is this caught by the linter as well? If not, it should

aicioara:
I've seen you opened quite a few tickets, so just not to forget, in case this is not covered yet:
- Comments should be enforced to always begin with either Capital letter or lower letter (we have mixture)
- I suggest we should not have inline comments like this (but up to you)
```typescript
doSomethin(); // comment on same line
```

plottable.d.ts
bluong: This probably happened due to the numericAxis.ts change

src/components/axes/numericAxis.ts
bluong: I am unsure how to tell the linter to look for the above.  The above seems very specific to Plottable

jtlan: Not really, this seems to be the way Typescript recommends doing this:

http://www.typescriptlang.org/Handbook#modules-splitting-across-files

PR #1885, opened 2015-04-23
[TSLint Rule] - adding no-debugger rule

bluong:
Disallowing the usage of debuggers in the source code.


aicioara:
@bluong what does this do? `console.log()` is already forbidden

jtlan:
@aicioara It prohibits the `debugger` statement.

PR #1886, opened 2015-04-23
[TSLint Rule] - adding no-switch-case-fall-through rule

bluong:
Disallowing having cases in a switch statement to fall into each other

Neat:
```typescript
switch(foo) {
    case 1:
        doStuff(foo);
        break;
     case 2:
         doStuff2(foo);
}
```

Messy:
```typescript
switch(foo) {
    case 1:
        doStuff(foo);
    case 2:
         doStuff2(foo);
}
```

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -24,6 +24,7 @@
     "no-duplicate-variable": true,
     "no-empty": true,
     "no-eval": true,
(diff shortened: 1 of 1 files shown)


bluong:
develop --.-;;;

PR #1888, opened 2015-04-23
[TSLint Rule] - adding no-unused-variable rule

bluong:
Disallowing the declaration of variables that are not used.

Neat:
```typescript
var a = "foo";
return doStuff(a);
```

Messy:
```typescript
var a = "foo";
var b = "bar";
return doStuff(a);
```

NOTE: There is an interesting spot in testUtils where tslint thinks that those functions are unused but indeed they are...?


jtlan:
If we `export` the functions, does that solve the problem?

jtlan:
What if we put the utility methods in a `module`?

PR #1891, opened 2015-04-23
[TSLint Rule] - adding no-var-requires rule

bluong:
Disallowing the usage of using require to be stored into a variable instead of importing.

Messy:
```typescript
var module = require("moduleName");
```


jtlan:
Does this affect us at all?

aicioara:
If we used something like Browserify we would, but since we don't, I believe this is useless (and slows the `tslint` as well)

bluong:
All of this is valid.  Closing

PR #1895, opened 2015-04-24
[DO NOT MERGE] Add CallbackSet.

jtlan:
A set of callbacks. You can call them all with `callCallbacks()`.

Question 1: Should we be using [ES6 Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead? I need to extend it to add `callCallbacks()`, so I don't think that would work. We'd also need to get a shim.

Question 2: Do we need the generic `Set` at all? I expect we'll want it if we convert `Dataset`s to be added by direct-reference, but maybe we should just write out the `push()` and `splice()` lines when it comes to that.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -345,6 +345,27 @@ declare module Plottable {
 
 
 declare module Plottable {
+    module Utils {
+        class Set<T> {
+            constructor();
+            add(value: T): Set<T>;
+            remove(value: T): Set<T>;
+            values(): T[];
+        }
+    }
+}
+
+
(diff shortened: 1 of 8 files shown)


bluong:
I think we're moving toward an approach of don't expose too much unless necessary.  In accordance, if we are only using Set for CallbackSet, just make the CallbackSet and if we want to use Set in the future, split that logic out in that time

bluong:
How does that sound?

jtlan:
Makes sense to me.

Issue #696, opened 2014-07-11
Interactions should have disable() and enable() calls


jtlan:
There should be some mechanism by which Interactions can be temporarily disabled.
We might also consider having some mechanism to disconnect Interactions.

jtlan:
The fact that `Interaction`s have different architectures internally is screwing this up. As it stands, each `Interaction` would need a separate implementation of this feature. It might still be worth doing that for now, and then redoing it later.

crmorford:
Hey, I noticed the milestone for this keeps getting pushed back. I'm going to assign it for real next week. How long do you expect this to take?

jtlan:
@crmorford This is dependent on the `Dispatcher` refactors I'm currently working on.

By the way, what's a good API here?

```Typescript
// separate calls
public enable(): AbstractInteraction;
public disable(): AbstractInteraction;
```

or

```Typescript
// getter-setter
public enabled(): boolean;
public enabled(doEnable: boolean): AbstractInteraction;
```

crmorford:
the getter-setter matches what I'm used to in Plottable. I prefer that.

PR #1896, opened 2015-04-24
Pan zoom interaction

### LINKED ISSUE #1434: Registering Interaction.PanZoom before rendering breaks
Repro: http://jsfiddle.net/x7va002L/
Attempt to pan the chart.

Expected: `Interaction.Panzoom` works.
Actual: The chart zooms away as soon as the user tries to pan

The workaround is to register the `Interaction.PanZoom` after rendering, or to call `resetZoom()` after rendering, but the user shouldn't have to do this.

### LINKED ISSUE #1502: Implement Interaction.Pan

### LINKED ISSUE #1581: Implement Interaction.Zoom
The flip side of the coin of #1502. It might be that the best way to do this is to have some sort of `zoomTo` method on `Scale`s, and `Interaction.Zoom` simply triggers that after a mousewheel event.

### LINKED ISSUE #1727: PanZoom + Modified Log gets 'stuck'
A picture's worth 1000 words... (zooming in)
 screen shot 2015-03-10 at 10 06 23 am
screen shot 2015-03-10 at 10 07 28 am'screen shot 2015-03-10 at 10 07 34 am
screen shot 2015-03-10 at 10 07 02 am

bluong:
`Interaction.PanZoom` has been completely reworked and rearchitected to follow the new architecture we have for `Interaction`s.

Example JSFiddle: http://jsfiddle.net/bluong63/s2pzbx4p/1/

Usage of `Interaction.PanZoom` should still be the same:
```typescript
var xScale;
var yScale;
var plot;

// xScale and yScale are optional if a user does not want to have panning/zooming on those scales
var panZoomInteraction = new Plottable.Interaction.PanZoom(xScale, yScale);

plot.registerInteraction(panZoomInteraction);
```

Note that this reworking and rearchitecting means that `Interaction.PanZoom` is a lot more integrated with our codebase so its interaction with scales should be a lot less buggy.

Upgrades in behavior:
* `Interaction.PanZoom` will now work with `Scale.ModifiedLog`s neatly.
* `Interaction.PanZoom` does not have to be registered after the plot is rendered.

Closes #1727 
Closes #1581 
Closes #1502 
Closes #1434


src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,115 @@ export module Interaction {
...
+        (ids, idToPoint, e) => this._handleTouchMove(ids, idToPoint, e));
+      this._touchDispatcher.onTouchEnd("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handleTouchEnd(ids, idToPoint, e));
     }
 
-    public _requiresHitbox() {
-      return true;
+    private _handleTouchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      ids.forEach((id) => {
+        if (this._touchIds.size() === 2) {
```
jtlan: Shouldn't this check be outside of the `forEach`? And we should probably do `>= 2`, in the case of more than 2 touches.

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,115 @@ export module Interaction {
...
     }
 
-    private _rerenderZoomed() {
-      // HACKHACK since the d3.zoom.x modifies d3 scales and not our TS scales, and the TS scales have the
-      // event listener machinery, let's grab the domain out of the d3 scale and pipe it back into the TS scale
-      if (this._xScale) {
-        var xDomain = (<any> this._xScale)._d3Scale.domain();
-        this._xScale.domain(xDomain);
+    private _handleTouchMove(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      if (this._touchIds.size() !== 2) {
```
jtlan: `< 2`, since if there are at least two it will work.

src/interactions/panZoomInteraction.ts
bluong: This is so if a user has a finger on the screen and then adds another one at a later time, both are picked up.

src/interactions/panZoomInteraction.ts
bluong: There currently is the invariant that size should be between 0 and 2

jtlan:
I pulled this just now, and it seems to work?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,143 @@ export module Interaction {
...
+        (ids, idToPoint, e) => this._handleTouchStart(ids, idToPoint, e));
+      this._touchDispatcher.onTouchMove("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handleTouchMove(ids, idToPoint, e));
+      this._touchDispatcher.onTouchEnd("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handleTouchEnd(ids, idToPoint, e));
+    }
+
+    private _handleTouchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      ids.forEach((id) => {
+        if (this._touchIds.size() === 2) {
```
jtlan: What's going on here? This should be outside the loop:
```Typescript
if (this._touchIds.size() === 2) {
  this._dragInteraction.onDrag(null);
} else {
  ids.forEach((id) => this._touchIds.set(id.toString(), this._translateToComponentSpace(idToPoint[id])));
}
```

Also, the condition should probably be something like `this._touchIds.size() > 1`, right?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,143 @@ export module Interaction {
...
+        this._touchIds.remove(id.toString());
+      });
+      if (this._touchIds.size() < 2) {
+        this._setupDragInteraction();
+      }
     }
 
-    public _requiresHitbox() {
-      return true;
+    private static magnify<D>(scale: Scale.AbstractQuantitative<D>, magnifyAmount: number, centerValue: number) {
```
jtlan: Probably `magnifyScale` is a better name, since the method isn't on the `Scale`.

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,143 @@ export module Interaction {
...
     }
 
-    private _rerenderZoomed() {
-      // HACKHACK since the d3.zoom.x modifies d3 scales and not our TS scales, and the TS scales have the
-      // event listener machinery, let's grab the domain out of the d3 scale and pipe it back into the TS scale
-      if (this._xScale) {
-        var xDomain = (<any> this._xScale)._d3Scale.domain();
-        this._xScale.domain(xDomain);
-      }
+    private static translate<D>(scale: Scale.AbstractQuantitative<D>, translateAmount: number) {
```
jtlan: Probably `translateScale` is a better name, since the method isn't on the `Scale`.

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,143 @@ export module Interaction {
...
+    private _handleTouchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      ids.forEach((id) => {
+        if (this._touchIds.size() === 2) {
+          return;
+        }
+
+        this._touchIds.set(id.toString(), this._translateToComponentSpace(idToPoint[id]));
+      });
+      if (this._touchIds.size() === 2) {
+        this._dragInteraction.onDrag(null);
```
jtlan: Instead of blanking the `onDrag` callback, would it make more sense to have some sort of conditional-check inside the callback that skips actions if there is more than one touch?

Might also be good to label this with a `HACKHACK` on #696.

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,143 @@ export module Interaction {
...
 
-    /**
-     * Sets the scales back to their original domains.
-     */
-    public resetZoom() {
-      // HACKHACK #254
-      this._zoom = d3.behavior.zoom();
-      if (this._xScale) {
-        this._zoom.x((<any> this._xScale)._d3Scale);
+    private _handleTouchMove(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
```
jtlan: Maybe rename this to indicate that it's intended to handle two- (or more-) finger touches?

jtlan:
From what I can see by pulling the branch, the functionality seems correct, but there's some weirdness in the way the drag interaction is enabled/disabled in the code. Also, the tests should be re-enabled, of course.

src/interactions/panZoomInteraction.ts
bluong: Please put the comment after the relevant code.  The part here obscures the relevant part (which is mainly a fault of github).

That aside, The check is placed afterwards since there is the invariant that `this._touchIds.size() <= 2`.  Thus, this if statement is trying to hold that invariant

The disabling happens only if after the registration there are more than 1 registered touch

src/interactions/panZoomInteraction.ts
bluong: Should I have the function be only destructive?  Or should I make it a pure function?

If pure, then I was thinking of magnifyScaleTransform?

src/interactions/panZoomInteraction.ts
bluong: Should I have the function be only destructive?  Or should I make it a pure function?

If pure, then I was thinking of translateScaleTransform?

src/interactions/panZoomInteraction.ts
bluong: This is a good point.  It might lead to a much more readable architecture.  I'll try it out.

src/interactions/panZoomInteraction.ts
bluong: Rename how so?

src/interactions/panZoomInteraction.ts
bluong: Any extra Touches beyond the second one are completely ignored, unless you want different functionality.

src/interactions/panZoomInteraction.ts
jtlan: I just thought it was weird that single-finger touches get grabbed by `Drag` but there's no way to indicate that multitouch is handled here only. `_handlePinch`?

src/interactions/panZoomInteraction.ts
jtlan: In this case though, the method returns, rather than processing the first two touches, right?

bluong: There should never be a situation where there are more than 2 touches, but I do agree that I think `< 2`is better.

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,140 @@ export module Interaction {
...
+      });
     }
 
-    private _rerenderZoomed() {
-      // HACKHACK since the d3.zoom.x modifies d3 scales and not our TS scales, and the TS scales have the
-      // event listener machinery, let's grab the domain out of the d3 scale and pipe it back into the TS scale
-      if (this._xScale) {
-        var xDomain = (<any> this._xScale)._d3Scale.domain();
-        this._xScale.domain(xDomain);
+    private _handlePinchMove(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
```
jtlan: `_handlePinch`?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,140 @@ export module Interaction {
...
+        this._xScale.domain(PanZoom.magnify(this._xScale, oldDiff / newDiff, oldAvgX));
+        this._xScale.domain(PanZoom.translate(this._xScale, oldAvgX - newAvgX));
+      }
+      if (this._yScale != null && newDiff !== 0 && oldDiff !== 0) {
+        this._yScale.domain(PanZoom.magnify(this._yScale, oldDiff / newDiff, oldAvgY));
+        this._yScale.domain(PanZoom.translate(this._yScale, oldAvgY - newAvgY));
       }
     }
+
+    private _handlePinchFinish(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
```
jtlan: Probably `_handlePinchEnd` for consistency?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,140 @@ export module Interaction {
...
+      var secondTouchPoint = points[1];
+
+      var leftX = Math.min(firstTouchPoint.x, secondTouchPoint.x);
+      var rightX = Math.max(firstTouchPoint.x, secondTouchPoint.x);
+      var topY = Math.min(firstTouchPoint.y, secondTouchPoint.y);
+      var bottomY = Math.max(firstTouchPoint.y, secondTouchPoint.y);
+
+      var oldAvgX = (leftX + rightX) / 2;
+      var oldAvgY = (bottomY + topY) / 2;
+      var oldDiff = Math.sqrt(Math.pow(rightX - leftX, 2) + Math.pow(bottomY - topY, 2));
```
jtlan: Is there a way to indicate that the preceding logic applies to the "old" points? Maybe factor the average and diff calculations out into a helper method?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,140 @@ export module Interaction {
...
+          this._yScale.domain(PanZoom.magnify(this._yScale, zoomAmount, translatedP.y));
+        }
+      }
+    }
+
+    private _setupDragInteraction() {
+      var lastDragPoint: Point;
+      this._dragInteraction.onDragStart(() => lastDragPoint = null);
+      this._dragInteraction.onDrag((startPoint, endPoint) => {
+        if (this._touchIds.size() === 2) {
```
jtlan: `>= 2`?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,140 @@ export module Interaction {
...
 
-    public _requiresHitbox() {
-      return true;
+    private _handlePinchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      ids.forEach((id) => {
+        if (this._touchIds.size() === 2) {
+          return;
+        }
+
+        this._touchIds.set(id.toString(), this._translateToComponentSpace(idToPoint[id]));
```
jtlan: I see, so `this._touchIds` never has more than 2 values in it. That's kind of confusing...

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,150 @@ export module Interaction {
...
+        (ids, idToPoint, e) => this._handlePinchFinish(ids, idToPoint, e));
+      this._touchDispatcher.onTouchCancel("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handlePinchFinish(ids, idToPoint, e));
     }
 
-    public _requiresHitbox() {
-      return true;
+    private _handlePinchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
+      ids.forEach((id) => {
+        if (this._touchIds.size() === 2) {
```
jtlan: I get that `===` suffices, but I think the logic would be easier to grok if it was `>=`.

Also since it doesn't actually process the pinch gesture, `_handleTouchStart` is probably the correct name. Sorry to flip-flop on this one...

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,150 @@ export module Interaction {
...
+      var topY = Math.min(firstTouchPoint.y, secondTouchPoint.y);
+      var bottomY = Math.max(firstTouchPoint.y, secondTouchPoint.y);
+
+      return Math.sqrt(Math.pow(rightX - leftX, 2) + Math.pow(bottomY - topY, 2));
+    }
 
-      if (this._yScale) {
-        var yDomain = (<any> this._yScale)._d3Scale.domain();
-        this._yScale.domain(yDomain);
+    private _handlePinchFinish(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
```
jtlan: Same here, `handleTouchEnd` is probably more descriptive.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      var startPoint = { x: SVG_WIDTH / 4, y: SVG_HEIGHT / 4 };
+      var startPoint2 = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 2 };
+      triggerFakeTouchEvent( "touchstart", eventTarget, [startPoint, startPoint2], [0, 1] );
+
+      var endPoint = { x: SVG_WIDTH * 3 / 4, y: SVG_HEIGHT * 3 / 4 };
+      triggerFakeTouchEvent( "touchmove", eventTarget, [endPoint], [1] );
+      triggerFakeTouchEvent( "touchend", eventTarget, [endPoint], [1] );
+      assert.deepEqual( xScale.domain(), [SVG_WIDTH / 16, SVG_WIDTH * 5 / 16], "xScale pans to the correct domain via drag (touch)" );
+      assert.deepEqual( yScale.domain(), [SVG_HEIGHT / 16, SVG_HEIGHT * 5 / 16], "yScale pans to the correct domain via drag (touch)" );
```
jtlan: Wrong assert fail messages here?

src/interactions/panZoomInteraction.ts
```diff
@@ -20,56 +29,150 @@ export module Interaction {
...
+        (ids, idToPoint, e) => this._handlePinch(ids, idToPoint, e));
+      this._touchDispatcher.onTouchEnd("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handlePinchFinish(ids, idToPoint, e));
+      this._touchDispatcher.onTouchCancel("Interaction.PanZoom" + this.getID(),
+        (ids, idToPoint, e) => this._handlePinchFinish(ids, idToPoint, e));
     }
 
-    public _requiresHitbox() {
-      return true;
+    private _handlePinchStart(ids: number[], idToPoint: { [id: number]: Point; }, e: TouchEvent) {
```
jtlan: Also since it doesn't actually process a pinch gesture, `_handleTouchStart` is probably the correct name for this method. Sorry to flip-flop on this one...

src/interactions/panZoomInteraction.ts
jtlan: Oh, I see what's going on here, since there's a line in this closure that also modifies `this._touchIDs`. The logic might be clearer if we used a `for` loop:
```Typescript
for (var i = 0; i < 2; i++) {
  this._touchIds.set(id[i].toString(), this. _translateToComponentSpace(idToPoint[id[i]]);
}
```

jtlan:
One last place where changing the loop logic makes it easier to understand.

crmorford:
nice work :)

test/interactions/panZoomInteractionTests.ts
```diff
@@ -0,0 +1,113 @@
...
+        svg.remove();
+      });
+
+      it("dragging to outside the component will translate the scale correctly (touch)", () => {
+        var startPoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        var endPoint = { x: -SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        triggerFakeTouchEvent("touchstart", eventTarget, [startPoint]);
+        triggerFakeTouchEvent("touchmove", eventTarget, [endPoint]);
+        triggerFakeTouchEvent("touchend", eventTarget, [endPoint]);
+        assert.deepEqual(xScale.domain(), [SVG_WIDTH / 2, SVG_WIDTH], "xScale pans to the correct domain via drag (mouse)");
```
jtlan: Wrong assert message.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -0,0 +1,113 @@
...
+        var endPoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT * 3 / 4 };
+        triggerFakeTouchEvent("touchstart", eventTarget, [startPoint]);
+        triggerFakeTouchEvent("touchmove", eventTarget, [endPoint]);
+        triggerFakeTouchEvent("touchend", eventTarget, [endPoint]);
+        assert.deepEqual(xScale.domain(), [-SVG_WIDTH / 8, SVG_WIDTH * 3 / 8], "xScale pans to the correct domain via drag (touch)");
+        assert.deepEqual(yScale.domain(), [-SVG_HEIGHT / 4, SVG_HEIGHT / 4], "yScale pans to the correct domain via drag (touch)");
+        svg.remove();
+      });
+
+      it("dragging to outside the component will translate the scale correctly (touch)", () => {
```
jtlan: No equivalent test for mouse?

jtlan:
[-1], tests only.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -0,0 +1,123 @@
...
+        svg.remove();
+      });
+
+      it("dragging to outside the component will translate the scale correctly (mouse)", () => {
+        var startPoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        var endPoint = { x: -SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        triggerFakeMouseEvent("mousedown", eventTarget, startPoint.x, startPoint.y);
+        triggerFakeMouseEvent("mousemove", eventTarget, endPoint.x, endPoint.y);
+        triggerFakeMouseEvent("mouseend", eventTarget, endPoint.x, endPoint.y);
+        assert.deepEqual(xScale.domain(), [SVG_WIDTH / 2, SVG_WIDTH], "xScale pans to the correct domain via drag (mouse)");
```
jtlan: Probably should test the `yScale` too.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -0,0 +1,123 @@
...
+        svg.remove();
+      });
+
+      it("dragging to outside the component will translate the scale correctly (touch)", () => {
+        var startPoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        var endPoint = { x: -SVG_WIDTH / 2, y: SVG_HEIGHT / 4 };
+        triggerFakeTouchEvent("touchstart", eventTarget, [startPoint]);
+        triggerFakeTouchEvent("touchmove", eventTarget, [endPoint]);
+        triggerFakeTouchEvent("touchend", eventTarget, [endPoint]);
+        assert.deepEqual(xScale.domain(), [SVG_WIDTH / 2, SVG_WIDTH], "xScale pans to the correct domain via drag (touch)");
```
jtlan: Maybe do a similar check for `yScale`?

PR #1900, opened 2015-04-24
Remove Interactive module

rcchen:


jtlan:
NOQE rename, I think. Make sure tests pass.

PR #1906, opened 2015-04-24
Fix for timeaxis bug

aicioara:
Now merged to develop.

The hotfix was only applied to the release atm

@jtlan is this how you do it?


aicioara:
@crmorford not sure what's up with the build (it's probably some branching fun), probably should not have merged

PR #1922, opened 2015-04-26
[DO NOT MERGE] simplifying _computeLayout()  API

jtlan:
I did this on **develop** instead of the **apiRefactor** branch because I wanted us to be able to regression-test this easily. Once y'all sign off on it I'll merge this to **apiRefactor** instead.

Originally I was going to do a more involved layout-engine refactor, but that's out of scope. Instead, I focused on paring down the `_SpaceRequest` interface to something more reasonable. Later on, it may be helpful to have it have this signature:
```Typescript
type SpaceRequest {
  minWidth: number;
  maxWidth: number;
  minHeight: number;
  maxHeight: number;
}
```

Not sure if "`minWidth`" and "`maxWidth`" are the best names here, since the intention is closer to something like "minimum acceptable width" and "ideal width". For example: a `Plot` would return this:
```Typescript
{
  minWidth: 0,
  maxWidth: Infinity,
  minHeight: 0,
  maxHeight: Infinity
}
```
since it can draw in tiny spaces but is willing to absorb any free space.

### DIFF
--- a/src/components/axes/categoryAxis.ts
+++ b/src/components/axes/categoryAxis.ts
@@ -41,7 +41,7 @@ export module Axis {
       var heightRequiredByTicks = this._isHorizontal() ? this._maxLabelTickLength() + this.tickLabelPadding() + this.gutter() : 0;
 
       if (this._scale.domain().length === 0) {
-        return {width: 0, height: 0, wantsWidth: false, wantsHeight: false };
+        return { width: 0, height: 0 };
       }
 
       var categoryScale: Scale.Category = <Scale.Category> this._scale;
@@ -55,11 +55,12 @@ export module Axis {
                                           offeredHeight,
                                           fakeScale,
                                           categoryScale.domain());
+
+      var desiredWidth = textResult.usedWidth + widthRequiredByTicks;
+      var desiredHeight = textResult.usedHeight + heightRequiredByTicks;
       return {
-        width : textResult.usedWidth  + widthRequiredByTicks,
-        height: textResult.usedHeight + heightRequiredByTicks,
-        wantsWidth : !textResult.textFits,
-        wantsHeight: !textResult.textFits
+        width: desiredWidth,
+        height: desiredHeight
       };
     }
 
--- a/src/components/table.ts
+++ b/src/components/table.ts
@@ -228,39 +228,42 @@ export module Component {
     private _determineGuarantees(offeredWidths: number[], offeredHeights: number[]): _LayoutAllocation {
       var requestedWidths  = _Util.Methods.createFilledArray(0, this._nCols);
       var requestedHeights = _Util.Methods.createFilledArray(0, this._nRows);
-      var layoutWantsWidth  = _Util.Methods.createFilledArray(false, this._nCols);
-      var layoutWantsHeight = _Util.Methods.createFilledArray(false, this._nRows);
+      var columnNeedsWidth  = _Util.Methods.createFilledArray(false, this._nCols);
+      var rowNeedsHeight = _Util.Methods.createFilledArray(false, this._nRows);
       this._rows.forEach((row: AbstractComponent[], rowIndex: number) => {
         row.forEach((component: AbstractComponent, colIndex: number) => {
           var spaceRequest: _SpaceRequest;
           if (component != null) {
             spaceRequest = component._requestedSpace(offeredWidths[colIndex], offeredHeights[rowIndex]);
           } else {
-            spaceRequest = {width: 0, height: 0, wantsWidth: false, wantsHeight: false};
+            spaceRequest = {width: 0, height: 0};
           }
+
 
           var allocatedWidth = Math.min(spaceRequest.width, offeredWidths[colIndex]);
           var allocatedHeight = Math.min(spaceRequest.height, offeredHeights[rowIndex]);
+          var componentNeedsWidth = spaceRequest.width > offeredWidths[colIndex];
+          var componentNeedsHeight = spaceRequest.height > offeredHeights[rowIndex];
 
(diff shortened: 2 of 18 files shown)


bluong:
Can you better explain why you developed this branch using develop as the base branch?

What you could've done differently is possibly:

 * Create the computeLayoutAPI branch from apiRefactor
 * Merge develop directly to computeLayoutAPI (skipping the apiRefactor base compeltely)

src/components/axes/categoryAxis.ts
```diff
@@ -55,11 +55,12 @@ export module Axis {
...
+
+      var desiredWidth = textResult.usedWidth + widthRequiredByTicks;
+      var desiredHeight = textResult.usedHeight + heightRequiredByTicks;
       return {
-        width : textResult.usedWidth  + widthRequiredByTicks,
-        height: textResult.usedHeight + heightRequiredByTicks,
-        wantsWidth : !textResult.textFits,
-        wantsHeight: !textResult.textFits
+        width: desiredWidth,
+        height: desiredHeight
```
bluong: Might as well inline the expressions no?

src/components/table.ts
```diff
@@ -228,39 +228,42 @@ export module Component {
...
+          var componentNeedsHeight = spaceRequest.height > offeredHeights[rowIndex];
 
-          requestedWidths [colIndex] = Math.max(requestedWidths [colIndex], allocatedWidth );
+          requestedWidths[colIndex] = Math.max(requestedWidths[colIndex], allocatedWidth);
           requestedHeights[rowIndex] = Math.max(requestedHeights[rowIndex], allocatedHeight);
-          layoutWantsWidth [colIndex] = layoutWantsWidth [colIndex] || spaceRequest.wantsWidth;
-          layoutWantsHeight[rowIndex] = layoutWantsHeight[rowIndex] || spaceRequest.wantsHeight;
+
+          columnNeedsWidth[colIndex] = columnNeedsWidth[colIndex] || componentNeedsWidth;
+          rowNeedsHeight[rowIndex] = rowNeedsHeight[rowIndex] || componentNeedsHeight;
```
bluong: Let's do a bit of variable reorganization?

I'm thinking of
```typescript
var allocatedWidth...
var allocatedHeight...
requestedWidths[colIndex]...
requestedHeights[rowIndex]...
var componentNeedsWidth...
var componentNeedsHeight...
...rest of code
```

The main thing I wanted to point out was you created this componenNeedsWidth variable but it is used... 4 lines later.  I think it'll be less confusing to read if you place the variable more next to the line where it is used.

src/core/interfaces.ts
```diff
@@ -48,8 +48,6 @@ module Plottable {
   export type _SpaceRequest = {
     width: number;
     height: number;
-    wantsWidth: boolean;
-    wantsHeight: boolean;
```
bluong: The pull request description does not match this type signature

src/scales/categoryScale.ts
```diff
@@ -142,7 +142,10 @@ export module Scale {
     }
 
     public copy(): Category {
-      return new Category(this._d3Scale.copy());
+      return new Category(this._d3Scale.copy())
+                  .range(this.range())
+                  .innerPadding(this.innerPadding())
+                  .outerPadding(this.outerPadding());
```
bluong: Why this change?  I think it makes sense to do this but why now?

bluong:
The scale change is quite a bit out of scope and if it should happen here, we should probably have a whole sweep

jtlan:
@bluong wrote:
>Can you better explain why you developed this branch using develop as the base branch?
>
>What you could've done differently is possibly:
>
> * Create the computeLayoutAPI branch from apiRefactor
> * Merge develop directly to computeLayoutAPI (skipping the apiRefactor base compeltely)

That breaks the overlaying tests, which I was using to see what the differences were before/after.

src/components/axes/categoryAxis.ts
jtlan: I think we should use this style for object literals; it's a lot easier to read.

src/core/interfaces.ts
jtlan: No it doesn't. I was asking for suggestions before I pulled the trigger on refactoring the names.

jtlan:
> The scale change is quite a bit out of scope and if it should happen here, we should probably have a whole sweep.

Yeah, I fixed that part when I was thinking to do a much larger refactor.

src/components/axes/categoryAxis.ts
bluong: is it a lot easier to read...
```typescript
var desiredWidth = {{desiredWidthCode}}
var desiredHeight = {{desiredHeightCode}}
return { width: desiredWidth, height: desiredHeight };
```
vs
```typescript
return { width: {{desiredWidthCode}}, height: {{desiredHeightCode}} };
```

src/scales/categoryScale.ts
bluong: I think we should separate out this functionality unless necessary.  Unless you want to make a new pull request with this change

bluong:
All the comments above are still valid.  The change mainly looks good, but I don't really see any advantages here except for a lot more clarity.  Can you make the cahnges and then target to apiRefactor?

jtlan:
`Scale.Category` issue filed at #1931. I'll copy that commit to a different PR.

src/components/axes/categoryAxis.ts
jtlan: Oh, the *expressions*; that makes sense. I think each *property* should be on a different line though.

jtlan:
Closing to make new PR aimed at **apiRefactor**. Still going to base the branch on `develop` so it's easier to test.

Issue #1915, opened 2015-04-25
Use reference equality rather than keys when registering objects/callbacks


jtlan:
Prime examples:
- `addDataset()` on `Plot`
- `onEvent()` on `Dispatcher`s and `Interaction`s.

Also includes a refactor to use `offEvent(callback)` to deregister `callback` rather than `offEvent(key, null)`.

- [x] Implement `CallbackSet`
- [x] `onDomainChange()` for `Scale`s -- currently `Scale`s use `Broadcaster`s
- [x] `Dispatcher`s also use `Broadcaster`s.
- [x] Remove `Broadcaster` from datasets
- [x] Remove `Broadcaster` (after the previous three items are complete)
- [x] `Interaction`s currently use `onEvent(null)` instead of `offEvent()` to remove callbacks.
- [x] <del> `addDataset()` + `removeDataset()`: should take in the `Dataset` directly instead of a key (this one might be hard enough to warrant a separate issue)</del> --> #1974
- [x] `domainer` use ids
- [x] Remove `PlottableObject` because we no longer need the IDs. 

One thing to remember is  
```Typescript
var clipPathId = (<any> c)._boxContainer.node().getElementsByTagName("clipPath")[0].getAttribute("id");
```
does fail on PhantomJS. It works on all other browsers though. Reverted it to 
```typescript
var clipPathId = (<any>c)._boxContainer[0][0].firstChild.id;
```

aicioara:
Picking this up

jtlan:
@aicioara , I've taken on #1974. It's as hard as I was afraid it might be...

aicioara:
The ID is now only used on the scale `domainer`

PR #1923, opened 2015-04-26
Implement CallbackSet.

jtlan:
It's ... a set of callbacks. It can be used to implement direct-reference for callback registration, rather than the key model currently used on `Broadcaster`.

We can't use d3.set because it only accepts string keys.
[ES6 set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) isn't available yet, and we'd still have to extend it to enable calling the callbacks.

Hits the first checkbox on #1915.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -277,6 +277,19 @@ declare module Plottable {
 
 
 declare module Plottable {
+    module Utils {
+        class CallbackSet<CB extends Function> {
+            constructor();
+            add(value: CB): CallbackSet<CB>;
+            remove(value: CB): CallbackSet<CB>;
+            values(): CB[];
+            callCallbacks(...args: any[]): CallbackSet<CB>;
+        }
+    }
+}
+
+
+declare module Plottable {
     type Formatter = (d: any) => string;
     var MILLISECONDS_IN_ONE_DAY: number;
     module Formatters {
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -278,6 +278,11 @@ declare module Plottable {
 
 declare module Plottable {
     module Utils {
+        /**
+         * A set of callbacks which can be all invoked at once.
+         * Each callback exists at most once in the set (based on reference equality).
+         * Ideally, all callbacks should have the same signature.
+         */
         class CallbackSet<CB extends Function> {
             constructor();
             add(value: CB): CallbackSet<CB>;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -281,7 +281,7 @@ declare module Plottable {
         /**
          * A set of callbacks which can be all invoked at once.
          * Each callback exists at most once in the set (based on reference equality).
-         * Ideally, all callbacks should have the same signature.
+         * All callbacks should have the same signature.
          */
         class CallbackSet<CB extends Function> {
             constructor();
--- a/src/utils/callbackSet.ts
+++ b/src/utils/callbackSet.ts
@@ -1,0 +1,38 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+  export module Utils {
+    export class CallbackSet<CB extends Function> {
+      private _values: CB[];
+
+      constructor() {
+        this._values = [];
+      }
+
+      public add(value: CB) {
+        if (this._values.indexOf(value) === -1) {
+          this._values.push(value);
+        }
+        return this;
+      }
+
+      public remove(value: CB) {
+        var index = this._values.indexOf(value);
+        if (index !== -1) {
+          this._values.splice(index, 1);
+        }
+        return this;
+      }
+
+      public values() {
+        return this._values;
+      }
+
+      public callCallbacks(...args: any[]) {
+        // no fat-arrow notation to set "this" to current "this" context
+        this.values().forEach(function(callback) { callback.apply(this, args); });
+        return this;
+      }
+    }
+  }
+}
--- a/src/utils/callbackSet.ts
+++ b/src/utils/callbackSet.ts
@@ -34,7 +34,7 @@ module Plottable {
       }
 
       public callCallbacks(...args: any[]) {
-        this.values().forEach(function(callback) {
+        this.values().forEach((callback) => {
           callback.apply(this, args);
         });
         return this;
--- a/src/utils/callbackSet.ts
+++ b/src/utils/callbackSet.ts
@@ -2,6 +2,11 @@
 
 module Plottable {
   export module Utils {
+    /**
+     * A set of callbacks which can be all invoked at once.
+     * Each callback exists at most once in the set (based on reference equality).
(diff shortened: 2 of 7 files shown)


plottable.d.ts
```diff
@@ -277,6 +277,19 @@ declare module Plottable {
 
 
 declare module Plottable {
+    module Utils {
+        class CallbackSet<CB extends Function> {
+            constructor();
+            add(value: CB): CallbackSet<CB>;
+            remove(value: CB): CallbackSet<CB>;
```
bluong: Note that a bunch of Java API and I think a bunch of other API have add/remove actually return a boolean.  I will note that because of our style, I think it is correct in this case that we return `this` and I actually personally disagree with the conventional API.  I am mainly noting that I think this is outside convention.  Not incorrect, but a tiny bit unconventional.

plottable.d.ts
```diff
@@ -277,6 +277,19 @@ declare module Plottable {
...
+            values(): CB[];
+            callCallbacks(...args: any[]): CallbackSet<CB>;
```
bluong: I think a `has(value: CB)` would probably be really valuable.  Is this not the case?

src/utils/callbackSet.ts
```diff
@@ -0,0 +1,38 @@
...
+  export module Utils {
+    export class CallbackSet<CB extends Function> {
+      private _values: CB[];
+
+      constructor() {
+        this._values = [];
+      }
+
+      public add(value: CB) {
+        if (this._values.indexOf(value) === -1) {
```
bluong: `Set`s are usually known to work within O(1)... but this is clearly not the case.  Is it alright if we go against the convention for time complexities for `Set`s?

src/utils/callbackSet.ts
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return this;
+      }
+
+      public values() {
+        return this._values;
+      }
+
+      public callCallbacks(...args: any[]) {
+        // no fat-arrow notation to set "this" to current "this" context
```
bluong: This is an extremely common trick that we use within our codebase, and I'm not sure if it makes sense to list this fact every time we use the "function" for `this` scenario.  That said, I'm not sure if it's intuitive enough to pick up.

How do we inform a reader that we are using `function` for the specific reason of achieving the right `this` context without using a comment? (Ideally we would use single-arrow... but that does not exist)

src/utils/callbackSet.ts
```diff
@@ -0,0 +1,38 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+  export module Utils {
+    export class CallbackSet<CB extends Function> {
```
bluong: Are we ok with exposing CallbackSet to the user?

bluong:
The biggest questions I have regard the exposure of CallbackSet as well as the fact that the implementation is not conventional on an abstract level.

plottable.d.ts
rcchen: Is there a need for chaining to work with CallbackSets?

src/utils/callbackSet.ts
rcchen: Agree with @bluong as there is an inherent assumption for any operations on a `Set`.

plottable.d.ts
bluong: Chaining is something that is pretty consistent in our codebase, and we should press for consistency where we can.  I think one thing that is really consistent with our library and other modern libraries is the idea of returning `this` whenever we are explicitly setting state / otherwise would return a `void`

plottable.d.ts
jtlan: I think `has` is more idiomatic. I don't think we actually need it though, since this is a strict-equality set.

src/utils/callbackSet.ts
jtlan: I could call it `UniqueArray` but it's really just a `Set`. Can't get O(1) (amortized) runtime because there's no hashing. Also n is probably <100 here so I doubt this will be much of a bottleneck.

src/utils/callbackSet.ts
jtlan: If I don't `export` it the other modules can't see it...

aicioara:
The code I was talking about is something like this
```javascript
foo = function(){console.log("foo")};
$(document).on("hello", foo);
$(document).dispatch("hello");
```

jtlan:
@aicioara wrote:
>The code I was talking about is something like this
```javascript
foo = function(){console.log("foo")};
$(document).on("hello", foo);
$(document).dispatch("hello");
```

Attaching things to `document` seems like it might lead to problems cleaning up.

plottable.d.ts
bluong: Well `has` or `contains` is mainly for checking if the callback is there.  If it doesn't need to exist then sure we don't need it.

src/utils/callbackSet.ts
bluong: Yes, exporting this class is the one way for us to use it.  However, it means that now this CallbackSet now falls under our public API.  So my main question is...  Are we ok with exposing this, since if we do, it's part of our public API.

aicioara:
```javascript
document.addEventListener('foo', function(e) {
  console.log(e.detail);
})

var ev = document.createEvent('Event');
ev.initEvent('foo', true, true);
ev.detail = "It Worked";

document.dispatchEvent(ev);
```

jtlan:
After some discussion with @ashwinr, `CallbackSet` seems the correct way to go for this refactor. I think it does a better job of preserving the semantics we want for callbacks.

Going to hand this off to @aicioara, who expressed interest in solving this part of the refactor.

aicioara:
Sounds good, taking this from here. Will address all issues above myself

aicioara:
Not much to address yet. Rebased on refactor Api, added some comments, and improved style of the function for now

src/utils/callbackSet.ts
```diff
@@ -0,0 +1,45 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+  export module Utils {
+    /**
+     * A set of callbacks which can be all invoked at once.
+     * Each callback exists at most once in the set (based on reference equality).
+     * Ideally, all callbacks should have the same signature.
```
jtlan: Isn't it not even an "ideally"?

I think the reference equality is worth noting, but I don't think we need to explain that each callback exists at most once -- that's the definition of a "set", after all.

aicioara: This is not an exposed API point (or will not be), hence I prefer clarity, but if you want, I can remove some of that information

aicioara:
Assigning @bluong since @jtlan wrote the code anyway

src/utils/callbackSet.ts
bluong: Remove this comment probably?

PR #1925, opened 2015-04-27
Refactor colon whitespace

aicioara:
We no longer have space before `:`. Also updated the tslint rules to enforce that

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -464,7 +464,7 @@ export module Plots {
     /**
      * Sets the hover mode for hover interactions. There are two modes:
      *     - "point": Selects the bar under the mouse cursor (default).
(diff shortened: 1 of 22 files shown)


src/components/plots/barPlot.ts
```diff
@@ -464,7 +464,7 @@ export module Plots {
     /**
      * Sets the hover mode for hover interactions. There are two modes:
      *     - "point": Selects the bar under the mouse cursor (default).
-     *     - "line" : Selects any bar that would be hit by a line extending
+     *     - "line": Selects any bar that would be hit by a line extending
```
jtlan: I'm surprised that it affects comments.

aicioara: The tslint fix was not manual, but a regex, hence comments were included

aicioara:
Waiting for all checks, then merging

PR #1926, opened 2015-04-27
[Dispatcher.Touch] - adding onTouchCancel

bluong:


jtlan:
If this fixes the problems we're having with PanZoom, then we should merge it.

PR #1933, opened 2015-04-29
requestedSpace() API

jtlan:
Re-targeted from #1922.

Still based on **develop** to make testing easier. I'll merge in **apiRefactor** after approval.

`_SpaceRequest` now looks like this:
```Typescript
export type _SpaceRequest {
  minWidth: number;
  minHeight: number;
}
```

Testing: There might be some minor layout changes; mostly we just need to make sure everything still looks reasonable, particularly `Axis.Category` and `Legend`.

### DIFF
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -94,10 +94,10 @@ export module Component {
       this._isSetup = true;
     }
 
-    public _requestedSpace(availableWidth : number, availableHeight: number): _SpaceRequest {
+    public _requestedSpace(availableWidth: number, availableHeight: number): _SpaceRequest {
       return {
-        width: 0,
-        height: 0
+        minWidth: 0,
+        minHeight: 0
       };
     }
 
@@ -149,8 +149,8 @@ export module Component {
     protected _getSize(availableWidth: number, availableHeight: number) {
       var requestedSpace = this._requestedSpace(availableWidth, availableHeight);
       return {
-        width: this._isFixedWidth()  ? Math.min(availableWidth , requestedSpace.width)  : availableWidth,
-        height: this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.height) : availableHeight
+        width: this._isFixedWidth()  ? Math.min(availableWidth , requestedSpace.minWidth)  : availableWidth,
+        height: this._isFixedHeight() ? Math.min(availableHeight, requestedSpace.minHeight) : availableHeight
       };
     }
 
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -95,7 +95,10 @@ export module Component {
     }
 
     public _requestedSpace(availableWidth : number, availableHeight: number): _SpaceRequest {
-      return {width: 0, height: 0, wantsWidth: false, wantsHeight: false};
+      return {
+        width: 0,
+        height: 0
+      };
     }
 
     /**
--- a/src/components/abstractComponent.ts
+++ b/src/components/abstractComponent.ts
@@ -32,7 +32,6 @@ export module Component {
     private _yOffset = 0;
     private _cssClasses: string[] = ["component"];
     private _removed = false;
-    private _usedLastLayout = false;
 
     /**
      * Attaches the Component as a child of a given a DOM element. Usually only directly invoked on root-level Components.
@@ -168,19 +167,7 @@ export module Component {
 
     public _doRender() {/* overwrite */}
 
-    public _useLastCalculatedLayout(): boolean;
-    public _useLastCalculatedLayout(useLast: boolean) : AbstractComponent;
-    public _useLastCalculatedLayout(useLast?: boolean) : any {
-      if (useLast == null) {
-        return this._usedLastLayout;
-      } else {
-        this._usedLastLayout = useLast;
-        return this;
-      }
-    }
-
     public _invalidateLayout() {
-      this._useLastCalculatedLayout(false);
       if (this._isAnchored && this._isSetup) {
         if (this._isTopLevelComponent) {
           this._scheduleComputeLayout();
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -184,9 +184,7 @@ export module Component {
       var wantsFitMoreEntriesInRow = estimatedLayout.rows.length > desiredNumRows;
       return {
         width : this._padding + longestRowLength,
-        height: acceptableHeight,
-        wantsWidth: offeredWidth < desiredWidth || wantsFitMoreEntriesInRow,
-        wantsHeight: offeredHeight < desiredHeight
+        height: acceptableHeight
       };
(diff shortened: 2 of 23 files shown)


src/components/abstractComponent.ts
```diff
@@ -95,7 +95,10 @@ export module Component {
     }
 
     public _requestedSpace(availableWidth : number, availableHeight: number): _SpaceRequest {
-      return {width: 0, height: 0, wantsWidth: false, wantsHeight: false};
+      return {
+        width: 0,
+        height: 0
+      };
```
bluong: How do we feel about this style vs

`return { width: 0, height: 0 }; `

bluong: I personally like the above only if that style has the object literal fit on a single line neatly

src/components/legend.ts
```diff
@@ -184,9 +184,7 @@ export module Component {
       var wantsFitMoreEntriesInRow = estimatedLayout.rows.length > desiredNumRows;
       return {
         width : this._padding + longestRowLength,
-        height: acceptableHeight,
-        wantsWidth: offeredWidth < desiredWidth || wantsFitMoreEntriesInRow,
-        wantsHeight: offeredHeight < desiredHeight
+        height: acceptableHeight
```
bluong: Lots of dead code above.

src/components/table.ts
```diff
@@ -228,39 +228,49 @@ export module Component {
...
+          var allocatedHeight = Math.min(spaceRequest.height, offeredHeights[rowIndex]);
           requestedHeights[rowIndex] = Math.max(requestedHeights[rowIndex], allocatedHeight);
-          layoutWantsWidth [colIndex] = layoutWantsWidth [colIndex] || spaceRequest.wantsWidth;
-          layoutWantsHeight[rowIndex] = layoutWantsHeight[rowIndex] || spaceRequest.wantsHeight;
+
+          var componentNeedsWidth = spaceRequest.width > offeredWidths[colIndex];
+          columnNeedsWidth[colIndex] = columnNeedsWidth[colIndex] || componentNeedsWidth;
+
+          var componentNeedsHeight = spaceRequest.height > offeredHeights[rowIndex];
+          rowNeedsHeight[rowIndex] = rowNeedsHeight[rowIndex] || componentNeedsHeight;
```
bluong: I think I see the point of the pull request now.  Is this pull request essentially trying the generalize the idea of wantsWidth/Height across all components into these lines here?

bluong:
Correct the PR description and let's see what happens if you merge in apiRefactor

src/components/abstractComponent.ts
jtlan: The "sometimes one-liner" seems less enforceable. While putting every property on a different line is less compact, I think it makes it clearer what the properties are (and by extension, makes it easier to identify the type of object being returned).

jtlan:
Don't want to merge in **apiRefactor** until after testing is complete, otherwise the overlaying tests stop working. Alternatively, we can fix all the overlaying tests in a separate PR first...

src/components/table.ts
jtlan: The actual objective is to simplify the return signature of `requestedSpace()`, although the way that is accomplished is to generalize `wantsWidth` and `wantsHeight`.

bluong:
Assigning to @crmorford since we are going to want a QE pass over this

jtlan:
Wow, this branch and PR were totally misnamed :(

bluong:
If only github had support for renaming branches on the fly -.-;

crmorford:
Double-axes (triple here) look very different, which is a little concerning since that's a major use case on slate
screen shot 2015-04-29 at 5 19 35 pm

jtlan:
Had to change the layout logic for `Table` and `Label`, since with the streamlined API `minWidth` and `minHeight` now refer to "minimum acceptable values", so:
* `Label` has to request enough to show all its rows
* `Table` has to request more space if its `Component`s don't have enough.

Also got rid of `_useLastCalculatedLayout`; instead, `Table` checks if the last-calculated layout will fit in the final offer received in `_computeLayout()`, and only reuses the last layout if it would fit.

src/components/legend.ts
```diff
@@ -167,26 +167,15 @@ export module Component {
 
     public _requestedSpace(offeredWidth: number, offeredHeight: number): _SpaceRequest {
       var estimatedLayout = this._calculateLayoutInfo(offeredWidth, offeredHeight);
-      var rowLengths = estimatedLayout.rows.map((row: string[]) => {
+
+      var estimatedRowLengths = estimatedLayout.rows.map((row: string[]) => {
         return d3.sum(row, (entry: string) => estimatedLayout.entryLengths.get(entry));
       });
+      var longestEstimatedRowLength = _Util.Methods.max(estimatedRowLengths, 0);
```
bluong: Note a big deal, but how much trouble do you think it would be to inline estimatedRowLengths

src/components/table.ts
```diff
@@ -225,49 +225,62 @@ export module Component {
...
-            spaceRequest = {width: 0, height: 0, wantsWidth: false, wantsHeight: false};
+            spaceRequest = {
+              minWidth: 0,
+              minHeight: 0
+            };
           }
 
-          var allocatedWidth = Math.min(spaceRequest.width, offeredWidths[colIndex]);
-          var allocatedHeight = Math.min(spaceRequest.height, offeredHeights[rowIndex]);
+          var allocatedWidth = isFinalOffer ? Math.min(spaceRequest.minWidth, offeredWidths[colIndex]) : spaceRequest.minWidth;
```
bluong: I guess I don't follow the reason for this change.  Why is it that the offeredWidths don't make an impact if this isn't the final offer?

bluong:
Mainly questioning what I see as the probably the most important line in this PR

src/components/table.ts
jtlan: If they did, `Table`s would never get to request more width than they were initially offered.

jtlan:
After we're done debating on this PR I'm going to rename the branch, otherwise it will be very confusing for anyone coming along later.

src/components/legend.ts
jtlan: Probably easier to read with them defined separately.

src/components/table.ts
bluong: The way I'm reading it, does the component's space request take precedent over the offered space at the final stage?  

I guess I'm wondering what happens in the scenario of sum(minWidths of components) > total width available ?

Based on my reading, allocatedWidth would be > total width available no?

jtlan: `allocatedWidth` <= totalWidth if the offer is final, otherwise it might be bigger.

bluong: Hmm consider maybe renaming at this point?  The functionality sort of makes sense but not intuitively

jtlan:
Version with correct branch name at #1962.

PR #1935, opened 2015-04-29
[TSLint Rule] - adding one-line check-whitespace rule

bluong:
This rule checks to make sure there is whitespace in front of specified tokens that are specified via the one-line rule.

Example specified token: `{` after an if like `if (foo) {`

Messy:
```typescript
if (foo){
...
```

Neat:
```typescript
if (foo) {
...
```


aicioara:
Looks good

PR #1936, opened 2015-04-29
[TSLint Rule] - adding call-signature rule

bluong:
This rule enforces having no space in front of the type definition in function signatures.

Messy:
```typescript
public foo() : number {
...
```

Neat:
```typescript
public foo(): number {
...
```


aicioara:

PR #1938, opened 2015-04-29
[TSLint Rule] - adding property-declaration rule

bluong:
This rule enforces that no whitespace is used before the colon in property declarations.

Messy:
```typescript
class Foo {
  public bar : number;
...
```

Neat:
```typescript
class Foo {
  public bar: number;
...
```


aicioara:

PR #1939, opened 2015-04-29
[TSLint Rule] - adding variable-declaration rule

bluong:
This rule enforces no whitespace in front of colons for arbitrary variable declarations.

Messy:
```typescript
var foo : number;
```

Neat:
```typescript
var foo: number;
```


aicioara:

PR #1941, opened 2015-04-29
[TSLint Rule] - adding ban rule on assert.equal

bluong:
Enforcing that `assert.equal` should not be used.  In situations that `assert.equal` is used, `assert.strictEqual` is probably more accurate.

Messy:
```typescript
assert.equal(foo, 5, "foo is equal to 5!");
```

Neat:
```typescript
assert.strictEqual(foo, 5, "foo is equal to 5!");
```


test/components/plots/barPlotTests.ts
```diff
@@ -64,18 +64,18 @@ describe("Plots", () => {
         var bar1 = d3.select(bars[0][1]);
         assert.closeTo(TestMethods.numAttr(bar0, "width"), xScale.rangeBand(), 1, "bar0 width is correct");
         assert.closeTo(TestMethods.numAttr(bar1, "width"), xScale.rangeBand(), 1, "bar1 width is correct");
-        assert.equal(bar0.attr("height"), "100", "bar0 height is correct");
-        assert.equal(bar1.attr("height"), "150", "bar1 height is correct");
+        assert.strictEqual(+bar0.attr("height"), 100, "bar0 height is correct");
+        assert.strictEqual(+bar1.attr("height"), 150, "bar1 height is correct");
         assert.closeTo(TestMethods.numAttr(bar0, "x"), 111, 1, "bar0 x is correct");
```
aicioara: I would rather do string comparison than coercing

test/components/plots/barPlotTests.ts
```diff
@@ -483,16 +483,16 @@ describe("Plots", () => {
...
+        assert.strictEqual(+bar1.attr("x"), 150, "bar1 x is correct");
 
         var baseline = renderArea.select(".baseline");
-        assert.equal(baseline.attr("x1"), "200", "the baseline is in the correct horizontal position");
-        assert.equal(baseline.attr("x2"), "200", "the baseline is in the correct horizontal position");
-        assert.equal(baseline.attr("y1"), "0", "the baseline starts at the top of the chart");
-        assert.equal(baseline.attr("y2"), SVG_HEIGHT, "the baseline ends at the bottom of the chart");
+        assert.strictEqual(+baseline.attr("x1"), 200, "the baseline is in the correct horizontal position");
+        assert.strictEqual(+baseline.attr("x2"), 200, "the baseline is in the correct horizontal position");
+        assert.strictEqual(+baseline.attr("y1"), 0, "the baseline starts at the top of the chart");
```
aicioara: This is a good example for my suggestion above. `+null === 0`

test/components/plots/gridPlotTests.ts
```diff
@@ -14,36 +14,36 @@ describe("Plots", () => {
...
-      assert.equal(cellAV.attr("x"), "0", "cell 'AV' x coord is correct");
-      assert.equal(cellAV.attr("y"), "100", "cell 'AV' y coord is correct");
-      assert.equal(cellAV.attr("fill"), "#ffffff", "cell 'AV' color is correct");
-
-      assert.equal(cellBV.attr("height"), "100", "cell 'BV' height is correct");
-      assert.equal(cellBV.attr("width"), "200", "cell 'BV' width is correct");
-      assert.equal(cellBV.attr("x"), "200", "cell 'BV' x coord is correct");
-      assert.equal(cellBV.attr("y"), "100", "cell 'BV' y coord is correct");
-      assert.equal(cellBV.attr("fill"), "#777777", "cell 'BV' color is correct");
+      assert.strictEqual(cellAU.attr("height"), "100", "cell 'AU' height is correct");
```
aicioara: And also, it is unconsistent here with the above

test/components/plots/gridPlotTests.ts
```diff
@@ -100,13 +100,13 @@ describe("Plots", () => {
...
       for (var i = 0; i < cells.length; i++) {
         var cell = d3.select(cells[i]);
-        assert.equal(cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
-        assert.equal(cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
-        assert.equal(cell.attr("width"), CELL_WIDTH, "Cell width is correct");
-        assert.equal(cell.attr("height"), CELL_HEIGHT, "Cell height is correct");
+        assert.strictEqual(+cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
+        assert.strictEqual(+cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
+        assert.strictEqual(+cell.attr("width"), CELL_WIDTH, "Cell width is correct");
+        assert.strictEqual(+cell.attr("height"), CELL_HEIGHT, "Cell height is correct");
```
aicioara: better convert `CELL_HEIGHT` to string here than the other way round

test/components/plots/plotTests.ts
```diff
@@ -34,24 +34,24 @@ describe("Plots", () => {
       var r = new CountingPlot();
       r.addDataset("foo", dFoo);
 
-      assert.equal(1, r.renders, "initial render due to addDataset");
+      assert.strictEqual(1, r.renders, "initial render due to addDataset");
```
aicioara: Not sure if part of this commit, but as a TODO, this is inconsistent. First value is the given, second is the expected (at least in the rest of the testing codebase)

test/scales/scaleTests.ts
```diff
@@ -141,8 +141,8 @@ describe("Scales", () => {
       var scale = new Plottable.Scale.Linear();
       scale.autoDomain();
       var d = scale.domain();
-      assert.equal(d[0], 0);
-      assert.equal(d[1], 1);
+      assert.strictEqual(d[0], 0);
+      assert.strictEqual(d[1], 1);
```
aicioara: Again, not part of this PR, but these asserts should also have a meaningful explanation as the 3rd argument

test/components/plots/barPlotTests.ts
bluong: Since we are comparing numbers, we should do everything in terms of numbers.

One argument for this from my view is that if we wanted to do any math, then we should definitely have numbers.  Otherwise we will have to do a bunch of coercion, going from string to number to calculate and then back to string

aicioara:
Mainly just convert numbers to strings as second argument, rather than coercing the first argument with the hacky `+`. Rest looks fine (although the last few files were not nicely highlighted by GitHub (not sure why). Open TODO issues for the other 2 issues

test/components/plots/barPlotTests.ts
bluong: I am unsure if a precedent should be set here, but I can try to enforce a precedent after we do an audit over the testing framework.

After some discussion / thinking, I think we're running into a semantical difference in understanding this test.

bar0.attr("height") returns the "height" attribute of the bar0 selection.  However, this is not actually the height of the element.  It is literally just the value that is ripped from the DOM, which is actually returned as a string (since primitives are easily coercible to strings).

The actual height of the element comes after a layer of parsing the value that is gotten from the DOM.  The idea of a height is represented as a number, so we must parse this information and extract the number value from it.  This numerical value is actually the height of the bar.

Thus, this comes back to our understanding of the test message "bar1 height is correct".  If this means the height of the element, then we should be using the numerical value.  If this means the "height" attribute of the element, then we should be using the string value.

What do you think?

test/components/plots/plotTests.ts
bluong: We will need to do an audit soon over the testing infrastructure, as we have very little rules over it and well we should

test/components/plots/gridPlotTests.ts
bluong: Beyond scope of the PR.  This shouldn't be setting a precedent.  We should however decide to set one once we get to an audit of the framework.

test/components/plots/barPlotTests.ts
aicioara: As @jtlan was nicely pointing out to me at some point, we should think of our testing as a set of axioms which are build on each other. This means that we assume that a height attribute of `"100"` is to correspond to its actual height. (if this is not yet covered, it should be).

I can definitely see your point here, but I can't go over the fact that `+null === 0`, which will make some of the tests (see below) useless. See for yourself
```typescript
+d3.select('svg').attr('foo') === 0 // returns true
```

I believe that we are being too pedantic about semantics in this case. The tests have to be correct in the first place. Then we can be pedantic about them.

aicioara: As a side note, if you coerce strings to numbers properly, I am happy with the change. Using `+` for coercing is unacceptable in this case. (IMO).

test/components/plots/gridPlotTests.ts
```diff
@@ -100,13 +100,13 @@ describe("Plots", () => {
...
+      assert.strictEqual(cells.length, data.length);
       for (var i = 0; i < cells.length; i++) {
         var cell = d3.select(cells[i]);
-        assert.equal(cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
-        assert.equal(cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
-        assert.equal(cell.attr("width"), CELL_WIDTH, "Cell width is correct");
-        assert.equal(cell.attr("height"), CELL_HEIGHT, "Cell height is correct");
+        assert.strictEqual(+cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
+        assert.strictEqual(+cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
+        assert.strictEqual(+cell.attr("width"), CELL_WIDTH, "Cell width is correct");
```
aicioara: You have coercing here, hence unconsistent. Did not really get your argument, why was this out of the scope of the PR. Stringify the right handside

test/components/timeAxisTests.ts
```diff
@@ -112,11 +112,11 @@ describe("TimeAxis", () => {
     scale.domain(["2010", "2014"]);
     axis.renderTo(svg);
     var firstTick = d3.select(".tick-mark");
-    assert.equal(firstTick.attr("x1"), 0, "xPos (x1) of first end tick is at the beginning of the axis container");
-    assert.equal(firstTick.attr("x2"), 0, "xPos (x2) of first end tick is at the beginning of the axis container");
+    assert.strictEqual(+firstTick.attr("x1"), 0, "xPos (x1) of first end tick is at the beginning of the axis container");
+    assert.strictEqual(+firstTick.attr("x2"), 0, "xPos (x2) of first end tick is at the beginning of the axis container");
```
aicioara: ?

test/components/timeAxisTests.ts
```diff
@@ -112,11 +112,11 @@ describe("TimeAxis", () => {
...
     var lastTick = d3.select(d3.selectAll(".tick-mark")[0].pop());
-    assert.equal(lastTick.attr("x1"), width, "xPos (x1) of last end tick is at the end of the axis container");
-    assert.equal(lastTick.attr("x2"), width, "xPos (x2) of last end tick is at the end of the axis container");
+    assert.strictEqual(+lastTick.attr("x1"), width, "xPos (x1) of last end tick is at the end of the axis container");
+    assert.strictEqual(+lastTick.attr("x2"), width, "xPos (x2) of last end tick is at the end of the axis container");
```
aicioara: ?

test/core/componentTests.ts
```diff
@@ -210,17 +210,17 @@ describe("Component behavior", () => {
     assert.isTrue(normalizeClipPath((<any> c)._element.attr("clip-path")) === expectedClipPathURL,
                   "the element has clip-path url attached");
     var clipRect = (<any> c)._boxContainer.select(".clip-rect");
-    assert.equal(clipRect.attr("width"), 100, "the clipRect has an appropriate width");
-    assert.equal(clipRect.attr("height"), 100, "the clipRect has an appropriate height");
+    assert.strictEqual(+clipRect.attr("width"), 100, "the clipRect has an appropriate width");
+    assert.strictEqual(+clipRect.attr("height"), 100, "the clipRect has an appropriate height");
```
aicioara: ?

test/testMethods.ts
```diff
@@ -89,10 +89,10 @@ module TestMethods {
   };
 
   export function assertWidthHeight(el: D3.Selection, widthExpected: number, heightExpected: number, message: string) {
-    var width = el.attr("width");
-    var height = el.attr("height");
-    assert.equal(width, widthExpected, "width: " + message);
-    assert.equal(height, heightExpected, "height: " + message);
+    var width = +el.attr("width");
+    var height = +el.attr("height");
```
aicioara: ?

test/tests.js
```diff
@@ -85,10 +85,10 @@ var TestMethods;
     TestMethods.assertPointsClose = assertPointsClose;
     ;
     function assertWidthHeight(el, widthExpected, heightExpected, message) {
-        var width = el.attr("width");
-        var height = el.attr("height");
-        assert.equal(width, widthExpected, "width: " + message);
-        assert.equal(height, heightExpected, "height: " + message);
+        var width = +el.attr("width");
+        var height = +el.attr("height");
```
aicioara: ?

test/tests.js
```diff
@@ -4123,13 +4123,13 @@ describe("Plots", function () {
...
-            assert.equal(cells.length, data.length);
+            assert.strictEqual(cells.length, data.length);
             for (var i = 0; i < cells.length; i++) {
                 var cell = d3.select(cells[i]);
-                assert.equal(cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
-                assert.equal(cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
-                assert.equal(cell.attr("width"), CELL_WIDTH, "Cell width is correct");
-                assert.equal(cell.attr("height"), CELL_HEIGHT, "Cell height is correct");
+                assert.strictEqual(+cell.attr("x"), i * CELL_WIDTH, "Cell x coord is correct");
+                assert.strictEqual(+cell.attr("y"), i * CELL_HEIGHT, "Cell y coord is correct");
```
aicioara: ?

test/tests.js
```diff
@@ -6566,16 +6566,16 @@ describe("Component behavior", function () {
         var normalizeClipPath = function (s) { return s.replace(/"/g, ""); };
         assert.isTrue(normalizeClipPath(c._element.attr("clip-path")) === expectedClipPathURL, "the element has clip-path url attached");
         var clipRect = c._boxContainer.select(".clip-rect");
-        assert.equal(clipRect.attr("width"), 100, "the clipRect has an appropriate width");
-        assert.equal(clipRect.attr("height"), 100, "the clipRect has an appropriate height");
+        assert.strictEqual(+clipRect.attr("width"), 100, "the clipRect has an appropriate width");
```
aicioara: ?

PR #1946, opened 2015-04-30
[Axis.Numeric] - Removing showEndTicklabel

bluong:
From my reading of this and removal of the appropriate code, it seems like this endpoint... did absolutely nothing.  Removing this dead code

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2155,30 +2155,6 @@ declare module Plottable {
              * @returns {Numeric} The calling Axis.Numeric.
              */
             tickLabelPosition(position: string): Numeric;
-            /**
-             * Gets whether or not the tick labels at the end of the graph are
-             * displayed when partially cut off.
-             *
-             * @param {string} orientation Where on the scale to change tick labels.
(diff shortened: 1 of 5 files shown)


aicioara:
Looked through this. I am not really sure who uses what from our public API, so can't really +1 it. Maybe @jtlan should have an opinion. 
Anyway, since we delete this, why do we keep `abstractAxis.showEndTickLabels()` ? 

I am also curious what was it supposed to do, can't really make it work anyway.

aicioara:
Oh, I see, this is the setter/getter of a field that is never really actually used. I am now wondering why is this the case, when did it become obsolete and why was it not removed right then in the first place?

Anyway, ideally we should either make it work or delete `abstractAxis.showEndTickLabels()` as well, as it is in the exact same situation.

jtlan:
It became obsolete because we put a CSS class on the end ticks, so you could hide it like this:
```CSS
```

jtlan:
`abstractAxis.showEndTickLabels()` does seem to be used though, so we should probably leave it. Maybe we should move it to `Axes.Numeric` though?

PR #1949, opened 2015-05-02
StrictEqualityAssociativeArray --> Map

jtlan:
(recreating branch to remove unrelated commit)

Seriously, it's just a shim for an [ES6 Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

Although we're removing `Broadcaster`, I'm going to need `Utils.Map` to fix `Scale` autodomain calculations.

### DIFF
--- a/src/core/broadcaster.ts
+++ b/src/core/broadcaster.ts
@@ -23,7 +23,7 @@ export module Core {
    * The listeners are called synchronously.
    */
   export class Broadcaster<L> extends Core.PlottableObject {
-    private _key2callback = new Utils.StrictEqualityAssociativeArray();
+    private _key2callback = new Utils.Map();
     private _listenable: L;
 
     /**
@@ -92,7 +92,7 @@ export module Core {
      * @returns {Broadcaster} The calling Broadcaster
      */
     public deregisterAllListeners() {
-      this._key2callback = new Utils.StrictEqualityAssociativeArray();
+      this._key2callback = new Utils.Map();
     }
   }
 }
--- a/src/core/broadcaster.ts
+++ b/src/core/broadcaster.ts
@@ -23,7 +23,7 @@ export module Core {
    * The listeners are called synchronously.
    */
   export class Broadcaster<L> extends Core.PlottableObject {
-    private _key2callback = new Utils.Map();
+    private _key2callback = new Utils.Map<any, Function>();
     private _listenable: L;
 
     /**
@@ -92,7 +92,7 @@ export module Core {
      * @returns {Broadcaster} The calling Broadcaster
      */
     public deregisterAllListeners() {
-      this._key2callback = new Utils.Map();
+      this._key2callback = new Utils.Map<any, Function>();
     }
   }
 }
--- a/src/core/dataset.ts
+++ b/src/core/dataset.ts
@@ -8,7 +8,7 @@ module Plottable {
   export class Dataset extends Core.PlottableObject {
     private _data: any[];
     private _metadata: any;
-    private _accessor2cachedExtent: Utils.StrictEqualityAssociativeArray;
+    private _accessor2cachedExtent: Utils.Map;
     public broadcaster: Core.Broadcaster<Dataset>;
 
     /**
@@ -25,7 +25,7 @@ module Plottable {
       super();
       this._data = data;
       this._metadata = metadata;
-      this._accessor2cachedExtent = new Utils.StrictEqualityAssociativeArray();
+      this._accessor2cachedExtent = new Utils.Map();
       this.broadcaster = new Core.Broadcaster(this);
     }
 
@@ -47,7 +47,7 @@ module Plottable {
         return this._data;
       } else {
         this._data = data;
-        this._accessor2cachedExtent = new Utils.StrictEqualityAssociativeArray();
+        this._accessor2cachedExtent = new Utils.Map();
         this.broadcaster.broadcast();
         return this;
       }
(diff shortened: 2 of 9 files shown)


src/core/dataset.ts
```diff
@@ -8,7 +8,7 @@ module Plottable {
   export class Dataset extends Core.PlottableObject {
     private _data: any[];
     private _metadata: any;
-    private _accessor2cachedExtent: Utils.StrictEqualityAssociativeArray;
+    private _accessor2cachedExtent: Utils.Map<any, any>;
```
bluong: One of these values is not an `any` but an `_Accessor` no?

jtlan: Indeed, it should be a `Utils.Map<_Accessor, any[]>`.

src/core/broadcaster.ts
```diff
@@ -92,7 +92,7 @@ export module Core {
      * @returns {Broadcaster} The calling Broadcaster
      */
     public deregisterAllListeners() {
-      this._key2callback = new Utils.StrictEqualityAssociativeArray();
```
bluong: I believe broadcasters push in a string as a key right?  Up to you if you want to enforce this since Broadcasters are going away.

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Set a new key/value pair in the store.
+     *
+     * @return {boolean} True if key already in store, false otherwise
+     */
+    public set(key: K, value: V) {
+      if (key !== key) {
+        throw new Error("NaN may not be used as a key to Map");
+      }
```
bluong: Why this enforcement?

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+          this._keyValuePairs[i].value = value;
+          return true;
+        }
+      }
+      this._keyValuePairs.push({
+        key: key,
+        value: value
+      });
+      return false;
+    }
```
bluong: Judging from this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set

map is chainable and doesn't return booleans in this manner

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+module Plottable {
+export module Utils {
+  /**
+   * Shim for the ES6 map (although NaN is not allowed as a key).
+   */
+  export class Map<K, V> {
+    private _keyValuePairs: { key: K; value: V; }[] = [];
+
+    /**
+     * Set a new key/value pair in the store.
```
bluong: Let's change the doc so that it doesn't say store but map

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+      }
+      this._keyValuePairs.push({
+        key: key,
+        value: value
+      });
+      return false;
+    }
+
+    /**
+     * Get a value from the store, given a key.
```
bluong: Same as above

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+      for (var i = 0; i < this._keyValuePairs.length; i++) {
+        if (this._keyValuePairs[i].key === key) {
+          return this._keyValuePairs[i].value;
+        }
+      }
+      return undefined;
+    }
+
+    /**
+     * Test whether store has a value associated with given key.
```
bluong: Same as above.  Also I like "Check" instead of "Test"

bluong: Let's stick with terminology similar to mdn to be consistent
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+      return false;
+    }
+
+    public values(): V[] {
+      return this._keyValuePairs.map((pair) => pair.value);
+    }
+
+    public keys(): K[] {
+      return this._keyValuePairs.map((pair) => pair.key);
+    }
```
bluong: MDN returns an iterator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/keys

Do we want to differ here?

src/utils/map.ts
```diff
@@ -0,0 +1,87 @@
...
+        if (this._keyValuePairs[i].key === key) {
+          return true;
+        }
+      }
+      return false;
+    }
+
+    public values(): V[] {
+      return this._keyValuePairs.map((pair) => pair.value);
+    }
```
bluong: Same as above

bluong:
My main concern is that we are inconsistent from the referenced MDN documentation.  I think we should be consistent here so that if we can be consistent with whatever happens in the future in ES6

src/utils/map.ts
jtlan: There's a lot of places in the codebase where we `forEach()` over the keys. Should this maybe not be named `Map`?

jtlan:
This will probably be easier after @aicioara gets rid of `Broadcaster.

aicioara:
@jtlan watch #1915 for when this will happen.

aicioara:
@jtlan Broadcasters no longer exist, feel free to reopen if needed

aicioara:
added `-2` as I am not sure what is the status of this

aicioara:
Too outdated to be worth continuing from here. Started fresh here https://github.com/palantir/plottable/pull/1983

PR #1950, opened 2015-05-03
Add in shim for ES6 Set.

jtlan:
Became the superclass of CallbackSet...


bluong:
Both the builds failed.  Correct it and we should be ok

jtlan:
Looks like Travis is failing because the file was accidentally named "*S*et.ts". My git isn't recognizing the case-change when I try to rename it so I'm going to do a force-push.

PR #1951, opened 2015-05-03
Utils.min() and Utils.max() now return defaultValue in all bad cases.

jtlan:
Previously, a bad accessor-function or a weird input array would cause `undefined` to be returned. Since the entire point of `Utils.min()` and `Utils.max()` was to be safer than the d3 versions, they now return
the default value whenever `undefined` would be returned.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -206,45 +206,33 @@ export module Utils {
     }
 
     /**
-     * Computes the max value from the array.
-     *
-     * If type is not comparable then t will be converted to a comparable before computing max.
+     * Applies the accessor, if provided, to each element of `array` and returns the maximum value.
+     * If no maximum value can be computed, returns defaultValue.
      */
-    export function max<C>(arr: C[], default_val: C): C;
-    export function max<T, C>(arr: T[], acc: (x?: T, i?: number) => C, default_val: C): C;
-    export function max(arr: any[], one: any, two?: any): any {
-      if (arr.length === 0) {
-        if (typeof(one) !== "function") {
-          return one;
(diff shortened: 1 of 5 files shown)


jtlan:
I think this should be [NOQE] since the unit tests on this part of the functionality are reasonably comprehensive.

src/utils/utils.ts
```diff
@@ -206,45 +206,33 @@ export module Utils {
...
+    export function max(array: any[], firstArg: any, secondArg?: any): any {
+      var accessor = typeof(firstArg) === "function" ? firstArg : null;
+      var defaultValue = accessor == null ? firstArg : secondArg;
       /* tslint:disable:ban */
-      var acc = typeof(one) === "function" ? one : typeof(two) === "function" ? two : undefined;
-      return acc === undefined ? d3.max(arr) : d3.max(arr, acc);
+      var maxValue = accessor == null ? d3.max(array) : d3.max(array, accessor);
       /* tslint:enable:ban */
+      return maxValue !== undefined ? maxValue : defaultValue;
     }
```
bluong: Side note I think we will want to rewrite this so that defaultValue is always the second argument.  I think it's becoming a smell if we have a signature that says firstArg/secondArg and we have to decipher which is which  like the structure we have now.

jtlan: It was probably done this way to keep with the [d3 signature](https://github.com/mbostock/d3/wiki/Arrays#d3_min):
```Typescript
min<T>(arr: T[]): T;
min<T, U>(arr: T[], map: (v?: T, i?: number) => U): U;
```

PR #1954, opened 2015-05-04
anchor() and computeLayout() refactor.

jtlan:
Removed underscores.

`computeLayout()` API has been changed to:
```Typescript
public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number);
```


src/components/component.ts
```diff
@@ -33,18 +32,18 @@ module Plottable {
     private _usedLastLayout = false;
 
     /**
-     * Attaches the Component as a child of a given a DOM element. Usually only directly invoked on root-level Components.
+     * Attaches the Component as a child of a given a D3 selection.
```
rcchen: Funky grammar here. Also we should probably standardize references to "D3 selection" or "Selection" and just put "D3.Selection" every time so there's zero ambiguity about what sort of selection it is.

src/components/component.ts
```diff
@@ -33,18 +32,18 @@ module Plottable {
...
      *
-     * @param {D3.Selection} element A D3 selection consisting of the element to anchor under.
+     * @param {D3.Selection} selection The Selection containing the Element to anchor under.
      */
```
rcchen: @return annotation now that it's returning this?

jtlan: Good catch. Thanks!

PR #1955, opened 2015-05-04
Replace string constants with static members of Alignment class

rcchen:
Moving from `if (alignment === "top") {` to become `if (alignment === Alignment.TOP) {`

Typescript does not have support for non-number based enums so this was an acceptable substitute that I found (http://stackoverflow.com/questions/15490560/create-an-enum-with-string-values-in-typescript)

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -1,6 +1,15 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
+  export module Components {
+    export class Alignment {
+      static TOP = "top";
+      static BOTTOM = "bottom";
+      static LEFT = "left";
+      static RIGHT = "right";
+      static CENTER = "center";
+    }
+  }
   export class Component extends Core.PlottableObject {
     protected _element: D3.Selection;
     protected _content: D3.Selection;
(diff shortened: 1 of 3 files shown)


src/components/component.ts
```diff
@@ -589,4 +589,11 @@ module Plottable {
       return this._backgroundContainer;
     }
   }
+  export class Alignment {
+    static TOP = "top";
+    static BOTTOM = "bottom";
+    static LEFT = "left";
+    static RIGHT = "right";
+    static CENTER = "center"; 
+  }
```
bluong: This is basically an enum.  Let's use an enum instead?

src/components/component.ts
```diff
@@ -589,4 +589,11 @@ module Plottable {
       return this._backgroundContainer;
     }
   }
+  export class Alignment {
+    static TOP = "top";
+    static BOTTOM = "bottom";
+    static LEFT = "left";
+    static RIGHT = "right";
+    static CENTER = "center";
+  }
```
bluong: It might be advisable to split this out into a file or place it into a separate file?  Do we use the idea of Alignment with the same set of variables anywhere else in the code base?

Also, I think we should put this class at the top of the file?

src/components/component.ts
```diff
@@ -589,4 +589,11 @@ module Plottable {
       return this._backgroundContainer;
     }
   }
+  export class Alignment {
+    static TOP = "top";
+    static BOTTOM = "bottom";
+    static LEFT = "left";
+    static RIGHT = "right";
+    static CENTER = "center";
```
bluong: Probably all public static variables?

PR #1956, opened 2015-05-04
Register extents using ExtentProvider interface instead of using a key.

jtlan:
Extents are now cached inside the `Plot` instead of the `Scale`.
These changes create a polling-based mechanism for extents, rather than a push-based mechanism.


src/components/plots/plot.ts
```diff
@@ -62,7 +66,7 @@ module Plottable {
       super._anchor(element);
       this._animateOnNextRender = true;
       this._dataChanged = true;
-      this._updateScaleExtents();
+      this._updateAllExtents();
```
bluong: rename to _updateExtents?

src/components/plots/plot.ts
```diff
@@ -54,6 +56,8 @@ module Plottable {
       this.clipPathEnabled = true;
       this.classed("plot", true);
       this._key2PlotDatasetKey = d3.map();
+      this._attrToExtents = d3.map();
+      this._extentProvider = (scale: Scale<any, any>) => this._extentsForScale(scale);
```
bluong: I'm wondering if this would be better as an instance function instead of an instance variable

src/components/plots/plot.ts
```diff
@@ -179,22 +183,33 @@ module Plottable {
...
-        this._datasetKeysInOrder.forEach((key) => {
-          existingScale._removeExtent(this.getID().toString() + "_" + key, attrToSet);
-          existingScale.broadcaster.deregisterListener(this);
-        });
+      accessor = Utils.Methods.accessorize(accessor);
+      this._projections[attrToSet] = {accessor: accessor, scale: scale, attribute: attrToSet};
+      this._updateExtentsForAttr(attrToSet);
+
+      if (previousScale) {
+        var listeningTopreviousScale = false;
```
bluong: I feel this variable can be summed up in a `.some` call

src/components/plots/plot.ts
```diff
@@ -179,22 +183,33 @@ module Plottable {
...
+        for (var attr in this._projections) {
```
bluong: I'm not too much a fan of `for in` since I believe there's some really weird javascript thing that you need to do this correctly...  Maybe rely more on d3 since they handle it well I believe?

src/components/plots/plot.ts
```diff
@@ -261,26 +276,52 @@ module Plottable {
...
-          var plotMetadata = plotDatasetKey.plotMetadata;
-          var extent = dataset._getExtent(projector.accessor, projector.scale._typeCoercer, plotMetadata);
-          var scaleKey = this.getID().toString() + "_" + key;
-          if (extent.length === 0 || !this._isAnchored) {
-            projector.scale._removeExtent(scaleKey, attr);
-          } else {
-            projector.scale._updateExtent(scaleKey, attr, extent);
+    protected _updateAllExtents() {
+      d3.keys(this._projections).forEach((attr: string) => { this._updateExtentsForAttr(attr); });
+      var scales: Scale<any, any>[] = [];
```
bluong: Would a filter make this cleaner?

src/components/plots/plot.ts
```diff
@@ -261,26 +276,52 @@ module Plottable {
...
+        if (scale != null && scales.indexOf(scale) === -1) {
+          scales.push(scale);
+        }
+      });
+      scales.forEach((scale) => scale._autoDomainIfAutomaticMode());
+    }
+
+    private _updateExtentsForAttr(attr: string) {
+      var accessor = this._projections[attr].accessor;
+      var scale = this._projections[attr].scale;
```
bluong: I think this variable is never used

src/components/plots/plot.ts
```diff
@@ -261,26 +276,52 @@ module Plottable {
...
+    protected _extentsForAttr(attr: string) {
+      return this._attrToExtents.get(attr);
+    }
+
+    private _extentsForScale<D>(scale: Scale<D, any>): D[][] {
+      if (!this._isAnchored) {
+        return [];
+      }
+      var allSetsOfExtents: D[][][] = [];
+      for (var attr in this._projections) {
```
bluong: Same as above for the weirdness of `for in`

test/components/plots/plotTests.ts
```diff
@@ -465,8 +465,8 @@ describe("Plots", () => {
       plot.project("x", "x", x);
       plot.project("y", "y", y);
       plot.renderTo(svg);
-      svg.remove();
       assert.equal(recordedTime, 20, "additionalPaint passed appropriate time argument");
+      svg.remove();
```
bluong: Whoa... the previous version looked really incorrect....

test/core/domainerTests.ts
```diff
@@ -11,7 +11,9 @@ describe("Domainer", () => {
   });
 
   it("pad() works in general case", () => {
-    scale._updateExtent("1", "x", [100, 200]);
+    var provider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[100, 200]];
```
bluong: I think the explicit typing is unneccesary?

test/core/domainerTests.ts
```diff
@@ -22,7 +24,9 @@ describe("Domainer", () => {
     var f = d3.time.format("%x");
     var d1 = f.parse("06/02/2014");
     var d2 = f.parse("06/03/2014");
-    timeScale._updateExtent("1", "x", [d1, d2]);
+    var provider: Plottable.Scales.ExtentProvider<Date> = (scale: Plottable.Scale<Date, number>) => [[d1, d2]];
```
bluong: Same as above

test/core/domainerTests.ts
```diff
@@ -36,7 +40,9 @@ describe("Domainer", () => {
 
   it("pad() works on log scales", () => {
     var logScale = new Plottable.Scales.Log();
-    logScale._updateExtent("1", "x", [10, 100]);
+    var provider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[10, 100]];
```
bluong: Same as above

test/core/domainerTests.ts
```diff
@@ -109,15 +112,17 @@ describe("Domainer", () => {
...
 
   it("paddingException(n) works on dates", () => {
-    var a = new Date(2000, 5, 5);
-    var b = new Date(2003, 0, 1);
-    domainer.pad().addPaddingException(a);
+    var startDate = new Date(2000, 5, 5);
+    var endDate = new Date(2003, 0, 1);
     var timeScale = new Plottable.Scales.Time();
-    timeScale._updateExtent("1", "x", [a, b]);
+    var provider: Plottable.Scales.ExtentProvider<Date> = (scale: Plottable.Scale<Date, number>) => [[startDate, endDate]];
```
bluong: Same as above

test/core/domainerTests.ts
```diff
@@ -150,15 +155,16 @@ describe("Domainer", () => {
...
-    var b = new Date(2000, 5, 5);
-    var c = new Date(2000, 5, 6);
-    var d = new Date(2003, 0, 1);
-    domainer.addIncludedValue(b);
+    var includedDate = new Date(2000, 5, 5);
+    var startDate = new Date(2000, 5, 6);
+    var endDate = new Date(2003, 0, 1);
     var timeScale = new Plottable.Scales.Time();
-    timeScale._updateExtent("1", "x", [c, d]);
+    var provider: Plottable.Scales.ExtentProvider<Date> = (scale: Plottable.Scale<Date, number>) => [[startDate, endDate]];
```
bluong: Same as above

test/core/domainerTests.ts
```diff
@@ -150,15 +155,16 @@ describe("Domainer", () => {
...
+    timeScale.addExtentProvider(provider);
```
bluong: Hey, if you just plug in the provider inline-style since you don't use the variable anywhere

test/scales/scaleTests.ts
```diff
@@ -50,7 +40,8 @@ describe("Scales", () => {
     });
 
     it("scale autoDomain flag is not overwritten without explicitly setting the domain", () => {
-      scale._updateExtent("1", "x", d3.extent(data, (e) => e.foo));
+      var fooProvider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [d3.extent(data, (e) => e.foo)];
+      scale.addExtentProvider(fooProvider);
```
bluong: Same as above

bluong:
Just in general, can you remove the explicit typing to Plottable.Scales.ExtentProvider<number> and just let the typing be implicit?

Also, since I believe the providers are not being used anywhere, I think the majority of them can be inlined without loss of complexity

src/components/plots/plot.ts
jtlan: I could pass `extentsForScale` directly as the `ExtentProvider`, but the `this`-context would be wrong.

src/components/plots/plot.ts
jtlan: Yeah, that's a good idea.

src/components/plots/plot.ts
jtlan: As per our offline discussion, going to switch to `Object.keys()`.

src/components/plots/plot.ts
jtlan: Can you explain how I might use that?

src/components/plots/plot.ts
jtlan: --> `Object.keys()`.

test/core/domainerTests.ts
jtlan: I think the compiler was annoyed at me for some reason, but at any rate it won't hurt.

test/core/domainerTests.ts
jtlan: I'll probably exceed the line-length limit :(

src/components/plots/plot.ts
bluong: Explain the `this`context being incorrect?

src/components/plots/plot.ts
jtlan: Only works for arrays, not `Object`s.

src/components/plots/plot.ts
jtlan: The line just below this...

src/components/plots/plot.ts
```diff
@@ -261,26 +276,52 @@ module Plottable {
...
-            projector.scale._updateExtent(scaleKey, attr, extent);
+    protected _updateAllExtents() {
+      d3.keys(this._projections).forEach((attr: string) => { this._updateExtentsForAttr(attr); });
+      var scales: Scale<any, any>[] = [];
+      d3.keys(this._projections).forEach((attr: string) => {
+        var scale = this._projections[attr].scale;
+        if (scale != null && scales.indexOf(scale) === -1) {
+          scales.push(scale);
+        }
+      });
```
bluong: I should have commented here to start with.

Maybe...

Object.keys(this._projections).map((attr) => this._projectionrs[attr].scale).filter((scale) => scale != null).reduce((scale, uniqScales) => if (uniqScales.indexOf(scale) === -1) { return uniqScales.concat(scale); });

Lots and lots of pure transformations

test/core/domainerTests.ts
bluong: Does this not work?
timeScale.addExtentProvider((scale) => [[startDate, endDate]]);

src/components/plots/plot.ts
jtlan: if I pass `this._extentsForScale` directly, it will be invoked with `this` equal to whatever it is inside the `CallbackSet` inside the `Scale`. Since `_extentsForScale()` depends on `this` being the `Plot`, that won't work.

src/components/plots/plot.ts
jtlan: Maybe have a `_allScales()` helper method.

src/components/plots/plot.ts
bluong: Oh I was referring that _extentProvider maybe should not be an instance variable but a function instead, cause now we have this _extentProvider instance variable running around that can be set or gotten

src/components/plots/plot.ts
bluong: So let's turn the object into an array.

```typescript
var listeningToPreviousScale = Object.keys(this._projections).some((attr) => this._projections[attr].scale === previousScale);
```

On another note... javascript (or we) needs a `.values()` endpoint

src/components/plots/plot.ts
bluong: Still would like to see transformations if possible as I think it could be easier to read

bluong: That said, if you think the resultant is not readable, leave as is

test/scales/scaleTests.ts
```diff
@@ -94,21 +84,29 @@ describe("Scales", () => {
...
-      assert.isTrue((<any> scale)._autoDomainAutomatically, "autoDomain enabled4");
-      assert.closeTo(scale.domain()[0], -20, 0.1, "only the bar accessor is active");
-      assert.closeTo(scale.domain()[1], 1, 0.1, "only the bar accessor is active");
-      scale._updateExtent("2", "x", d3.extent(data, (e) => e.foo));
-      assert.isTrue((<any> scale)._autoDomainAutomatically, "autoDomain enabled5");
-      assert.closeTo(scale.domain()[0], 0, 0.1, "the bar accessor was overwritten");
-      assert.closeTo(scale.domain()[1], 5, 0.1, "the bar accessor was overwritten");
+    it("addExtentProvider()", () => {
+      var posProvider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[0, 10]];
+      scale.addExtentProvider(posProvider);
```
bluong: Can we inline the Provider here?  I don't really see the purpose of separating into two lines

```typescript
scale.addExtentProvider(() => [[0, 10]]);
```

test/scales/scaleTests.ts
```diff
@@ -94,21 +84,29 @@ describe("Scales", () => {
...
+      scale.autoDomain();
+      assert.deepEqual(scale.domain(), [0, 10], "scale domain accounts for first provider");
+
+      var negProvider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[-10, 0]];
+      scale.addExtentProvider(negProvider);
```
bluong: Same as above

test/scales/scaleTests.ts
```diff
@@ -94,21 +84,29 @@ describe("Scales", () => {
...
+      scale.autoDomain();
+      assert.deepEqual(scale.domain(), [-10, 10], "scale domain accounts for second provider");
+    });
+
+     it("removeExtentProvider()", () => {
+      var posProvider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[0, 10]];
+      scale.addExtentProvider(posProvider);
+      var negProvider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => [[-10, 0]];
+      scale.addExtentProvider(negProvider);
```
bluong: Same as above

test/scales/scaleTests.ts
```diff
@@ -463,11 +462,15 @@ describe("Scales", () => {
...
+      var providedExtents = [[0, base / 2]];
+      var provider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => providedExtents;
+      scale.addExtentProvider(provider);
+      scale.autoDomain();
       var ticks = scale.ticks();
       assert.operator(ticks.length, ">", 0);
 
-      scale._updateExtent("1", "x", [-base * 2, base * 2]);
+      providedExtents = [[-base * 2, base * 2]];
+      scale.autoDomain();
```
bluong: Same as above

test/scales/scaleTests.ts
```diff
@@ -499,8 +503,12 @@ describe("Scales", () => {
     });
 
     it("ticks() is always non-empty", () => {
-      [[2, 9], [0, 1], [1, 2], [0.001, 0.01], [-0.1, 0.1], [-3, -2]].forEach((domain) => {
-        scale._updateExtent("1", "x", domain);
+      var desiredExtents: number[][] = [];
+      var provider: Plottable.Scales.ExtentProvider<number> = (scale: Plottable.Scale<number, number>) => desiredExtents;
+      scale.addExtentProvider(provider);
```
bluong: Same as above.

src/components/plots/plot.ts
jtlan: I have to keep a reference to the wrapped function, because reference equality is used to add and remove `ExtentProvider`s (the entire point of this PR).

src/components/plots/plot.ts
jtlan: Adding `_scales()`, which returns all scales currently used in projections.

test/scales/scaleTests.ts
jtlan: Can't inline it here, since I need the reference to be able to remove the `ExtentProvider` later.

bluong:
Random tab character

PR #1957, opened 2015-05-04
Dispatchers no longer use broadcasters

aicioara:
Instead they use the `CallbackSet`

This is tickbox 3 of #1915 

As part of this commit:
- Dispatchers use the callback set instead of the broadcaster
- Dispatcher no longer wrap the callbacks to match the broadcaster signature (mainly because this would ruin reference equality for deletion)
- Removing a callback is done using the `off` prefix (ex. `onKeyDown(callback)` <-> `offKeyDown(callback)`
- Passing null to a `on` function no longer removes the callback (as it would have no idea which callback to remove)
- All tests have been changed to reflect these changes, with minimum change to the levels of abstractions in each unit test
- Updated all comments and added comments for the new methods


aicioara:
Ready for review. Not sure what should be the generic signature of the `callbackSet`. (everywhere). At the moment it is just `Function`, but did not really give it a thought yet.

src/dispatchers/abstractDispatcher.ts
```diff
@@ -3,21 +3,22 @@
...
-      if (!this._connected) {
-        Object.keys(this._event2Callback).forEach((event: string) => {
-          var callback = this._event2Callback[event];
-          document.addEventListener(event, callback);
-        });
-        this._connected = true;
+      if (this._connected) {
+        return;
       }
+      Object.keys(this._event2Callback).forEach((event: string) => {
```
jtlan: Just realized this might be better done with a `for ... in` loop.

jtlan: Never mind, apparently `Object.keys` is vastly superior:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

aicioara:
Merged with apiRefactor. Part of the merge was 
- `callbackSet.remove()` -> `callbackSet.delete()`
- existence of the new `touchDispatcher.onTouchCancel` which had to get away from using broadcasters

src/dispatchers/dispatcher.ts
```diff
@@ -30,21 +31,14 @@ module Plottable {
...
-      if (callback === null) { // remove listener if callback is null
-        b.deregisterListener(key);
-        this._disconnect();
-      } else {
-        this._connect();
-        b.registerListener(key, this._getWrappedCallback(callback));
-      }
+    protected _unsetCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function) {
+      callbackSet.delete(callback);
+      this._disconnect();
```
bluong: Based on our switching of API names, do you think we should change these names?  Not a huge deal.

src/dispatchers/keyDispatcher.ts
```diff
@@ -34,30 +34,34 @@ export module Dispatchers {
...
-      this._keydownBroadcaster = new Core.Broadcaster(this);
-      this._broadcasters = [this._keydownBroadcaster];
+      this._keydownCallbackSet = new Utils.CallbackSet();
+      this._callbackSets = [this._keydownCallbackSet];
     }
 
-    protected _getWrappedCallback(callback: Function): Core.BroadcasterCallback<Dispatchers.Key> {
-      return (d: Dispatchers.Key, e: KeyboardEvent) => callback(e.keyCode, e);
+    /**
+     * Registers a callback to be called whenever a key is pressed,
```
jtlan: Ends with period.

src/dispatchers/keyDispatcher.ts
```diff
@@ -6,7 +6,7 @@ export module Dispatchers {
 
   export class Key extends Dispatcher {
     private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
-    private _keydownBroadcaster: Core.Broadcaster<Dispatchers.Key>;
+    private _keydownCallbackSet: Utils.CallbackSet<Function>;
```
bluong: Does `Utils.CallbackSet<KeyCallback>` work instead?

src/dispatchers/keyDispatcher.ts
```diff
@@ -34,30 +34,34 @@ export module Dispatchers {
...
+     */
+    public onKeyDown(callback: KeyCallback): Key {
+      this._setCallback(this._keydownCallbackSet, callback);
+      return this;
     }
 
     /**
-     * Registers a callback to be called whenever a key is pressed,
-     * or removes the callback if `null` is passed as the callback.
+     * Removes the callback to be called whenever a key is pressed,
```
jtlan: Should end with period.

jtlan:
Thought:
There's a lot of variables named `_eventnameCallbackSet`. But since the typing system already tells us that they're `CallbackSet`s, we could probably just name the variable `_eventnameCallbacks`.

src/dispatchers/mouseDispatcher.ts
```diff
@@ -8,11 +8,12 @@ export module Dispatchers {
...
-    private _downBroadcaster: Core.Broadcaster<Dispatchers.Mouse>;
-    private _upBroadcaster: Core.Broadcaster<Dispatchers.Mouse>;
-    private _wheelBroadcaster: Core.Broadcaster<Dispatchers.Mouse>;
-    private _dblClickBroadcaster: Core.Broadcaster<Dispatchers.Mouse>;
+
+    private _moveCallbackSet: Utils.CallbackSet<Function>;
+    private _downCallbackSet: Utils.CallbackSet<Function>;
+    private _upCallbackSet: Utils.CallbackSet<Function>;
+    private _wheelCallbackSet: Utils.CallbackSet<Function>;
+    private _dblClickCallbackSet: Utils.CallbackSet<Function>;
```
jtlan: We can get more specific here and probably use `<MouseCallback>` for the generic.

src/dispatchers/keyDispatcher.ts
```diff
@@ -34,30 +34,34 @@ export module Dispatchers {
...
-      return (d: Dispatchers.Key, e: KeyboardEvent) => callback(e.keyCode, e);
+    /**
+     * Registers a callback to be called whenever a key is pressed,
+     *
+     * @param {KeyCallback} callback
+     * @return {Dispatcher.Key} The calling Dispatcher.Key.
+     */
+    public onKeyDown(callback: KeyCallback): Key {
+      this._setCallback(this._keydownCallbackSet, callback);
+      return this;
```
bluong: Err... looking back at the old implementation, I'm now not too confident in this switch of implementation.  Can you do a quick check to see if onKeyDown works as intended?

src/dispatchers/mouseDispatcher.ts
bluong: Let's specify these function types?

jtlan:
Minor [-1] on using more specific generics.

test/dispatchers/mouseDispatcherTests.ts
```diff
@@ -48,17 +48,17 @@ describe("Dispatchers", () => {
...
       assert.isTrue(cb1Called, "callback 1 was called on mousemove");
       assert.isTrue(cb2Called, "callback 2 was called on mousemove");
 
       cb1Called = false;
       cb2Called = false;
-      md.onMouseMove("callback1", null);
+      md.offMouseMove(cb1);
       triggerFakeMouseEvent("mousemove", target, targetX, targetY);
-      assert.isFalse(cb1Called, "callback was not called after blanking");
+      // assert.isFalse(cb1Called, "callback was not called after blanking");
```
bluong: Fix comment please

src/dispatchers/mouseDispatcher.ts
```diff
@@ -45,109 +46,151 @@ export module Dispatchers {
...
-     * @param {any} key The key associated with the callback.
-     *                  Key uniqueness is determined by deep equality.
      * @param {MouseCallback} callback A callback that takes the pixel position
      *                                     in svg-coordinate-space.
      *                                     Pass `null` to remove a callback.
      * @return {Dispatcher.Mouse} The calling Dispatcher.Mouse.
      */
-    public onDblClick(key: any, callback: MouseCallback): Dispatchers.Mouse {
-      this._setCallback(this._dblClickBroadcaster, key, callback);
+    public onDblClick(callback: MouseCallback): Dispatchers.Mouse {
```
bluong: Dispatcher endpoints currently reflect the event they are listening to.  Not sure if we should change this endpoint name

aicioara: I agree with @bluong here. The DOM event is called `dblclick` and is set here

```typescript
      this._event2Callback["dblclick"] = (e: MouseEvent) => this._measureAndBroadcast(e, this._dblClickCallbacks);
```

src/dispatchers/keyDispatcher.ts
aicioara: As discussed on a separate thread, this actually works and has unit test attached to it

aicioara:
Waiting for build complete and merging...

PR #1963, opened 2015-05-05
Fix all the overlaying quicktests.

jtlan:
Let's try to keep these functional going forward.


jtlan:
Also, while fixing these, discovered that `DragBoxLayer` is no longer working correctly.

PR #1964, opened 2015-05-05
Scales no longer use broadcasters

aicioara:
Instead they use `callbackSet`.

This is tickbox2 of #1915 

As part of this commit:
- Fixed few comments from dispatchers where broadcasters were still referenced
- Switched to using a `callbackSet` everywhere 
- Reshuffled public and private fields for `scale.ts`
- removed `scaleDomainCoordinator` as it was not used and because `less is more`

Again, I was not sure if the `<Function>` is the right generic type for the callback. Feel free to comment. In particular, you might want to look at `axis.ts:59` vs `plotTests.ts:75`


jtlan:
I was thinking that `onDomainChange()` /  `offDomainChange()` would be better names for the methods, to align with the "onEvent"/"offEvent" naming scheme we use elsewhere.

aicioara:
You mean instead of `registerListener` / `deregisterListener`?

jtlan:
@aicioara wrote:
> You mean instead of `registerListener` / `deregisterListener`?
Yes, I think `onDomainChange()` / `offDomainChange()` might be more appropriate, although `Scales.Category` also broadcasts when the padding changes...

Also, it might be good to have the `Scale` pass itself to the callbacks.

src/components/axes/axis.ts
```diff
@@ -28,6 +28,8 @@ module Plottable {
     private _gutter = 15;
     private _showEndTickLabels = false;
 
+    private _rescaleFunctionWrapper: Function;
```
bluong: Type the function to be more specific

src/components/gridlines.ts
```diff
@@ -8,6 +8,8 @@ export module Components {
     private _xLinesContainer: D3.Selection;
     private _yLinesContainer: D3.Selection;
 
+    private _renderFunctionWrapper: Function;
```
bluong: Type the function to be more specific

src/components/interpolatedColorLegend.ts
```diff
@@ -17,6 +17,8 @@ export module Components {
     private _lowerLabel: D3.Selection;
     private _upperLabel: D3.Selection;
 
+    private _redrawFunctionWrapper: Function;
```
bluong: Type the function to be more specific

src/components/legend.ts
```diff
@@ -25,6 +25,8 @@ export module Components {
     private _writer: SVGTypewriter.Writers.Writer;
     private _symbolFactoryAccessor: (datum: any, index: number) => SymbolFactory;
 
+    private _redrawFunctionWrapper: Function;
```
bluong: Type the function to be more specific

src/components/plots/plot.ts
```diff
@@ -40,6 +40,8 @@ module Plottable {
     protected _animateOnNextRender = true;
     private _nextSeriesIndex: number;
 
+    private _renderFunctionWrapper: Function;
```
bluong: Type the function to be more specific

src/components/plots/xyPlot.ts
```diff
@@ -7,6 +7,9 @@ module Plottable {
     private _autoAdjustXScaleDomain = false;
     private _autoAdjustYScaleDomain = false;
 
+    private _adjustYDomainOnChangeFromXFunctionWrapper: Function;
+    private _adjustXDomainOnChangeFromYFunctionWrapper: Function
```
bluong: Type the function to be more specific

Also, missing semicolon (tslint should've warned against this)

bluong:
Do you think function wrapper is a proper name?

src/dispatchers/mouseDispatcher.ts
```diff
@@ -196,7 +196,7 @@ export module Dispatchers {
 
     /**
      * Computes the mouse position from the given event, and if successful
-     * calls broadcast() on the supplied Broadcaster.
+     * calls all the callbacks in the provided callbackSet.
      */
     private _measureAndBroadcast(event: MouseEvent, callbackSet: Utils.CallbackSet<MouseCallback>) {
```
bluong: Consider renaming this function name

src/dispatchers/touchDispatcher.ts
```diff
@@ -165,7 +165,7 @@ export module Dispatchers {
 
     /**
      * Computes the Touch position from the given event, and if successful
-     * calls broadcast() on the supplied Broadcaster.
+     * calls all the callbacks in the provided callbackSet.
      */
     private _measureAndBroadcast(event: TouchEvent, callbackSet: Utils.CallbackSet<TouchCallback>) {
```
bluong: Consider renaming this function name

src/scales/scale.ts
```diff
@@ -41,6 +42,18 @@ module Plottable {
...
+        this._callbacks.add(callback);
+    }
+
+    public deregisterListener(callback: Function) {
+        this._callbacks.delete(callback);
+    }
+
+    public broadcast() {
+        this._callbacks.callCallbacks(this);
+    }
```
bluong: This API is not really apt for scales as @jtlan has said

src/components/plots/xyPlot.ts
aicioara: I run tslint every time before CR. Not caught.

aicioara:
New update, following the CR:
- The type of the callback is now `ScaleCallback<S extends Scale<any, any>>`
- The api points on the scale are `onUpdate`, `offUpdate`, and `dispatchUpdate`
- `dispatchUpdate` is now protected
- renamed `measureAndBroadcast` to `measureAndDispatch`

src/components/axes/axis.ts
```diff
@@ -54,12 +55,13 @@ module Plottable {
 
       this.formatter(formatter);
 
-      this._scale.broadcaster.registerListener(this, () => this._rescale());
+      this._rescaleCallback = (scale) => this._rescale();
```
bluong: I don't think the scale argument is really that useful here (as for other similar code)

aicioara: ofc not, but it needs to respect the signature now

aicioara: Or does it not? Apparently I can omit the (scale) argument, but that seems weird. Should I remove it where irrelevant?

jtlan:
Wouldn't `update()` be a better name than `dispatchUpdate()`?

aicioara:
I don't see `update()` as being informative enough. I see questions like "update who? the scale? But the listeners receive the update". Seems weird

src/scales/scale.ts
```diff
@@ -1,6 +1,11 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
+
+  export interface ScaleCallback<S extends Scale<any, any>> {
+    (scale: S): any;
```
bluong: Noticed that none of the callbacks actually utilize the scale except for in test code.  We should probably not have the scale be an enforced argument?

test/scales/scaleTests.ts
```diff
@@ -8,12 +8,12 @@ describe("Scales", () => {
...
     };
     var scale = new Plottable.Scales.Linear();
-    scale.broadcaster.registerListener(null, testCallback);
+    scale.onUpdate(testCallback);
     var scaleCopy = scale.copy();
     assert.deepEqual(scale.domain(), scaleCopy.domain(), "Copied scale has the same domain as the original.");
     assert.deepEqual(scale.range(), scaleCopy.range(), "Copied scale has the same range as the original.");
-    assert.notDeepEqual(scale.broadcaster, scaleCopy.broadcaster,
-                              "Broadcasters are not copied over");
+    assert.notDeepEqual((<any>scale)._callbacks, (<any>scaleCopy)._callbacks,
```
bluong: What's the need for this assert?

aicioara: Callback sets are not copied over. Sounds legit. Was here before as well

src/components/axes/axis.ts
bluong: I personally like removing it where irrelevant as that's less to read.

There is an argument to be made that all the arguments should always be provided (less lookup), but I dunno, I value the cleanliness

src/scales/scale.ts
```diff
@@ -41,6 +47,18 @@ module Plottable {
...
+    public onUpdate(callback: ScaleCallback<Scale<D, R>>) {
+        this._callbacks.add(callback);
+    }
+
+    public offUpdate(callback: ScaleCallback<Scale<D, R>>) {
+        this._callbacks.delete(callback);
+    }
+
+    protected _dispatchUpdate() {
+        this._callbacks.callCallbacks(this);
```
bluong: All the above has 4 space indentation

jtlan:
API looks good.

aicioara:
To document the discussions we had:

scale.dispatchUpdate rename
---
`scale.update()` vs `scale.dispatchUpdate()` vs `scale._dispatchUpdate()`. Since this is a protected method, it should be underscored. Also, dispatchUpdate is more informative than update, as for a developer standpoint, using `update` might seem like the something is being updated, while in fact the update is just propagated at this point. The update is in fact a domain change (or padding change at the moment of this conversation)
Decision: call it `scale._dispatchUpdate()`

 `ScaleCallback` being a function which receives 1 argument, a `Scale<D, R>`. 
---
The arguments brought up were
- `when in doubt, leave it out`. We were not sure if we will need that argument passed or not, so we might well just leave it out and add it later. However, this will actually be an API break for the TypeScript consumers, as they will now have to provide generics to the Callback type
- The scale was actually used in the (now) deleted `ScaleDomainCoordinator`. Also some old users might be relying on it as it is the easiest way to figure out who invoked the callback. Users who were relying on this might need to apply nasty workarounds to regain the old behavior.
Decision: keep the scale as the only parameter of the callback

test/scales/scaleTests.ts
jtlan: Not quite right, since we can't deep-equal test `Set`s. Really we want to ensure that the copied scale has no existing callbacks:
```Typescript
assert.lengthOf((<any> scaleCopy)._callbacks.values(), 0, "copied scale has no callbacks attached");
```
Alternatively, force an update on `scaleCopy` and see if the callbacks get triggered.

PR #1965, opened 2015-05-05
Component.origin() returned cloned point.

jtlan:
As before, to avoid accidentally mutating the internal origin.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -3822,7 +3822,10 @@ var Plottable;
          * @return {Point} The x-y position of the Component relative to its parent.
          */
         Component.prototype.origin = function () {
-            return this._origin;
+            return {
+                x: this._origin.x,
+                y: this._origin.y
+            };
         };
         /**
          * Gets the origin of the Component relative to the root <svg>.
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -538,7 +538,10 @@ module Plottable {
      * @return {Point} The x-y position of the Component relative to its parent.
      */
     public origin(): Point {
-      return this._origin;
+      return {
+        x: this._origin.x,
+        y: this._origin.y
+      };
     }
 
     /**
--- a/test/core/componentTests.ts
+++ b/test/core/componentTests.ts
@@ -403,6 +403,18 @@ describe("Component behavior", () => {
   describe("origin methods", () => {
     var cWidth = 100;
     var cHeight = 100;
+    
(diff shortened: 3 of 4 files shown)


bluong:
Do you think a test here is worth it?

aicioara:
I agree with the existence of the test. Every time we fix a bug, we should add a test to capture it, otherwise the test coverage does not evolve.

However, I believe that the test is too specific. It is testing an implementation detail, instead of the bug itself. How about writing a test that reproduces the actual bug fixed here.

jtlan:
@aicioara wrote:
> However, I believe that the test is too specific. It is testing an implementation detail, instead of the bug itself. How about writing a test that reproduces the actual bug fixed here.

I disagree. If I write a test involving `Group`s, `Interaction`s, and `Component`s, there's so much going on that it's not really a unit test any more. The core of the problem is that the origin of a `Component` should not be modified by any changes made to the result of the `origin()` call. Would this be better?

```Typescript
var originCall1 = component.origin();
var originalX = originCall1.x;
originCall1.x += 10;
assert.strictEqual(originCall1.x, originalX + 10, "can modify returned point");
assert.strictEqual(originalX, component.origin().x, "modifications to returned point don't affect actual origin");
```

?

bluong:
I think the test should more test about the security of the original object.

In essence, if I do
```typescript
var origin = component.origin();
origin.x = 10;
```

This does not necessarily mean that component.origin().x is 10.

jtlan:
@bluong wrote:
>I think the test should more test about the security of the original object.

>In essence, if I do
>```typescript
var origin = component.origin();
origin.x = 10;
```

>This does not necessarily mean that component.origin().x is 10.

So, you are agreeing with me?

**Comment by bluong** [comment]
I agree with your test case except for this part
```typescript
assert.strictEqual(originCall1.x, originalX + 10, "can modify returned point");
```

I don't think this statement really means anything.

And I would modify this statement here
```typescript
assert.strictEqual(originalX, component.origin().x, "modifications to returned point don't affect actual origin")
```

to either one of the following
```typescript
assert.strictEqual(originCall1.x, component.origin().x + 10, "modifications to returned point don't affect actual origin")
```

or 

```typescript
assert.notEqual(originCall1.x, component.origin().x, "modifications to returned point don't affect actual origin")
```

aicioara:
You were actually right, it's more about semantics. Now I actually understood what was being tested.

PR #1966, opened 2015-05-05
[Plots.Pie] - Replacing project calls with separate accessor API

bluong:
Before the usage to set the value accessor, inner radius accessor, and the outer radius accessor was, respectively:

```typescript
piePlot.project("value", "dataKey");
piePlot.project("inner-radius", 10);
piePlot.project("outer-radius", 20);
```

This should really be split out into accessors like the following..

```typescript
piePlot.sectorValue((d) => d.value);
piePlot.innerRadius(10);
piePlot.outerRadius(20);
```

In accordance, plots should have the ideas of scales as that is what drives Plottable.  In accordance, there could be this usage:

```typescript
var linearScale = new Plottable.Scales.Linear();
piePlot.outerRadius((d) => d.value, linearScale);
```

Because of the above setter endpoints, the getters are currently of this form

```typescript
// Returns an object where
// the `accessor` attribute returns the accessor of the innerRadius
// and where `scale` returns the scale that the value is backing
// (this parameter is optional for when the value isn't backing a scale)
piePlot.innerRadius();
```

Note 1: The names of the API are up for debate.  I personally am probably more in line of something like
```typescript
piePlot.sectorValue((d) => d.value);
```
See that the word accessor has been removed.  This also semantically allows for something like...
```typescript
piePlot.innerRadius(0);
```

EDIT: We should go with the simpler idea of removing the accessor suffix.

Note 2: For reviewers, note that basically the corresponding Drawer has gotten a tiny bit more messy but at the same time a lot of responsibility has been lifted from it.  I am unsure of the architecture of Drawers as of the moment but I am hopefully removing pieces from it as I grunge it up a bit to get our API into a steady state.


jtlan:
I would prefer not having "accessor" in the names.

src/components/plots/piePlot.ts
```diff
@@ -15,6 +15,9 @@ export module Plots {
   export class Pie extends Plot {
 
     private _colorScale: Scales.Color;
+    private _innerRadiusAccessor: _Accessor;
+    private _outerRadiusAccessor: _Accessor;
+    private _valueAccessor: _Accessor;
```
jtlan: Shouldn't we use some sort of `Map` for the accessors?

src/components/plots/piePlot.ts
```diff
@@ -24,6 +27,8 @@ export module Plots {
     constructor() {
       super();
       this._colorScale = new Scales.Color();
+      this._innerRadiusAccessor = () => 0;
+      this._outerRadiusAccessor = () => Math.min(this.width(), this.height()) / 2;
```
jtlan: We should use the actual `propertyAccessor()` calls here, to avoid inconsistencies.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +70,39 @@ export module Plots {
 
       return allPlotData;
     }
+
+    public valueAccessor(): _Accessor;
+    public valueAccessor(valueAccessor: _Accessor): Plots.Pie;
```
jtlan: Should take a scale.

jtlan:
I think the `propertyAccessor()` methods should take a `Scale`. This makes their semantics line up better with `project()`/`attr()`:

```Typescript
public propertyAccessor(accessor: _Accessor, scale?: Scale);
```

Not sure what you get back if you invoke it with no args though. Maybe have this?
```Typescript
export type AccessorScaleBinding {
  accessor: _Accessor;
  scale?: Scale;
}
```

src/components/plots/piePlot.ts
bluong: What would the key be to the map?

src/components/plots/piePlot.ts
bluong: The setter you mean?

bluong:
The thing I was mainly concerned about having a scale arg is about retrieving the scale.

@jtlan Your proposal sounds like it would work... the main thing I'm concerned about is that users will need to take a second step in order to get to either pieces of information.

Another important part is that what are the ranges of these scales?  For the inner, outer radius I guess I can set it to min(componentWidth, componentHeight)/2  What about value?

src/components/plots/piePlot.ts
```diff
@@ -2,6 +2,12 @@
 
 module Plottable {
 export module Plots {
+
+  export interface PieValueBinding<D> {
```
jtlan: Can you use whatever the generalized name will be for this, so we can see what the final API might look like on other `Plot`s?

src/components/plots/piePlot.ts
```diff
@@ -67,6 +77,73 @@ export module Plots {
...
+      if (innerRadiusScale == null) {
+        return this._innerRadiusScale;
+      }
+      Pie._replaceScaleBinding(this._innerRadiusScale, innerRadiusScale, this._renderCallback);
+      this._innerRadiusScale = innerRadiusScale;
+      this._render();
+      return this;
+    }
+
+    public innerRadiusScaledAccessor(): _Accessor {
```
jtlan: This is quite a mouthful. What's the point of this call?

src/components/plots/piePlot.ts
```diff
@@ -67,6 +77,73 @@ export module Plots {
 
       return allPlotData;
     }
+
+    public valueAccessor(): _Accessor;
+    public valueAccessor(valueAccessor: _Accessor): Plots.Pie<D>;
+    public valueAccessor(valueAccessor?: _Accessor): any {
```
jtlan: Why "`valueAccessor()`" here, but "`innerRadius()`" later?

jtlan:
I see some inconsistencies in the code up for review around whether or not "accessor" is in the method names.

It also feels clunky to have to set the accessor and the `Scale` independently. While it's nice to be able to control them separately, I think the clunkiness would be more obvious on a class other than `Pie`:

```Typescript
var linePlot = new Plottable.Plots.Line();
linePlot.x((d) => d.x).xScale(xScale).y((d) => d.y).yScale(yScale);
```

jtlan:
Also, I think the signature for `_Accessor` should change to this:
```
export type Accessor = (datum: any, index: number, dataset: Dataset) => any;
```

Also, maybe `datum` should be an `Object`.

bluong:
I think datum's should indeed be `Object`s.  At the very least they shouldn't be `Function`s which can be passed in.

I think _Accessor` should actually be an interface like the following...

```typescript
export interface Accessor<V> {
  (datum: Object, index: number, dataset: Dataset) => V;
}
```

jtlan:
@bluong wrote:
>I think datum's should indeed be `Object`s. At the very least they shouldn't be `Function`s which can be passed in.

The question is whether we want to allow `number`s or `String`s.

src/components/plots/piePlot.ts
```diff
@@ -2,6 +2,12 @@
 
 module Plottable {
 export module Plots {
+
+  export interface ValueScaleBinding<V, D, R> {
+    value: V | _Accessor | D;
```
jtlan: Shouldn't the value be either `R` or `D`? `V` isn't used anywhere else.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +79,82 @@ export module Plots {
...
+
+    public scaledInnerRadiusAccessor(): _Accessor {
+      return Pie._scaledValueAccessor(this._innerRadius, this._innerRadiusScale);
+    }
+
+    public scaledOuterRadiusAccessor(): _Accessor {
+      return Pie._scaledValueAccessor(this._outerRadius, this._outerRadiusScale);
+    }
+
+    public scaledSectorValueAccessor(): _Accessor {
```
jtlan: Why are this and the above methods public?

jtlan:
I don't really see the point of the methods that provide "scaled value accessors", since if the user were to query the binding they could compute that themselves.

I'm also unsure of the wisdom of adding getters at this time. While setters should have getters, we're also pretty low on time to finish this refactor, and the getter could be added later. We also haven't gotten any requests for the getter functionality.

src/components/plots/piePlot.ts
bluong: Observe the usage on drawers.  This is essentially the best way for anyone to have access to the actual value that will be piped into the drawn plot

src/components/plots/piePlot.ts
bluong: For `innerRadius` and `outerRadius` you want your `value`s to be `number`s or `Accessor`s to a number.  Thus, how would you encapsulate that into a type?

bluong:
Note that we need some sort of `getter`s or else the appropriate `Drawer` has no way of accessing those important accessors

jtlan:
Makes sense. But the "scaled accessor" convenience methods seem extraneous when the user could easily assemble it themselves.

bluong:
One thing that is also important to note is that the scales attached currently have no range attached to them.  Should we attach a range to these scales?

bluong:
> Makes sense. But the "scaled accessor" convenience methods seem extraneous when the user could easily assemble it themselves.

It's not that easy.  Note the implementation of the "scaled accessor" methods.  The alternative is that I copy + paste these branching if statement throughout the drawers which seems really clunky

jtlan:
What if the value was always an function?

src/components/plots/piePlot.ts
```diff
@@ -2,6 +2,12 @@
 
 module Plottable {
 export module Plots {
+
+  export interface ValueScaleBinding<D, R> {
+    value: _Accessor;
```
jtlan: Probably should call this an `AccessorScaleBinding` then?

src/components/plots/piePlot.ts
jtlan: That doesn't seem correct to me. `Drawer`s are essentially an implementation detail, so we're adding public API points for no good reason except that some internal user needs them. The logic required to generate the `_scaledValueAccessor` isn't that complex either.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +79,82 @@ export module Plots {
...
+
+    public scaledOuterRadiusAccessor(): _Accessor {
+      return Pie._scaledValueAccessor(this._outerRadius, this._outerRadiusScale);
+    }
+
+    public scaledSectorValueAccessor(): _Accessor {
+      return Pie._scaledValueAccessor(this._sectorValue, this._sectorValueScale);
+    }
+
+    private static _scaledValueAccessor<V, SD, SR>(value: _Accessor, scale: Scale<SD, SR>): _Accessor {
```
jtlan: Is the `V` still needed?

src/components/plots/piePlot.ts
```diff
@@ -67,6 +79,82 @@ export module Plots {
...
+      return scale == null ?
+               value :
+               (d: any, i: number, u: any, m: Plots.PlotMetadata) => scale.scale(value(d, i, u, m));
+    }
+
+    private static _constructBinding<V, D, R>(value: V | _Accessor, scale: Scale<D, number>) {
+      return scale == null ? { accessor: value } : { accessor: value, scale: scale };
```
jtlan: Why is this necessary? It's seems like there'd be less misdirection if we just wrote
```Typescript
binding = {
  accessor: accessor,
  scale: scale
};
```
every time we needed it. If `scale` is `null` or `undefined` it still works correctly.

bluong: Oh I was nervous on sending scale through.  I'll adjust accordingly

src/components/plots/piePlot.ts
bluong: Based on this, no.  Adjusting.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +79,82 @@ export module Plots {
 
       return allPlotData;
     }
+
+    public sectorValue(): AccessorScaleBinding<D, number>;
+    public sectorValue(sectorValue: number | _Accessor): Plots.Pie<D>;
```
jtlan: This suggests that maybe `_Accessor` should use a generic, so you can specify that it returns number:
```Typescript
public sectorValue(sectorValue: number | _Accessor<number>): Plots.Pie<D>;
```

bluong: Yup, I'm planning on making `_Accessor`s generic.  That'll add a bit more generics to this file, but it'll make things a lot more precise.

jtlan:
I think it's weird to have `_scaledAccessor` and the associated public methods on the `Plot`. Let's move the weirdness to the `Drawer`.

src/components/plots/piePlot.ts
```diff
@@ -24,12 +39,33 @@ export module Plots {
     constructor() {
       super();
       this._colorScale = new Scales.Color();
+      this._innerRadius = { accessor: () => 0 };
```
jtlan: Why not just use the setter?

src/components/plots/piePlot.ts
```diff
@@ -12,9 +18,18 @@ export module Plots {
...
     private _colorScale: Scales.Color;
+    private _innerRadius: AccessorScaleBinding<D, number>;
+    private _innerRadiusExtents: any[];
+    private _innerRadiusExtentProvider: Scales.ExtentProvider<any>;
+    private _outerRadius: AccessorScaleBinding<D, number>;
+    private _outerRadiusExtents: any[];
+    private _outerRadiusExtentProvider: Scales.ExtentProvider<any>;
+    private _sectorValue: AccessorScaleBinding<D, number>;
+    private _sectorValueExtents: any[];
+    private _sectorValueExtentProvider: Scales.ExtentProvider<any>;
```
jtlan: It might be better to use some sort of map to the extents, rather than having tons of new variables.

jtlan:
Take a look at **plot.ts** and see how the accessor-`Scale` bindings (projections) are set there. It's way easier to use a dictionary or `D3.Map` instead of carrying around tons of private variables.

src/components/plots/piePlot.ts
```diff
@@ -12,9 +18,15 @@ export module Plots {
...
    *   "value" - Accessor to extract the value determining the proportion of each slice to the total
    */
-  export class Pie extends Plot {
+  export class Pie<D> extends Plot {
 
     private _colorScale: Scales.Color;
+    private static _INNER_RADIUS_KEY = "inner-radius";
+    private static _OUTER_RADIUS_KEY = "outer-radius";
+    private static _SECTOR_VALUE_KEY = "sector-value";
+    private _key2DataExtents: D3.Map<any[]>;
```
jtlan: Since this is mirroring the `_projections` objects, we should make it clear:
```
private _propertyExtents: D3.Map<any []>;
private _attrExtents: D3.Map<any []>; // for DOM attributes
```

src/components/plots/piePlot.ts
```diff
@@ -12,9 +18,15 @@ export module Plots {
...
+    private _key2DataBindings: D3.Map<AccessorScaleBinding<any, any>>;
```
jtlan: and this one would be
```Typescript
private _propertyBindings: D3.Map<AccessorScaleBinding<any, any>;
```

src/components/plots/piePlot.ts
```diff
@@ -67,6 +91,102 @@ export module Plots {
...
+      this._render();
+      return this;
+    }
+
+    protected _updateExtents() {
+      super._updateExtents();
+      this._key2DataExtents.keys().forEach((dataAttr) => this._updateDataScaleExtents(dataAttr));
+    }
+
+    private _dataExtentsForScale<D>(scale: Scale<D, any>): D[][] {
```
jtlan: Going the the parallels drawn above, we could just override `_extentsForScale()` from **plot.ts** to apply to both `_projections` and `_propertyBindings`, putting the extents into each appropriate array.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +91,102 @@ export module Plots {
...
+          var extents = this._key2DataExtents.get(dataAttr);
+          if (extents != null) {
+            allSetsOfExtents.push(extents);
+          }
+        }
+      });
+      return d3.merge(allSetsOfExtents);
+    }
+
+    private _updateDataScaleExtents(dataAttr: string) {
```
jtlan: By parallels with `_extentsForAttr()`, it would be `_extentsForProperty()`.

src/components/plots/piePlot.ts
```diff
@@ -67,6 +91,102 @@ export module Plots {
...
+      this._key2DataExtents.set(dataAttr, this._datasetExtents(dataAttr));
+      var dataScale = this._key2DataBindings.get(dataAttr).scale;
```
jtlan: This line needs to combine with the below `_datasetExtents()` method to form `_updateExtentsForProperty()`, the parallel to `_updateExtentsForAttr()`.

jtlan:
[-1] Try for structural parallels to how `attr`s are handled in **plot.ts**.

jtlan:
API and implementation look solid. Fix the overlaying tests?

PR #1967, opened 2015-05-05
TSLint breaks

aicioara:
Just an experiment...

PR #1968, opened 2015-05-05
Render refactor

jtlan:
Stage 1: renaming `_render()` to `render()`.

Next up: making `_doRender()` `protected` or `private`.


src/core/renderController.ts
```diff
@@ -22,8 +22,8 @@ export module Core {
    * ```
    */
   export module RenderControllers {
-    var _componentsNeedingRender: {[key: string]: Component} = {};
-    var _componentsNeedingComputeLayout: {[key: string]: Component} = {};
+    var _componentsNeedingRender = new Utils.Set<Component>();
+    var _componentsNeedingComputeLayout = new Utils.Set<Component>();
```
aicioara: Not part of this CR, but I thought we decided (or was brought up during the API refactor) that we instantiate all instance variables in the constructor as opposed to the declaration. Are we still iterating on that idea?

src/core/renderController.ts
```diff
@@ -91,34 +91,23 @@ export module Core {
     export function flush() {
       if (_animationRequested) {
         // Layout
-        var toCompute = d3.values(_componentsNeedingComputeLayout);
-        toCompute.forEach((c) => c.computeLayout());
+        _componentsNeedingComputeLayout.values().forEach((c: Component) => c.computeLayout());
```
aicioara: I really don't like single character variable names. We should probably move away from them

src/core/renderController.ts
```diff
@@ -91,34 +91,23 @@ export module Core {
...
 
-        // Top level render.
-        // Containers will put their children in the toRender queue
-        var toRender = d3.values(_componentsNeedingRender);
-        toRender.forEach((c) => c._render());
+        // Top level render; Containers will put their children in the toRender queue
+        _componentsNeedingRender.values().forEach((c: Component) => c.render());
```
aicioara: same

src/core/renderController.ts
```diff
@@ -91,34 +91,23 @@ export module Core {
...
         _isCurrentlyFlushing = true;
-
-        // Finally, perform render of all components
-        var failed: {[key: string]: Component} = {};
-        Object.keys(_componentsNeedingRender).forEach((k) => {
+        var failed = new Utils.Set<Component>();
+        _componentsNeedingRender.values().forEach((c: Component) => {
           try {
-            _componentsNeedingRender[k]._doRender();
+            c._doRender();
```
aicioara: same

src/core/renderController.ts
```diff
@@ -91,34 +91,23 @@ export module Core {
...
           } catch (err) {
-            // using setTimeout instead of console.log, we get the familiar red
-            // stack trace
-            setTimeout(() => {
-              throw err;
-            }, 0);
-            failed[k] = _componentsNeedingRender[k];
+            // throw error with timeout to avoid interrupting further renders
+            window.setTimeout(() => { throw err; }, 0);
```
aicioara: no need for `{` `}`

aicioara: Does this pass tslint? I thought we enforce newline after `{`. I prefer the old way.

aicioara:
Minor style things. Mainly personal preference.

src/core/renderController.ts
jtlan: This isn't a `class`, so there's no `constructor`. The `module` just happens to have some variables in it.

src/core/renderController.ts
jtlan: Okay, I'll make the change.

src/core/renderController.ts
aicioara: Fair enough

src/core/renderController.ts
bluong: If you're worried about character constraints, feel free to remove the typing at the function declaration stage.  We should probably do this moving forward

aicioara:
Also, to save you some time with the build, you should also cherry-pick my first commit from here https://github.com/palantir/plottable/pull/1967

aicioara:
@jtlan you need another merge

aicioara:
Once the build passes

PR #1971, opened 2015-05-06
Removed Broadcasters

aicioara:
This is tickbox 4 and 5 of #1915 

As part of this commit
- Datasets no longer use broadcasters, Instead they use `callbackSet`.
- added test to `Dataset.offUpdate()`

Very similar to #1964


test/components/plots/plotTests.ts
```diff
@@ -36,21 +36,21 @@ describe("Plots", () => {
 
       assert.strictEqual(1, r.renders, "initial render due to addDataset");
 
-      dFoo.broadcaster.broadcast();
+      (<any>dFoo)._dispatchChange();
```
jtlan: Since `_dispatchChange()` isn't public, the better way to trigger an update might be something like
```Typescript
dFoo.data(dFoo.data());
```

src/core/dataset.ts
```diff
@@ -26,7 +29,19 @@ module Plottable {
...
+    public onUpdate(callback: DatasetCallback) {
+      this._callbacks.add(callback);
+    }
+
+    public offUpdate(callback: DatasetCallback) {
+      this._callbacks.delete(callback);
+    }
+
+    private _dispatchChange() {
+      this._callbacks.callCallbacks(this);
```
jtlan: This is a one-line private function thats only used twice. Should we maybe jus do `this._callbacks.callCallbacks(this)` whenever we want to dispatch a change?

jtlan:
Questions around style.

test/components/plots/plotTests.ts
aicioara: Looks like a hack. It's not like we are not accessing private fields anyway

bluong: Maybe a better solution is to actually change the data around manually.

```typescript
dFoo.data(dFoo.data().concat({foo: "foo"}))
```

aicioara: Oh, is this for the line below? okay

bluong: @aicioara You are correct in that we do use private fields in our test code and I do think that we should not disallow ourselves from doing so (There are valid arguments on both ends).  However, I think in this case, we can trigger our changes in a very direct way (use the `data` setter) and that if we can use the public API, then we should

PR #1972, opened 2015-05-06
Plot computes extents instead of Dataset.

jtlan:
`_getExtent()` and `_computeExtent()` removed from `Dataset`.

### DIFF
--- a/test/core/metadataTests.ts
+++ b/test/core/metadataTests.ts
@@ -188,27 +188,6 @@ describe("Metadata", () => {
     svg.remove();
   });
 
-  it("_getExtent works as expected with plot metadata", () => {
-    var svg = TestMethods.generateSVG(400, 400);
-    var metadata = {foo: 11};
-    var id = (d: any) => d;
-    var dataset = new Plottable.Dataset(data1, metadata);
-    var a = (d: any, i: number, u: any, m: any) => d.x + u.foo + m.foo;
-    var plot = new Plottable.Plot().project("a", a, xScale);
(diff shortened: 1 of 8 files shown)


aicioara:
Will do in the morning

bluong:
It looks like functionality is mostly kept except there is no caching.  Will there be any caching in this PR?

jtlan:
Cacheing is already there: the `_attrToExtents` field caches the extent for a given attribute and is only updated with `_updateExtentsForAttr()`.

test/core/metadataTests.ts
```diff
@@ -188,27 +188,6 @@ describe("Metadata", () => {
...
-      };
-    };
-    plot.addDataset(dataset);
-    plot.renderTo(svg);
-    assert.deepEqual(dataset._getExtent(a, id), [16, 17], "plot metadata is reflected in extent results");
-    dataset.metadata({foo: 0});
-    assert.deepEqual(dataset._getExtent(a, id), [5, 6], "plot metadata is reflected in extent results after change user metadata");
-    svg.remove();
-  });
-
```
aicioara: Shouldn't this functionality be now tested in `plotTests.ts`? in particular the `plot._computeExtent()` method

jtlan: `_computeExtent()` is private. There are a number of indirect tests of it by adding test data and seeing the resultant domains of the scales, which I think is better.

aicioara: Fair enough

PR #1973, opened 2015-05-06
RenderController cleanup

jtlan:
Removed an extra loop (commit 1).
Moved it up a level (commit 2):
`Plottable.Core.RenderControllers` --> `Plottable.RenderController` (singleton)
`Plottable.Core.RenderControllers.RenderPolicies` --> `Plottable.RenderPolicies`

### DIFF
--- a/src/core/renderController.ts
+++ b/src/core/renderController.ts
@@ -93,9 +93,6 @@ export module Core {
         // Layout
         _componentsNeedingComputeLayout.values().forEach((component: Component) => component.computeLayout());
 
-        // Top level render; Containers will put their children in the toRender queue
-        _componentsNeedingRender.values().forEach((component: Component) => component.render());
-
         _isCurrentlyFlushing = true;
         var failed = new Utils.Set<Component>();
         _componentsNeedingRender.values().forEach((component: Component) => {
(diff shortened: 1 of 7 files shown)


src/core/renderController.ts
```diff
@@ -93,9 +91,6 @@ export module Core {
         // Layout
         _componentsNeedingComputeLayout.values().forEach((component: Component) => component.computeLayout());
 
-        // Top level render; Containers will put their children in the toRender queue
-        _componentsNeedingRender.values().forEach((component: Component) => component.render());
```
bluong: Why was this line remove?  This seems like a very important line?

jtlan: Explained already:

Calling `render()` only enqueues the `Component` for rendering, so it can be done synchronously. Calling `render()` on a `ComponentContainer` immediately enqueues all its children, so looping over `_componentsNeedingRender` and forcing them to re-enqueue does nothing.

jtlan: Turns out it's actually really important for when a child of a `ComponentContainer` gets updated and calls `redraw()`. Digging into *why* now...

PR #1975, opened 2015-05-06
Created consistency on events by using onEvent and offEvent APIs

aicioara:
Point 6 of #1915 

As part of this PR
- Renamed `interactionTests` to `keyInteractionTests` because that is what was tested in that file
- We can now have multiple callbacks attached to one particular event of an interaction
- The callbacks are now removed using the `off` keyword (ex `ofDrag()`), instead of passing null
- The callbacks can no longer be retrieved because we don't want to expose our `CallbackSet`
- touched the `dragBoxLayer` as well, since it was using on/off drag interactions
- added tests to cover the fact that we can now have multiple callbacks attached to an event
- modified tests to no longer be able to retrieve the callback
- typed the callback signatures (ex `ClickCallback = (p: Point) => any;`

Disclaimer: Since this was a pretty long PR, there might be some inconsistencies in the way I structured tests or commented on stuff. Please signal any if you see.


test/components/dragBoxLayerTests.ts
```diff
@@ -162,22 +174,95 @@ describe("Interactive Components", () => {
...
+
+      var target = dbl.background();
+      TestMethods.triggerFakeDragSequence(target, startPoint, endPoint);
+
+      assert.isTrue(callbackDragStart1Called, "the callback 1 for drag start was called");
+      assert.isTrue(callbackDragStart2Called, "the callback 2 for drag start was called");
+      assert.isTrue(callbackDrag1Called, "the callback 1 for drag was called");
+      assert.isTrue(callbackDrag2Called, "the callback 2 for drag was called");
+      assert.isTrue(callbackDragEnd1Called, "the callback 1 for drag end was called");
+      assert.isTrue(callbackDragEnd2Called, "the callback 2 for drag end was called");
```
bluong: Can we move these assert statements so that they are right after their corresponding action?

Like
```typescript
dbl.onDragStart...
assert.isTrue(callbackDragStart1Called...
...
```

test/interactions/dragInteractionTests.ts
```diff
@@ -158,9 +172,87 @@ describe("Interactions", () => {
...
+      TestMethods.triggerFakeMouseEvent("mouseup", target, endPoint.x, endPoint.y);
+
+      assert.isFalse(startCallback1Called, "callback 1 was disconnected from drag start interaction");
+      assert.isTrue(startCallback2Called, "callback 2 is still connected to the drag start interaction");
+
+      assert.isFalse(moveCallback1Called, "callback 1 was disconnected from drag interaction");
+      assert.isTrue(moveCallback2Called, "callback 2 is still connected to the drag interaction");
+
+      assert.isFalse(endCallback1Called, "callback 1 was disconnected from the drag end interaction");
+      assert.isTrue(endCallback2Called, "callback 2 is still connected to the drag end interaction");
```
bluong: Same issues as above

bluong:
Mainly an issue on the design of the tests from an ordering standpoint

test/components/dragBoxLayerTests.ts
aicioara: Their corresponding dragging action is line 235. I don't really want to invoke it 3 times

test/interactions/dragInteractionTests.ts
aicioara: addressed

aicioara:
Addressed one of @bluong tests and improved comment consistency

aicioara:
Waiting for build to pass, then merging

PR #1976, opened 2015-05-06
[apiRefactor] _doRender/render renames

rcchen:
In `component.ts`:
* `public _doRender()` -> `protected _render()`
* `render()` -> `render(immediate: boolean)`

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -163,7 +163,16 @@ module Plottable {
       };
     }
 
-    public render() {
+    /**
+     * Queues the Component for rendering.
+     * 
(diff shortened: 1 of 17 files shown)


src/components/component.ts
```diff
@@ -163,7 +163,16 @@ module Plottable {
       };
     }
 
-    public render() {
+    /**
+     * Queues the Component for rendering.
+     * 
+     * @param immediate boolean Immediately renders the Component.
```
jtlan: "immediately" might be a better name for the param, and we probably don't need a description in that case.

Also need a @return

jtlan:
[minor] docs

PR #1979, opened 2015-05-06
_requestedSpace -> requestedSpace

rcchen:
Based in `component.ts` where `_requestedSpace` -> `requestedSpace`


bluong:
Easy one

PR #1981, opened 2015-05-06
Refactor Domainer to no longer use ids

aicioara:
Task 8 of #1915 

As part of this PR:
- Using `StrictEqualityAssociativeArray` (soon `Plottable.Map`) to keep track of the mapping between components and `exception`/`value`
- Using reference equality to register padding exceptions and included values on the domainers
- the unregistered paddingExceptions and includedValues were not used anywhere (except in testing), hence removed them
- As a result of the above, simplified the definitions of `addPaddingException`, `removePaddingException`, `addIncludedValue`, `removeIncludedValue`
- As a result of the above, `key` and `value` are passed in this order to the function
- Amended comments


jtlan:
I'd like to remove `Domainer`s later, but this looks good since we don't need the IDs anymore.

PR #1983, opened 2015-05-06
Renamed StrictEqualityAssociativeArray to Map

aicioara:


jtlan:
Not strictly the same as the ES6 map, but I think it's the best we can do unless we want to take a shimming library (the actual spec might also not be possible in ES5).

jtlan:
Can we make this take generics?
```Typescript
Map<K, V>
```

bluong:
Just in general if we want to rework a class or so, having `any` types in the final version is generally a bad smell.  Even for our general `data` arrays, we might actually type those things as `Object`s.

Be on the lookout for `any`s and get rid of them if you can.

aicioara:
Solved. With this, I realized that our map is not actually a map. For example for `paddingException`, it should work both numbers and dates

src/core/dataset.ts
```diff
@@ -11,7 +11,7 @@ module Plottable {
   export class Dataset extends Core.PlottableObject {
     private _data: any[];
     private _metadata: any;
-    private _accessor2cachedExtent: Utils.StrictEqualityAssociativeArray;
+    private _accessor2cachedExtent: Utils.Map<any, any>;
```
jtlan: Actually, this variable is no longer used.

src/core/domainer.ts
```diff
@@ -6,8 +6,8 @@ module Plottable {
     private _doNice = false;
     private _niceCount: number;
     private _padProportion = 0.0;
-    private _paddingExceptions = new Utils.StrictEqualityAssociativeArray();
-    private _includedValues = new Utils.StrictEqualityAssociativeArray();
+    private _paddingExceptions: Utils.Map<any, any>;
+    private _includedValues: Utils.Map<any, any>;
```
jtlan: `Domainer` as a whole probably needs to take a generic (the `<D>` on a `Scale<D, R>`), but we can put that aside for now.

PR #1987, opened 2015-05-07
Use a D3.Map for the attribute-projector bindings.

jtlan:
Based on discussions in #1966, we're using a parallel structure for attributes and properties.

Signature of `_Projection` should be changed to match `_AccessorScaleBinding` from #1966.

Also, added **d3.d.ts** to **reference.ts** so Typescript tooling can actually resolve D3 methods.

### DIFF
--- a/src/components/plots/plot.ts
+++ b/src/components/plots/plot.ts
@@ -32,7 +32,7 @@ module Plottable {
 
     protected _renderArea: D3.Selection;
     protected _projections: { [attrToSet: string]: _Projection; } = {};
-    protected _attrToExtents: D3.Map<any[]>;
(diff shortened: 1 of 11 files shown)


src/components/plots/plot.ts
```diff
@@ -265,13 +265,14 @@ module Plottable {
      * Updates the extents associated with each attribute, then autodomains all scales the Plot uses.
      */
     protected _updateExtents() {
-      Object.keys(this._projections).forEach((attr: string) => { this._updateExtentsForAttr(attr); });
+      this._attrBindings.forEach((attr) => { this._updateExtentsForAttr(attr); });
```
bluong: curly branches in the forEach can be removed as well as the semicolon

PR #1989, opened 2015-05-07
ComponentContainer should only enqueue children if not rendering immediately

jtlan:


bluong:
I'll look over this if I have time, but I think @rcchen has more context on this matter.

jtlan:
Thinking this isn't quite the right way to do this now.

jtlan:
Better fix in the works.

PR #1991, opened 2015-05-07
Upgrade to plot

bluong:
Now that the prototyping of the new property API seems to have passed review in `Plots.Pie`, let's upgrade this to the parent level so that the other plots can derive this architecture


src/components/plots/plot.ts
```diff
@@ -267,6 +278,8 @@ module Plottable {
     protected _updateExtents() {
       Object.keys(this._projections).forEach((attr: string) => { this._updateExtentsForAttr(attr); });
       this._scales().forEach((scale) => scale._autoDomainIfAutomaticMode());
+      this._propertyExtents.forEach((property) => this._updateExtentsForProperty(property));
+      this._propertyScales().forEach((scale) => scale._autoDomainIfAutomaticMode());
```
jtlan: This seems silly. `_scales()` should just return all `Scale`s, including those associated with properties.

Splitting them also causes a problem where if a scale is used for both a property and an attribute you'll get two `autodomain()` calls, the first of which is wrong.

src/components/plots/plot.ts
```diff
@@ -540,5 +563,44 @@ module Plottable {
       return !(pixelPoint.x < 0 || pixelPoint.y < 0 ||
         pixelPoint.x > this.width() || pixelPoint.y > this.height());
     }
+
+    protected _updateExtentsForProperty(property: string) {
```
jtlan: Maybe we can merge this with `_updateExtentsForAttr()`:
```Typescript
protected _updateExtentsFor(key: string, attrOrPropertyBinding: D3.Map<Plots.AccessorScaleBinding<any, any>>)
```

src/components/plots/plot.ts
```diff
@@ -540,5 +563,44 @@ module Plottable {
...
+      var accScaleBinding = this._propertyBindings.get(property);
+      if (accScaleBinding.accessor == null) { return; }
```
jtlan: Is this a worthwhile safety check? It seems like in this case, instead of returning immediately, the method should set the extents to `[]` -- otherwise `_propertyExtents` will contain outdated information.

Also, I'm surprised TSLint didn't yell about the single-line `if`.

src/components/plots/plot.ts
```diff
@@ -540,5 +563,44 @@ module Plottable {
...
+      var coercer = (accScaleBinding.scale != null) ? accScaleBinding.scale._typeCoercer : (d: any) => d;
+      this._propertyExtents.set(property, this._datasetKeysInOrder.map((key) => {
+        var plotDatasetKey = this._key2PlotDatasetKey.get(key);
+        var dataset = plotDatasetKey.dataset;
+        var plotMetadata = plotDatasetKey.plotMetadata;
+        return this._computeExtent(dataset, accScaleBinding.accessor, coercer, plotMetadata);
+      }));
+    }
+
+    protected _replacePropertyScale(oldScale: Scale<any, any>, newScale: Scale<any, any>) {
```
jtlan: Is there a way to generalize this to work on the attributes too?

src/components/plots/plot.ts
```diff
@@ -540,5 +563,44 @@ module Plottable {
...
+
+        if (newScale != null) {
+          newScale.onUpdate(this._renderCallback);
+          newScale.addExtentProvider(this._extentProvider);
+          newScale._autoDomainIfAutomaticMode();
+        }
+      }
+    }
+
+    private _propertyScales() {
```
jtlan: Combine with `_scales()`, as mentioned above.

jtlan:
[-1] Combine with the existing objects for attributes, when possible.

src/components/plots/plot.ts
bluong: This is worthwhile since there could be accessors that are not necessarily defined before this is called.

I believe for example this is called in `addDataset` but the property may not be set before that call

src/components/plots/plot.ts
```diff
@@ -266,22 +276,23 @@ module Plottable {
...
       this._scales().forEach((scale) => scale._autoDomainIfAutomaticMode());
     }
 
-    private _updateExtentsForAttr(attr: string) {
-      var binding = this._attrBindings.get(attr);
-      var accessor = binding.accessor;
-      var scale = binding.scale;
-      var coercer = (scale != null) ? scale._typeCoercer : (d: any) => d;
-      var extents = this._datasetKeysInOrder.map((key) => {
+    protected _updateExtentsForKey(key: string, ifAttr: boolean) {
```
jtlan: Invocations of this method might be more readable if you passed in the `d3.Map` instead of a boolean.

jtlan:
I see a number of places where there's code like this:
```Typescript
this._attrBindings.set(attr, binding);
this._updateExtentsForKey(attr, true);
this._replaceScale(previousScale, binding.scale);
```

Are these methods always used together? If so, it seems like it might be worth combining them.

src/components/plots/plot.ts
```diff
@@ -195,6 +193,12 @@ module Plottable {
         var fn = scale ? (d: any, i: number, dataset: Dataset, m: Plots.PlotMetadata) => scale.scale(accessor(d, i, dataset, m)) : accessor;
         h[attr] = fn;
       });
+      var propertyProjectors = this._propertyToProjectors();
```
jtlan: Going by this line, maybe the method should just be called `_propertyProjectors()`?

jtlan:
[-1] Questions about whether some methods could be combined.

src/components/plots/plot.ts
```diff
@@ -163,30 +173,29 @@ module Plottable {
...
+      this._setupAttr(attrToSet, accessor, scale);
       this.render(); // queue a re-render upon changing projector
       return this;
     }
 
+    protected _setupProperty(property: string, value: any, scale: Scale<any, any>) {
+      this._setupKey(property, value, scale, false);
+    }
+
+    private _setupAttr(property: string, value: any, scale: Scale<any, any>) {
```
jtlan: should be `attr` and not `property`.

src/components/plots/plot.ts
```diff
@@ -163,30 +173,29 @@ module Plottable {
...
+      this._setupKey(property, value, scale, true);
+    }
+
+    private _setupKey(key: string, value: any, scale: Scale<any, any>, ifAttr: boolean) {
```
jtlan: I mentioned earlier that the `boolean` is confusing. Can we split this into two methods? I understand that there will be some repetition, but I think the code will be much clearer. Just move the logic up to `_setupAttr()` and `setupProperty()`

src/components/plots/plot.ts
```diff
@@ -163,30 +173,29 @@ module Plottable {
...
-        scale.onUpdate(this._renderCallback);
-        scale.addExtentProvider(this._extentProvider);
-        scale._autoDomainIfAutomaticMode();
-      }
+      this._setupAttr(attrToSet, accessor, scale);
       this.render(); // queue a re-render upon changing projector
       return this;
     }
 
+    protected _setupProperty(property: string, value: any, scale: Scale<any, any>) {
```
jtlan: not sure "setup" is the right verb here. Since we're setting a binding, maybe "`_bindProperty()`"?

src/components/plots/plot.ts
```diff
@@ -266,25 +287,26 @@ module Plottable {
...
       this._scales().forEach((scale) => scale._autoDomainIfAutomaticMode());
     }
 
-    private _updateExtentsForAttr(attr: string) {
-      var binding = this._attrBindings.get(attr);
-      var accessor = binding.accessor;
-      var scale = binding.scale;
-      var coercer = (scale != null) ? scale._typeCoercer : (d: any) => d;
-      var extents = this._datasetKeysInOrder.map((key) => {
+    private _updateExtentsForKey(key: string, ifAttr: boolean) {
```
jtlan: I think this one should take the binding `Map` and the extent `Map` instead of a `boolean` -- makes it more readable when invoked.

src/components/plots/plot.ts
```diff
@@ -540,5 +572,34 @@ module Plottable {
...
+    private static _scaledAccessor<SD, SR>(accScaleBinding: Plots.AccessorScaleBinding<SD, SR>): _Accessor {
+      return accScaleBinding.scale == null ?
+               accScaleBinding.accessor :
+               (d: any, i: number, dataset: Dataset, m: Plots.PlotMetadata) =>
+                 accScaleBinding.scale.scale(accScaleBinding.accessor(d, i, dataset, m));
+    }
+
+    private _generatePropertyToProjectors(): AttributeToProjector {
+      var attrToProjector: AttributeToProjector = {};
+      this._propertyBindings.forEach((key, binding) => attrToProjector[key] = Plot._scaledAccessor(binding));
```
jtlan: If `_scaledAccessor` is only used here, can we inline it?

jtlan:
I think the `Scale` argument on the property-setters could simply be named `scale`, instead of `propertyNameScale`.

jtlan:
[-1] Comments about restructuring for clarity.

src/components/plots/plot.ts
```diff
@@ -163,30 +173,29 @@ module Plottable {
...
 
+    protected _bindProperty(property: string, value: any, scale: Scale<any, any>) {
+      this._bindKey(property, value, scale, this._propertyBindings, this._propertyExtents);
+    }
+
+    private _bindAttr(attr: string, value: any, scale: Scale<any, any>) {
+      this._bindKey(attr, value, scale, this._attrBindings, this._attrExtents);
+    }
+
+    private _bindKey(key: string, value: any, scale: Scale<any, any>,
```
jtlan: maybe just `_bind`?

PR #1993, opened 2015-05-07
Removed Core.PlottableObject

### LINKED ISSUE #1915: Use reference equality rather than keys when registering objects/callbacks

aicioara:
Closes #1915 
As part of this commit: 
- Clip Path no longer uses the object ID, but instead we have a DOM method which generates unique IDs for this particular purpose
- Altered the Clip Path test a bit. Tried to keep it as close to the original as possible
- Removed the object

To remember is that 
```Typescript
var clipPathId = (<any> c)._boxContainer.node().getElementsByTagName("clipPath")[0].getAttribute("id");
```
does fail on PhantomJS. It works on all other browsers though. Reverted it to 
```typescript
var clipPathId = (<any>c)._boxContainer[0][0].firstChild.id;
```
for now


jtlan:
Probably `_generateClipPath()` should no-op if called more than once, but that's probably better saved for a later PR.

test/core/componentTests.ts
```diff
@@ -220,15 +220,6 @@ describe("Component behavior", () => {
     svg.remove();
   });
 
-  it("componentID works as expected", () => {
```
jtlan: Maybe add a replacement test for `getUniqueClipPathId()`?

src/utils/domUtils.ts
```diff
@@ -127,6 +127,11 @@ export module Utils {
       }
       return null; // not in the DOM
     }
+
+    var _latestClipPathId = 0;
+    export function getUniqueClipPathId() {
+      return ++_latestClipPathId;
```
jtlan: Maybe this should return a `string` that includes a prefix so it can be used directly in `_generateClipPath()`.

src/components/component.ts
```diff
@@ -339,9 +339,10 @@ module Plottable {
       // They don't need the current URL in the clip path reference.
       var prefix = /MSIE [5-9]/.test(navigator.userAgent) ? "" : document.location.href;
       prefix = prefix.split("#")[0]; // To fix cases where an anchor tag was used
-      this._element.attr("clip-path", "url(\"" + prefix + "#clipPath" + this.getID() + "\")");
+      var clipPathId = Utils.DOM.getUniqueClipPathId();
+      this._element.attr("clip-path", "url(\"" + prefix + "#clipPath" + clipPathId + "\")");
```
jtlan: Now that I think about it, our `clipPath` IDs should include "plottable" in the name somewhere to avoid collisions.

jtlan:
[-1] Add in a test for the new method, and maybe change the prefixing to make it more usable in its intended role.

test/core/componentTests.ts
```diff
@@ -203,13 +203,13 @@ describe("Component behavior", () => {
   it("clipPath works as expected", () => {
     assert.isFalse(c.clipPathEnabled, "clipPathEnabled defaults to false");
     c.clipPathEnabled = true;
-    var expectedClipPathID = c.getID();
     c.anchor(svg);
     c.computeLayout({ x: 0, y: 0 }, 100, 100);
     c.render();
+    var expectedClipPathID = (<any>c)._boxContainer[0][0].firstChild.id;
```
jtlan: Since this was peeled directly off of the DOM, it's not really "expected" anymore -- it's the actual value.

Selecting this will be more robust if you do
```Typescript
(<any> c)._boxContainer.select("clipPath").attr("id");
```

jtlan:
[minor] Test semantics.

test/core/componentTests.ts
aicioara: it is expected. it is taken from another component. The current component is `expected` to have the same id as the other component

aicioara: and tried that, it does not work on d3

test/core/componentTests.ts
```diff
@@ -203,13 +203,13 @@ describe("Component behavior", () => {
   it("clipPath works as expected", () => {
     assert.isFalse(c.clipPathEnabled, "clipPathEnabled defaults to false");
     c.clipPathEnabled = true;
-    var expectedClipPathID = c.getID();
     c.anchor(svg);
     c.computeLayout({ x: 0, y: 0 }, 100, 100);
     c.render();
+    var clipPathId = (<any>c)._boxContainer[0][0].firstChild.id;
```
jtlan:
```Typescript
var clipPathId = (<any> c)._boxContainer.node().getElementsByTagName("clipPath")[0].getAttribute("id");
```
works.

aicioara: not on PhantomJS

PR #1995, opened 2015-05-07
No longer pass key in to the Drawer

jtlan:
Wasn't being used for anything!


jtlan:
Going to do this later after the `project()` refactor.

PR #1997, opened 2015-05-08
Reversed the semantics of attaching an interaction 

aicioara:
```typescript
var interaction = new Plotable.Interactions.Pointer();
var component = new Plottable.Component();

component.registerInteraction(interaction); // old
interaction.attachTo(component); //new
```

As part of this commit there are various side effects:
- Having an `onAnchor` and `offAnchor` public methods on the `Component`
- have a symmetric `detachFrom(component)` on `Interaction`
- detaching Interaction requires unsubscribing from `Dispatcher`s, hence callbacks have to be now instance variables
- there was no test coverage on attaching / detaching interactions. Even now the test coverage is pretty low, as it is very hard to test this at an abstract level (abstract interaction has no dispatchers attached), hence only click interaction is tested
- hence created an `interactionTests.ts` file


src/components/component.ts
```diff
@@ -67,6 +70,36 @@ module Plottable {
...
+    public onAnchor(callback: AnchorCallback) {
+      if (this._isAnchored) {
+        callback(this);
+      }
+      this._onAnchorCallbacks.add(callback);
+      return this;
+    }
+
+    /**
+     * Removes a callback to be called on anchoring the Component to the DOM.
```
bluong: Let's match the on/offAnchor docs exactly.

`onAnchor` has the word current.  Let's remove that

src/components/component.ts
```diff
@@ -67,6 +70,36 @@ module Plottable {
...
+     * The callback is identified by reference equality.
+     *
+     * @param {AnchorCallback} callback The callback to be removed
```
bluong: Let's have a better parallel between the doc between on/offAnchor

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
      */
     protected _componentToListenTo: Component;
 
-    public _anchor(component: Component) {
+    private _anchorCallback = (component: Component) => this._anchor(component);
+
+    private _isAnchored: Boolean;
+
+    /* tslint:disable: */
```
jtlan: Is this still needed? I tried pulling the branch and removing the `disable`, and tslint still passed.

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
...
+    // tslint disable does not really work
+    protected _anchor(component: Component) {
+      this._isAnchored = true;
+    }
+
+    protected _unanchor() {
+      this._isAnchored = false;
+    }
+    /* tslint:enable */
```
bluong: What is with the tslint messages here?

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
...
+     * attached to a Component, it first detaches itself from the old component.
+     *
+     * @param {Component} component The component to which to attach the interaction.
+     *
+     * @return {Interaction}
+     */
+    public attachTo(component: Component) {
+      if (this._componentToListenTo) {
+        this.detachFrom(this._componentToListenTo);
+      }
```
bluong: Possibly a premature optimization but should this no-op  if
component === this._componentToListenTo

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
      */
     protected _componentToListenTo: Component;
```
bluong: Should we rename this to just `_component`?

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
...
+
+    /**
+     * Detaches current interaction from the Component. Interaction can be reused.
+     *
+     * @param {Component} component The component to which to attach the interaction.
+     *
+     * @return {Interaction}
+     */
+    public detachFrom(component: Component) {
+      this._unanchor();
```
bluong: Possibly `_unAnchor()`

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      callbackCalled = false;
+      clickInteraction.detachFrom(component);
+
+      TestMethods.triggerFakeMouseEvent("mousedown", component.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      TestMethods.triggerFakeMouseEvent("mouseup", component.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      assert.isFalse(callbackCalled, "callback was removed from component and should not be called");
+
+      svg.remove();
+    });
```
bluong: Is there a way you can test on the base `Interaction` class?

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,94 @@
...
+      var svg2 = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var component1 = new Plottable.Component();
+      var component2 = new Plottable.Component();
+
+      component1.renderTo(svg1);
+      component2.renderTo(svg2);
+
+      var clickInteraction = new Plottable.Interactions.Click();
+
+      var callbackCalled: Boolean;
```
bluong: Let's be consistent with the rest of the codebase and use the primitive types.  I am unsure but I thought tslint should ban this

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,94 @@
...
+      TestMethods.triggerFakeMouseEvent("mouseup", component1.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      assert.isFalse(callbackCalled, "Round 2 callback not called for component 1");
+
+      callbackCalled = false;
+      TestMethods.triggerFakeMouseEvent("mousedown", component2.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      TestMethods.triggerFakeMouseEvent("mouseup", component2.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      assert.isTrue(callbackCalled, "Round 2 callback called for component 2");
+
+      // no detach, as this shorthand should also work
+      clickInteraction.attachTo(component1);
```
bluong: Instead of making a comment here, can we instead have a test to prove this or it should be decently obvious what you are doing here anyways that the comment is uncecessary

bluong:
A few nitpicky things in the inplementation.  Mostly issues on documentation and test structuring.

src/interactions/interaction.ts
```diff
@@ -11,8 +11,53 @@ module Plottable {
      */
     protected _componentToListenTo: Component;
 
-    public _anchor(component: Component) {
+    private _anchorCallback = (component: Component) => this._anchor(component);
```
jtlan: We'll probably want some sort of mechanism to detach the `Interaction` if the `Component` is `destroy()`ed, so we don't have loose references to the `Component` floating around and preventing garbage collection. However, that's an existing bug, so we can tackle it after the API refactor.

src/interactions/interaction.ts
jtlan: Looks like a previous commit had the blocks empty...? I tried pulling the branch in its current state and it worked fine without the tslint stuff.

test/interactions/interactionTests.ts
jtlan: It looks like he would have to add a dispatcher or something. Would it be enough to spy on `_anchor()` and `_unanchor()`, or `<any>`-cast to check `_componentToListenTo`?

test/interactions/interactionTests.ts
jtlan: We should probably tslint-ban the non-primitive `Boolean`, `Number`, and `String` types until we can think of a good use for them.

src/components/component.ts
aicioara: Fixed

src/components/component.ts
aicioara: I believe fixed, but not sure what you mean

src/interactions/interaction.ts
aicioara: @jtlan Intentional leftover, so I don't forget. Basically I need a no-op function (kind of abstract) in the base class. But the tslint does not allow me to do that. So I needed to invent an instance variable which is never ever used. Do you think it is alright? Also tslint disable does not work

src/interactions/interaction.ts
aicioara: I believe we're trying to be too smart with that change. The user should not do that in the first place I believe.

src/interactions/interaction.ts
aicioara: Not suggestive enough without prior knowledge of how this class works. maybe `_componentAttachedTo` would work (doing this change)

src/interactions/interaction.ts
aicioara: `unanchor` is a word in itself (well kind of, unanchored is a word). Camel case when we mean 2 words only

test/interactions/interactionTests.ts
aicioara: I'd like to avoid writing extra tests to test shorthand for the same thing. Tests are already running for 6 seconds (or more on SauceLabs). I'd also want to make it obvious that the detach was intentional. Not sure if it is without the comment.

src/interactions/interaction.ts
aicioara: We probably need an onDestroy callback as well. Will think about this

src/components/component.ts
```diff
@@ -67,6 +70,38 @@ module Plottable {
...
       }
       this._isAnchored = true;
+      this._onAnchorCallbacks.callCallbacks(this);
+      return this;
+    }
+
+    /**
+     * Adds a callback to be called on anchoring the Component to the DOM.
+     * If the component is already anchored, the callback is called immediately.
+     * To later deregister, use offAnchor(callback).
```
jtlan: I don't think it's worth cross-referencing the other method, since the point of the `onEvent()`/`offEvent()` naming scheme is to make things sufficiently intuitive for users.

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,110 @@
+///<reference path="../testReference.ts" />
+
+var assert = chai.assert;
+
+describe("Interactions", () => {
+  describe("Interactions", () => {
```
jtlan: Why the double nesting?

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,110 @@
...
+      clickInteraction.detachFrom(component);
+
+      TestMethods.triggerFakeMouseEvent("mousedown", component.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      TestMethods.triggerFakeMouseEvent("mouseup", component.content(), SVG_WIDTH / 2, SVG_HEIGHT / 2);
+      assert.isFalse(callbackCalled, "callback was removed from component and should not be called");
+
+      svg.remove();
+    });
+
+    it("interactions are reusable", () => {
```
jtlan: I think it would be better to split this into three tests:
- `attachTo()`
- `detachFrom()`
- `attachTo() disconnects from previous Component`

src/components/component.ts
aicioara: Then I did not understand @bluong 's comment

src/interactions/interaction.ts
```diff
@@ -9,10 +9,52 @@ module Plottable {
      * "foreground" and "background" elements where it can draw things,
      * e.g. crosshairs.
      */
-    protected _componentToListenTo: Component;
+    protected _componentAttachedTo: Component;
```
jtlan: As @bluong mentioned, "`_component`" is a sufficient name.

jtlan:
[-1] Test semantics.

test/interactions/interactionTests.ts
aicioara: I actually meant `Interaction` for the second one. Do you mind double nesting with this renaming?

src/interactions/interaction.ts
aicioara: The commenting system here is pretty hard to follow. Pasting my argument here as well

Not suggestive enough without prior knowledge of how this class works. maybe `_componentAttachedTo` would work (doing this change)

test/interactions/interactionTests.ts
aicioara: I disagree. I believe it's extra boilerplate and extra time consumption for testing (due to generating SVGs, attaching them to the DOM). Also the `detachFrom` test includes the `attachTo` and `attachTo() disconnects from previous Component` is copy+pasting half of the behavior of item 2

bluong: I think I agree with Justin in this case.  At our current stage, we should I think be not concerned with the performance of our tests but rather the correctness + maintainability of them.  We shouldn't group a bunch of tests into 1 big test as that detracts from comprehension and independence of the tests

aicioara: Performance of the tests should always be a concern. The entire dev process gets slower with slower tests. What is even worse is that we never refactor tests. We always refactor code.

bluong: Yes it is indeed a concern, but I feel that there is a higher priority on us having confidence that our library works in the first place and the place that builds that confidence is the tests

aicioara:
The two main points addressed were
`_componentAttachedTo` -> `_component`
- pro: axis does the same for the scale (it does not call it scale attached to)
- cons: _componentAttachedTo is more descriptive. We should allow naming to be self explainatory. We should not read the code (or comments) to figure out what an instance variable is / should do

Separate the attach/detach test into multiple tests
- pro: modularity of the test. concepts of unit tests. other purity stuff...
- pro: I can test attach more thoroughly (throw junk at it, see what it does)
- cons: boilerplate
- cons: time consuming for the tests. Performance of the tests should always be a concern. The entire dev process gets slower with slower tests. What is even worse is that we never refactor tests. We always refactor code.
- cons: we cannot have `detach` without `attach` so it makes no sense to separate the two methods in testing. 
- cons: the fix for this is 1 extra test, which is exactly a subset of the tests I wrote already (well, almost). The existing tests will not shrink anyway.
- cons: there is no modularity anyway. If the attach fails, you can see that in the assert now. If it fails in the split, it will just fail multiple tests, so it is hard to see which one is the issue anyway, because tests are not hierarchical. It's not like the tests know which axioms they are dependent on and where they have been asserted

test/interactions/interactionTests.ts
jtlan: @aicioara wrote:
>What is even worse is that we never refactor tests. We always refactor code.
We probably *should* refactor tests, and I think that should be something we do immediately after finishing the API refactor.

test/interactions/interactionTests.ts
```diff
@@ -0,0 +1,110 @@
+///<reference path="../testReference.ts" />
+
+var assert = chai.assert;
+
+describe("Interactions", () => {
+  describe("Interaction", () => {
```
jtlan: I was going to say that we didn't need this inner `describe()` at all, since anything on the top level (directly under `describe("Interactions")`) could probably be safely assumed to test the root class.

aicioara: It's a convenience. If you want to run just the tests in this file, you can click on the `Interaction` in the tests interface. (I very often want to do that). If you put it at the root, there is no way to do it

jtlan: Okay, that makes a lot of sense. It might be good to come up with some way to identify the base class in tests, but I think we can save that decision for when we revisit tests.

test/interactions/interactionTests.ts
aicioara: if we _should_ revisit the tests, then why not bring up the discussion again then and leave the tests like this for now.

bluong:
Decently important preface - Our testing infrastructure for essentially the last year (or more) has essentially been using a pure additive structure.  (Add a feature?  Add a test.), but it hasn't necessarily been made concise in almost any way.

One main instance is that there are many tests that are trying to prove a fact that other tests have already proven in a much shorter, performant fashion.  (scatterPlots work on quadratic accessors! or something to that effect)

Thus, we definitely need to improve that piece in our codebase for the reasons you've described but it's also important on our end that we cleanup the tests so we have a much better understanding of what our library actually does

In accordance with the above preface, I still think the test should be broken up, mainly because I see the test name and I don't think it's very unitized. "interactions are reusable"

The test names should point near directly at the fact they are trying to prove and the test code should follow

test/interactions/interactionTests.ts
bluong: The tests interface is extremely thin if you look at the url.  It literally does a grep for the string you pressed and runs those according tests

test/interactions/interactionTests.ts
dbt: If you know something is the right thing to do, why wait?

the performance of the tests is a relatively small concern, and we can address performance issues if they extend the suite running time by a significant fraction.

If you're worried about duplication of code between the attach and detach tests, you can pull it out into a local function that you invoke both times.

aicioara: @dbt because I don't believe it is the right thing to do. However, I was a minority so made the change.

plottable.d.ts
```diff
@@ -1502,6 +1503,24 @@ declare module Plottable {
...
+         * If the component is already anchored, the callback is called immediately.
+         *
+         * @param {AnchorCallback} callback The callback to be called on Anchor
+         *
+         * @return {Component}
+         */
+        onAnchor(callback: AnchorCallback): Component;
+        /**
+         * Removes a callback to be called on anchoring the Component to the DOM.
+         * The callback is identified by reference equality.
```
bluong: Do you think the above line is necessary?

plottable.d.ts
```diff
@@ -1502,6 +1503,24 @@ declare module Plottable {
          */
         anchor(selection: D3.Selection): Component;
         /**
+         * Adds a callback to be called on anchoring the Component to the DOM.
+         * If the component is already anchored, the callback is called immediately.
+         *
+         * @param {AnchorCallback} callback The callback to be called on Anchor
```
bluong: lowercase anchor

plottable.d.ts
```diff
@@ -1502,6 +1503,24 @@ declare module Plottable {
...
+         *
+         * @return {Component}
+         */
+        onAnchor(callback: AnchorCallback): Component;
+        /**
+         * Removes a callback to be called on anchoring the Component to the DOM.
+         * The callback is identified by reference equality.
+         *
+         * @param {AnchorCallback} callback The callback to be removed
```
bluong: I'm trying to find a better parallel between the `onAnchor` and `offAnchor` docs and it mostly looks good.

For this param part, would it be too vague to say
The callback to be added / removed ?

I guess another way to word it is that the parallel doc for `offAnchor` are a little odd, `The callback to be called off Anchor`

plottable.d.ts
```diff
@@ -3680,15 +3692,26 @@ declare module Plottable {
...
+        protected _componentAttachedTo: Component;
+        protected _anchor(component: Component): void;
+        protected _unanchor(): void;
         /**
-         * It maintains a 'hitBox' which is where all event listeners are
-         * attached. Due to cross- browser weirdness, the hitbox needs to be an
-         * opaque but invisible rectangle.  TODO: We should give the interaction
-         * "foreground" and "background" elements where it can draw things,
-         * e.g. crosshairs.
+         * Attaches current interaction to a Component. If the interaction was already
```
bluong: Let's put the second sentence on the second line

plottable.d.ts
```diff
@@ -3680,15 +3692,26 @@ declare module Plottable {
...
+         * attached to a Component, it first detaches itself from the old component.
```
bluong: Let's put the second sentence on the second line

plottable.d.ts
aicioara: It depends.

plottable.d.ts
bluong: I don't this the word `current` here is nice in the doc... I would remove the word completely or replace it with `this`

plottable.d.ts
```diff
@@ -3680,15 +3692,26 @@ declare module Plottable {
...
+         *
+         * @param {Component} component The component to which to attach the interaction.
+         *
+         * @return {Interaction}
+         */
+        attachTo(component: Component): Interaction;
+        /**
+         * Detaches current interaction from the Component. Interaction can be reused.
```
bluong: "This interaction can be reused." and also consider placing this on a second line

plottable.d.ts
```diff
@@ -3680,15 +3692,26 @@ declare module Plottable {
...
+         *
+         * @param {Component} component The component to which to attach the interaction.
```
bluong: "The component from which to detach the interaction."

plottable.d.ts
aicioara: I liked the `callback to be added / removed` idea. Changing to that to have better symmetry

bluong:
Docs docs docs

PR #2000, opened 2015-05-08
restructure piePlot to use old arcDrawer architecture

bluong:
Restoring _Drawers back to their previous state.  Utiliize generateAttrToProjector to stay as consistent with the current codebase as possible

### DIFF
--- a/src/components/plots/piePlot.ts
+++ b/src/components/plots/piePlot.ts
@@ -74,11 +74,13 @@ export module Plots {
       var defaultFillFunction = (d: any, i: number) => this._colorScale.scale(String(i));
       attrToProjector["fill"] = attrToProjector["fill"] || defaultFillFunction;
 
+      this._propertyBindings.forEach((key, binding) => attrToProjector[key] = Pie._scaledAccessor(binding));
+
       return attrToProjector;
     }
 
     protected _getDrawer(key: string) {
-      return new Plottable.Drawers.Arc(key, this).setClass("arc");
+      return new Plottable.Drawers.Arc(key).setClass("arc");
     }
 
     public getAllPlotData(datasets = this.datasets()): Plots.PlotData {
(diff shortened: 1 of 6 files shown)


src/components/plots/piePlot.ts
```diff
@@ -74,11 +74,13 @@ export module Plots {
       var defaultFillFunction = (d: any, i: number) => this._colorScale.scale(String(i));
       attrToProjector["fill"] = attrToProjector["fill"] || defaultFillFunction;
 
+      this._propertyBindings.forEach((key, binding) => attrToProjector[key] = Pie._scaledAccessor(binding));
```
jtlan: Properly, the property-bindings should go on before the attr-bindings, so that `attr()` takes precedence over the property setters. We can emulate that functionality here by checking if `attrToProjector[key]` already exists and not setting the property in that case.

jtlan:
[minor] Small change to have `attr()` eventually take precedence over the property calls.

src/components/plots/piePlot.ts
```diff
@@ -69,7 +69,11 @@ export module Plots {
     }
 
     protected _generateAttrToProjector(): AttributeToProjector {
-      var attrToProjector = super._generateAttrToProjector();
+      var attrToProjector: AttributeToProjector = {};
+      this._propertyBindings.forEach((key, binding) => attrToProjector[key] = Pie._scaledAccessor(binding));
+
+      var superAttrToProjector = super._generateAttrToProjector();
+      Object.keys(superAttrToProjector).forEach((key) => attrToProjector[key] = superAttrToProjector[key]);
```
jtlan: Eventually we'll probably want something closer to this in `Plot`:
```Typescript
  var attrToProjector = this._propertiesToProjectors(); // loads up an attrToProjector with the correct projectors for properties
this._attrBindings.forEach((attr, binding) => {
  var projector = binding.accessor;
  if (binding.scale != null) {
    projector = (datum: any, index: number, dataset: Dataset, m: PlotMetadata) => binding.scale.scale(accessor(datum, index, dataset, m);
  }
  attrToProjector[attr] = projector;
});
}

jtlan:
When you port this up to the `Plot` level, use something like the future-facing version in my comment.

jtlan:
[+1]

PR #2001, opened 2015-05-08
Remove unused QuantitativeScale methods.

jtlan:
Reducing API surface area. Removed:

`interpolate()`
`rangeRound()`
`clamp()`


jtlan:
There weren't even tests for these...

PR #2002, opened 2015-05-08
Get rid of QuantitativeScale.numTicks()

jtlan:
Follow-up to #2001. This change may be kind of controversial...

The main problem with `numTicks(n)` was that users were never guaranteed to get exactly `n` ticks out of the `Scale`, so it didn't really give the control it claimed to give. Users could only use it as an approximate method for controlling the tick density -- but in that case, wouldn't it make more sense to have a method that actually controlled tick density, instead of claiming to set the exact number of ticks?


bluong:
I personally mainly am ok with removing functionality in general as long as ask ourselves "What will be the consequences of this for the user?"  If this function was just never used by the user and we find that users are never going to use this functionality then sure let's remove it.

However, I distinctly remember somewhere having a usecase for this.  We definitely should first investigate their usage and how to accomodate before taking any drastic action

rcchen:
Code-wise this is :+1: but leaving open for usage discussion.

jtlan:
Holding to get some user input.

crmorford:
It's not the cleanest way to do it, so it should come out. We can redo it in a better way if customers ask for this functionality again.

PR #2005, opened 2015-05-08
Fix the logic in RenderController.

jtlan:
>"The loop was kind of pointless."

@jtlan, a155f04

In fact, the loop was super critical, but because the logic was so convoluted, it was difficult to tell. New logic for `RenderController`'s `flush()`:

* Save off current `Set` of `Component`s that have requested a render.
* Create a new `Set` for `Component`s that request a render while we're drawing.
* Draw all `Component`s in the saved-off `Set`.
* If there are `Component`s that have requested a render while we were drawing, request another frame.

### DIFF
--- a/src/core/renderController.ts
+++ b/src/core/renderController.ts
@@ -23,7 +23,6 @@ module Plottable {
     var _componentsNeedingRender = new Utils.Set<Component>();
     var _componentsNeedingComputeLayout = new Utils.Set<Component>();
     var _animationRequested: boolean = false;
-    var _isCurrentlyFlushing: boolean = false;
     export var _renderPolicy: RenderPolicies.RenderPolicy = new RenderPolicies.AnimationFrame();
 
     export function setRenderPolicy(policy: string | RenderPolicies.RenderPolicy): void {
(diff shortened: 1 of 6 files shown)


aicioara:
tslint first

src/core/renderController.ts
```diff
@@ -89,23 +84,24 @@ module Plottable {
     export function flush() {
       if (_animationRequested) {
         // Layout
-        _componentsNeedingComputeLayout.values().forEach((component: Component) => component.computeLayout());
+        _componentsNeedingComputeLayout.values().forEach((component) => component.computeLayout());
```
aicioara: type the function. It's weird that `tsc` sometimes complains about this, sometimes it does not

jtlan: Visual Studio Code detects the type correctly...

aicioara: Actually I looked around the codebase. We have a mixture anyway. I prefer the typed version in the future, just because TSC enforces it sometimes (for example if you wanted to extract the function into an instance variable somewhere)

jtlan: From offline:
Going forward, we'll avoid typing things when the tools can correctly infer the types.

PR #2007, opened 2015-05-09
Removed the any from the scales

aicioara:
Also removed the coercer and no longer accepting coercing (for example passing numbers formatted as strings when numbers are required)

### DIFF
--- a/src/scales/quantitativeScale.ts
+++ b/src/scales/quantitativeScale.ts
@@ -71,9 +71,9 @@ module Plottable {
     /**
      * Gets a set of tick values spanning the domain.
      *
-     * @returns {any[]} The generated ticks.
+     * @returns {D[]} The generated ticks.
      */
-    public ticks(): any[] {
+    public ticks(): D[] {
       return this._tickGenerator(this);
     }
 
@@ -81,7 +81,7 @@ module Plottable {
      * Given a domain, expands its domain onto "nice" values, e.g. whole
      * numbers.
      */
-    public _niceDomain(domain: any[], count?: number): any[] {
+    public _niceDomain(domain: D[], count?: number): D[] {
       return this._d3Scale.copy().domain(domain).nice(count).domain();
     }
 
@@ -115,7 +115,7 @@ module Plottable {
       }
     }
 
-    public _defaultExtent(): any[] {
+    public _defaultExtent(): number[] {
       return [0, 1];
     }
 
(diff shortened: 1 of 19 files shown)


aicioara:
Self reviewing for a second

test/scales/timeScaleTests.ts
```diff
@@ -15,24 +15,14 @@ describe("TimeScale tests", () => {
       var time2 = scale.domain()[1].valueOf();
       assert.strictEqual(time2, secondDate, "first value of domain set correctly");
     }
-    checkDomain(["10/1/2014", "11/1/2014"]);
-    checkDomain(["October 1, 2014", "November 1, 2014"]);
-    checkDomain(["Oct 1, 2014", "Nov 1, 2014"]);
+    checkDomain([new Date("10/1/2014"), new Date("11/1/2014")]);
+    checkDomain([new Date("October 1, 2014"), new Date("November 1, 2014")]);
+    checkDomain([new Date("Oct 1, 2014"), new Date("Nov 1, 2014")]);
```
aicioara: Unsure if we still want this test. It is actually checking JS coercing behavior (which we should assume correct)

jtlan: Yeah, if we don't have the coercer, there's no point in testing this.

jtlan:
I like it! I think we should also make sure that users downstream aren't going to be strongly affected by the removal of this functionality. @crmorford, @istarnx, any thoughts?

jtlan:
@aicioara Please fix any Overlaying Tests this change may have broken. A number of them seem to depend on the `string` coercer logic on `Scales.Category()`.

istarnx:
@jtlan I scanned our code and don't think this'll be painful for us. Is there anything I'm missing? What's the major user impact here?

aicioara:
Fixing overlaying tests

src/scales/quantitativeScale.ts
```diff
@@ -115,8 +114,8 @@ module Plottable {
       }
     }
 
-    public _defaultExtent(): any[] {
-      return [0, 1];
+    public _defaultExtent(): D[] {
+      throw Error("The quantitative scale itself does not have a default extent");
```
bluong: This sorta basically disallows QuantitativeScales to be instantiated and used on its own.  Is this ok?

aicioara: This is a good point. Now the problem is: 
1. We want to remove `any` everywhere it is possible. 
2. There is no default extent when there is no concrete type.

Also, a sign that there is a problem is the underscore in the name. I suggest we remove this method altogether. It is only used in the `domainer.ts`

aicioara: opened https://github.com/palantir/plottable/issues/2015 for this issue

PR #2008, opened 2015-05-10
Reworked ComponentContainer inheritance

jtlan:
* `ComponentContainer` no longer assumes a single list of `Components`.
 * Moved implementation details associated with the single list down to `Group`.
 * Moved testing down to subclass level.

Renamed methods:
 * `addComponent()` --> `append()` on `Group`:
```Typescript
class Group {
  public add(component: Component, prepend = false);
}
```
 * `addComponent()` --> `add()` on `Table`:
```Typescript
class Table {
  public add(component: Component, row: number, col: number);
}
```
Removed unused methods from `ComponentContainer`:
 - `detachAll()`
 - `empty()`

Added new method:
 - `has(component: Component)`


src/components/componentContainer.ts
```diff
@@ -6,54 +6,36 @@ module Plottable {
    * It will not do anything if instantiated directly.
    */
   export class ComponentContainer extends Component {
-    private _components: Component[] = [];
```
rcchen: Why is this being removed? Aren't classes that extend ComponentContainer all based on some base Component[]?

rcchen: Never mind just looked at the table code and this makes sense.

src/components/group.ts
```diff
@@ -52,12 +53,55 @@ export module Components {
...
 
     public fixedHeight(): boolean {
-      return this.components().every((c) => c.fixedHeight());
+      return this._components().every((c) => c.fixedHeight());
     }
+
+    /**
+     * @return {Component[]} The Components in this Group.
+     */
+    public components(): Component[] {
```
rcchen: Should this also be available on `Table`?

src/components/componentContainer.ts
```diff
@@ -6,54 +6,36 @@ module Plottable {
...
-     * Removes the specified Component from the ComponentContainer
-     * 
-     * @param c Component the Component to remove.
+     * Removes the specified Component from the ComponentContainer.
      */
-    public remove(c: Component) {
-      var removeIndex = this._components.indexOf(c);
-      if (removeIndex >= 0) {
-        this.components().splice(removeIndex, 1);
+    public remove(component: Component) {
```
rcchen: Let's move this to `Group`/`Table`; seems strange for there to be `remove`/`_remove` here and not `add`/`_add`

src/components/componentContainer.ts
```diff
@@ -6,54 +6,36 @@ module Plottable {
...
-      } else {
-        this.components().push(c);
-      }
-      c._parent(this);
-      if (this._isAnchored) {
-        c.anchor(this._content);
-      }
-      this.redraw();
-      return true;
+    protected _remove(component: Component) {
```
rcchen: See above.

src/components/group.ts
jtlan: The implementation for `Table` should probably return `Component[][]`. I was thinking we'd add it later.

src/components/componentContainer.ts
jtlan: The signatures of the two `add()` methods are different, which is why `add()` doesn't exist at the `ComponentContainer` level.

src/components/group.ts
```diff
@@ -52,12 +53,55 @@ export module Components {
...
+      return this._componentList;
+    }
+
+    /**
+     * Adds a Component to the Group.
+     * 
+     * @param {Component} component
+     * @param {boolean} prepend If true, prepends the Component. If false, appends it.
+     */
+    public add(component: Component, prepend = false) {
```
rcchen: We should try and think of a better way to handle the boolean flag for prepend/append (ideally not use a boolean flag here at all).

rcchen: I'm leaning towards a stringy parameter but now feel that prepend/append are not the correct set of verbs for this because there's no sense of front or back in a group, only top and bottom. Maybe "above" and "below", though that runs the (potential) risk of confusion in a table layout.

rcchen: The public API points we provide are actually `.above()`/`.below()` now that I think about it. What do you think about changing the signature of this API point to `add(component: Component, position = "above")`

src/components/table.ts
```diff
@@ -75,43 +79,42 @@ export module Components {
...
-     */
-    public removeComponent(component: Component) {
-      super.remove(component);
+    protected _remove(component: Component) {
       for (var r = 0; r < this._nRows; r++) {
         for (var c = 0; c < this._nCols; c++) {
           if (this._rows[r][c] === component) {
             this._rows[r][c] = null;
-            return;
+            return true;
```
rcchen: I know we're not hitting on optimization yet, but it might be worth adding a dictionary/hashmap lookup structure { component: (i, j) } to table so we can avoid this n^2 traversal.

src/components/group.ts
jtlan: I think it'd make more sense to just use `above()`/`below()`, or maybe `append()`/`prepend()`.

src/components/table.ts
jtlan: `(this._nRows * this._nCols) < 100` here for most reasonable cases, so I'm not really concerned. Adding a dictionary means having to duplicate the information, which introduces an entire class of synchronization bugs.

src/components/group.ts
rcchen: Preference for the former.

jtlan: Semantics are weird though:
```Typescript
```

plottable.d.ts
```diff
@@ -1415,7 +1415,7 @@ declare module Plottable {
 
 
 declare module Plottable {
-    type AnchorCallback = (component: Component) => any;
```
bluong: Should this actually be called `ComponentCallback`?  Do we want to use this signature for anything other than anchoring?

plottable.d.ts
```diff
@@ -1598,8 +1598,23 @@ declare module Plottable {
...
+        /**
+         * Removes a callback to be called when th Component is detach()-ed.
+         * The callback is identified by reference equality.
+         *
+         * @param {ComponentCallback} callback The callback to be removed.
+         * @return {Component} The calling Component.
+         */
+        offDetach(callback: ComponentCallback): Component;
+        parent(): ComponentContainer;
+        parent(parentElement: ComponentContainer): any;
```
bluong: Should the above 2 be exposed?  If so, why?

Lastly, why does the second signature return an `any`

bluong: I personally like the parameter to be called `parent` to mirror the API endpoint

plottable.d.ts
```diff
@@ -1662,40 +1677,34 @@ declare module Plottable {
...
+        protected _remove(component: Component): boolean;
         /**
-         * Returns true iff the ComponentContainer is empty.
+         * Returns a list of components in the ComponentContainer.
          *
-         * @returns {boolean} Whether the calling ComponentContainer is empty.
+         * @returns {Component[]} the contained Components
          */
-        empty(): boolean;
+        protected _components(): Component[];
```
bluong: Are we not exposing `components` anymore?

plottable.d.ts
```diff
@@ -1727,6 +1736,19 @@ declare module Plottable {
...
+             */
+            components(): Component[];
+            protected _components(): Component[];
+            /**
+             * Adds a Component to the Group.
+             *
+             * @param {Component} component
+             * @param {boolean} prepend If true, prepends the Component. If false, appends it.
+             */
+            add(component: Component, prepend?: boolean): Group;
```
bluong: I remember there was talk if we like this siganture.  Ringing the bell that `boolean` parameters are bad

src/components/component.ts
```diff
@@ -479,29 +480,51 @@ module Plottable {
...
-      }
-      this._isAnchored = false;
-      this._parentElement = null;
+    /**
+     * Adds a callback to be called when th Component is detach()-ed.
+     *
+     * @param {ComponentCallback} callback The callback to be added.
+     * @return {Component} The calling Component.
+     */
+    public onDetach(callback: ComponentCallback) {
```
bluong: Should we follow the convention that @aicioara presented that if the `Component` is already detached, call the callback immediately?

src/components/componentContainer.ts
```diff
@@ -6,90 +6,77 @@ module Plottable {
...
+     * Carry out the actual removal of a Component.
+     * Implementation dependent on the type of container.
+     * 
+     * @return {boolean} true if the Component was successfully removed, false otherwise.
      */
-    public empty() {
-      return this._components.length === 0;
+    protected _remove(component: Component) {
+      return false;
     }
```
bluong: Do we want to follow fluent convention here or typical `remove` convention here?

src/components/group.ts
```diff
@@ -3,6 +3,7 @@
 module Plottable {
 export module Components {
   export class Group extends ComponentContainer {
+    private _componentList: Component[] = [];
```
bluong: Is `_components` ok?

bluong: What's the difference between this and `_components()`

src/components/group.ts
```diff
@@ -52,12 +53,54 @@ export module Components {
...
+    /**
+     * @return {Component[]} The Components in this Group.
+     */
+    public components(): Component[] {
+      return this._componentList.slice();
+    }
+
+    protected _components(): Component[] {
+      return this._componentList;
+    }
```
bluong: The protected version and the public version are so similar yet so different...  This could be easily confusing for extenders

src/components/table.ts
```diff
@@ -52,12 +52,16 @@ export module Components {
...
         row.forEach((component, colIndex) => {
           if (component != null) {
-            this.addComponent(component, rowIndex, colIndex);
+            this.add(component, rowIndex, colIndex);
           }
         });
       });
     }
 
+    protected _components(): Component[] {
```
bluong: Minor, but I don't think you need the `Component[]` explicit typing

src/components/table.ts
```diff
@@ -75,43 +79,40 @@ export module Components {
...
+      if (!this.has(component)) {
+        component.detach();
 
-      if (currentComponent) {
-        component = component.above(currentComponent);
-      }
+        var currentComponent = this._rows[row] && this._rows[row][col];
+        if (currentComponent) {
+          component = component.above(currentComponent);
+        }
```
bluong: So do we agree that `Table`s do magic with `Group`s?

src/components/table.ts
bluong: `return true;` or `return this;`?

jtlan: Internal method, returns a `boolean` that says whether or not the removal was successful.

plottable.d.ts
jtlan: Also used for `detach()`-ing.

plottable.d.ts
jtlan: Setter functionality has to be exposed for `ComponentContainer` to get at it.

plottable.d.ts
jtlan: Only on `Group`.

src/components/component.ts
jtlan: That doesn't see quite right. I'm also not sure that should be the case for the `anchor()` callback. Instead we should have an `anchored()` method or `inDOM()` method.

src/components/componentContainer.ts
jtlan: See above note. `_remove()` is an internal method that returns a boolean based on whether or not the removal was successful (similar to other list-like APIs).

src/components/group.ts
jtlan: `_componentList` is the internal list; `_components()` is the method called by the parent class to get the full list of `Component`s. `components()` is the public-facing version.

src/components/table.ts
jtlan: No, let's not do that any more.

src/components/component.ts
bluong: @aicioara What do you think?

src/components/componentContainer.ts
```diff
@@ -6,90 +6,77 @@ module Plottable {
...
-     * @returns {boolean} Whether the calling ComponentContainer is empty.
+     * Carry out the actual removal of a Component.
+     * Implementation dependent on the type of container.
+     * 
+     * @return {boolean} true if the Component was successfully removed, false otherwise.
      */
-    public empty() {
-      return this._components.length === 0;
+    protected _remove(component: Component) {
+      return false;
```
bluong: Containers can never be `remove`d?

src/components/group.ts
```diff
@@ -52,12 +44,43 @@ export module Components {
...
+    /**
+     * @return {Component[]} The Components in this Group.
+     */
+    public components(): Component[] {
+      return this._componentList.slice();
     }
+
+    protected _components(): Component[] {
+      return this._componentList;
+    }
```
bluong: Still noting that the above implementation will be really confusing for extenders

src/components/componentContainer.ts
jtlan: No, `remove()` removes a `Component` from the container. `destroy()` is the method you're thinking of.

src/components/component.ts
aicioara: The reason behind my choice of `onAnchor()`: For `onAnchor`, calling it immediately was a sensible choice. Also, it avoided a lot of hustle of checking the `_isAnchored` flag and made the code prettier.

Here the semantics are a bit different. A component that has not been attached yet is basically detached, but that does not mean that it was detached. To draw a parallel, a component that is anchored means that the component has gone through the anchor process, hence calling the callback is sensible. The difference is the initial state.

src/components/component.ts
```diff
@@ -477,29 +431,51 @@ module Plottable {
...
+     */
+    public offDetach(callback: ComponentCallback) {
+      this._onDetachCallbacks.delete(callback);
+      return this;
+    }
 
-      this.detach();
-      this._parentElement = parentElement;
+    public parent(): ComponentContainer;
+    public parent(parent: ComponentContainer): any;
```
bluong: Should return this `ComponentContainer` no?

src/components/componentContainer.ts
```diff
@@ -6,90 +6,75 @@ module Plottable {
...
-     * @returns {ComponentContainer} The calling ComponentContainer
+     * Invokes a callback on each Component in the ComponentContainer.
      */
-    public detachAll() {
-      // Calling c.remove() will mutate this._components because the component will call this._parent._removeComponent(this)
-      // Since mutating an array while iterating over it is dangerous, we instead iterate over a copy generated by Arr.slice()
-      this.components().slice().forEach((c: Component) => c.detach());
-      return this;
+    protected _forEach(callback: (component: Component) => any) {
+      throw new Error("_forEach() is not implemented on ComponentContainer");
```
bluong: If only we had AbstractClasses :/

src/components/group.ts
```diff
@@ -20,26 +17,32 @@ export module Components {
     constructor(components: Component[] = []) {
       super();
       this.classed("component-group", true);
-      components.forEach((c: Component) => this.add(c));
+      components.forEach((c: Component) => this.append(c));
+    }
+
+    protected _forEach(callback: (component: Component) => any) {
```
bluong: callback: `ComponentCallback`?

bluong: Note to be a big careful as I think `callback`'s purpose is different from `ComponentCallback` and it's not obvious that this is the fact.

The thing that points this out is that `forEach` callbacks usually take in an index / the whole array but that doesn't apply at all to `ComponentCallback`s

src/components/group.ts
```diff
@@ -20,26 +17,32 @@ export module Components {
...
-      this.add(c, !below);
-      return this;
-    }
-
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.components().forEach((c) => {
-        c.computeLayout({ x: 0, y: 0 }, this.width(), this.height());
+      this._forEach((component) => {
+        component.computeLayout({ x: 0, y: 0 }, this.width(), this.height());
```
bluong: I forget the semantics , should (0, 0) be enterred here or `this.origin()` ?

src/components/table.ts
```diff
@@ -75,43 +99,38 @@ export module Components {
...
-      var currentComponent = this._rows[row] && this._rows[row][col];
-
-      if (currentComponent) {
-        component = component.above(currentComponent);
-      }
-
-      if (super.add(component)) {
+      if (!this.has(component)) {
+        var currentComponent = this._rows[row] && this._rows[row][col];
+        if (currentComponent) {
```
bluong: `currentComponent != null` please

test/components/labelTests.ts
```diff
@@ -93,9 +93,8 @@ describe("Labels", () => {
   it("centered text in a table is positioned properly", () => {
     var svg = TestMethods.generateSVG(400, 400);
     var label = new Plottable.Components.Label("X");
-    label.classed(Plottable.Components.Label.TITLE_LABEL_CLASS, true);
-    var t = new Plottable.Components.Table().addComponent(label, 0, 0)
-                                 .addComponent(new Plottable.Component(), 1, 0);
+    var t = new Plottable.Components.Table().add(label, 0, 0)
+                                 .add(new Plottable.Component(), 1, 0);
```
bluong: Was the title label class important?  If so, add in.  If not, feel free to leave it out

bluong:
Only small things at this point

src/components/component.ts
jtlan: Nice catch, thanks.

src/components/group.ts
jtlan: (0, 0).

test/components/labelTests.ts
jtlan: Not important, I think, seeing as the test still passes.

PR #2004, opened 2015-05-08
[XYPlot] - Adding .x / .y endpoints

bluong:
```typescript
.project("x"/"y"...
```

is now a thing from the past!  Now the API looks like...

```typescript
public x(): Plots.AccessorScaleBinding<X, number>;
public y(): Plots.AccessorScaleBinding<Y, number>;
public x(x: number | _Accessor): XYPlot<X, Y>;
public y(y: number | _Accessor): XYPlot<X, Y>;
public x(x: X | _Accessor, xScale: Scale<X, number>): XYPlot<X, Y>;
public y(y: Y | _Accessor, yScale: Scale<Y, number>): XYPlot<X, Y>;
```

PR #2010, opened 2015-05-11
Move autoranging logic into attrToExtents

jtlan:
(Based off of #2007, with overlaying fixes from #2009 overlaid on it).

Setting the Y-scale's domain based on over what's visible on X (or vice versa) is now accomplished by filtering the data prior to computing extents. This allowed the removal of `_normalizeDatasets` and better expresses the logic behind the feature.

Now to pick better names for the methods that enable it...


jtlan:
@bluong This will probably conflict a bit with #2004 (which I expect will get there first), but it surprisingly it also conflicts less than I expected when I started writing the feature.

src/components/plots/plot.ts
```diff
@@ -270,24 +270,25 @@ module Plottable {
...
-        return this._computeExtent(dataset, accessor, coercer, plotMetadata);
+        return this._computeExtent(dataset, accessor, plotMetadata, filter);
       });
       this._attrExtents.set(attr, extents);
     }
 
-    private _computeExtent(dataset: Dataset, accessor: _Accessor, typeCoercer: (d: any) => any, plotMetadata: any): any[] {
+    private _computeExtent(dataset: Dataset, accessor: _Accessor, plotMetadata: any, filter?: _Accessor): any[] {
       var data = dataset.data();
+      if (filter) {
```
bluong: `filter != null`

Just noting that I'm starting to align more to the explicit `filter != null` rather than just the `filter` mainly due to a more stricter typing

For example, `projectorsReady()` used to have an implementation that looked like...

```typescript
return this._projections["x"] && this._projections["y"];
```

which meant that `projectionsReady()` had a return type of... `_Projection`... not `boolean`...

src/components/plots/stackedAreaPlot.ts
```diff
@@ -69,6 +69,13 @@ export module Plots {
       return this;
     }
 
+    protected _updateExtentsForAttr(attr: string) {
+      super._updateExtentsForAttr(attr);
+      if ((attr === "x" || attr === "y") && this._projectorsReady()) {
+        this._updateStackExtents();
+      }
```
bluong: Why can't we leverage StackedPlot to do this?

src/components/plots/stackedBarPlot.ts
```diff
@@ -73,12 +73,15 @@ export module Plots {
...
+      super._updateExtentsForAttr(attr);
+      if ((attr === "x" || attr === "y") && this._projectorsReady()) {
+        this._updateStackExtents();
+      }
     }
 
-    protected _normalizeDatasets<A, B>(fromX: boolean): {a: A; b: B}[] {
-      return Stacked.prototype._normalizeDatasets.call(this, fromX);
+    protected _getPlotMetadataForDataset(key: string): StackedPlotMetadata {
+      return Stacked.prototype._getPlotMetadataForDataset.call(this, key);
```
bluong: Strange did stackedBarPlot never get the plotMetadata from StackedPlot before?  Why is this new code?

src/components/plots/stackedBarPlot.ts
```diff
@@ -73,12 +73,15 @@ export module Plots {
       return this;
     }
 
-    protected _getPlotMetadataForDataset(key: string): StackedPlotMetadata {
-      return Stacked.prototype._getPlotMetadataForDataset.call(this, key);
+    protected _updateExtentsForAttr(attr: string) {
+      super._updateExtentsForAttr(attr);
+      if ((attr === "x" || attr === "y") && this._projectorsReady()) {
+        this._updateStackExtents();
+      }
```
bluong: Code should be on the StackedPlot

src/components/plots/stackedPlot.ts
```diff
@@ -62,10 +62,16 @@ module Plottable {
...
+      var filter = this._isVertical ? this._yFilter() : this._xFilter();
+
       var maxStackExtent = Utils.Methods.max<string, number>(this._datasetKeysInOrder, (k: string) => {
         var dataset = this._key2PlotDatasetKey.get(k).dataset;
         var plotMetadata = <Plots.StackedPlotMetadata>this._key2PlotDatasetKey.get(k).plotMetadata;
-        return Utils.Methods.max<any, number>(dataset.data(), (datum: any, i: number) => {
+        var data = dataset.data();
+        if (filter != null) {
+          data = data.filter((d, i) => filter(d, i, dataset, plotMetadata));
+        }
```
bluong: Can we clean the above with a ternary?

```typescript
var data = filter != null ? dataset.data() : dataset.data().filter((d, i) => filter(d, i, dataset, plotMetadata));
```

src/components/plots/stackedPlot.ts
```diff
@@ -74,7 +80,11 @@ module Plottable {
       var minStackExtent = Utils.Methods.min<string, number>(this._datasetKeysInOrder, (k: string) => {
         var dataset = this._key2PlotDatasetKey.get(k).dataset;
         var plotMetadata = <Plots.StackedPlotMetadata>this._key2PlotDatasetKey.get(k).plotMetadata;
-        return Utils.Methods.min<any, number>(dataset.data(), (datum: any, i: number) => {
+        var data = dataset.data();
+        if (filter != null) {
+          data = data.filter((d, i) => filter(d, i, dataset, plotMetadata));
+        }
```
bluong: Same as above

src/components/plots/xyPlot.ts
```diff
@@ -71,6 +71,49 @@ module Plottable {
...
     }
 
+    protected _updateExtentsForAttr(attr: string) {
+      if (attr === "x") {
+        return super._updateExtentsForAttr(attr, this._xFilter());
+      } else if (attr === "y") {
+        return super._updateExtentsForAttr(attr, this._yFilter());
+      }
+      return super._updateExtentsForAttr(attr);
+    }
```
bluong: Hmm... I'm getting a really a nervous feeling about this design...  Can't put my finger on it, but this seems really weird...

bluong: Would a better design be... to have `_updateExtentsForAttr` to expose an endpoint that this can override?  Now we are putting the idea of filtering on the super class when it doesn't seem like it should exist there

src/components/plots/xyPlot.ts
```diff
@@ -158,69 +201,26 @@ module Plottable {
...
     public showAllData() {
+      this._updateExtentsForAttr("x");
       this._xScale.autoDomain();
-      if (!this._autoAdjustYScaleDomain) {
-        this._yScale.autoDomain();
-      }
+      this._updateExtentsForAttr("y");
+      this._yScale.autoDomain();
+      return this;
     }
```
bluong: So are we keeping this endpoint?

src/components/plots/xyPlot.ts
```diff
@@ -71,6 +71,49 @@ module Plottable {
...
+          if (xScale != null) {
+            return (d, i, dataset, metadata) => {
+              var range = xScale.range();
+              return Utils.Methods.inRange(xScale.scale(xAccessor(d, i, dataset, metadata)), range[0], range[1]);
+            };
+          }
+        }
+      }
+      return null;
+    }
```
bluong: The function above seems to be basically a lot of repeats of `_xFilter`.  Can we just have a single filter and then have `_xFilter` / `_yFilter` just call this general function?  We'd save about 10 lines or so of code with this

src/components/plots/xyPlot.ts
```diff
@@ -158,69 +201,26 @@ module Plottable {
...
-    private _adjustDomainOverVisiblePoints<A, B>(values: {a: A; b: B}[], filterFn: (v: any) => boolean): B[] {
-      var bVals = values.filter(v => filterFn(v.a)).map(v => v.b);
-      var retVal: B[] = [];
-      if (bVals.length !== 0) {
-        retVal = [Utils.Methods.min<B>(bVals, null), Utils.Methods.max<B>(bVals, null)];
+        this._updateExtentsForAttr("x");
+        this._xScale.autoDomain();
       }
-      return retVal;
     }
```
bluong: Assuming this code works... hurrah for removal of somewhat confusing code

src/components/plots/stackedAreaPlot.ts
jtlan: Bugs me a bit since we'd be tapping in to a `protected` method...

src/components/plots/stackedBarPlot.ts
jtlan: It's not new, there's just a git display problem. See above, lines 76~78, for the old location of the code.

src/components/plots/stackedPlot.ts
jtlan: We could, but I'm not sure that's cleaner. I find in general that ternaries that don't evaluate to simply a constant value are kind of hard to read. The way I wrote the code reflects this logic:
 - get the data from the `Dataset`
 - If there is a filter, use it to filter the data.

src/components/plots/xyPlot.ts
jtlan: The tricky part is really you need to apply a filter inside `_computeExtent`, but you don't know what the filter is unless you have a the key being set. Maybe we should have a `D3.Map` of filters, and just try to apply the filter for a particular attribute every time? This way we don't need to override `_updateExtentsForAttr()` on `XYPlot`.

bluong: I think you're along the lines of what I'm suggesting.  I'm not asking to override _updateExtentsForAttr but instead setup a filter injection point inside the parent implementation that xyPlot can inject into

src/components/plots/xyPlot.ts
jtlan: (I assume that you're referring to the fact that `_xFilter()` and `_yFilter()` look similar.) I think that was the mistake we made with `normalizeDatasets():. Reluctant to repeat code, instead we added an obscure-looking method whose logic is hard to follow. The logic only applies to X being filtered based on Y, and Y being filtered based on X, so I don't think it's worth being generalized.

src/components/plots/xyPlot.ts
jtlan: It won't work with just calling `autoDomain()` on both scales, since we actually have to change the extents for this to work. Not sure if we want to keep it though, since the effect could be accomplished by shutting filtering off, calling autoDomain(), then turning the filtering back on.

src/components/plots/stackedAreaPlot.ts
bluong: Yes but now we have duplication of logic when they should derive from a single point of truth

src/components/plots/xyPlot.ts
bluong: I remember there used to be some confusion on the point of this endpoint.  If this is useful to users, then keep it.  Otherwise don't.

src/components/plots/xyPlot.ts
bluong: We'd save a bunch of lines as well as we won't run the risk of desync later down the line.  Note that the only pieces of information that define `_xFilter()` / `_yFilter()` are the key used as well as if the instance variable is `this._autoAdjustYScaleDomain` or `this._autoAdjustXScaleDomain`.  All the rest of the code is generalizable and probably should be.

src/components/plots/xyPlot.ts
jtlan: The map of filters is kind of annoying because the filter now has to look up the scale each time it's executed:
```Typescript
var filterByX = (datum: any, index: number, dataset: Dataset, metadata: Plots.PlotMetadata) => {
  var xBinding = this._attrBindings.get("x");
  if (xBinding != null) {
    var xAccessor = xBinding.accessor;
    var xScale = xBinding.scale;
    if (xScale != null) {
      var range = xScale.range();
      return Utils.Methods.inRange(xScale.scale(xAccessor(d, i, dataset, metadata)), range[0], range[1]);
    }
  }
  return true;
}
```
This is because at the time `automaticallyAdjustYScaleOverVisiblePoints(true) is called, the projectors may not have been set up correctly.

src/components/plots/stackedAreaPlot.ts
jtlan: We lose our type safety though, since we have to do this:
```Typescript
protected _updateExtentsForAttr(attr: string) {
  return (<any> Stacked.prototype)._updateExtentsForAttr.call(this, attr);
}
```
I think the fact that we're running into diamond problems is a strong sign that we've structured the stacked inheritance incorrectly.

bluong: `Plots.Stacked` should from my gathering be some sort of useful utility.  What we're running into from my view is that `Plots.StackedBar`/`Plots.StackedArea` need to extend from their respective parents but needs to utilize some object or thing in order to figure out offsets.

src/components/plots/plot.ts
```diff
@@ -270,20 +270,23 @@ module Plottable {
...
-    private _updateExtentsForAttr(attr: string) {
+    protected _updateExtentsForAttr(attr: string, filter?: _Accessor) {
       var binding = this._attrBindings.get(attr);
       var accessor = binding.accessor;
       var extents = this._datasetKeysInOrder.map((key) => {
         var plotDatasetKey = this._key2PlotDatasetKey.get(key);
         var dataset = plotDatasetKey.dataset;
         var plotMetadata = plotDatasetKey.plotMetadata;
-        return this._computeExtent(dataset, accessor, plotMetadata);
+        return this._computeExtent(dataset, accessor, plotMetadata, filter);
```
bluong: Err I'm unsure what you're meaning by a map of filters, but I was referring that this `filter` argument here should not be inputted via an argument but instead through possibly like a `getFilterForAttr()` protected endpoint

src/components/plots/plot.ts
```diff
@@ -33,6 +33,7 @@ module Plottable {
     protected _renderArea: D3.Selection;
     protected _attrBindings: D3.Map<_Projection>;
     protected _attrExtents: D3.Map<any[]>;
+    protected _attrFilters: D3.Map<_Accessor>;
```
bluong: How high in the hierarchy should filters go?  If it makes sense to exist here, then sure it should exist here.  However, if it should exist at the `XYPlot` level we should try to push to make sure it stays there

src/components/plots/plot.ts
```diff
@@ -270,20 +272,24 @@ module Plottable {
       this._scales().forEach((scale) => scale._autoDomainIfAutomaticMode());
     }
 
-    private _updateExtentsForAttr(attr: string) {
+    protected _updateExtentsForAttr(attr: string) {
       var binding = this._attrBindings.get(attr);
       var accessor = binding.accessor;
+      var filter = this._attrFilters.get(attr);
```
bluong: The comment about the _attrFIlters instance variable connects the implementation of filter here

src/components/plots/plot.ts
```diff
@@ -270,20 +272,24 @@ module Plottable {
...
       });
       this._attrExtents.set(attr, extents);
     }
 
-    private _computeExtent(dataset: Dataset, accessor: _Accessor, plotMetadata: any): any[] {
+    private _computeExtent(dataset: Dataset, accessor: _Accessor, plotMetadata: any, filter?: _Accessor): any[] {
       var data = dataset.data();
+      if (filter) {
+        data = data.filter((d, i) => filter(d, i, dataset, plotMetadata));
+      }
```
bluong: Somewhat neater but not as performant is to have a filter that passes everything through it as the default.

src/components/plots/plot.ts
```diff
@@ -323,6 +329,21 @@ module Plottable {
...
+          var accessor = attrBinding.accessor;
+          var scale = attrBinding.scale;
+          if (scale != null) {
+            var range = scale.range();
+            return Utils.Methods.inRange(scale.scale(accessor(datum, index, dataset, metadata)), range[0], range[1]);
+          }
+        }
+        return true;
+      };
+    }
```
bluong: Same comment about the instance variables ties to this implementation.  Also, it is nice that it works for all attributes, but does it make sense to?

bluong:
Please consider the implementation you had before because I think it was cleaner before and the scope of that change was smaller than this current change.

I think if you use proper subclassing in the previous implementation, you should arrive at a cleaner much better scoped answer.

test/testReference.ts
```diff
@@ -26,6 +26,7 @@
 ///<reference path="components/selectionBoxLayerTests.ts" />
 
 ///<reference path="components/plots/plotTests.ts" />
+///<reference path="components/plots/XYPlotTests.ts" />
```
bluong: xyPlotTests Let's stay with the convention that filenames should start with a lowercase letter

bluong:
filenaming

jtlan:
Enough has changed on **apiRefactor** that I thought it was best to recreate this based on the new logic. The remake is at #2030.

Issue #2014, opened 2015-05-11
Extent Computation needs to take scale type into account


jtlan:
Because coercers were removed in #2007, the `_computeExtent()` method on `Plot` now needs to account for the intended use of the values returned by the `Accessor` and stringify if necessary.

dbt:
is this closed with #2091?

aicioara:
No, they are not related.

PR #2016, opened 2015-05-11
deleted unnecessary tests

crmorford:


bluong:
Quicktests are from my view are QE domain.  If QE feels that these tests are unnecessary, I'm good with it.

quicktests/overlaying/tests/basic/grid_plot.js
```diff
@@ -1,32 +0,0 @@
-
-function makeData() {
-  "use strict";
-
-    var data = [{name: "jon", y: 10, type: "q1"}, {name: "dan", y: 2, type: "q1"}, {name: "zoo", y: 1, type: "q1"}];
```
jtlan: This part caught #2014, but I think we should accomplish that by having messy/wrong-type data in other tests.

jtlan:
As @bluong said, you're the master of these, so do what you think is appropriate. I think there's a few key things some of them are testing that we should keep in mind as we build newer tests.

aicioara:
Should we delete them or, instead, just make them easier to follow (have a notion of loading, better comparison highlight, etc)?

bluong:
@aicioara I think the approach that we're going with is along the lines of these tests are not necessarily useful and that even if they are useful, their usefulness is limited and probably should be replaced by better tests.  From what I hear, @crmorford is looking into better tests than the ones being deleted.

aicioara:
@bluong But should we not first implement the solution then delete the obsolete? (Just asking I probably have no context)

bluong:
I would defer to @crmorford here...

I think the point of this pr is remove the extraneous cruft in the quicktests branch.

crmorford:
These aren't catching any bugs. No reason to keep them :(

PR #2017, opened 2015-05-11
[Interaction] Interaction fix

### LINKED ISSUE #2012: Detaching Interactions when they're already detached causes errors
Calling `detachFrom()` on an `Interaction` that's not currently attached to anything throws errors:

Repro: http://jsfiddle.net/zLerabxg/

**Actual**: An error appears in the console:
screen shot 2015-05-11 at 11 54 17 am
**Expected**: No error should be thrown, and nothing else should happen either.

Calling `detachFrom()` twice produces a slightly different error:
screen shot 2015-05-11 at 11 56 10 am
http://jsfiddle.net/zLerabxg/1/

aicioara:
Closes #2012
Calling `detachFrom` twice used to throw an error. Fixed the error and added comprehensive testing for the testcase. Waiting for a merge, then opening this up

Before 1: http://jsfiddle.net/zLerabxg/
After 1:http://jsfiddle.net/zLerabxg/2/

Before 2:http://jsfiddle.net/zLerabxg/1/
After 2: http://jsfiddle.net/zLerabxg/3/

### DIFF
--- a/src/interactions/interaction.ts
+++ b/src/interactions/interaction.ts
@@ -44,7 +44,9 @@ module Plottable {
      * @return {Interaction}
      */
     public detachFrom(component: Component) {
-      this._unanchor();
+      if (this._isAnchored) {
+        this._unanchor();
+      }
       this._componentAttachedTo = null;
       component.offAnchor(this._anchorCallback);
 
--- a/test/interactions/interactionTests.ts
+++ b/test/interactions/interactionTests.ts
@@ -71,7 +71,7 @@ describe("Interactions", () => {
       svg.remove();
     });
 
-    it("calling detachFrom on a detached Interaction has no effect", () => {
+    it("calling detachFrom() on a detached Interaction has no effect", () => {
       var svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
       var component = new Plottable.Component();
 
--- a/test/interactions/interactionTests.ts
+++ b/test/interactions/interactionTests.ts
@@ -71,6 +71,34 @@ describe("Interactions", () => {
       svg.remove();
     });
 
+    it("calling detach on a detached interaction has no effect", () => {
+      var svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var component = new Plottable.Component();
+
+      var clickInteraction = new Plottable.Interactions.Click();
+
+      assert.doesNotThrow(() => {
+        clickInteraction.detachFrom(component);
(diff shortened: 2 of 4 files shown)


src/interactions/interaction.ts
```diff
@@ -44,7 +44,9 @@ module Plottable {
      * @return {Interaction}
      */
     public detachFrom(component: Component) {
-      this._unanchor();
+      if (this._isAnchored) {
```
jtlan: Wouldn't it make more sense if this `if`-block encompassed all of the logic, rather than just `_unanchor()`? Although I don't really expect anything bad to happen.

aicioara: Not really because we want detachFrom to revert the interaction to its original state (not be anchored to anything). Say this bit of code
```typescript
var interaction = new Plottable.Interaction();
var component = new Plottable.Component();
interaction.attachTo(component); // nothing anchored yet
interaction.detachFrom(component);

assert(interaction._componentAttachedTo === null)

Comment (jtlan) [inline src/interactions/interaction.ts]: I see, that makes sense.

File: test/interactions/interactionTests.ts [hunk 2]
```diff
@@ -71,6 +71,34 @@ describe("Interactions", () => {
...
+
+      assert.doesNotThrow(() => {
+        clickInteraction.detachFrom(component);
+      }, "Detaching an interaction which was not attached should not throw");
+
+      clickInteraction.attachTo(component);
+      clickInteraction.detachFrom(component);
+      assert.doesNotThrow(() => {
+        clickInteraction.detachFrom(component);
+      }, "calling detaching a component twice should not throw");
```
jtlan: "calling detach() on a Component twice should not throw an error"

test/interactions/interactionTests.ts
```diff
@@ -71,6 +71,34 @@ describe("Interactions", () => {
...
+
+      component.renderTo(svg);
+
+      clickInteraction.attachTo(component);
+      clickInteraction.detachFrom(component);
+      assert.doesNotThrow(() => {
+        clickInteraction.detachFrom(component);
+      }, "calling detaching a component twice should not throw even when attached to SVG");
```
jtlan: "calling detachFrom() twice should not throw and error even if the Component is anchored"

test/interactions/interactionTests.ts
```diff
@@ -71,6 +71,34 @@ describe("Interactions", () => {
       svg.remove();
     });
 
+    it("calling detachFrom on a detached Interaction has no effect", () => {
```
jtlan: parens :(

PR #2019, opened 2015-05-12
Other properties

bluong:
Adding `x1`/`x2`/`y1`/`y2` endpoints to `Plots.Rectangle` as well as `symbol` and `size` to `Plots.Scatter`


jtlan:
Looks reasonable.

PR #2023, opened 2015-05-12
[timeAxis & timeScale] TimeAxis configurations no longer expose d3.time

aicioara:
Instead we defined our own enum called TimeInterval, which we expose to the public API.
The reasoning behind this change is that we do not want d3 implementation details (internals) to leak out to the user.

### DIFF
--- a/src/components/axes/timeAxis.ts
+++ b/src/components/axes/timeAxis.ts
@@ -2,6 +2,11 @@
 
 module Plottable {
 export module Axes {
+
+  export enum TimeInterval {
+    second, minute, hour, day, month, year
+  };
+
   /**
    * Defines a configuration for a time axis tier.
    * For details on how ticks are generated see: https://github.com/mbostock/d3/wiki/Time-Scales#ticks
@@ -531,6 +536,22 @@ export module Axes {
         }
       });
     }
+
+    private _getD3TimeInterval(timeInterval: TimeInterval) {
+      switch(timeInterval) {
+        case TimeInterval.second:
+          return d3.time.second;
+        case TimeInterval.minute:
+          return d3.time.minute;
+        case TimeInterval.hour:
+          return d3.time.hour;
+        case TimeInterval.day:
+          return d3.time.day;
+        case TimeInterval.month:
+          return d3.time.month;
+      }
+    }
+
   }
 }
 }
--- a/src/components/axes/timeAxis.ts
+++ b/src/components/axes/timeAxis.ts
@@ -1,12 +1,12 @@
 ///<reference path="../../reference.ts" />
 
 module Plottable {
+
+export enum TimeInterval {
+  second, minute, hour, day, month, year
+};
+
 export module Axes {
-
-  export enum TimeInterval {
-    second, minute, hour, day, month, year
-  };
-
   /**
    * Defines a configuration for a time axis tier.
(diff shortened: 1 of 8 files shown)


src/utils/formatters.ts
```diff
@@ -190,6 +190,31 @@ module Plottable {
...
+        return d3.time.day;
+      case TimeInterval.week:
+        return d3.time.week;
+      case TimeInterval.month:
+        return d3.time.month;
+      case TimeInterval.year:
+        return d3.time.year;
+      default:
+        Utils.Methods.warn("TimeInterval specified does not exist");
+        return d3.time.year;
```
aicioara: The reason this default is here is the rendering logic. If this function fails and throws an exception, the exception is going to be caught by https://github.com/palantir/plottable/blob/timeD3/src/core/renderController.ts#L95 and the renderer is going to loop infinitely. We need to either keep this default or fix the renderer. 

In order to test the renderer fix, leave out lines 209 and 210 (the year case).

src/components/axes/timeAxis.ts
```diff
@@ -1,6 +1,11 @@
 ///<reference path="../../reference.ts" />
 
 module Plottable {
+
+export enum TimeInterval {
+  second, minute, hour, day, week, month, year
```
jtlan: Is it possible to use a "`string`y enum", so that the values are human-readable after compilation?

```Typescript
export module TimeInterval {
  export var second = "second";
  export var minute = "minute";
  ...
}
```

bluong: `string`y enums also have the advantage of using the raw string values for those insistent on using raw strings (usually Javascript consumers)

aicioara: yeah, but that's not going to be an enum anymore. Can do something like this http://stackoverflow.com/questions/15490560/create-an-enum-with-string-values-in-typescript

src/utils/formatters.ts
jtlan: Okay, I see the problem. Would it make sense for there to be a `try`-`catch` around invocations of `timeIntervalToD3Time()`?

How come the infinite loop didn't happen before?

src/components/axes/timeAxis.ts
jtlan: Is there a problem with it not being an actual `enum`? The type information will still check out for Typescript consumers, and for Javascript consumers it's about the same, except with readable `string` values.

src/utils/formatters.ts
aicioara: Probably nothing failed inside the try catch. If something fails (say this method has a bug, or the (JavaScript) user passes a wrong value into the axisConfiguration), then infinite loop (I believe). It happened when I mistyped some things in this method.

src/components/axes/timeAxis.ts
aicioara: actual enums are numbers only (afaik)

src/components/axes/timeAxis.ts
```diff
@@ -1,6 +1,17 @@
 ///<reference path="../../reference.ts" />
 
 module Plottable {
+
+export class TimeInterval {
```
jtlan: Are we worried that people can instantiate a `TimeInterval`? Not that there's anything they could do with it, but making `TimeInterval` a `module ` would prevent that right?

jtlan:
Great, this feels more readable. We can also change the time library we use without affecting the API.

PR #2025, opened 2015-05-12
Repaired renderer

aicioara:
Issue:  The try, catch block inside the `rendererController.flush()`. If anything inside here (and there's a lot in her) fails and throws an exception, the exception is going to be caught by https://github.com/palantir/plottable/blob/timeD3/src/core/renderController.ts#L95 and not thrown immediately. Also, the renderer is going to loop infinitely, as it keeps rescheduling the component to a render. Removed the rescheduling bit. 

Reasoning behind it: From a discussion with @jtlan , the render should be (and I hope is) a pure function (no side effect), hence it should not modify the model. Since it does not modify the model, nothing changes. Since nothing changes, the component render will fail forever.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -949,8 +949,7 @@ var Plottable;
                 case Plottable.TimeInterval.year:
                     return d3.time.year;
                 default:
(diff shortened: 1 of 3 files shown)


aicioara:
Waiting for the rest of the base to merge

PR #2029, opened 2015-05-13
Repaired RenderController again

jtlan:
b98a94b687de50b87143e2e64a001b4b7e46d4b2 caused a problem by removing the declaration of a new `_componentsToRender` `Set` before the loop where all the `Components` were drawn. That change caused a bug because `Component`s that were enqueued by another `Component`'s `render()` call would never be redrawn.

Added a test to catch this bug.

### DIFF
--- a/test/core/renderControllerTests.ts
+++ b/test/core/renderControllerTests.ts
@@ -1,0 +1,23 @@
+///<reference path="../testReference.ts" />
+
+var assert = chai.assert;
+
+describe("RenderController", () => {
+  it("Components whose render() is triggered by another Component's render() will be drawn", () => {
+    var link1 = new Plottable.Component();
+    var svg1 = TestMethods.generateSVG();
+    var link2 = new Plottable.Component();
+    var svg2 = TestMethods.generateSVG();
+    link2.renderTo(svg2);
+
+    (<any> link1)._render = () => link2.render();
+    var link2Rendered = false;
+    (<any> link2)._render = () => link2Rendered = true;
+
+    link1.renderTo(svg1);
+    assert.isTrue(link2Rendered, "dependent Component was eventually drawn");
+
+    svg1.remove();
+    svg2.remove();
+  });
+});
--- a/test/core/renderControllerTests.ts
+++ b/test/core/renderControllerTests.ts
@@ -16,7 +16,7 @@ describe("RenderController", () => {
     (<any> link2)._render = () => link2Rendered = true;
 
     link1.render();
-    assert.isTrue(link2Rendered, "dependent Component was eventually drawn");
+    assert.isTrue(link2Rendered, "dependent Component was render()-ed");
 
     svg1.remove();
     svg2.remove();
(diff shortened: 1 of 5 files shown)


test/core/renderControllerTests.ts
```diff
@@ -0,0 +1,23 @@
...
+describe("RenderController", () => {
+  it("Components whose render() is triggered by another Component's render() will be drawn", () => {
+    var link1 = new Plottable.Component();
+    var svg1 = TestMethods.generateSVG();
+    var link2 = new Plottable.Component();
+    var svg2 = TestMethods.generateSVG();
+    link2.renderTo(svg2);
+
+    (<any> link1)._render = () => link2.render();
+    var link2Rendered = false;
```
bluong: Err the semantics are a bit confusing why is it ordered such that...
```typescript
link2.renderTo(svg2)
```
Doesn't this mean that link2 is rendered, but if so, then why is
```typescript
var link2Rendered = false;
```
directly afterwards?

jtlan: Good catch. I should probably do `link2.anchor(svg2)` instead.

test/core/renderControllerTests.ts
```diff
@@ -0,0 +1,24 @@
...
+    var link2 = new Plottable.Component();
+    var svg2 = TestMethods.generateSVG();
+    link2.anchor(svg2).computeLayout();
+
+    (<any> link1)._render = () => link2.render();
+    var link2Rendered = false;
+    (<any> link2)._render = () => link2Rendered = true;
+
+    link1.render();
+    assert.isTrue(link2Rendered, "dependent Component was eventually drawn");
```
bluong: remove the "eventually".  This implies that you are waiting around for the `Component` to be drawn but as the render is running immediately the term throws that fact off.

test/core/renderControllerTests.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    svg1.remove();
+    svg2.remove();
+  });
```
bluong: The test mostly looks good.  I think ideally you should be above to connect more directly to the `RenderController` so that we have a much more unitized test, but that said if you think this test is suitable enough, it's a good concise test.

jtlan: There's not much value in checking the internal value of the `RenderController`; this indirect method really gets at the actual bug.

aicioara:
I would have liked a more real world scenario (no private method invocation). Maybe the quicktest which was failing? I still could not see how the previous PR broke the render (which was the particular edge case). However, if you believe the real world scenario is not "unit" enough, I think this should work.

jtlan:
@aicioara wrote:
>I would have liked a more real world scenario (no private method invocation). Maybe the quicktest which was failing? I still could not see how the previous PR broke the render (which was the particular edge case). However, if you believe the real world scenario is not "unit" enough, I think this should work.

The description of the bug is basically:
>"`Components` that enqueue for render() during flushing are still rendered"

So this is basically the smallest possible repro. I agree that I'd rather not mock `protected` fields to do this -- so let me come up with a better way to write it.

aicioara:
Sounds good

PR #2030, opened 2015-05-13
Autorange using extents

jtlan:
(Remake of #2010)

Previously, some goofy filtering was done to set the Y domain as a function of points visible on X, or vice versa.

This commit causes the dependent `Scale`'s extent to be computed with a filter based on what points are visible on the independent `Scale`.

Probably we should give the methods that enable the behavior more reasonable names...


src/components/plots/plot.ts
```diff
@@ -268,24 +269,40 @@ module Plottable {
...
+        extents: D3.Map<any[]>, filter?: _Accessor) {
       var accScaleBinding = bindings.get(key);
       if (accScaleBinding.accessor == null) { return; }
       extents.set(key, this._datasetKeysInOrder.map((key) => {
         var plotDatasetKey = this._key2PlotDatasetKey.get(key);
         var dataset = plotDatasetKey.dataset;
         var plotMetadata = plotDatasetKey.plotMetadata;
-        return this._computeExtent(dataset, accScaleBinding.accessor, plotMetadata);
+        return this._computeExtent(dataset, accScaleBinding.accessor, plotMetadata, filter);
       }));
```
bluong: The above code looks fine but it is a bit messy... However I can't think of much to clean it up...

That said, I feel that having an optional parameter in filter is messy since filter could now be `null`, `undefined` or valid

Should `_updateExtentsForAttr()` for now pass in null and then have `_updateExtentsForKey()` have the parameter be required?

Alternative clean solutions are preferred but I'm unsure

jtlan: Passing in `null` can signify explicitly that no filter is wanted for attributes, but I'm not sure how easy it is to read that intention from looking at the code.

src/components/plots/xyPlot.ts
```diff
@@ -49,6 +49,10 @@ module Plottable {
       }
 
       this._bindProperty(XYPlot._X_KEY, x, xScale);
+      if (this._autoAdjustYScaleDomain) {
+        this._updateExtentsForProperty("y");
+        this.y().scale.autoDomain();
```
bluong: `this.y().scale` can be null.  Please guard against this

src/components/plots/xyPlot.ts
```diff
@@ -63,11 +67,39 @@ module Plottable {
       }
 
       this._bindProperty(XYPlot._Y_KEY, y, yScale);
+      if (this._autoAdjustXScaleDomain) {
+        this._updateExtentsForProperty("x");
+        this.x().scale.autoDomain();
```
bluong: Same as above

src/components/plots/xyPlot.ts
```diff
@@ -175,69 +207,26 @@ module Plottable {
...
-      }));
-    }
-
-    private _adjustDomainOverVisiblePoints<A, B>(values: {a: A; b: B}[], filterFn: (v: any) => boolean): B[] {
-      var bVals = values.filter(v => filterFn(v.a)).map(v => v.b);
-      var retVal: B[] = [];
-      if (bVals.length !== 0) {
-        retVal = [Utils.Methods.min<B>(bVals, null), Utils.Methods.max<B>(bVals, null)];
+        this._updateExtentsForProperty("x");
+        this.x().scale.autoDomain();
```
bluong: Same as above

src/components/plots/xyPlot.ts
```diff
@@ -175,69 +207,26 @@ module Plottable {
...
+      this.y().scale.autoDomain();
+      return this;
     }
 
     private _adjustYDomainOnChangeFromX() {
       if (!this._projectorsReady()) { return; }
       if (this._autoAdjustYScaleDomain) {
-        this._adjustDomainToVisiblePoints<X, Y>(this.x().scale, this.y().scale, true);
+        this._updateExtentsForProperty("y");
+        this.y().scale.autoDomain();
```
bluong: Same as above

bluong:
Currently the way the code is set up, xScales / yScales are always defined... but that may not be the case in the future.  Thus please null check in the above cases

PR #2031, opened 2015-05-13
[Scale] - Removing copy() function from scales

bluong:


src/components/axes/categoryAxis.ts
```diff
@@ -135,13 +129,15 @@ export module Axes {
      * @param {string[]} ticks The strings that will be printed on the ticks.
      */
     private _measureTicks(axisWidth: number, axisHeight: number, scale: Scales.Category, ticks: string[]) {
+      var axisSpace = this._isHorizontal() ? axisWidth : axisHeight;
+      var expectedRangeBand = axisSpace / (2 * scale.outerPadding() + (ticks.length - 1) * scale.innerPadding() + ticks.length);
```
aicioara: Why is this something that is expected? Should it be just rangeBand? Also, I can't figure out what is going on over here, which means the math is not well explained / not well encapsulated. I had to take pen and paper and still can't figure out if this is correct actually.

aicioara:
Rest is good

src/components/axes/categoryAxis.ts
bluong: It's `expectedRangeBand` because the rangeBand of the scale cannot be calculated until the range is set.  Thus, it is an expected `rangeBand` of the scale "if it had the range given by the `axisSpace variable`.

The math behind it is yeah a bit complexish:  I'll explain below.
 
The entirety of a `Scales.Category` can consist of some number of bands as well as paddings in-between the bands as well as paddings between the edge of the scale and the outer bands.

Thus,
let Wb = bandWidth
let Wo = outerPaddingWidth
let Wi = innerPaddingWidth
let Nb = number of bands
let Wt = total width

Thus, we get the following equation:
Wt = 2 * Wo + (Nb - 1) * Wi + Nb * Wb

Now by definition of `Scales.Category`, the width of each of the paddings is equivalent to the respective padding proportion on the scale multiplied by the bandWidth.

So,
let Pi = innerPadding = Wi / Wb
let Po = outerPadding = Wo / Wb

And thus,
Wt = 2 * Po * Wb + (Nb - 1) * Pi * Wb + Nb * Wb

And since we are solving for Wb, we just need to do a bit of algebra...
Wb = Wt / (2 * Po + (Nb - 1) * Pi + Nb)

aicioara: When I read inner padding, I was assuming a size (like innerPaddingWidth) (note this is exactly CSS meaning of padding). So I believe that is why the entire thing is confusing. I suggest either comment something similar to the above on the method or extract bits of the equation into variables

plottable.js
```diff
@@ -4945,12 +4903,16 @@ var Plottable;
              */
             Category.prototype._measureTicks = function (axisWidth, axisHeight, scale, ticks) {
                 var _this = this;
+                var axisSpace = this._isHorizontal() ? axisWidth : axisHeight;
+                var totalOuterPadding = 2 * scale.outerPadding();
```
aicioara: totalOuterPaddingRatio
Padding to me means a concrete measure, while this is the ratio between padding width and band width

bluong:
Squishing this PR into another PR since our Travis is in a bit of chaos...

PR #2034, opened 2015-05-13
[Table] - Hiding _IterateLayoutResult type

bluong:
Note the underscore in front.  We should probably not expose this type

### DIFF
--- a/src/components/table.ts
+++ b/src/components/table.ts
@@ -9,7 +9,7 @@ export module Components {
     wantsHeightArr: boolean[];
   }
 
-  export type _IterateLayoutResult = {
+  type _IterateLayoutResult = {
     colProportionalSpace: number[];
     rowProportionalSpace: number[];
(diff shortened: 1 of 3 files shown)


src/components/table.ts
```diff
@@ -9,7 +9,7 @@ export module Components {
     wantsHeightArr: boolean[];
   }
 
-  export type _IterateLayoutResult = {
+  type _IterateLayoutResult = {
```
rcchen: Should we just inline this and remove the type? It's only being used in one place

bluong: If it's only being used in one place, it's probably way not worth it to abstract it out to a type

rcchen: Two places in table.ts. I stand corrected; this is fine as is.

bluong: The type is being used to encapsulate the large object type for an instance variable.  We should actually probably keep the type.

PR #2046, opened 2015-05-13
Fix bug on Scales.ModifiedLog

jtlan:
Domain wasn't being set in the constructor with a method ... aaaah!!

Default domain is now [0, base].

Moved tests to a separate file.


src/scales/modifiedLogScale.ts
```diff
@@ -88,11 +88,11 @@ export module Scales {
...
-      return this.untransformedDomain;
+      return this._untransformedDomain;
     }
 
     protected _setDomain(values: number[]) {
-      this.untransformedDomain = values;
+      this._untransformedDomain = values;
       var transformedDomain = [this.adjustedLog(values[0]), this.adjustedLog(values[1])];
       this._d3Scale.domain(transformedDomain);
       this._dispatchUpdate();
```
aicioara: instead of lines 97 and 98 -> super._setDomain(transformedDoman)

jtlan: Good catch.

aicioara:
How come this was not causing a bug? Wasn't dispatchUpdate important? And if it was causing a bug, we should add a new test for this

jtlan:
@aicioara wrote:
> How come this was not causing a bug? Wasn't dispatchUpdate important? And if it was causing a bug, we should add a new test for this

It wasn't `disptachUpdate` that was the problem, it was the fact that the domain wasn't actually being set correctly. The use of an internal D3 scale didn't really help that at all. I'll add a test for the specific scenario, but it's going to look a bit weird.

jtlan:
Also the previous bug was that the default extent was `[0, 1]`, which didn't produce the correct behavior that the tests depended on -- except that the scale wasn't set up properly. Now that the scale's default extent is `[0, base]`, the problem won't occur any more.

PR #2049, opened 2015-05-13
[Component] - deleting the fixedHeight/fixedWidth flags

bluong:
The idea of fixedHeight and fixedWidth should only be overwritten by overriding the functions, not the flag instance variables as well.


bluong:
Squishing this PR into another PR since our Travis is in a bit of chaos...

Issue #2064, opened 2015-05-14
`Plots.Line` and `Plots.Area` should take a generic for Y


jtlan:
Currently the signature for the two classes look like this:
```Typescript
export class Line<X> extends XYPlot<X, number>
export class Area<X> extends Line<X>
```

But the type shown on the Y-axis doesn't have to be a `number`, so the signatures should take a second generic.

bluong:
Currently we are running with the idea that the `Y` domain should always be numbers

PR #2068, opened 2015-05-14
Generic accessor

bluong:
* `_Accessor` is now exposed as `Accessor` and will be referred to as such
* `Accessor` now takes in a generic type param to signify the return type of the Accessor function
* `Plots.Pie` no longer requires a generic type as the types are applied to the specific property functions

### DIFF
--- a/src/components/plots/areaPlot.ts
+++ b/src/components/plots/areaPlot.ts
@@ -29,9 +29,9 @@ export module Plots {
     }
 
     public y0(): Plots.AccessorScaleBinding<number, number>;
-    public y0(y0: number | _Accessor): Area<X>;
-    public y0(y0: number | _Accessor, y0Scale: Scale<number, number>): Area<X>;
-    public y0(y0?: number | _Accessor, y0Scale?: Scale<number, number>): any {
+    public y0(y0: number | Accessor): Area<X>;
+    public y0(y0: number | Accessor, y0Scale: Scale<number, number>): Area<X>;
+    public y0(y0?: number | Accessor, y0Scale?: Scale<number, number>): any {
       if (y0 == null) {
         return this._propertyBindings.get(Area._Y0_KEY);
       }
--- a/src/components/plots/areaPlot.ts
+++ b/src/components/plots/areaPlot.ts
@@ -29,9 +29,9 @@ export module Plots {
     }
 
     public y0(): Plots.AccessorScaleBinding<number, number>;
-    public y0(y0: number | Accessor): Area<X>;
-    public y0(y0: number | Accessor, y0Scale: Scale<number, number>): Area<X>;
-    public y0(y0?: number | Accessor, y0Scale?: Scale<number, number>): any {
+    public y0(y0: number | Accessor<number>): Area<X>;
+    public y0(y0: number | Accessor<number>, y0Scale: Scale<number, number>): Area<X>;
+    public y0(y0?: number | Accessor<number>, y0Scale?: Scale<number, number>): any {
       if (y0 == null) {
         return this._propertyBindings.get(Area._Y0_KEY);
       }
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -48,17 +48,14 @@ export module Plots {
       return [{attrToProjector: this._generateAttrToProjector(), animator: this._getAnimator("cells")}];
     }
 
-    public x(): Plots.AccessorScaleBinding<any, number>;
-    public x(x: number | Accessor): Grid;
-    public x(x: any | Accessor, scale: Scale<any, number>): Grid;
-    public x(x?: number | Accessor | any, scale?: Scale<any, number>): any {
+    public x(x?: number | Accessor<number> | any | Accessor<any>, scale?: Scale<any, number>): any {
       if (x == null) {
         return super.x();
       }
       if (scale == null) {
-        super.x(<number | Accessor> x);
+        super.x(<number | Accessor<number>> x);
       } else {
-        super.x(<any | Accessor> x, scale);
+        super.x(<any | Accessor<any>> x, scale);
         if (scale instanceof Scales.Category) {
           this.x1((d, i, dataset, m) => scale.scale(this.x().accessor(d, i, dataset, m)) - scale.rangeBand() / 2);
           this.x2((d, i, dataset, m) => scale.scale(this.x().accessor(d, i, dataset, m)) + scale.rangeBand() / 2);
@@ -69,17 +66,14 @@ export module Plots {
       return this;
     }
 
-    public y(): Plots.AccessorScaleBinding<any, number>;
-    public y(y: number | Accessor): Grid;
-    public y(y: any | Accessor, scale: Scale<any, number>): Grid;
-    public y(y?: number | Accessor | any, scale?: Scale<any, number>): any {
+    public y(y?: number | Accessor<number> | any, scale?: Scale<any, number>): any {
       if (y == null) {
         return super.y();
       }
       if (scale == null) {
-        super.y(<number | Accessor> y);
+        super.y(<number | Accessor<number>> y);
       } else {
-        super.y(<any | Accessor> y, scale);
+        super.y(<any | Accessor<any>> y, scale);
         if (scale instanceof Scales.Category) {
           this.y1((d, i, dataset, m) => scale.scale(this.y().accessor(d, i, dataset, m)) - scale.rangeBand() / 2);
           this.y2((d, i, dataset, m) => scale.scale(this.y().accessor(d, i, dataset, m)) + scale.rangeBand() / 2);
--- a/src/components/plots/gridPlot.ts
+++ b/src/components/plots/gridPlot.ts
@@ -49,16 +49,16 @@ export module Plots {
     }
 
     public x(): Plots.AccessorScaleBinding<any, number>;
-    public x(x: number | _Accessor): Grid;
-    public x(x: any | _Accessor, scale: Scale<any, number>): Grid;
-    public x(x?: number | _Accessor | any, scale?: Scale<any, number>): any {
+    public x(x: number | Accessor): Grid;
+    public x(x: any | Accessor, scale: Scale<any, number>): Grid;
+    public x(x?: number | Accessor | any, scale?: Scale<any, number>): any {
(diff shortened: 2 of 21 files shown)


src/components/plots/areaPlot.ts
```diff
@@ -29,9 +29,9 @@ export module Plots {
     }
 
     public y0(): Plots.AccessorScaleBinding<number, number>;
-    public y0(y0: number | _Accessor): Area<X>;
-    public y0(y0: number | _Accessor, y0Scale: Scale<number, number>): Area<X>;
-    public y0(y0?: number | _Accessor, y0Scale?: Scale<number, number>): any {
+    public y0(y0: number | Accessor<number>): Area<X>;
+    public y0(y0: number | Accessor<number>, y0Scale: Scale<number, number>): Area<X>;
+    public y0(y0?: number | Accessor<number>, y0Scale?: Scale<number, number>): any {
```
jtlan: This will be changed when #2064 is done, right?

src/components/plots/gridPlot.ts
```diff
@@ -48,17 +48,14 @@ export module Plots {
       return [{attrToProjector: this._generateAttrToProjector(), animator: this._getAnimator("cells")}];
     }
 
-    public x(): Plots.AccessorScaleBinding<any, number>;
-    public x(x: number | _Accessor): Grid;
-    public x(x: any | _Accessor, scale: Scale<any, number>): Grid;
-    public x(x?: number | _Accessor | any, scale?: Scale<any, number>): any {
+    public x(x?: number | Accessor<number> | any | Accessor<any>, scale?: Scale<any, number>): any {
```
jtlan: Shouldn't we specify that the `Accessor` outputs the same type as the `Scale` takes in?
And what happened to the other signatures?

src/components/plots/plot.ts
```diff
@@ -283,12 +284,12 @@ module Plottable {
       this._updateExtentsForKey(property, this._propertyBindings, this._propertyExtents, this._filterForProperty(property));
     }
 
-    protected _filterForProperty(property: string): _Accessor {
+    protected _filterForProperty(property: string): Accessor<any> {
```
jtlan: `Accessor<boolean>` is the correct type here, since it's a filter.

src/components/plots/gridPlot.ts
jtlan: I see the problem, it's because at the top we have:
```Typescript
export class Grid extends Rectangle<any, any>
```
We should probably maintain the generics:
```Typescript
export class Grid<X, Y> extends Rectangle<X, Y>
```

src/components/plots/gridPlot.ts
```diff
@@ -69,17 +66,14 @@ export module Plots {
       return this;
     }
 
-    public y(): Plots.AccessorScaleBinding<any, number>;
-    public y(y: number | _Accessor): Grid;
-    public y(y: any | _Accessor, scale: Scale<any, number>): Grid;
-    public y(y?: number | _Accessor | any, scale?: Scale<any, number>): any {
+    public y(y?: number | Accessor<number> | any, scale?: Scale<any, number>): any {
```
jtlan: Same comments regarding using a generic here, by having the `Grid` class also accept generics.

src/components/plots/piePlot.ts
```diff
@@ -98,10 +98,10 @@ export module Plots {
...
     }
 
-    public outerRadius(): AccessorScaleBinding<D, number>;
-    public outerRadius(outerRadius: number | _Accessor): Plots.Pie<D>;
-    public outerRadius(outerRadius: D | _Accessor, scale: Scale<D, number>): Plots.Pie<D>;
-    public outerRadius(outerRadius?: number | _Accessor | D, scale?: Scale<D, number>): any {
+    public outerRadius<O>(): AccessorScaleBinding<O, number>;
+    public outerRadius(outerRadius: number | Accessor<number>): Plots.Pie;
+    public outerRadius<O>(outerRadius: O | Accessor<O>, scale: Scale<O, number>): Plots.Pie;
+    public outerRadius<O>(outerRadius?: number | Accessor<number> | O | Accessor<O>, scale?: Scale<O, number>): any {
```
jtlan: Would it make more sense if `innerRadius()` and `outerRadius()` took the same generic type?
```Typescript
```

src/components/plots/linePlot.ts
```diff
@@ -23,7 +23,7 @@ export module Plots {
       this._defaultStrokeColor = new Scales.Color().range()[0];
     }
 
-    protected _rejectNullsAndNaNs(d: any, i: number, dataset: Dataset, plotMetadata: any, accessor: _Accessor) {
+    protected _rejectNullsAndNaNs(d: any, i: number, dataset: Dataset, plotMetadata: any, accessor: Accessor<X> | Accessor<number>) {
```
jtlan: This function has a goofy signature. Is it possible to just inline it where it's used?

src/components/plots/plot.ts
```diff
@@ -142,9 +142,10 @@ module Plottable {
     }
 
     public attr<D>(attr: string): Plots.AccessorScaleBinding<D, number | string>;
-    public attr(attr: string, attrValue: number | string | _Accessor): Plot;
-    public attr<D>(attr: string, attrValue: D | _Accessor, scale: Scale<D, number | string>): Plot;
-    public attr<D>(attr: string, attrValue?: number | string | _Accessor | D, scale?: Scale<D, number | string>): any {
+    public attr(attr: string, attrValue: number | string | Accessor<number> | Accessor<string>): Plot;
+    public attr<D>(attr: string, attrValue: D | Accessor<D>, scale: Scale<D, number | string>): Plot;
+    public attr<D>(attr: string, attrValue?: number | string | Accessor<number> | Accessor<string> | D | Accessor<D>,
```
jtlan: Here you used `<D>` for the generic, presumably because `<D>` corresponds to "domain" on the `Scale`. Elsewhere, you used a letter that corresponded to the property being set, such as `sectorValue<S>`. Do we want to use `<A>` here instead, or is it different because it's `attr()` and not a property-setter?

src/components/plots/plot.ts
```diff
@@ -283,12 +284,12 @@ module Plottable {
...
     }
 
-    protected _filterForProperty(property: string): _Accessor {
+    protected _filterForProperty(property: string): Accessor<any> {
       return null;
     }
 
     private _updateExtentsForKey(key: string, bindings: D3.Map<Plots.AccessorScaleBinding<any, any>>,
-        extents: D3.Map<any[]>, filter: _Accessor) {
+        extents: D3.Map<any[]>, filter: Accessor<any>) {
```
jtlan: `Accessor<boolean>` for the filter here.

src/components/plots/plot.ts
```diff
@@ -299,7 +300,7 @@ module Plottable {
       }));
     }
 
-    private _computeExtent(dataset: Dataset, accessor: _Accessor, plotMetadata: any, filter: _Accessor): any[] {
+    private _computeExtent(dataset: Dataset, accessor: Accessor<any>, plotMetadata: any, filter: Accessor<any>): any[] {
```
jtlan: `Accessor<boolean>` for the filter here.

jtlan:
[-1] Some comments on the types of generics used in various places.

src/components/plots/areaPlot.ts
bluong: Correct

src/components/plots/piePlot.ts
bluong: It's an unnecessary restriction in my opinion.  Also, to truly enforce this, I believe you need the generic on the plot level rather the function level.

src/components/plots/plot.ts
bluong: I like `<A>` let's use it.

src/components/plots/gridPlot.ts
bluong: Technically out of scope but let's see if I can make the conversion without any problems

src/components/plots/piePlot.ts
jtlan: It's more of a "using the same name suggests to the user that they should be the same type"

src/components/plots/gridPlot.ts
```diff
@@ -16,16 +16,16 @@ export module Plots {
...
      * to use for each grid cell.
      */
-    constructor(xScale: Scale<any, any>, yScale: Scale<any, any>) {
+    constructor(xScale: Scale<X, any>, yScale: Scale<Y, any>) {
       super(xScale, yScale);
       this.classed("grid-plot", true);
 
       // The x and y scales should render in bands with no padding for category scales
       if (xScale instanceof Scales.Category) {
-        xScale.innerPadding(0).outerPadding(0);
```
jtlan: How did this even work before...?

src/components/plots/rectanglePlot.ts
```diff
@@ -63,10 +63,10 @@ export module Plots {
...
     }
 
-    public x1(): AccessorScaleBinding<X, number>;
-    public x1(x1: number | _Accessor): Plots.Rectangle<X, Y>;
-    public x1(x1: X | _Accessor, scale: Scale<X, number>): Plots.Rectangle<X, Y>;
-    public x1(x1?: number | _Accessor | X, scale?: Scale<X, number>): any {
+    public x1<X1>(): AccessorScaleBinding<X1, number>;
+    public x1(x1: number | Accessor<number>): Plots.Rectangle<X, Y>;
+    public x1<X1>(x1: X1 | Accessor<X1>, scale: Scale<X1, number>): Plots.Rectangle<X, Y>;
+    public x1<X1>(x1?: number | Accessor<number> | X1 | Accessor<X1>, scale?: Scale<X1, number>): any {
```
jtlan: Why `<X1>` and `<X2>`? Don't they have to be the same type as the generic `<X>` on the `Plots.Rectangle`?

jtlan:
[-1] Questions on the generic type used on `Rectangle`.

src/components/plots/gridPlot.ts
bluong: Cause xScale was of type Scale<any, any>

jtlan: That doesn't necessarily imply that it has the methods on `Scales.Category`...

PR #2069, opened 2015-05-14
[Scale] - Remove d3 scale input to constructor

bluong:
We shouldn't be exposing the fact that D3 backs our scales in our API.  That's purely an implementation detail


src/scales/categoryScale.ts
```diff
@@ -3,7 +3,6 @@
 module Plottable {
 export module Scales {
   export class Category extends Scale<string, number> {
-    protected _d3Scale: D3.Scale.OrdinalScale;
```
aicioara: Is there a good reason why this was deleted? I see two reasons to keep it:
- When reading code, you don't have to navigate to a different file to find the variable
- You don't need those ugly casts (which get uglier in the next files) every time you want to use the scale.

If we decide to keep it, this comment applies to all future files (including those where the `protected _d3Scale` was not present in the first place (for consistency)

bluong: So, unless I'm mistaken the technique that is used here is called "shadowing" where `this._d3Scale` in this context shares the same reference as the one in the parent class.

One problem I have with this technique is that if the parent class were to change it's instance variable name, we can easily enter a scenario where the child class's shadowed instance variable is accidentally not also renamed and thus we have 2 instance variables when we should have one.

To give an example, I think this has caused a bug in the past that was really hard to detect.  The `_isVertical` flag I believe used to be named `isVertical`.  However, for some reason it was used in quite a few places in the hierarchy.  Someone (probably I) changed `isVertical` to `_isVertical` and forgot to change this fact on some of the child classes.  Thankfully I think we wrote our code such that we didn't run into user-facing bugs, but this sounds like something that could easily be dangerous in a quiet way which is quite scary to me.

What do you think?

aicioara: That's why Java has `@Override`. Hmm...

aicioara: Can we go around this? I really don't like the casts. Do something like
`this._d3Scale = <cast_here>super._d3Scale`
This will solve that issue

src/components/plots/gridPlot.ts
```diff
@@ -59,9 +59,10 @@ export module Plots {
         super.x(<number | _Accessor> x);
       } else {
         super.x(<any | _Accessor> x, scale);
+        var catScale = (<Scales.Category> <any> scale);
```
aicioara: move it inside the `if`

src/components/plots/gridPlot.ts
```diff
@@ -80,9 +81,10 @@ export module Plots {
         super.y(<number | _Accessor> y);
       } else {
         super.y(<any | _Accessor> y, scale);
+        var catScale = (<Scales.Category> <any> scale);
```
aicioara: same

src/components/plots/gridPlot.ts
```diff
@@ -22,10 +22,10 @@ export module Plots {
 
       // The x and y scales should render in bands with no padding for category scales
       if (xScale instanceof Scales.Category) {
-        xScale.innerPadding(0).outerPadding(0);
+        <Scales.Category> xScale.innerPadding(0).outerPadding(0);
```
aicioara: Why was this cast not necessary before?

src/scales/interpolatedColorScale.ts
```diff
@@ -10,7 +10,7 @@ export module Scales {
    *
    * By default it generates a linear scale internally.
    */
-  export class InterpolatedColor extends Scale<number, string> {
+  export class InterpolatedColor extends Scale<number, number> {
```
aicioara: Why this change?

src/scales/quantitativeScale.ts
```diff
@@ -33,29 +18,23 @@ module Plottable {
...
      */
     public invert(value: number): D {
-      return <any> this._d3Scale.invert(value);
-    }
-
-    public domain(): D[];
-    public domain(values: D[]): QuantitativeScale<D>;
-    public domain(values?: D[]): any {
-      return super.domain(values); // need to override type sig to enable method chaining:/
+      return null;
```
aicioara: Isn't `throw` a better alternative?

src/scales/scale.ts
```diff
@@ -93,7 +89,7 @@ module Plottable {
      * @returns {R} The range value corresponding to the supplied domain value.
      */
     public scale(value: D): R {
-      return this._d3Scale(value);
+      return null;
```
aicioara: Same, isn't throw a better alternative?

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
...
         this._domainModificationInProgress = true;
-        this._d3Scale.domain(values);
+        this._setActualDomain(values);
         this._dispatchUpdate();
         this._domainModificationInProgress = false;
       }
     }
 
+    protected _setActualDomain(values: D[]) {
+      // Subclasses should override
```
aicioara: Throw?

src/scales/scale.ts
```diff
@@ -158,13 +158,21 @@ module Plottable {
...
         return this;
       }
     }
 
+    protected _getRange(): R[] {
+      return [];
+    }
+
+    protected _setRange(values: R[]) {
+      // Subclasses should override
```
aicioara: Throw?

src/scales/timeScale.ts
```diff
@@ -47,6 +46,38 @@ export module Scales {
...
+    public invert(value: number) {
+      return this._d3Scale.invert(value);
+    }
+
+    public getDefaultTicks(): Date[] {
+      return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+    }
+
+    public _niceDomain(domain: Date[], count?: number) {
+      return Utils.D3Scale.niceDomain(<D3.Scale.QuantitativeScale> <D3.Scale.Scale> this._d3Scale, domain, count);
```
aicioara: Is there any workaround for this? Double casting is so ugly.

test/components/numericAxisTests.ts
```diff
@@ -307,8 +307,8 @@ describe("NumericAxis", () => {
     var yLabel = new Plottable.Components.Label("LABEL", "left");
     yLabel.classed(Plottable.Components.Label.AXIS_LABEL_CLASS, true);
     var barPlot = new Plottable.Plots.Bar(xScale, yScale);
-    barPlot.x((d) => d.x, xScale);
-    barPlot.y((d) => d.y, yScale);
+    barPlot.x((d: any) => d.x, xScale);
+    barPlot.y((d: any) => d.y, yScale);
```
aicioara: Why is this needed now and not before?

aicioara:
Regarding the `throw`s, at least log a warning, if not really throw. Returning `null` is ugly
Regarding the `why is it needed`, I was just curious

src/scales/timeScale.ts
bluong: Unless I'm mistaken `D3.Scale.TimeScale` is actually a child of `D3.Scale.QuantitativeScale` but the d3.d.ts file does not say this point.  Let me check/verify

src/components/plots/gridPlot.ts
bluong: Unsure...

test/components/numericAxisTests.ts
bluong: Compilation error now.  I'm actually unsure why it fails now and not before

src/scales/interpolatedColorScale.ts
bluong: After diving into the code, it forced me to make this contract.  Looking at it again, this does look odd.  @rcchen do you have any context on why the compiler would force me to make this decision?

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
...
       if (!this._domainModificationInProgress) {
         this._domainModificationInProgress = true;
-        this._d3Scale.domain(values);
+        this._setActualDomain(values);
         this._dispatchUpdate();
         this._domainModificationInProgress = false;
       }
     }
 
+    protected _setActualDomain(values: D[]) {
```
jtlan: What's this? This name is really confusing, since there's already a `_setDomain()` method.

bluong: This is specifically to set the domain of usually the backing D3 scale.  I was completely unsure of the name as it's basically `_setDomain()` of the backing scale but I wasn't sure of how to go about it.  I mainly didn't want to rewrite the previous implementation of `setDomain()` over and over as that can lead to inconsistencies.

src/scales/categoryScale.ts
```diff
@@ -142,7 +142,23 @@ export module Scales {
...
+    protected _setBackingDomain(values: string[]) {
+      this._d3Scale.domain(values);
+    }
+
+    protected _getRange() {
+      return this._d3Scale.range();
+    }
+
+    protected _setRange(values: number[]) {
+      this._d3Scale.range(values);
```
aicioara: This looks so Java-ish (i.e. verbose)... but sure

src/scales/linearScale.ts
```diff
@@ -13,15 +14,46 @@ export module Scales {
...
+      return this._d3Scale.invert(value);
+    }
+
+    public getDefaultTicks(): number[] {
+      return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+    }
+
+    public _niceDomain(domain: number[], count?: number): number[] {
+      return Utils.D3Scale.niceDomain(this._d3Scale, domain, count);
+    }
```
aicioara: Move `public` above `protected`

src/scales/linearScale.ts
```diff
@@ -13,15 +14,46 @@ export module Scales {
...
+    public invert(value: number) {
+      return this._d3Scale.invert(value);
+    }
+
+    public getDefaultTicks(): number[] {
+      return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+    }
+
+    public _niceDomain(domain: number[], count?: number): number[] {
+      return Utils.D3Scale.niceDomain(this._d3Scale, domain, count);
```
aicioara: `public` underscored?

src/scales/modifiedLogScale.ts
```diff
@@ -97,6 +99,10 @@ export module Scales {
       super._setDomain(transformedDomain);
     }
 
+    protected _setBackingDomain(values: number[]) {
```
aicioara: Move protected after public

src/scales/modifiedLogScale.ts
```diff
@@ -204,6 +210,18 @@ export module Scales {
...
+      return this._d3Scale.range();
+    }
+
+    protected _setRange(values: number[]) {
+      this._d3Scale.range(values);
+    }
+
+    public getDefaultTicks(): number[] {
+      return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+    }
```
aicioara: `public` < `protected`

src/scales/quantitativeScale.ts
```diff
@@ -33,29 +18,23 @@ module Plottable {
...
      */
     public invert(value: number): D {
-      return <any> this._d3Scale.invert(value);
-    }
-
-    public domain(): D[];
-    public domain(values: D[]): QuantitativeScale<D>;
-    public domain(values?: D[]): any {
-      return super.domain(values); // need to override type sig to enable method chaining:/
+      throw new Error("subclass should override _niceDomain");
```
aicioara: Capitalize Sublcass

src/scales/quantitativeScale.ts
```diff
@@ -33,29 +18,23 @@ module Plottable {
...
+      }
+      super._setDomain(values);
     }
 
     /**
      * Gets ticks generated by the default algorithm.
      */
     public getDefaultTicks(): D[] {
-        return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+      throw new Error("subclass should override _niceDomain");
```
aicioara: capitalize + function name

src/scales/quantitativeScale.ts
```diff
@@ -72,7 +51,7 @@ module Plottable {
      * numbers.
      */
     public _niceDomain(domain: D[], count?: number): D[] {
-      return this._d3Scale.copy().domain(domain).nice(count).domain();
+      throw new Error("subclass should override _niceDomain");
```
aicioara: capitalize

src/scales/scale.ts
```diff
@@ -93,7 +89,7 @@ module Plottable {
      * @returns {R} The range value corresponding to the supplied domain value.
      */
     public scale(value: D): R {
-      return this._d3Scale(value);
+      throw new Error("subclasses should override functionality");
```
aicioara: capitalize. Now we're not consistent, everywhere above it was stated the function name in the error

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
       }
     }
 
-    protected _getDomain() {
-      return this._d3Scale.domain();
+    protected _getDomain(): D[] {
```
aicioara: I don't think you need the return type in the function definition. Check the `plottable.d.ts` if it is autogenerated anyway

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
...
+      throw new Error("subclasses should override _getDomain");
```
aicioara: Capitalize

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
...
         this._domainModificationInProgress = true;
-        this._d3Scale.domain(values);
+        this._setBackingDomain(values);
         this._dispatchUpdate();
         this._domainModificationInProgress = false;
       }
     }
 
+    protected _setBackingDomain(values: D[]) {
+      throw new Error("subclasses should override _setBackingDomain");
```
aicioara: Capitalize

src/scales/scale.ts
```diff
@@ -158,13 +158,21 @@ module Plottable {
...
-        return this._d3Scale.range();
+        return this._getRange();
       } else {
-        this._d3Scale.range(values);
+        this._setRange(values);
         return this;
       }
     }
 
+    protected _getRange(): R[] {
```
aicioara: Not sure if the return type is needed (as above)

src/scales/scale.ts
```diff
@@ -158,13 +158,21 @@ module Plottable {
...
+      throw new Error("subclasses should override _getRange");
+    }
+
+    protected _setRange(values: R[]) {
```
aicioara: Not sure if the return type is needed (as above)

src/scales/timeScale.ts
```diff
@@ -47,6 +46,38 @@ export module Scales {
...
+      return this._d3Scale.invert(value);
+    }
+
+    public getDefaultTicks(): Date[] {
+      return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
+    }
+
+    public _niceDomain(domain: Date[], count?: number) {
+      return Utils.D3Scale.niceDomain(this._d3Scale, domain, count);
+    }
```
aicioara: Same issue with `public` > `protected`, public underscored. That's why I said it is super verbose and prone to repetition

src/scales/timeScale.ts
```diff
@@ -47,6 +46,38 @@ export module Scales {
...
+
+    public scale(value: Date): number {
+      return this._d3Scale(value);
+    }
+
+    protected _getDomain() {
+      return this._d3Scale.domain();
+    }
+
+    protected _setBackingDomain(values: Date[]) {
```
aicioara: method name does not make sense. What is a backing domain?

src/scales/scale.ts
```diff
@@ -122,19 +118,23 @@ module Plottable {
...
       if (!this._domainModificationInProgress) {
         this._domainModificationInProgress = true;
-        this._d3Scale.domain(values);
+        this._setBackingDomain(values);
         this._dispatchUpdate();
         this._domainModificationInProgress = false;
       }
     }
 
+    protected _setBackingDomain(values: D[]) {
```
aicioara: All of these should be documented in the superclass (here). As a developer I am not even sure what is the contract I am getting into if I decide to extend the classes and I have to implement these methods.

src/scales/linearScale.ts
bluong: We currently don't have a rule surrounding this.  I'm also unsure about the usefulness of this rule...

src/scales/linearScale.ts
bluong: `_niceDomain` has not been cleared from the codebase.  I believe @jtlan 's `Domainer` changes is planning to address this

src/scales/modifiedLogScale.ts
bluong: As above

src/scales/modifiedLogScale.ts
bluong: As above

src/scales/timeScale.ts
bluong: As above

src/scales/timeScale.ts
bluong: Since there's gotta be something backing the domain of the scale but it is never explicitly set anywhere, the domain had to actually be set somehwere.  There is an implementation of `_setDomain()` that has to tap into how to actually set the domain of the scale, and I'm unsure of how this name should be.

aicioara: @bluong  What is the difference between `_setDomain()` and `_setBackingDomain()` and why do we need both?

aicioara:
Document at least `setBackingDomain()`

src/scales/scale.ts
bluong: I'll get to documentation after we get our API into a good state.  I'm mainly trying to focus on functionality and once we have the refactor into a somewhat stable state, I'll go document this.

src/scales/timeScale.ts
bluong: `_setDomain()` has the job of making sure that the domain of the scale is set to the according value.  However, this does not necessarily concern setting the domain of whatever backing scale.

I do note that the naming is confusing, and I'm completely open to suggestions on how to change the design of how to implement this idea.

src/scales/linearScale.ts
```diff
@@ -13,15 +14,46 @@ export module Scales {
...
+
+    public scale(value: number) {
+      return this._d3Scale(value);
+    }
+
+    protected _getDomain() {
+      return this._d3Scale.domain();
+    }
+
+    protected _setBackingDomain(values: number[]) {
```
aicioara: Rename to `_setBackingScaleDomain()`

PR #2070, opened 2015-05-14
Add min() and max() on QuantitativeScale.

jtlan:
<del>Set/get restrictions on the results of `autoDomain()`, without having to subclass `Domainer`.</del>

**Edit 2015-05-15 1:30 PDT**: get/set min and max values of the domain set by `autoDomain()`.


jtlan:
Removing [apiRefactor] since this is a feature (although on the **apiRefactor** branch). QE should review this.

src/scales/linearScale.ts
```diff
@@ -17,6 +16,8 @@ export module Scales {
     constructor(scale: D3.Scale.LinearScale);
     constructor(scale?: any) {
       super(scale == null ? d3.scale.linear() : scale);
+      this.autoMin(-Infinity);
```
aicioara: This name does not tell me anything. I believe from our previous discussions it refers to the feature where the autoDomain is bounded by some min or max value. If that is the case, call it `autoDomainMin`. or even better `autoDomainMinValue`.

src/scales/quantitativeScale.ts
```diff
@@ -23,7 +25,26 @@ module Plottable {
     }
 
     protected _getExtent(): D[] {
-      return this._domainer.computeDomain(this._getAllExtents(), this);
+      var computedExtent = this._domainer.computeDomain(this._getAllExtents(), this);
+      var e0 = computedExtent[0];
+      var e1 = computedExtent[1];
```
aicioara: call them extentLeft and extentRight (or something more explicit)

src/scales/quantitativeScale.ts
```diff
@@ -23,7 +25,26 @@ module Plottable {
...
+          e1 = this._autoMin;
+        }
+      }
+      if (this._autoMax != null) {
+        if (e0 > this._autoMax) {
+          e0 = this._autoMax;
+        }
+        if (e1 > this._autoMax) {
+          e1 = this._autoMax;
+        }
```
aicioara: You may prefer Math.min(e1, this.autoMax). Less code, but I am okay with either

src/scales/quantitativeScale.ts
```diff
@@ -42,6 +63,40 @@ module Plottable {
...
+    /**
+     * Gets the minimum domain value when autoDomain()-ing. 
+     */
+    public autoMin(): D;
+    /**
+     * Sets the minimum domain value when autoDomain()-ing.
+     */
+    public autoMin(value: D): QuantitativeScale<D>;
+    public autoMin(value?: D): any {
+      if (value === undefined) {
```
aicioara: `value == null` to match the rest of our codebase

src/scales/quantitativeScale.ts
```diff
@@ -42,6 +63,40 @@ module Plottable {
...
+        return this._autoMin;
+      }
+      this._autoMin = value;
+      this._autoDomainIfAutomaticMode();
```
aicioara: Side issue, why do we use both `auto` and `automatic` in the name of this function?

src/scales/quantitativeScale.ts
```diff
@@ -42,6 +63,40 @@ module Plottable {
...
+      return this;
+    }
+
+    /**
+     * Gets the maximum domain value when autoDomain()-ing.
```
aicioara: The comment is somewhat vague. It's rather the `domain maximum value constraint`

src/scales/timeScale.ts
```diff
@@ -16,6 +16,9 @@ export module Scales {
     constructor(scale?: D3.Scale.LinearScale) {
       // need to cast since d3 time scales do not descend from QuantitativeScale scales
       super(scale == null ? (<any>d3.time.scale()) : scale);
+      // Minimum and maximum dates; http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1
+      this.autoMin(new Date(-8640000000000000));
+      this.autoMax(new Date(8640000000000000));
```
aicioara: Magic number. Extract it into a static private variable.

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,58 @@
...
+      var scale = new Plottable.Scales.Linear();
+      assert.strictEqual(scale.autoMax(), Infinity, "autoMax() defaults to +Infinity");
+      var desiredDomain = [-10, 10];
+      scale.addExtentsProvider((scale: Plottable.Scales.Linear) => [desiredDomain]);
+      scale.autoMax(0);
+      assert.strictEqual(scale.domain()[0], desiredDomain[0], "lower end of domain was set to desired value");
+      assert.strictEqual(scale.domain()[1], 0, "upper end of domain was set to autoMax() value");
+    });
+  });
+});
```
aicioara: We should avoid doing such things in future PRs. It's very hard to spot the differences. We should open separate PRs for code movement (without altering)

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,58 @@
...
+      scale.domain([0, 10]);
+      var ticks = scale.ticks();
+      assert.closeTo(ticks.length, 10, 1, "ticks were generated correctly with default generator");
+      scale.tickGenerator((scale) => scale.getDefaultTicks().filter(tick => tick % 3 === 0));
+      ticks = scale.ticks();
+      assert.deepEqual(ticks, [0, 3, 6, 9], "ticks were generated correctly with custom generator");
+    });
+  });
+
+  describe("Linear", () => {
```
aicioara: Extract into own file.

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,58 @@
...
+      scale.autoDomain();
+      assert.deepEqual(scale.domain(), [0, 1]);
+    });
+
+    it("autoMin()", () => {
+      var scale = new Plottable.Scales.Linear();
+      assert.strictEqual(scale.autoMin(), -Infinity, "autoMin() defaults to -Infinity");
+      var desiredDomain = [-10, 10];
+      scale.addExtentsProvider((scale: Plottable.Scales.Linear) => [desiredDomain]);
+      scale.autoMin(0);
```
aicioara: Extract the 0 into variable `constrainedMinimum`

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    it("autoMax()", () => {
+      var scale = new Plottable.Scales.Linear();
+      assert.strictEqual(scale.autoMax(), Infinity, "autoMax() defaults to +Infinity");
+      var desiredDomain = [-10, 10];
+      scale.addExtentsProvider((scale: Plottable.Scales.Linear) => [desiredDomain]);
+      scale.autoMax(0);
+      assert.strictEqual(scale.domain()[0], desiredDomain[0], "lower end of domain was set to desired value");
+      assert.strictEqual(scale.domain()[1], 0, "upper end of domain was set to autoMax() value");
```
aicioara: Same as above

test/scales/timeScaleTests.ts
```diff
@@ -2,41 +2,67 @@
...
+
+    it("autoMin()", () => {
+      var scale = new Plottable.Scales.Time();
+      // Minimum and maximum dates; http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1
+      assert.strictEqual(scale.autoMin().getTime(), -8640000000000000, "default autoMin() is smallest possible date");
+      var desiredDomain = [new Date(-1000), new Date(1000)];
+      scale.addExtentsProvider((scale: Plottable.Scales.Time) => [desiredDomain]);
+      scale.autoMin(new Date(0));
+      var domain = scale.domain();
+      assert.strictEqual(domain[0].getTime(), 0, "lower end of domain was set to autoMin() value");
```
aicioara: Same as above

test/scales/timeScaleTests.ts
```diff
@@ -2,41 +2,67 @@
...
-    scale.domain([new Date(2000, 0, 1, 0, 0, 0, 0), new Date(2000, 0, 1, 0, 1, 0, 0)]);
-    ticks = scale.tickInterval(Plottable.TimeInterval.second);
-    assert.strictEqual(ticks.length, 61, "generated correct number of ticks");
+    it("autoMax()", () => {
+      var scale = new Plottable.Scales.Time();
+      // Minimum and maximum dates; http://ecma-international.org/ecma-262/5.1/#sec-15.9.1.1
+      assert.strictEqual(scale.autoMax().getTime(), 8640000000000000, "default autoMax() is largest possible date");
+      var desiredDomain = [new Date(-1000), new Date(1000)];
+      scale.addExtentsProvider((scale: Plottable.Scales.Time) => [desiredDomain]);
+      scale.autoMax(new Date(0));
```
aicioara: Same as above

src/scales/quantitativeScale.ts
jtlan: `Math.min` only works on `number`s, so I am using `<` and `>` to have it work on `number`-like objects.

src/scales/quantitativeScale.ts
jtlan: `_autoDomainIfAutomaticMode()` is getting renamed, but I think "`autoDomain()`" is a good name.

src/scales/linearScale.ts
jtlan: Presumably it's a value, so I think `autoDomainMin` is fine.

aicioara: I don't think we should "presume" things in public methods. I still think about it as "auto domain minimum... what"?

src/scales/quantitativeScale.ts
aicioara: Fair enough

src/scales/quantitativeScale.ts
```diff
@@ -42,6 +51,44 @@ module Plottable {
...
+     */
+    public min(): D;
+    /**
+     * Sets the minimum domain value when autoDomain()-ing.
+     * 
+     * @returns {QuantitativeScale<D>} The calling QuantitativeScale.
+     */
+    public min(value: D): QuantitativeScale<D>;
+    public min(value?: D): any {
+      if (value === undefined) {
```
aicioara: `==` instead of `===`

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,56 @@
...
+      scale.domain([0, 10]);
+      var ticks = scale.ticks();
+      assert.closeTo(ticks.length, 10, 1, "ticks were generated correctly with default generator");
+      scale.tickGenerator((scale) => scale.getDefaultTicks().filter(tick => tick % 3 === 0));
+      ticks = scale.ticks();
+      assert.deepEqual(ticks, [0, 3, 6, 9], "ticks were generated correctly with custom generator");
+    });
+  });
+
+  describe("Linear", () => {
```
aicioara: Either extract into own file, or remove the `Linear` from the definition, since you are actually testing autoDomains & restrictions

src/scales/quantitativeScale.ts
jtlan: In this case, passing `null` is used to remove a `min()`. I'll add it to the docs so it's clear.

test/scales/quantitativeScaleTests.ts
jtlan: Seems like the tests above it also use a `Linear` scale, since currently it's not possible to test with a `QuantitativeScale` directly.

aicioara: Just call them Lienar then

src/scales/quantitativeScale.ts
aicioara: This is so weird as it does not happen anywhere else in our codebase. I suggest adding a third method called `removeMinMaxConstraints()` or something which removes constraints for both min and max

jtlan: `.parent()` on `Component` takes `null` as a valid param (but I also wrote that).

We do have a lot of places in our codebase with `=== undefined`, but I think that's an error. I'll make a ticket.

jtlan: Having a separate method for clearing the value seems odd, since we don't really do that elsewhere. Maybe we should define a `static` constant to signify no min or max?

jtlan:
Given the trouble we're having naming these, I think we have the wrong API for this feature.

PR #2072, opened 2015-05-15
[Plot] - Default keys

bluong:
Moving appropriate defaults from `generateAttrToProjector()` to the constructor so that appropriate getter calls to those attirbutes can be called safely.

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -14,7 +14,6 @@ export module Plots {
     private _labelFormatter: Formatter = Formatters.identity();
     private _labelsEnabled = false;
     private _hideBarsIfAnyAreTooWide = true;
-    private _defaultFillColor: string;
 
     /**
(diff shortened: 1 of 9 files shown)


src/components/plots/barPlot.ts
```diff
@@ -27,12 +26,13 @@ export module Plots {
...
       super(xScale, yScale);
       this.classed("bar-plot", true);
-      this._defaultFillColor = new Scales.Color().range()[0];
       this.animator("bars-reset", new Animators.Null());
       this.animator("bars", new Animators.Base());
       this.animator("baseline", new Animators.Null());
       this._isVertical = isVertical;
       this.baseline(0);
+      this.attr("fill", new Scales.Color().range()[0]);
+      this.attr("width", () => this._getBarPixelWidth());
```
jtlan: I see why this was so slow now...

PR #2075, opened 2015-05-15
[Plots.Bar] - BarPlots vertical by default, change to horizontal via `.orientation()`

### LINKED ISSUE #2063: `Plots.Bar` constructor should take string-y enum instead of boolean
Currently the `Plots.Bar` constructor takes a boolean to set the orientation:
```Typescript
var barPlot = new Plottable.Plots.Bar(xScale, yScale, true);
```

A `string`y `enum` would make it much clearer:
```Typescript
var barPlot = new Plottable.Plots.Bar("vertical");
```

bluong:
Going with the convention that boolean parameters are bad... The `isVertical` parameter has now been replaced with a mode string that should be either one of
`Plottable.Plots.Bars.Mode.HORIZONTAL = "horizontal"` or
`Plottable.Plots.Bars.Mode.VERTICAL = "vertical"`

Names above are up for debate

Fixes #2063


src/components/plots/barPlot.ts
```diff
@@ -2,6 +2,12 @@
 
 module Plottable {
 export module Plots {
+  export module Bars {
+    export class Mode {
+      public static VERTICAL = "vertical";
+      public static HORIZONTAL = "horizontal";
```
rcchen: Plottable.Plots.Bars.Mode.VERTICAL is quite a mouthful (also didn't we say we only want to go to like a depth of two modules or something like that?) Why don't you just put these constants under Plottable.Plots since most plots can be described as either HORIZONTAL or VERTICAL?

test/components/plots/barPlotTests.ts
```diff
@@ -443,7 +443,7 @@ describe("Plots", () => {
           {y: "B", x: 1} // duplicate Y-value
         ];
         dataset = new Plottable.Dataset(data);
-        barPlot = new Plottable.Plots.Bar(xScale, yScale, false);
+        barPlot = new Plottable.Plots.Bar(xScale, yScale, Plottable.Plots.Bars.Mode.HORIZONTAL);
```
rcchen: Basically an example of being too long.

rcchen:
I'm not convinced that there should even be a third parameter in the constructor. Why don't plots default to horizontal, and then get adjusted via `.orientation("horizontal"|"vertical")` on an as-needed basis?

src/components/plots/barPlot.ts
aicioara: how about `Plottable.Orientation.VERTICAL` ? This is not used for plots only. Will be used for Scales as well in the future

src/components/plots/clusteredBarPlot.ts
```diff
@@ -18,10 +18,10 @@ export module Plots {
      * @constructor
      * @param {Scale} xScale The x scale to use.
      * @param {Scale} yScale The y scale to use.
-     * @param {boolean} isVertical if the plot if vertical.
+     * @param {string} mode the mode of the plot.
```
aicioara: Isn't it an `Orientation` rather than a `mode`? Just semantics

src/components/plots/barPlot.ts
```diff
@@ -21,15 +25,15 @@ export module Plots {
      * @constructor
      * @param {Scale} xScale The x scale to use.
      * @param {Scale} yScale The y scale to use.
-     * @param {boolean} isVertical if the plot if vertical.
+     * @param {string} mode the mode of the plot.
```
rcchen: Remove this, it's not part of the constructor anymore.

src/components/plots/clusteredBarPlot.ts
rcchen: Remove this line as well.

src/components/plots/stackedBarPlot.ts
```diff
@@ -11,10 +11,10 @@ export module Plots {
      * @constructor
      * @param {Scale} xScale the x scale of the plot.
      * @param {Scale} yScale the y scale of the plot.
-     * @param {boolean} isVertical if the plot if vertical.
+     * @param {string} mode the mode of the plot.
```
rcchen: Remove this one.

rcchen:
Just small documentation-related issues, looking much cleaner now.

jtlan:
@rcchen wrote:
>I'm not convinced that there should even be a third parameter in the constructor. Why don't plots default to horizontal, and then get adjusted via `.orientation("horizontal"|"vertical")` on an as-needed basis?

Because changing the mode of the `Plot` after construction wrecks both `x()` and `y()` setters, and introduces complications with adding/removing baselines from both plots.

PR #2079, opened 2015-05-15
Don't enqueue Components that are already rendering.

jtlan:


src/components/componentContainer.ts
```diff
@@ -25,7 +25,7 @@ module Plottable {
     }
 
     public renderImmediately() {
-      this._forEach((c) => c.renderImmediately());
+      this._forEach((c) => c.render());
```
aicioara: Why this change?

jtlan: If the `ComponentContainer` calls `renderImmediately()`, the `Component` can be rendered twice via the following mechanism:
 1. `ComponentContainer`'s `render()` is called. It queues itself and also all its children.
 2. `ComponentContainer`'s `renderImmediately()` is called. This causes its children to render immediately.
 3. The children's `renderImmediately()` is called, causing it to draw a second time.

The better way to fix this is to have `Component`s eject themselves from the queue when `renderImmediately()` is called. If we're not in a rush I'll do that instead.

jtlan:
Closing for now; we'll come back for this later.

Issue #2057, opened 2015-05-14
Rename _autoDomainIfAutomaticMode()


jtlan:
`scale._autoDomainIfAutomaticMode()` is called by `Plot`s, so it needs to be `public`. We should pick a more usable name for it.

PR #2085, opened 2015-05-18
Move Domainer logic to Scales; delete Domainers.

jtlan:
`Domainer` logic properly belongs on the `QuantitativeScale`s anyway, rather than having switches inside the logic for different types of `QuantitativeScales`.

The `Domainer` unit tests all seemed to be "test `Domainer` in particular case involving `SomeOtherClass`". Broke up the `Domainer` tests and moved them to be with the tests for that other class.

`Scale`s now pad by default, and `nice()`-ing (rounding domain values to human-readable ones) has been rolled into the padding, since it's doubtful that someone would want just `nice()`-ing, or padding without `nice()`-ing.


quicktests/overlaying/tests/functional/titleLegend_change.js
```diff
@@ -24,10 +24,17 @@ function run(svg, data, Plottable) {
...
+  var yScale = new Plottable.Scales.Linear();
+  if (Plottable.Domainer != null) {
+    var domainer_X = new Plottable.Domainer().addPaddingException("overlaying", 0);
+    var domainer_Y = new Plottable.Domainer().addPaddingException("overlaying", 0);
+    xScale.domainer(domainer_X);
+    yScale.domainer(domainer_Y);
+  } else {
+    xScale.padProportion(0).addPaddingException("overlaying", 0);
+    yScale.padProportion(0).addPaddingException("overlaying", 0);
+  }
```
bluong: Do we want our quicktests to be as fluid like this going forward?  It'll guard against API breaks but it's also a bunch of clutter in our quicktests

src/components/plots/areaPlot.ts
```diff
@@ -54,23 +54,19 @@ export module Plots {
...
 
-    protected _updateYDomainer() {
-      super._updateYDomainer();
-
+    protected _updateYScale() {
       var extents = this._propertyExtents.get("y0");
-      var extent = Utils.Methods.flatten(extents);
-      var uniqExtentVals = Utils.Methods.uniq(extent);
+      var extent = Utils.Methods.flatten<number>(extents);
+      var uniqExtentVals = Utils.Methods.uniq<number>(extent);
```
bluong: Is the number cast here necessary?

src/components/plots/barPlot.ts
```diff
@@ -66,8 +66,7 @@ export module Plots {
         return this._baselineValue;
       }
       this._baselineValue = value;
-      this._updateXDomainer();
-      this._updateYDomainer();
+      this._updateValueScale();
```
bluong: `_updatePropertyScale()`?

src/components/plots/stackedAreaPlot.ts
```diff
@@ -75,15 +75,11 @@ export module Plots {
...
       var scale = <QuantitativeScale<any>> this.y().scale;
-      if (!scale._userSetDomainer) {
-        scale.domainer().addPaddingException(this, 0)
-                        .addIncludedValue(this, 0);
-        // prepending "AREA_PLOT" is unnecessary but reduces likely of user accidentally creating collisions
-        scale._autoDomainIfAutomaticMode();
-      }
+      scale.addPaddingException(this, 0);
+      scale.addIncludedValue(this, 0);
+      scale._autoDomainIfAutomaticMode();
```
bluong: How do we want to deal with `scale._autoDomainIfAutomaticMode()`

src/components/plots/xyPlot.ts
```diff
@@ -181,24 +176,6 @@ module Plottable {
...
-
-    protected _updateYDomainer() {
-      if (this.y().scale instanceof QuantitativeScale) {
-        var scale = <QuantitativeScale<any>> this.y().scale;
-        if (!scale._userSetDomainer) {
-          scale.domainer().pad().nice();
-        }
-      }
-    }
-
```
bluong: This looks important.  Is there somewhere that accounts for this?

src/scales/quantitativeScale.ts
```diff
@@ -4,9 +4,10 @@ module Plottable {
   export class QuantitativeScale<D> extends Scale<D, number> {
     protected static _DEFAULT_NUM_TICKS = 10;
     protected _d3Scale: D3.Scale.QuantitativeScale;
-    public _userSetDomainer = false;
-    private _domainer: Domainer = new Domainer();
     private _tickGenerator: Scales.TickGenerators.TickGenerator<D> = (scale: Plottable.QuantitativeScale<D>) => scale.getDefaultTicks();
+    private _padProportion = 0.05;
+    private _paddingExceptions: Utils.Map<any, D>;
+    private _includedValues: Utils.Map<any, D>;
```
bluong: Is there a better way we can better type these keys

src/scales/quantitativeScale.ts
```diff
@@ -20,10 +21,89 @@ module Plottable {
...
     protected _getExtent(): D[] {
-      return this._domainer.computeDomain(this._getAllExtents(), this);
+      var extents = this._getAllExtents().filter((extent) => extent.length > 0);
+      var defaultExtent = this._defaultExtent();
+      if (extents.length === 0) {
+        return defaultExtent;
+      }
+      var combinedExtent = [
+        Utils.Methods.min<D[], D>(extents, (extent) => extent[0], defaultExtent[0]),
+        Utils.Methods.max<D[], D>(extents, (extent) => extent[1], defaultExtent[1])
```
bluong: Is the explicit typing necessary?

src/scales/quantitativeScale.ts
```diff
@@ -20,10 +21,89 @@ module Plottable {
...
+    }
+
+    public padProportion(): number;
+    public padProportion(padProportion: number): QuantitativeScale<D>;
+    public padProportion(padProportion?: number): any {
+      if (padProportion == null) {
+        return this._padProportion;
+      }
+      if (padProportion < 0) {
+        throw Error("padProportion must be non-negative");
```
bluong: `throw new Error`?

test/components/plots/areaPlotTests.ts
```diff
@@ -17,6 +17,24 @@ describe("Plots", () => {
...
     });
+
+    it("adds a padding exception to the y scale at the constant y0 value", () => {
+      var svg = TestMethods.generateSVG(400, 400);
+      var xScale = new Plottable.Scales.Linear();
+      var yScale = new Plottable.Scales.Linear();
+      var constantY0 = 30;
+      yScale.addExtentsProvider((scale: Plottable.Scales.Linear) => [[constantY0, constantY0 + 10]]);
+      yScale.autoDomain();
+      assert.operator(yScale.domain()[0], "<", constantY0, "lower end of y Scale's domain is normally padded");
```
bluong: I don't really think the above few lines are necessary.  Note that the `plot` is never used before these lines yet this is in "areaPlotTests.ts"

bluong:
I think overall this is good as now we won't have to rely as much on the domainers to ensure that stuff is set correctly.  A few questions of replaced behavior

quicktests/overlaying/tests/functional/titleLegend_change.js
jtlan: I think we should make our quicktests polymorphic during the PR where the API break is taking place, and then restore it after that. We shouldn't be breaking the API after 1.0 of course, but it might be helpful in comparing before/after if we add a feature that doesn't work on the previous version.

src/components/plots/areaPlot.ts
jtlan: It's not a cast, `uniq()` takes a generic. If I don't specify that it's operating on `<number>` it returns an `<any>`.

src/components/plots/stackedAreaPlot.ts
jtlan: Future PR to deal with #2057.

I think `_autoDomainIfAutomaticMode()` is only ever invoked when:
* padding exceptions are added/removed
* included values are added/removed
* an ExtentsProvider is added/removed/updated

Given that there are methods that are invoked whenever the first three happen, we could just call `_autoDomainIfAutomaticMode()` in each of those methods.

Speaking of which, the `_autoDomainIfAutomaticMode()` is redundant here, since that gets invoked when `addPaddingException()` and `addIncludedValue()` are called.

src/components/plots/xyPlot.ts
jtlan: `Scale`s now pad by default, and `nice()`ing is included in the padding. I'll make a note of it in the PR description.

src/scales/quantitativeScale.ts
jtlan: Possibly, but that wasn't the main thrust of this PR. Thus far the keys are always `Plot`s, but keeping it open is fine.

Maybe there should be some sort of parallel structure for `ExtentsProviders`, but for padding exceptions?

src/scales/quantitativeScale.ts
jtlan: If I don't add the generic, `min()` and `max()` will return `<any>`s (actually, it's possible that the line won't compile at all).

src/scales/quantitativeScale.ts
jtlan: Good catch, thanks.

jtlan: Actually, it looks like invoking it with and without `new` do the same thing: http://es5.github.io/#x15.11.1

I think we should stick with one style though, so I'll add the `new`.

test/components/plots/areaPlotTests.ts
jtlan: I'm testing the behavior before/after attaching the `Scale` to the `Plots.Area` to make the intended behavior clear. I agree that it redundantly tests the padding behavior on `Scales.Linear` though.

quicktests/overlaying/tests/functional/titleLegend_change.js
bluong: So are you going to be making a PR right afterwards to reduce the clutter or?

bluong:
Just a question on the quicktest clutter / a unit test that has a bit of clutter / and `throw new Error()`

test/components/plots/areaPlotTests.ts
bluong: Can we cut down on the redundancy?  Still unsure why anything is tested before the plot is instantiated.  You can retrieve the previous value and then compare it with a value after the plot is instantiated / rendered and then test that

PR #2080, opened 2015-05-16
[RectanglePlot] Reimplemented to take on GridPlot functionality

rcchen:
* `Plots.Grid` is completely removed. Its functionality has been folded into `Plots.Rectangle`
* `Plots.Rectangle` takes in four main data endpoints: `.x()`, `.x1()`, `.y()`, `.y1()`

image

Fixes #2062

PR #2090, opened 2015-05-18
[RectanglePlot] Reimplemented to take on GridPlot functionality

### LINKED ISSUE #2062: `Plots.Grid` x/y methods
Currently `Plots.Grid` has 6 different methods for setting the X and Y properties:
```Typescript
x()
x1()
x2()
y()
y1()
y2()
```

Is it possible to trim this down? The user needs to be able to set the left and right sides on non-`Category` `Scales`, but should they also be able to specify the position of the data as distinct from those edges?

rcchen:
(Transferring from #2080 to retarget develop instead of apiRefactor)

* `Plots.Grid` is completely removed. Its functionality has been folded into `Plots.Rectangle`
* `Plots.Rectangle` takes in four main data endpoints: `.x()`, `.x2()`, `.y()`, `.y2()`

image

Fixes #2062


plottable.d.ts
```diff
@@ -2670,8 +2672,10 @@ declare module Plottable {
              * Constructs a RectanglePlot.
              *
              * A RectanglePlot consists of a bunch of rectangles. The user is required to
-             * project the left and right bounds of the rectangle (x1 and x2 respectively)
-             * as well as the bottom and top bounds (y1 and y2 respectively)
+             * project the left and right bounds of the rectangle (x and x1 respectively)
+             * as well as the bottom and top bounds (y and y1 respectively). If x1/y1 is
+             * not set, the plot will apply auto-centering logic to the extent of x/y (all
+             * values are treated as categories regardless of their scale)
```
bluong: I don't think the parentheses part is necessary.

src/components/plots/rectanglePlot.ts
```diff
@@ -21,8 +21,17 @@ export module Plots {
      */
     constructor(xScale: Scale<X, any>, yScale: Scale<Y, any>) {
       super(xScale, yScale);
+
+      this.animator("cells", new Animators.Null());
```
bluong: Up to you since we probably want to refine this endpoint anyways, but I think the string should instead be "rectangles"

src/components/plots/rectanglePlot.ts
```diff
@@ -33,26 +42,39 @@ export module Plots {
...
+      var yScale = this.y().scale;
 
-      // Generate height based on difference, then adjust for the correct y origin
-      attrToProjector["height"] = (d, i, dataset, m) => Math.abs(y2Attr(d, i, dataset, m) - y1Attr(d, i, dataset, m));
-      attrToProjector["y"] = (d, i, dataset, m) => {
-        return Math.max(y1Attr(d, i, dataset, m), y2Attr(d, i, dataset, m)) - attrToProjector["height"](d, i, dataset, m);
-      };
+      // In order to define a range from x, x1 has to also be set.
+      // If x1 is not set, auto-centering logic will be applied to x.
+      // The above also applies to y/y1.
```
bluong: Let's fold the last line comment here into the first 2.

As a side note, I don't really know if this comment is necessary

src/components/plots/rectanglePlot.ts
```diff
@@ -61,6 +83,18 @@ export module Plots {
...
+    public x(x: number | Accessor<number>): Plots.Rectangle<X, Y>;
+    public x(x: X | Accessor<X>, scale: Scale<X, number>): Plots.Rectangle<X, Y>;
+    public x(x?: number | Accessor<number> | X | Accessor<X>, scale?: Scale<X, number>): any {
+      if (x == null) {
+        return this._propertyBindings.get(Rectangle._X_KEY);
+      }
+      this._bindProperty(Rectangle._X_KEY, x, scale);
+      this.renderImmediately();
+      return this;
+    }
```
bluong: As said I'm unsure of this redeclaration.

Also, at the very least, there should be a `super` call to ensure no duplication of code

src/components/plots/rectanglePlot.ts
```diff
@@ -73,14 +107,14 @@ export module Plots {
...
+    public y(y: Y | Accessor<Y>, scale: Scale<Y, number>): Plots.Rectangle<X, Y>;
+    public y(y?: number | Accessor<number> | Y | Accessor<Y>, scale?: Scale<Y, number>): any {
+      if (y == null) {
+        return this._propertyBindings.get(Rectangle._Y_KEY);
       }
-      this._bindProperty(Rectangle._X2_KEY, x2, scale);
+      this._bindProperty(Rectangle._Y_KEY, y, scale);
       this.renderImmediately();
       return this;
     }
```
bluong: Same as above

bluong:
Mainly the redeclaration

crmorford:
fix the grid quicktest?

jtlan:
I'm going to chime in here again and say I strongly disagree with the `x1`/'y1' naming. I think with both `x` and `x1`, it can get confusing as to which is the "primary" x-value. With `x2`/`y2`, I think there will be less confusion.

bluong:
@jtlan Are you thinking that it should be instead a `x`/`x2`/`y`/`y2` combo?

rcchen:
I think it's just as confusing for there to not be `x1`/`y1` parameters if we use `x`/`x2`/`y`/`y2`. It gives me the sense that there's an `x1`/`y1` that I'm missing out on somewhere. Perhaps some sort of middle ground can be achieved here, but I hope you understand where the potential source of confusion comes from.

jtlan:
@rcchen wrote:
>It gives me the sense that there's an x1/y1 that I'm missing out on somewhere

My thoughts were that `x()` **is** the "x1". With both `x` and `x1` I think it's less clear which is the primary, and so a different type of confusion arises.

jtlan:
The SVG `<line>` element [takes x1, x2, y1, y2](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/line). We can't have both 1 and 2, so `x()` is going to replace one of them. If I see `x()` + `x1()`, I think a reasonable assumption is that `x()` has replaced `x2()`, which is incorrect. Logically `x()` is replacing `x1`, which leaves us with `x2()`.

quicktests/html/_grid.html
```diff
@@ -0,0 +1,177 @@
+
```
bluong: Where'd this file come from?

quicktests/overlaying/tests/realistic/grid.js
```diff
@@ -1,55 +0,0 @@
...
-  plot.x(timeFormatStart, xScale)
-  .y(function(d) { return d.team; }, yScale)
-  .x2(timeFormatEnd, xScale)
-  .attr("fill", function(d) { return d.fill; })
-  .attr("stroke", function(d) { return d.stroke; });
-
-  var table = new Plottable.Components.Table([[yAxis, plot],
-                                             [null, xAxis]]);
-  table.renderTo(svg);
-}
```
bluong: Why is this quicktest removed

bluong:
Just quicktest issues from my end

quicktests/overlaying/tests/realistic/grid.js
rcchen: There are no grids anymore just rectangles (it's a rename)

bluong:
Modify the PR description please

crmorford:
A lot of the quicktests broke since I last looked at this. I'm getting errors like "x1Attr is not a function", but I'm only using .x and .y in my category_grid quicktest.

rcchen:
@crmorford we changed the API so that it took `x()` and `x2()`; as a consequence, using the modified test on the develop branch no longer works (we don't deal with x1Attr anymore) but it works on this current branch.

crmorford:
I'm not using x1 or x2 on the category_grid though, just .x and .y

crmorford:
Whoops, that was all me! I'll look at this again now that I've figured out my test's problems.

PR #2094, opened 2015-05-18
Change callback return types from any to void.

### LINKED ISSUE #2071: Functions that return nothing should return `void`
In parts of our codebase we have things like this:
```Typescript
export type MouseCallback = (p: Point, event: MouseEvent) => any;
```

Since the return value isn't used anywhere, the type should return `void` instead of `any`.

dbt:
Closes #2071

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -465,7 +465,7 @@ declare module Plottable {
 
 
 declare module Plottable {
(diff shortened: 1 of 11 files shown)


jtlan:
[NOQE] because no JS changes.

PR #2095, opened 2015-05-18
Trying Browserstack

rcchen:


rcchen:
We don't need this anymore.

PR #2099, opened 2015-05-19
    rename automaticallyAdjust(X|Y)ScaleOverVisiblePoints in public api

### LINKED ISSUE #2065: Rename automaticallyAdjustYScaleOverVisiblePoints()
... and `automaticallyAdjustYScaleOverVisiblePoints()`. Two options:
```Typescript
autorangeX();
autorangeY();
autorangeNone();
```

or

```Typescript
autorange(property: string);
```

where *property* is an `string`-y enum of `"x"`, `"y"`, or `"none"`.

dbt:
Closes #2065.


src/components/plots/xyPlot.ts
```diff
@@ -123,33 +123,34 @@ module Plottable {
...
-     *
-     * @param {boolean} autoAdjustment The new value for the automatic adjustment domain for x scale.
-     * @returns {XYPlot} The calling XYPlot.
-     */
-    public automaticallyAdjustXScaleOverVisiblePoints(autoAdjustment: boolean): XYPlot<X, Y>  {
-      this._autoAdjustXScaleDomain = autoAdjustment;
-      this._adjustXDomainOnChangeFromY();
-      return this;
-    }
+     public autorange(scale: string): XYPlot<X, Y> {
```
jtlan: Indent off-by-one :(

jtlan: Also, you don't need to explicitly type the return here.

src/components/plots/xyPlot.ts
```diff
@@ -123,33 +123,34 @@ module Plottable {
...
 
-    /**
-     * Sets the automatic domain adjustment over visible points for y scale.
-     *
-     * If autoAdjustment is true adjustment is immediately performend.
-     *
-     * @param {boolean} autoAdjustment The new value for the automatic adjustment domain for y scale.
+    /** 
+     * Sets the automatic domain adjustment for visible points to operate against the X scale, Y scale, or neither.  If 'x' or 'y' is
+     * specified the adjustment is immediately performed.
```
bluong: I personally like the doc style of every sentence is on its own line

src/components/plots/xyPlot.ts
bluong: Do we want to consider a getter?

test/components/plots/stackedPlotTests.ts
```diff
@@ -198,7 +198,7 @@ describe("Plots", () => {
           .addDataset(dataset2);
       plot.x((d) => d.x, xScale)
           .y((d: any) => d.y, yScale);
-      (<any>plot).automaticallyAdjustYScaleOverVisiblePoints(true);
+      (<any>plot).autorange("y");
```
bluong: We can remove the anycasts now! :)

test/components/plots/stackedPlotTests.ts
```diff
@@ -210,7 +210,7 @@ describe("Plots", () => {
...
```
bluong: Same as above

test/components/plots/stackedPlotTests.ts
```diff
@@ -251,7 +251,7 @@ describe("Plots", () => {
           .addDataset(dataset2);
       plot.x((d) => d.x, yScale)
           .y((d: any) => d.y, yScale);
-      (<any>plot).automaticallyAdjustYScaleOverVisiblePoints(true);
+      (<any>plot).autorange("y");
```
bluong: Same as above

test/components/plots/stackedPlotTests.ts
```diff
@@ -263,7 +263,7 @@ describe("Plots", () => {
           .addDataset(dataset2);
       plot.x((d) => d.x, xScale)
           .y((d: any) => d.y, yScale);
-      (<any>plot).automaticallyAdjustYScaleOverVisiblePoints(true);
+      (<any>plot).autorange("y");
```
bluong: Same as above

src/components/plots/xyPlot.ts
```diff
@@ -123,31 +123,32 @@ module Plottable {
...
-
-    /**
-     * Sets the automatic domain adjustment over visible points for x scale.
-     *
-     * If autoAdjustment is true adjustment is immediately performend.
-     *
-     * @param {boolean} autoAdjustment The new value for the automatic adjustment domain for x scale.
+    /** 
+     * Sets the automatic domain adjustment for visible points to operate against the X scale, Y scale, or neither.  If 'x' or 'y' is
+     * specified the adjustment is immediately performed.
```
bluong: Personally like the doc style of every sentence gets its own line

src/components/plots/xyPlot.ts
```diff
@@ -123,31 +123,32 @@ module Plottable {
...
+    public autorange(scale: string) {
+      switch (scale) {
+        case "x":
+          this._autoAdjustXScaleDomain = true;
+          this._adjustXDomainOnChangeFromY();
+          break;
+        case "y":
+          this._autoAdjustYScaleDomain = true;
+          this._adjustYDomainOnChangeFromX();
+          break;
```
bluong: For both of these cases, shouldn't they set the other `_autoAdjust` instance variable to `false`?

Maybe it would serve better if we had a single instance variable to hold a state similar to how you are switching here

src/components/plots/xyPlot.ts
```diff
@@ -123,31 +123,40 @@ module Plottable {
...
+     * 'y' will adjust the y scale in relation to changes in the x domain.
+     * 
+     * 'none' means neither scale will change automatically.
+     * 
      * @returns {XYPlot} The calling XYPlot.
      */
-    public automaticallyAdjustXScaleOverVisiblePoints(autoAdjustment: boolean): XYPlot<X, Y>  {
-      this._autoAdjustXScaleDomain = autoAdjustment;
-      this._adjustXDomainOnChangeFromY();
+    public autorange(scale: string) {
```
bluong: We probably want to think about the name of the parameter to this function.  We found that the name `scale` may be bad as it hints that the input is a scale... which is not the case.  There was a talk about an identifier, but I think this is a soso name.

src/components/plots/xyPlot.ts
```diff
@@ -123,31 +123,40 @@ module Plottable {
...
+          this._autoAdjustXScaleDomain = false;
+          this._autoAdjustYScaleDomain = true;
+          this._adjustYDomainOnChangeFromX();
+          break;
+        case "none":
+          this._autoAdjustXScaleDomain = false;
+          this._autoAdjustYScaleDomain = false;
+          break;
+        default:
+          throw new Error("Invalid scale value '" + scale + "', must be 'x', 'y' or 'none'");
```
bluong: Whatever decision we decide to make in the parameter above should probably be reflected in the error message here

bluong:
Let's come to a conclusion about the parameter name quickly and this should be good to roll out.

dbt:
@crmorford 

This should be an API only change, but it might impact the functionality of auto range scales -- where you zoom or out the X axis and the Y axis adjusts to show the available points.

crmorford:
I can call .autorange(string) with string == "x" or "y" or "none", right? any other strings?
Does it make sense to have a unit test where it's given nonsense?

crmorford:
this console error is unfamiliar to me, not sure if it's caused by this. 
screen shot 2015-05-20 at 11 18 05 am

PR #2100, opened 2015-05-19
Improve typing on the map shim.

jtlan:
`keys()` and values() used to return `<any>`, which wasn't very helpful.

On a `Map<K, V>`, `keys()` now returns `K[]`, `values()` now returns `V[]`.

Also got rid of `Map.map()`, which besides being really goofy wasn't used anywhere.

### DIFF
--- a/src/utils/map.ts
+++ b/src/utils/map.ts
@@ -2,20 +2,15 @@
 
 module Plottable {
 export module Utils {
-  /**
-   * An associative array that can be keyed by anything (inc objects).
-   * Uses pointer equality checks which is why this works.
-   * This power has a price: everything is linear time since it is actually backed by an array...
-   */
   export class Map<K, V> {
     private _keyValuePairs: { key: K; value: V; }[] = [];
 
     /**
-     * Set a new key/value pair in the store.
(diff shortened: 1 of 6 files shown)


src/utils/map.ts
```diff
@@ -2,32 +2,27 @@
...
+     * @return {boolean} True if key already in Map, false otherwise
      */
     public set(key: K, value: V) {
       if (key !== key) {
         throw new Error("NaN may not be used as a key to the Map");
       }
       for (var i = 0; i < this._keyValuePairs.length; i++) {
-        if (this._keyValuePairs[i][0] === key) {
-          this._keyValuePairs[i][1] = value;
+        if (this._keyValuePairs[i].key === key) {
```
bluong: Shouldn't this be `this._keyValuePairs[i][key]`?  Wouldn't this look up the wrong entry in the pairs?

src/utils/map.ts
```diff
@@ -2,32 +2,27 @@
...
+          this._keyValuePairs[i].value = value;
```
bluong: Same as here

PR #2101, opened 2015-05-19
[Plots.ClusteredBar] - Removing ClusteredPlotMetadata

bluong:
In the path forward to erase `PlotMetadata` from the `Accessor` signature, we have to erase the usages of it in our code base to probably something more sensible.

Thus, `Plots.ClusteredBar` will simply hold an instance variable for the offsets and the accessors will reference it


src/components/plots/clusteredBarPlot.ts
```diff
@@ -34,25 +33,23 @@ export module Plots {
 
       var xAttr = attrToProjector["x"];
       var yAttr = attrToProjector["y"];
-      attrToProjector["x"] = (d: any, i: number, dataset: Dataset, m: ClusteredPlotMetadata) =>
-        this._isVertical ? xAttr(d, i, dataset, m) + m.position : xAttr(d, i, dataset, m);
-      attrToProjector["y"] = (d: any, i: number, dataset: Dataset, m: ClusteredPlotMetadata) =>
-        this._isVertical ? yAttr(d, i, dataset, m) : yAttr(d, i, dataset, m) + m.position;
+      attrToProjector["x"] = (d: any, i: number, dataset: Dataset, m: PlotMetadata) =>
```
jtlan: Shouldn't the ternary switch be outside the `Accessor`?
```Typescript
this._isVertical ? (d, i, dataset, m) => xAttr(d, i, dataset, m) + this._clusterOffsets.get(dataset) : (d, i, dataset, m) => xAttr(d, i, dataset, m);
```

src/components/plots/clusteredBarPlot.ts
```diff
@@ -34,25 +33,23 @@ export module Plots {
...
+        this._isVertical ? xAttr(d, i, dataset, m) + this._clusterOffsets.get(dataset) : xAttr(d, i, dataset, m);
```
jtlan: Shouldn't the ternary switch be outside the `Accessor`?
```Typescript
this._isVertical ? (d, i, dataset, m) => xAttr(d, i, dataset, m) + this._clusterOffsets.get(dataset)
                 : (d, i, dataset, m) => xAttr(d, i, dataset, m);
```

src/components/plots/clusteredBarPlot.ts
```diff
@@ -34,25 +33,23 @@ export module Plots {
...
+      attrToProjector["y"] = (d: any, i: number, dataset: Dataset, m: PlotMetadata) =>
+        this._isVertical ? yAttr(d, i, dataset, m) : yAttr(d, i, dataset, m) + this._clusterOffsets.get(dataset);
```
jtlan: Same comment about the ternary here.

src/components/plots/clusteredBarPlot.ts
```diff
@@ -34,25 +33,23 @@ export module Plots {
...
-        plotMetadata.position = innerScale.scale(key) - innerScale.rangeBand() / 2;
-      });
+      this.datasets().forEach((d, i) => this._clusterOffsets.set(d, innerScale.scale(String(i)) - innerScale.rangeBand() / 2));
     }
 
     private _makeInnerScale() {
       var innerScale = new Scales.Category();
-      innerScale.domain(this._datasetKeysInOrder);
+      var fakeDomain = this.datasets().map((d, i) => String(i));
+      innerScale.domain(fakeDomain);
```
jtlan: Maybe just inline `fakeDomain`?

test/components/plots/clusteredBarPlotTests.ts
```diff
@@ -80,7 +80,7 @@ describe("Plots", () => {
 
       // check that clustering is correct
       var innerScale = (<any>renderer)._makeInnerScale();
-      var off = innerScale.scale("_0");
+      var off = innerScale.scale("0");
       var width = xScale.rangeBand() / 2;
       assert.closeTo(TestMethods.numAttr(bar0, "x") + TestMethods.numAttr(bar0, "width") / 2, xScale.scale(bar0X) - width + off, 0.01
```
jtlan: Might be better if this test didn't rely on the inner scale.

jtlan:
[-1] Minor thing about how the Projectors are set up.

src/components/plots/clusteredBarPlot.ts
bluong: This should make it one step faster, let me look into it

test/components/plots/clusteredBarPlotTests.ts
bluong: We currently need to depend on the inner scale in order to determine the offsets for each bar.  While this leads to a less ideal testing infrastructure, we can fix the infrastructure later when we do an audit.

jtlan: I agree we can come back for it later. I just wanted to note that the test would probably be less fragile if the bars positions were checked "fuzzily" relative to each other, rather than a strict check of the position value.

jtlan:
[+1]

bluong:
Retargeting to the side branch

PR #2102, opened 2015-05-19
[Plots.Bar] - _getBarPixelWidth guard against invalid values

bluong:


aicioara:
Please add my test as well. (I am removing it on my side, since the build does not pass otherwise)

PR #2107, opened 2015-05-19
Remove unused color.ts

dbt:
Remove unnecessary Plottable.Core.Colors


bluong:
The file should also be removed from `reference.ts` which is also probably why the builds failed

PR #2110, opened 2015-05-20
Actually adding tests...but i just continued work on the branch

crmorford:


bluong:
jshint is complaining.  Please run
`grunt jshint` before pushing.

bluong:
While we do like tests, managing data files can be quite the hassle...  What do you think about what we should do with these data files? @jtlan

crmorford:
oops, i missed your comment for this push @bluong , but will do next time

quicktests/overlaying/tests/realistic/animals.js
```diff
@@ -0,0 +1,121 @@
+
+
```
bluong: Remove the above lines

bluong:
As an aside, we should probably up our jshint configurations so that our javascript files are decently neat.  I can see a random user stumbling upon these quicktests and we wouldn't want to present to them a decent mess.

jtlan:
@bluong wrote:
>While we do like tests, managing data files can be quite the hassle...  What do you think about what we should do with these data files? @jtlan 

We can just carry around the .csv files for now. Later on, we can try to work out a better way to do it.

crmorford:
closing for now, focusing on PRs

PR #2111, opened 2015-05-20
[Plot] - Remove PlotMetadata

bluong:
This field just isn't really something that should be exposed to the user.

### DIFF
--- a/src/core/interfaces.ts
+++ b/src/core/interfaces.ts
@@ -3,13 +3,13 @@ module Plottable {
    * Access specific datum property.
    */
   export interface Accessor<T> {
-    (datum: any, index: number, dataset: Dataset, plotMetadata: Plots.PlotMetadata): T;
+    (datum: any, index: number, dataset: Dataset): T;
   }
 
   /**
    * Retrieves scaled datum property.
    */
-  export type _Projector = (datum: any, index: number, dataset: Dataset, plotMetadata: Plots.PlotMetadata) => any;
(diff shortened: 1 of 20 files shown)


src/core/interfaces.ts
```diff
@@ -3,13 +3,13 @@ module Plottable {
...
   export interface Accessor<T> {
-    (datum: any, index: number, dataset: Dataset, plotMetadata: Plots.PlotMetadata): T;
+    (datum: any, index: number, dataset: Dataset): T;
   }
 
   /**
    * Retrieves scaled datum property.
    */
-  export type _Projector = (datum: any, index: number, dataset: Dataset, plotMetadata: Plots.PlotMetadata) => any;
+  export type _Projector = (datum: any, index: number, dataset: Dataset) => any;
```
jtlan: Do we still need the `_Projector` type, by the way?

jtlan:
[+1] Minor comment about `_Projector` type, but that probably belongs in a different PR anyway.

src/core/interfaces.ts
bluong: Now this is where type aliasing is probably handy.

`Accessor`s currently have the understanding that the scale has not affected their values whereas `_Projector`s do have that understanding that the scale has possibly affected the end value.

That said, if you think that differentiation is not enough of a merit for separation, possibly they can just be folded into `Accessor`

jtlan: Maybe `type`-name it something like `ScaledAccessor`, but only in the module where it's used?

PR #2112, opened 2015-05-20
Bring back TitleLabel and AxisLabel

jtlan:
(Reverts 4aff28b80635da2f7ac1ab6b22db88e2d3474e00)

Basically, the syntactic sugar of being able to say
```Typescript
var title = new Plottable.Components.TitleLabel("My Title");
```
instead of
```Typescript
var title = new Plottable.Components.Label("My Title").classed(Plottable.Components.Label.TITLE_LABEL_CLASS, true);
```

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2010,8 +2010,6 @@ declare module Plottable {
 declare module Plottable {
     module Components {
         class Label extends Component {
-            static TITLE_LABEL_CLASS: string;
-            static AXIS_LABEL_CLASS: string;
             /**
              * Creates a Label.
              *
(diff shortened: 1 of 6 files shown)


bluong:
While it does provide syntactic sugar to the user, I personally am afraid that constructing a whole class for something so simple seems a bit incorrect.  It mainly seems really odd to me.

One possible extension from my above feelings could be a diamond problem that could happen.  If we decide to extend `Label` and then we want to have that extended class also have the `Title` idea, then we have an easy diamond problem even though with the previous implementation, we just add the appropriate class.

That said, the current implementation is a bit annoying.  Is there any way we can do some word shortening?

dbt:
chatted in person.  Makes sense to consider these different types conceptually even though the implementation isn't very different.  The risk of the diamond problem isn't that high, and we can still expose that string constant if someone wants to subclass Label and have a multi inheritance of TitleLabel and Label.

bluong:
We just need to expose the class strings and we should be good now.

PR #2113, opened 2015-05-20
Api refactor

bluong:
Consists of three things that was leftover from an old branch
* Plottable `Scale`s are no longer necessarily backed by a D3 scale.  They may choose so by implementation.  What this means to a user... is not really anything at all.  Thus, a regression pass on Scales should be enough here.  Extending `Scale` should be significantly more relaxed, though.
* `Plots.Scatter` front loads some logic to the constructor, specifically it sets up the "size" and the "symbol" property at constructor time, meaning that the `size()` and the `symbol()` calls will return the defaults on instantiation.
* `Plots.Bar` has lost the third constructor argument, meaning that they will render vertical by default and can be switched via
`.orientation("horizontal")`


crmorford:
How can i test the Plots.Scatter changes? I tried to log .size() and .symbol() right after instantiation, but they looked no different from develop

crmorford:
all looks good

PR #2114, opened 2015-05-20
[Plot] - Remove plot metadata

bluong:
Something that was exposed to the user for not necessarily good reason... has now... disappeared!

PlotMetadata is no longer a thing.


crmorford:
Just checked out overlaying, everything looks fine.

Issue #1542, opened 2015-01-21
[Scale]  - Auto domain fails if only one end of the domain is set


bluong:
Let's say I use a Linear Scale on a Bar Plot.

If I don't do anything to the Scale, then it will autodomain nicely.

However, if I wanted to set a min/max boundary on the scale domain but not both, autodomaining will not happen.

We may want to support autodomaining on a single side of a scale instead of both

jtlan:
This seems more like the type of thing where the user would write a `Domainer` instead of us building out a fixed API point for it.

Speaking of which, `Domainer`s are pretty complicated...

jtlan:
I think I've got a decent way to solve this. Let the user set only one side of the domain if they want:
```
domainMin(value); // sets only the lower end
domainMax(value); // sets only the upper end
```
These methods are basically `domain(values)` applied to only one end of the `Scale`'s domain. The other end will still be set with the value retrieved by `autoDomain()`, unless that would cause the `Scale` to reverse. *Actually* calling `autoDomain()` will blank both the min and the max, the same way that calling `autoDomain()` overrides a domain set with `domain(values)`.

PR #2116, opened 2015-05-20
Allow setting only one end of the domain on QuantitativeScale

jtlan:
`domainMin()` sets the lower end of the domain.
`domainMax()` sets the upper end of the domain.

If only one is set, the other end of the scale will still be autodomained. If the domain would be flipped instead it is expanded using the same logic for expanding single-point domains.

If both `domainMin()` and `domainMax() are called, it's equivalent to calling `domain([min, max])`. Setting both can reverse the domain on `Scale`s that support reversal.

Calling `autoDomain()` will clear both set values, the same way calling `autoDomain()` overrides `domain()`.

Addresses #1542, although that won't be closed until this makes it to **develop**.

Demo fiddle for feature: http://jsfiddle.net/temt44gx/2/


src/scales/quantitativeScale.ts
```diff
@@ -119,7 +166,29 @@ module Plottable {
...
-      return super.domain(values); // need to override type sig to enable method chaining:/
+      if (values != null) {
+        this._domainMin = values[0];
+        this._domainMax = values[1];
+      }
+      return super.domain(values);
+    }
+
+    /**
+     * Sets just the lower end of the domain.
```
bluong: I think the word "just" is unnecessary.

src/scales/quantitativeScale.ts
```diff
@@ -119,7 +166,29 @@ module Plottable {
...
+     */
+    public domainMin(domainMin: D) {
+      this._domainMin = domainMin;
+      this._autoDomainIfAutomaticMode();
+      return this;
+    }
+
+    /**
+     * sets just the upper end of the domain.
```
bluong: Same as above.  Also, capitalize "sets" please

src/scales/quantitativeScale.ts
```diff
@@ -25,19 +27,64 @@ module Plottable {
...
-        Utils.Methods.min<D[], D>(extents, (extent) => extent[0], defaultExtent[0]),
-        Utils.Methods.max<D[], D>(extents, (extent) => extent[1], defaultExtent[1])
-      ];
-      var includedDomain = this._includeValues(combinedExtent);
-      var paddedDomain  = this._padDomain(includedDomain);
-      return paddedDomain;
+
+      if (this._domainMin != null) {
+        extent[0] = this._domainMin;
+      }
```
bluong: Fun times with boolean operations
`extent[0] = this._domainMin || extent[0];`

src/scales/quantitativeScale.ts
```diff
@@ -25,19 +27,64 @@ module Plottable {
...
+      if (this._domainMax != null) {
+        extent[1] = this._domainMax;
+      }
```
bluong: Same as above

bluong:
Logic mainly looks sound.  Can we ask the requesting user to see if this performs as requested?

src/scales/quantitativeScale.ts
jtlan: Right, I think the `if`-block style is more readable though.

jtlan:
[-1] bug when `domainMin()` is set to the original max value.

src/scales/quantitativeScale.ts
bluong: I personally am a fan of the suscinctness of the boolean operator though.

bluong: 0 is falsy -_-

PR #2118, opened 2015-05-20
[Plot] - removing generateProjectors

bluong:
This function basically became redundant now that we have a nice `attr()` getter as well as appropriate property getters


jtlan:
When this merges, ping me, and I'll update the docs with the recommended way to get this functionality.

PR #2120, opened 2015-05-20
Updated browser versions in Gruntfile

rcchen:
Browser updates
* Firefox has been updated from v30 to v37
* Chrome has been updated from v35 to v42


aicioara:
Did this here https://github.com/palantir/plottable/compare/unskipTests?expand=1 . Will open PR soon. Also, version should not be specified if we want latest version

PR #2092, opened 2015-05-18
[Label] Changed orientation(string) to angle(number)

rcchen:
We no longer use orientation strings, and instead pass the angle straight in. 

* `constructor(displayText = "", orientation = "horizontal")` -> `constructor(displayText = "")`
* `orientation(orientation: string)` -> `angle(angle: number)`

Fixes #2048

Issue #1921, opened 2015-04-26
Un-skip tests


jtlan:
A number of tests are currently being skipped, which makes it harder to see if we've screwed something up during the refactor. We should un-skip any skipped tests.

aicioara:
On it. Tests are failing for various reasons (old browser config in SauceLab, TODOs, weird measurements in FF), so trying to fix them all.

aicioara:
Could not eliminate one skiptest because of this https://github.com/palantir/plottable/issues/2003

aicioara:
Blocked by #2003

aicioara:
Reopen the PR #2127 and continue once unblocked

aicioara:
Eliminated all. Blocked by 
#2164
When done open PR here https://github.com/palantir/plottable/compare/unskipTests3?expand=1

PR #2121, opened 2015-05-20
Unskip tests

aicioara:
Unskips few tests. Created this now to unblock #2092
Part of #1921 
- Upgraded Firefox and Chrome on SauceLabs to the latest version
- Increased tolerance of FireFox pixel precision to 2

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -134,10 +134,10 @@ module.exports = function(grunt) {
 
   var browsers = [{
     browserName: "firefox",
-    version: "30"
+    platform: "linux"
   }, {
     browserName: "chrome",
-    version: "35"
+    platform: "linux"
   }, {
     browserName: "internet explorer",
     version: "9",
--- a/test/globalInitialization.ts
+++ b/test/globalInitialization.ts
@@ -13,7 +13,7 @@ before(() => {
   if (window.PHANTOMJS) {
     window.Pixel_CloseTo_Requirement = 2;
   } else if (isFirefox) {
-    //HACKHACK #2122
+    // HACKHACK #2122
     window.Pixel_CloseTo_Requirement = 2;
   } else {
     window.Pixel_CloseTo_Requirement = 0.5;
--- a/test/globalInitialization.ts
+++ b/test/globalInitialization.ts
@@ -13,7 +13,7 @@ before(() => {
   if (window.PHANTOMJS) {
     window.Pixel_CloseTo_Requirement = 2;
   } else if (isFirefox) {
-    window.Pixel_CloseTo_Requirement = 1;
+    window.Pixel_CloseTo_Requirement = 2;
   } else {
     window.Pixel_CloseTo_Requirement = 0.5;
   }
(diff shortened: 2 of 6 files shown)


Gruntfile.js
```diff
@@ -134,10 +134,10 @@ module.exports = function(grunt) {
 
   var browsers = [{
     browserName: "firefox",
-    version: "30"
+    platform: "linux"
   }, {
     browserName: "chrome",
-    version: "35"
+    platform: "linux"
```
bluong: Why "linux"?  At the very least we probably want to go something like... Windows?

rcchen: Yeah we should default to Windows for these.

aicioara: That was the original default. I tried to minimize the changes. You want Windows?

test/globalInitialization.ts
```diff
@@ -13,7 +13,7 @@ before(() => {
   if (window.PHANTOMJS) {
     window.Pixel_CloseTo_Requirement = 2;
   } else if (isFirefox) {
-    window.Pixel_CloseTo_Requirement = 1;
+    window.Pixel_CloseTo_Requirement = 2;
```
bluong: Why this change?

rcchen: What's the reason for this? (why is 2 a more correct number than 1)

Gruntfile.js
bluong: Most of our users will probably be on Windows.  We probably want to align to that fact.

test/globalInitialization.ts
aicioara: The issue is that the `SVGTypeWriter` uses `getBBox()` to figure out the width and height of a text element, while our tests use `getBoundingClientRect()` for the same purpose. There is no difference between the 2 values in IE9 and Chrome, but it differs in Mozilla by 2 pixels. This was the easiest change

bluong: As long as we think we have the right number here I think it makes sense

Gruntfile.js
aicioara: Did the change and there are more tests failing, due to pixel rounding differences (I believe)... I'd file another issue for this.

test/globalInitialization.ts
aicioara: We probably don't since that should be a ratio (be dependent on the size), while we use it as a constant

Gruntfile.js
aicioara: Opened this https://github.com/palantir/plottable/issues/2122

test/globalInitialization.ts
aicioara: + that thing is a hack

rcchen: Can you add the explanation that you had about `getBBox()` and `getBoundingClientRect()` to the comments on the source here and annotate it with `HACKHACK:`? That way this explanation won't be lost for the future.

Gruntfile.js
bluong: Why can't we switch to Windows and then just fix the tests in accordance in this PR.  I don't mark this currently as an upgrade

PR #2129, opened 2015-05-21
Remove Domainers.

### LINKED ISSUE #1542: [Scale]  - Auto domain fails if only one end of the domain is set

### LINKED ISSUE #2055: Move Domainer logic to Scales
There's only ever one `Domainer` per `QuantitativeScale` and vice versa, so it makes sense to move its logic to the `QuantitativeScale` itself.

### LINKED ISSUE #2057: Rename _autoDomainIfAutomaticMode()

jtlan:
Changes:

#### Moved `Domainer` logic to `Scale`s.
Deleted `Domainer`.
Close #2055.

#### Added `domainMin()` and `domainMax()` to `QuantitativeScale`
`domainMin()` sets the lower end of the domain.
`domainMax()` sets the upper end of the domain.

If only one is set, the other end of the scale will still be autodomained. If the domain would be flipped instead it is expanded using the same logic for expanding single-point domains.

If both `domainMin()` and `domainMax() are called, it's equivalent to calling `domain([min, max])`. Setting both can reverse the domain on `Scale`s that support reversal.

Calling `autoDomain()` will clear both set values, the same way calling `autoDomain()` overrides `domain()`.

Close #1542.
Demo fiddle for feature: http://jsfiddle.net/temt44gx/3/

#### Fixed autodomain logic
Adding or removing an `ExtentsProvider` (with `addExtentsProvider()` and `removeExtentsProvider()`) now causes the scale to `autoDomain()` if it's in automatic mode.

Made `_autoDomainIfAutomaticMode()` `protected`.

Made `_defaultExtent()` protected.

Close #2057.


crmorford:
Discussion requested:
If i set the min to 4.9 like this:
screen shot 2015-05-22 at 10 10 36 am
It still autodomains like the whole line is being shown. It's odd to me that the padding can be so much wider than the data shown.

jtlan:
Demo variant with `Scales.Time` on X: http://jsfiddle.net/2f0xycto/

crmorford:
Modified Log: make max a negative number and the scale flips
I tried -6 here
http://jsfiddle.net/temt44gx/6/

Issue #2048, opened 2015-05-13
Poor and unclear documentation of label orientation


istarnx:
Can you guys make it easier to understand the direction in which labels will render? 

This has no description for "orientation":
http://plottablejs.org/docs/classes/plottable.component.axislabel.html#constructor

...and this is unclear: ("The orientation of the Label (horizontal/left/right) (default = "horizontal")")
http://plottablejs.org/docs/classes/plottable.component.label.html#constructor

Really, we're talking about the rotation of the text:
- 0 deg (default, horizontal left-to-right), 90 deg (vertical top-to-bottom), and -90 deg (vertical bottom-to-top)

Ideally, these would be (self-documenting...) string constants with some description... somewhere. :)

jtlan:
Another possibility is to take a `number` the same way `tickLabelAngle()` (on `Axes.Category`) does.

PR #2132, opened 2015-05-21
Label constructor now takes angle as an optional parameter.

### LINKED ISSUE #2130: Label constructor should take angle
Missed this when #2092 went in, but `Label`s should take an angle in their constructor, the way they used to take an `orientation` in the constructor.

jtlan:
..the same way that the old constructor used to take the orientation as a parameter.

Close #2130.

### DIFF
--- a/src/components/label.ts
+++ b/src/components/label.ts
@@ -90,6 +90,12 @@ export module Components {
       if (angle == null) {
         return this._angle;
       } else {
+        angle %= 360;
+        if (angle > 180) {
+          angle -= 360;
+        } else if (angle < -180) {
+          angle += 360;
+        }
         if (angle === -90 || angle === 0 || angle === 90) {
           this._angle = angle;
         } else {
--- a/src/components/label.ts
+++ b/src/components/label.ts
@@ -14,12 +14,11 @@ export module Components {
     /**
      * Creates a Label.
      *
-     * A label is component that renders just text. The most common use of
-     * labels is to create a title or axis labels.
+     * A Label is a Component that draws a single line of text.
      *
      * @constructor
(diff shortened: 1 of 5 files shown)


crmorford:
Reason for this? 
If i see a number with the parameter name "angle" as a parameter, I'm going to expect to be able to put in anything 0-360. The strings felt discrete, this feels like it should be continuous and I find it a bit misleading.

Also confused about why don't 180 and 270 work? I tried 270 before -90, it felt more intuitive to me.

jtlan:
Originally the API was changed (in #2092) because of #2048. If later we want to support labels at tilted angles (45 degrees, etc), passing in the angle seems more expandable. It's also closer in line with the `tickLabelAngle()` API point on `Axes.Category`.

crmorford:
Hm okay I see the advantages if we want to support tilted angles later. 
But can we also handle 270 correctly? That's more natural to me.

jtlan:
>But can we also handle 270 correctly? That's more natural to me.
Yeah, we can add a parser that converts arbitrarily large numbers to a valid range.

src/components/label.ts
```diff
@@ -19,12 +19,13 @@ export module Components {
      *
      * @constructor
      * @param {string} displayText The text of the Label (default = "").
+     * @param {number} angle The rotation angle of the text (default = 0).
```
aicioara: Mention that 0 means horizontal. As a reader of this documentation, I have no idea what a rotation of 0 means in this library

aicioara: (I hope it means horizontal)

aicioara:
Rest looks good

Issue #1859, opened 2015-04-18
Plot.Grid pixelPoints in wrong spot


jtlan:
![screen shot 2015-04-17 at 4 47 32 pm](https://cloud.githubusercontent.com/assets/1440449/7212780/7e5976f6-e521-11e4-92e3-81b4293e3fdf.png)
http://jsfiddle.net/pgn8ms8u/

**Actual**: `pixelPoints` are at top-center of each cell.
**Expected**: `pixelPoints` are in center of each cell.

Note that this also happens with non-`Category` `Scale`s:
![screen shot 2015-04-17 at 5 03 35 pm](https://cloud.githubusercontent.com/assets/1440449/7212890/bd96ae9a-e523-11e4-9d68-c8d532bc0632.png)
http://jsfiddle.net/y5pkauk7/

... but in this case we should probably put the `pixelPoints` on whatevever the `y` values are.

aicioara:
http://jsfiddle.net/pgn8ms8u/6/ Rectangle plot is fixed

aicioara:
Same for the second example
http://jsfiddle.net/pgn8ms8u/7/

aicioara:
@crmorford can you confirm?

PR #2134, opened 2015-05-21
[Plot] - _getPixelPoint transferred from Drawers

bluong:
The logic that used to be on the Drawers now exist on the plot.

One notable upgrade is that `Plots.Grid` now have the pixel point aimed at the center of the cell instead of the top middle


jtlan:
Be sure to close #1859 when this makes it to **develop**.

jtlan:
Shouldn't the default implementation for `_getPixelPoint()` on `XYPlot`return
```Typescript
{ x: attrToProjector["x"](datum, index, dataset), y: attrToProjector["y"](datum, index, dataset) }
```
?

That would save some implementation on `Line` and `Scatter`.

jtlan:
We should probably pick a name for this method that doesn't include "get", but the rename is slightly out of scope here.

PR #2135, opened 2015-05-21
[Drawers.Area] - No longer draws the line with the area

bluong:
In consequence, `Plots.Area` holds a bunch of `Drawers.Line` that will draw in the `_additionalPaint()` step


src/components/plots/areaPlot.ts
```diff
@@ -50,6 +57,26 @@ export module Plots {
       }
     }
 
+    public addDataset(dataset: Dataset) {
+      // HACKHACK Drawers should take in a dataset instead of the key
+      var lineDrawer = new Drawers.Line("foo");
```
jtlan: `"foo"`?

jtlan:
I'm a little confused here. Why does this change require re-defining `getAllSelections()` and `getAllPlotData()` on `Plots.Area`?

bluong:
Note that `getAllSelections()` and `getAllPlotData()` are dependent on retrieving the element data from the `Drawer`s, however `Drawers.Area` no longer has information about drawing the line.

src/components/plots/areaPlot.ts
bluong: Note the HACKHACK

jtlan: ... but why is the key "foo" ?

bluong: I guess I could put in "".  I'll do that instead

jtlan:
@bluong wrote:
>Note that `getAllSelections()` and `getAllPlotData()` are dependent on retrieving the element data from the `Drawer`s, however `Drawers.Area` no longer has information about drawing the line.

What's in the `PlotData`'s `Selection` field? Both the `.line` and `.area` `<path>`s?

src/components/plots/areaPlot.ts
```diff
@@ -89,6 +116,37 @@ export module Plots {
...
+        var plotDatasetKey = this._key2PlotDatasetKey.get(datasetKey);
+        if (plotDatasetKey == null) { return; }
+        var drawer = plotDatasetKey.drawer;
+        var dataset = plotDatasetKey.dataset;
+        plotDatasetKey.dataset.data().forEach((datum: any, index: number) => {
+          var pixelPoint = this._pixelPoint(datum, index, dataset);
+          if (pixelPoint.x !== pixelPoint.x || pixelPoint.y !== pixelPoint.y) {
+            return;
+          }
+          allElements.splice(index * 2 + 1, 0, drawer._getSelection(index).node());
```
jtlan: Why is it spliced to this location? If the `.line` `<path>`s are supposed to be at the end (to parallel `getAllSelections`), they should just be `push`ed right?

jtlan:
Some confusion around what is returned as the `PlotData` `Selection`.

jtlan:
I'll [+1] this since I think it's a good step, but we should look into the `<path>`-`Drawer` approach later.

PR #2136, opened 2015-05-21
Extent computation takes Scale into account

### LINKED ISSUE #2014: Extent Computation needs to take scale type into account

aicioara:
Closes #2014 
Instead checking the type of the data, we are using scale to determine how an extent is being calculated.

QE notes:
Ideally nothing should be changed regarding our tests.
Try stringy numbers and numbery strings in the datasets. They should behave properly now


src/components/plots/plot.ts
```diff
@@ -265,28 +265,25 @@ module Plottable {
...
+      if (scale) {
+        return scale.getExtentFromDomain(mappedData);
       } else {
-        var extent = d3.extent(mappedData);
-        if (extent[0] == null || extent[1] == null) {
-          return [];
-        } else {
-          return extent;
-        }
+        return [];
```
jtlan: A minor perf improvement would be to move the scale check to the start of the method.

src/components/plots/plot.ts
```diff
@@ -265,28 +265,25 @@ module Plottable {
...
         data = data.filter((d, i) => filter(d, i, dataset));
       }
       var appliedAccessor = (d: any, i: number) => accessor(d, i, dataset);
       var mappedData = data.map(appliedAccessor);
-      if (mappedData.length === 0) {
-        return [];
-      } else if (typeof(mappedData[0]) === "string") {
-        return Utils.Methods.uniq(mappedData);
+
+      if (scale) {
```
jtlan: I think our current style is to use `!= null` for existence checks, although I'm aware of inconsistencies in the codebase.

src/scales/scale.ts
```diff
@@ -32,6 +32,10 @@ module Plottable {
       this._extentsProviders = new Utils.Set<Scales.ExtentsProvider<D>>();
     }
 
+    public getExtentFromDomain(data: D[]): D[] {
```
jtlan: We should probably:
* use a name without "get" in it
* Use "domain" in a way that's consistent with its use on `Scale`

In particular, since the argument is named `data` I would expect the method to be named something like `extentFromData()`.

jtlan:
[-1] I think we should rename the extent-computing method.

crmorford:
For some reason the animate_scatter overlaying test isn't working on this branch. It's not obvious to me why not.

crmorford:
in the interaction_BarHover quicktest change line 3 to   
```
return [{name: 1, y: 3}, {name: "Sam", y: 2}, {name: "Gollum", y: 4}];
```
and the hovering doesn't work anymore, after you hover over the leftmost bar. Not sure if this is expected or not.

aicioara:
Yup, it's a bug. Will fix

aicioara:
Opened this as well https://github.com/palantir/plottable/issues/2140

PR #2139, opened 2015-05-21
[Plots.Scatter] - animation starts at an empty path

bluong:
During the rework of Drawers and `_generateAttrToProjector()`, `Plots.Scatter` should be aligned with the idea that the projectors passed to the `Drawer`s are in fact the items that will be on the DOM node(s).

...And in doing so, I forgot to change the logic in the reset animation

QE: Animations were broken before for `Plots.Scatter`.  They should work now.


crmorford:
on my overlaying, this only works on the top layer.  i'm used to that happening with hitbox stuff, but not recently.

PR #2141, opened 2015-05-21
[Plots] - dont use renderImmediately()

bluong:
In the property's first authorings, they were using `renderImmediately` for not really a good reason.  Going with the more efficient `render()`

QE: Mainly a regression pass.  This should introduce performance boost in rendering in comparison to develop, but then against develop is currently in a less performant state against master


aicioara:
why is `render()` more efficient than `renderImmediately` ?

bluong:
`render()` is more efficient since it places this component into a queue in which it gets rendered with a bunch of other components in a batched manner.  Internally we should be using `render()` and not `renderImmediately()`.  This was initially written as an error on my end as I think I got confused between ambiguous function names.

`renderImmediately()` is supported for 2 reasons.  1 is that if the user wanted to synchronously render, instead of relying on a queue to render, then they can.  2 is that our `RenderController` needs some sort of endpoint on the `Component` to actually render in a synchronous manner when it is that `Component`'s turn to render.

Issue #1913, opened 2015-04-25
Refactor Drawers


jtlan:
`Drawer`s were originally created to compartmentalize part of the `Plot`'s logic in a reusable way. However, shifting information between `Plot`s and their `Drawer`s turned out to be troublesome and involved a whole bunch of "package-protected" methods.

We either need to come up with a good public-facing API for `Drawer`s, or re-integrate them back into the `Plot`s.

Close #1917 or edit (depending on the case) when done

crmorford:
What things did drawers used to do for the plots? so that I can make sure all of that still works.

bluong:
Done!

PR #2143, opened 2015-05-22
[Plots.Pie] - All attribute calculations are done at the plot

bluong:
In accordance to #1913, the calculations for the attributes should not be done at the `Drawer` level but at the plot level.  I'm starting with `Plots.Pie` here

### DIFF
--- a/src/components/plots/piePlot.ts
+++ b/src/components/plots/piePlot.ts
@@ -49,7 +49,7 @@ export module Plots {
       }
       this._startAngles.set(dataset, []);
       this._endAngles.set(dataset, []);
-      this._updatePieLayout();
+      this._updatePieAngles();
       super.addDataset(dataset);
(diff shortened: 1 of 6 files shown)


src/components/plots/piePlot.ts
```diff
@@ -11,6 +11,8 @@ export module Plots {
     private static _INNER_RADIUS_KEY = "inner-radius";
     private static _OUTER_RADIUS_KEY = "outer-radius";
     private static _SECTOR_VALUE_KEY = "sector-value";
+    private _startAngles: Utils.Map<Dataset, number[]>;
+    private _endAngles: Utils.Map<Dataset, number[]>;
```
jtlan: If only one `Dataset` is supported, couldn't these just be arrays instead of a `Map`?

PR #2144, opened 2015-05-22
[Drawers.Area] - No longer draws the line with the area v2

bluong:
Different implementation from #2135 where instead the `line` is not included in `getAllSelections()` and `getAllPlotData()`.

My reasoning for this approach is that... it is very consistent in our current iteration of Plottable that there is essentially a single(set of) DOM element(s) that we will need to set attributes on to convey what we call a plot...  This is consistent for everywhere except for... the "line"s on a `Plots.Bar` as well as the "baseline"s on a `Plots.Bar`.

It is perfectly fine that we decided to have these visually displayed from a visual design standpoint, but for what type of information we want our users to be directly tapping into, we probably want them to focus on the single DOM element that is the core visualization of our data (The "bars" on a `Plots.Bar` and the "area" on a `Plots.Area`).  Thus, I think that we should have `getAllSelections()` etc. either consistently include all the selections on the svg (not that useful) or focus directly on the core svg elements


bluong:
Decided we wanted the area path.  Closing

PR #2147, opened 2015-05-22
[Plots.Bar] Removed getBars() in favor of plotDataAt()/plotDataIn()

### LINKED ISSUE #2066: Rename and split `getBars`
`Plots.Bar` has a `getBars()` method:
```Typescript
public getBars(xValOrExtent: number | Extent, yValOrExtent: number | Extent): D3.Selection;
```

This should probably be split into two methods (one takes a point, another takes `Extent`s), and return `PlotData` to give the user more context. It should also be renamed so we can implement similar functionality on other `Plot`s:
```Typescript
plotDataAt(p: Point);
pltoDataInExtents(xExtent: Extent, yExtent: Extent);
```

rcchen:
We also now return `PlotData` so there is more context. See below for usage:
```typescript
// Data at a single pixel point
plot.plotDataAt({x: 2, y: 100});

// Data within a defined range
plot.plotDataIn({min: 200, max: 500}, {min: 0, max: 500});

// Data within bounds
plot.plotDataIn({topLeft: {x: 1, y: 1}, bottomRight: {x: 100, y: 100}})
```

Fiddle: http://jsfiddle.net/rcchen/tj79xsg4/1/

Fixes #2066


src/components/plots/barPlot.ts
```diff
@@ -238,38 +238,40 @@ export module Plots {
...
+      return this._getPlotData(p.x, p.y);
+    }
 
-      return d3.selectAll(bars);
+    /**
+     * Gets the {Plots.PlotData} that correspond to a given xRange/yRange
+     * 
+     * @param {Extent} xRange The specified range of x values
+     * @param {Extent} yRange The specified range of y values
+     * @return {Plots.PlotData} The plot data that corresponds to the {Extent}(s)
```
jtlan: "ranges" instead of "Extent(s)"

jtlan:
[-1] Minor docs thing.

Loving the new signatures.

plottable.d.ts
```diff
@@ -2767,16 +2767,34 @@ declare module Plottable {
...
+             *
+             * @param {Point} p The provided pixel position as a {Point}
+             * @return {Plots.PlotData} The plot data that corresponds to the {Point}.
+             */
+            plotDataAt(p: Point): {
+                data: any[];
+                pixelPoints: {
+                    x: number;
+                    y: number;
+                }[];
```
bluong: Can we explicitly type this as `Point`s or even better `PlotData`?

plottable.d.ts
```diff
@@ -2767,16 +2767,34 @@ declare module Plottable {
...
              *
-             * @param {number | Extent} xValOrExtent The pixel x position, or range of x values.
-             * @param {number | Extent} yValOrExtent The pixel y position, or range of y values.
-             * @returns {D3.Selection} The selected bar, or null if no bar was selected.
+             * @param {Extent} xRange The specified range of x values
+             * @param {Extent} yRange The specified range of y values
+             * @return {Plots.PlotData} The plot data that corresponds to the ranges
              */
-            getBars(xValOrExtent: number | Extent, yValOrExtent: number | Extent): D3.Selection;
+            plotDataIn(xRange: Extent, yRange: Extent): {
```
bluong: Same as above

test/components/plots/barPlotTests.ts
```diff
@@ -106,33 +106,33 @@ describe("Plots", () => {
       });
 
       it("getBars()", () => {
-        var bar: D3.Selection = barPlot.getBars(155, 150); // in the middle of bar 0
+        var bar: Plottable.Plots.PlotData = barPlot.plotDataAt({x: 155, y: 150}); // in the middle of bar 0
```
bluong: You can remove the explicit typing.

src/components/plots/barPlot.ts
```diff
@@ -237,38 +237,40 @@ export module Plots {
...
 
-      return d3.selectAll(bars);
+    /**
+     * Gets the {Plots.PlotData} that correspond to a given xRange/yRange
+     * 
+     * @param {Extent} xRange The specified range of x values
+     * @param {Extent} yRange The specified range of y values
+     * @return {Plots.PlotData} The plot data that corresponds to the ranges
+     */
+    public plotDataIn(xRange: Extent, yRange: Extent) {
```
bluong: To correlate with the other `plotData` API should we include a datasets parameter?  Possibly out of scope.

plottable.d.ts
```diff
@@ -2767,16 +2767,20 @@ declare module Plottable {
...
-            getBars(xValOrExtent: number | Extent, yValOrExtent: number | Extent): D3.Selection;
+            plotDataAt(p: Point): PlotData;
+            /**
+             * Gets the {Plots.PlotData} that correspond to a given xRange/yRange
+             *
+             * @param {Extent} xRange The specified range of x values
+             * @param {Extent} yRange The specified range of y values
+             * @return {Plots.PlotData} The plot data that corresponds to the ranges
+             */
+            plotDataIn(xRange: Extent, yRange: Extent): PlotData;
```
bluong: Do we want the names of the types of the parameters to be `Extent`s?  We should have the type name match the parameter name where we can

bluong:
Just names at this point

plottable.d.ts
rcchen: I vaguely recall an in-person discussion about this with @jtlan where we came to the conclusion that they should be named `xRange` and `yRange` because "range" is more or less the only part of an `Extent` that we care about. I could be wrong though..

crmorford:
So would 
plot.plotDataIn({min: 0, max: 200}, {min: 400, max: 500});
Look for data where:
1. x is inbetween 0-200 OR 400-500?
2. x is 0-200, OR y is inbetween 400-500?
3. x is 0-200, AND y is inbetween 400-500?

rcchen:
> So would `plot.plotDataIn({min: 0, max: 200}, {min: 400, max: 500});`
> Look for data where:
> 1. x is in between 0-200 OR 400-500?
> 2. x is 0-200, OR y is in between 400-500?
> 3. x is 0-200, AND y is in between 400-500?
(3)

crmorford:
That feels inconsistent with this fiddle http://jsfiddle.net/rcchen/tj79xsg4/. None of the data has an x value between 200 and 500, but 
var dataIn = plot.plotDataIn({min: 200, max: 500},{min: 0, max: 500}).data.length == 2

rcchen:
`plotDataIn` and `plotDataAt` work on pixel points, not actual chart values. The reason for this is that these methods are typically bound to `Interactions` such as `Drag` and `Point`

rcchen:
I just updated the JSFiddle with a mouse listener so you can verify that data is getting picked up from the pixel position of the mouse.

crmorford:
OH okay. Didn't realize it was pixel position. Thanks :)

crmorford:
Think about the workflow where you drag, and then want to use the bounds to find the data inside. Users are going to be writing code to turn bounds into an xrange and yrange over and over. 

It might make more sense to have plotDataIn take bounds as a parameter.

src/components/plots/barPlot.ts
```diff
@@ -237,38 +237,51 @@ export module Plots {
...
+     */
+    public plotDataIn(xRange: Range, yRange: Range): PlotData;
+    public plotDataIn(xRangeOrBounds: Range | Bounds, yRange?: Range): PlotData {
+      var dataXRange = (<Range> xRangeOrBounds);
+      var dataYRange: Range = yRange;
+      if (yRange == null) {
+        var bounds = (<Bounds> xRangeOrBounds);
+        dataXRange = { min: bounds.topLeft.x, max: bounds.bottomRight.x };
+        dataYRange = { min: bounds.topLeft.y, max: bounds.bottomRight.y };
+      }
```
bluong: Since it's consistent + it could save us in development downstream, let's have the
`if (yRange == null)` check at near the first line  I'm not really a fan of having the sanitation being done at the third line

bluong: I think something that looks a lot more intuitive is
```typescript
var dataXRange: Range;
var dataYRange: Range;
if (yRange == null)
  set ranges to according bounds
else 
  set ranges to arguments
```

I guess the thing I'm concerned about is that the casting is done in decently odd places, and I would like the casting to be done as consistently as possible

bluong:
Tslint issues.  Fix those and then pass to @crmorford

PR #2152, opened 2015-05-23
[Animators] Streamlining variables and consolidating API

### LINKED ISSUE #1916: Animator API needs some rework
<del>Mostly its inheritance structure is kind of weird.</del>

A bigger problem is that `Animator`s are added to `Plot`s using string keys, which can be kind of obscure.

rcchen:
An initial stab at the Animators API issue. All plots have now been standardized to `"main"` and `"reset"` keys for the `Animator.Baseline()` and `Animator.Null()` states, respectively. In addition, these stringy constants have been moved to `Plottable.Plots.Animator`

Fixes #1916


src/components/plots/plot.ts
```diff
@@ -23,6 +23,11 @@ module Plottable {
       accessor: Accessor<any>;
       scale?: Scale<D, R>;
     }
+
+    export class Animator {
```
jtlan: We should probably standardize around what we want to do for `string`y enums. I think it'd be better if this was an exported `static var` instead of a `class`.

src/components/plots/plot.ts
```diff
@@ -36,7 +41,10 @@ module Plottable {
     private _extentsProvider: Scales.ExtentsProvider<any>;
 
     protected _animate: boolean = false;
-    private _animators: Animators.PlotAnimatorMap = {};
+    private _animators: Animators.PlotAnimatorMap = {
+      "main": new Animators.Base(),
+      "reset": new Animators.Null()
```
jtlan: We should use the `static` values here.

rcchen: The compiler won't let me use references to variables as keys

jtlan: What if you construct it across multiple lines:
```Typescript
private _animators: {[animator: String]: PlotAnimator} = {};
_animators[Animator.MAIN] = new Animators.Base();
_animators[Animator.RESET] = new Animators.Null();
```

jtlan:
I think we should rename `PlotAnimator` to `Animators.Plot`, to keep with the style we use elsewhere.

src/components/plots/plot.ts
rcchen: It thinks that `_animators` is a duplicate variable. I ended up taking the first line of your suggestion and initializing the latter two in the constructor, where they probably should belong anyways.

src/animators/baseAnimator.ts
```diff
@@ -15,7 +15,7 @@ export module Animators {
    * min(maxIterativeDelay(),
    *   max(maxTotalDuration() - duration(), 0) / <number of iterations>)
    */
-  export class Base implements PlotAnimator {
+  export class Base implements Plot {
```
jtlan: Hmm, with the rename, this class declaration looks pretty weird. Maybe use `Animators.Plot`, even though it's slightly redundant?

src/components/plots/plot.ts
```diff
@@ -23,6 +23,11 @@ module Plottable {
       accessor: Accessor<any>;
       scale?: Scale<D, R>;
     }
+
+    export module Animator {
```
jtlan: Should this be inside the `Plot` class? So it reads `Plot.Animator.MAIN`

rcchen: Eh I'm going to disagree with this one and say it should be `Plots.Animator.MAIN`. Not much reason other than "it seemed strange when I was going through and refactoring to `Plot.Animator`", though a secondary reason might be avoiding [merging declarations for a class and a module](http://www.typescriptlang.org/Handbook#declaration-merging)

PR #2153, opened 2015-05-23
[Drawers.Line] - Logic moved to Plots.Line

bluong:
All of the logic is now on the `Plots.Line` as part of #1913


src/components/plots/linePlot.ts
```diff
@@ -4,6 +4,8 @@ module Plottable {
 export module Plots {
   export class Line<X> extends XYPlot<X, number> {
 
+    public static SELECTION_CLASS = "line";
```
jtlan: I don't think this is the right name for the field. People might grab the elements without using d3, for example.

src/components/plots/linePlot.ts
```diff
@@ -4,6 +4,8 @@ module Plottable {
 export module Plots {
   export class Line<X> extends XYPlot<X, number> {
 
+    public static PLOT_CLASS = "line";
```
jtlan: Do we have a consistent name for this concept across multiple `Plot` types?

src/components/plots/linePlot.ts
```diff
@@ -43,7 +48,22 @@ export module Plots {
...
+        var yProjector = Plot._scaledAccessor(this.y());
+
+        var definedProjector = (d: any, i: number, dataset: Dataset) => {
+          var positionX = xProjector(d, i, dataset);
+          var positionY = yProjector(d, i, dataset);
+          return positionX != null && positionX === positionX &&
+                 positionY != null && positionY === positionY;
+        };
+
+        attrToProjector["d"] = (datum: any, index: number, dataset: Dataset) => {
```
jtlan: Why is this logic in `_generateDrawSteps()` as well as in `_propertyProjectors()`?

src/components/plots/linePlot.ts
```diff
@@ -43,7 +48,22 @@ export module Plots {
       var drawSteps: Drawers.DrawStep[] = [];
       if (this._dataChanged && this._animate) {
         var attrToProjector = this._generateAttrToProjector();
-        attrToProjector["y"] = this._getResetYFunction();
+        var xProjector = Plot._scaledAccessor(this.x());
+        var yProjector = Plot._scaledAccessor(this.y());
+
+        var definedProjector = (d: any, i: number, dataset: Dataset) => {
```
jtlan: Do we still want to call these "projectors"?

test/components/plots/areaPlotTests.ts
```diff
@@ -59,7 +59,8 @@ describe("Plots", () => {
       renderArea = (<any> areaPlot)._renderArea;
     });
 
-    it("draws area and line correctly", () => {
+    // Need to deal with repercussions in area plot
+    it.skip("draws area and line correctly", () => {
```
jtlan: I assume these will be un-skipped in a future commit?

jtlan:
Question about factoring out repeated logic.

test/components/plots/areaPlotTests.ts
bluong: They will be fixed in the `Plots.Area` changes

src/components/plots/linePlot.ts
```diff
@@ -4,6 +4,8 @@ module Plottable {
 export module Plots {
   export class Line<X> extends XYPlot<X, number> {
 
+    public static PATH_CLASS = "line";
```
jtlan: Isn't this name going to cause a problem when we need a similar field on `Area`?

src/components/plots/linePlot.ts
```diff
@@ -21,6 +23,9 @@ export module Plots {
 
       this.attr("stroke", new Scales.Color().range()[0]);
       this.attr("stroke-width", "2px");
+      this.attr("vector-effect", "non-scaling-stroke");
+      this.attr("fill", "none");
+      this.attr("class", Line.PATH_CLASS);
```
jtlan: If the user sets their own class with `attr`, that will override this, correct?

src/drawers/lineDrawer.ts
```diff
@@ -12,25 +12,12 @@ export module Drawers {
...
-
-      return d3.svg.line()
-                   .x(xFunction)
-                   .y(yFunction)
-                   .defined(definedFunction);
+    public setup(line: D3.Selection) {
+      this._pathSelection = line.append("path")
+                                .classed(Line.PATH_CLASS, true)
+                                .style("fill", "none")
+                                .style("vector-effect", "non-scaling-stroke");
```
jtlan: Let's not have the `vector-effect` any more, since `non-scaling-stroke` doesn't work on IE. I'm not actually sure why the property was initially added.

jtlan:
Remove `non-scaling-stroke`?

PR #2156, opened 2015-05-23
Area plot

bluong:
All of the logic is now on the `Plots.Area` as part of #1913


src/components/plots/areaPlot.ts
```diff
@@ -6,6 +6,7 @@ export module Plots {
    * An AreaPlot draws a filled region (area) between the plot's projected "y" and projected "y0" values.
    */
   export class Area<X> extends Line<X> {
+    public static PLOT_CLASS = "area";
```
jtlan: We should use a consistent name between this and the `Plots.Line` version of this concept. Something like `LINE_PATH_CLASS` and `AREA_PATH_CLASS`?

src/components/plots/areaPlot.ts
```diff
@@ -77,6 +80,59 @@ export module Plots {
...
+      drawSteps.push({attrToProjector: this._generateLineAttrToProjector(), animator: this._getAnimator("main")});
+      return drawSteps;
+    }
+
+    private _generateLineAttrToProjector() {
+      var lineAttrToProjector = this._generateAttrToProjector();
+      var fillProjector = lineAttrToProjector["fill"];
+      lineAttrToProjector["stroke"] = fillProjector;
+      lineAttrToProjector["fill"] = () => "none";
+      lineAttrToProjector["vector-effect"] = () => "non-scaling-stroke";
```
jtlan: Probably we shouldn't have this anymore...?

src/components/plots/areaPlot.ts
```diff
@@ -77,6 +80,59 @@ export module Plots {
...
+      }
+      drawSteps.push({attrToProjector: this._generateLineAttrToProjector(), animator: this._getAnimator("main")});
+      return drawSteps;
+    }
+
+    private _generateLineAttrToProjector() {
+      var lineAttrToProjector = this._generateAttrToProjector();
+      var fillProjector = lineAttrToProjector["fill"];
+      lineAttrToProjector["stroke"] = fillProjector;
+      lineAttrToProjector["fill"] = () => "none";
```
jtlan: This probably should be a style, and not an attribute.

src/components/plots/areaPlot.ts
```diff
@@ -77,6 +80,59 @@ export module Plots {
...
+        drawSteps.push({attrToProjector: attrToProjector, animator: this._getAnimator("reset")});
+      }
+      drawSteps.push({attrToProjector: this._generateLineAttrToProjector(), animator: this._getAnimator("main")});
+      return drawSteps;
+    }
+
+    private _generateLineAttrToProjector() {
+      var lineAttrToProjector = this._generateAttrToProjector();
+      var fillProjector = lineAttrToProjector["fill"];
+      lineAttrToProjector["stroke"] = fillProjector;
```
jtlan: This overrides the user's settings if they called `attr("stroke", ...)`.

src/components/plots/areaPlot.ts
```diff
@@ -77,6 +80,59 @@ export module Plots {
...
+      lineAttrToProjector["fill"] = () => "none";
+      lineAttrToProjector["vector-effect"] = () => "non-scaling-stroke";
+      lineAttrToProjector["stroke-width"] = () => "2px";
```
jtlan: User setting overridden here too...

src/components/plots/areaPlot.ts
```diff
@@ -77,6 +80,59 @@ export module Plots {
...
+      lineAttrToProjector["class"] = () => Line.SELECTION_CLASS;
```
jtlan: ... and here.

jtlan:
A whole bunch of places where the user's settings using `attr()` are getting overridden.

src/components/plots/areaPlot.ts
bluong: Note that the name is `_generateLineAttrToProjector()`  I probably should've put tsdoc on this function, but it basically will infer the attributes of the line path based on the attributes placed on the `Plots.Area`

src/components/plots/areaPlot.ts
bluong: Some of this logic used to be on the drawer level.  What's the difference between having this ideas be a style vs an attribute

src/components/plots/areaPlot.ts
bluong: Note that this is all under `_generateLineAttrToProjector` meaning that basically the line path is basically inferring all the projectors based on the area's attrToProjector

src/components/plots/areaPlot.ts
jtlan: The area `<path>` should never have a stroke, so setting it as a CSS style overrides any `attr` choices.

src/components/plots/areaPlot.ts
jtlan: How does the user set the line properties then? What if they want a different stroke based on data properties?

bluong: Note that the core of the problem even existing on the current implementation of `Plots.Area` is that there is no way to specify the attributes for the line and area separately.  Both are forced to infer their attributes from `_generateAttrToProjector`.

I'll play around with adding logic to the `Drawer`s (this is the only way to achieve styling) so that we can force certain ideas for attributes.

src/components/plots/areaPlot.ts
```diff
@@ -141,6 +168,22 @@ export module Plots {
 
       return { data: allPlotData.data, pixelPoints: allPlotData.pixelPoints, selection: d3.selectAll(allElements) };
     }
+
+    protected _constructAreaProjector(xProjector: _Projector, yProjector: _Projector, y0Projector: _Projector) {
+      var definedProjector = (d: any, i: number, dataset: Dataset) => {
+        var positionX = Plot._scaledAccessor(this.x())(d, i, dataset);
+        var positionY = Plot._scaledAccessor(this.y())(d, i, dataset);
+        return positionX != null && positionX === positionX &&
+               positionY != null && positionY === positionY;
```
jtlan: Can we use `Utils.isValidNumber()` here?

src/components/plots/areaPlot.ts
```diff
@@ -141,6 +168,22 @@ export module Plots {
...
+      };
+      return (datum: any, index: number, dataset: Dataset) => {
```
jtlan: Should we specify that `datum` is an `any[]`?

src/components/plots/areaPlot.ts
```diff
@@ -141,6 +168,22 @@ export module Plots {
...
+        return d3.svg.area()
+                     .x((innerDatum, innerIndex) => xProjector(innerDatum, innerIndex, dataset))
+                     .y1((innerDatum, innerIndex) => yProjector(innerDatum, innerIndex, dataset))
+                     .y0((innerDatum, innerIndex) => y0Projector(innerDatum, innerIndex, dataset))
+                     .defined((innerDatum, innerIndex) => definedProjector(innerDatum, innerIndex, dataset))(datum, index);
```
jtlan: It might be easier to read this if we declared the D3 shape generator as a local var:
```Typescript
var areaGenerator = d3.svg.area()
                       .x((innerDatum, innerIndex) => xProjector(innerDatum, innerIndex, dataset))
                       .y((innerDatum, innerIndex) => this._getResetYFunction()(innerDatum, innerIndex, dataset))
                       .y0((innerDatum, innerIndex) => Plot._scaledAccessor(this.y0())(innerDatum, innerIndex, dataset))
                       .defined((innerDatum, innerIndex) => definedProjector(innerDatum, innerIndex, dataset));
return areaGenerator(datum, index);
```

src/components/plots/areaPlot.ts
bluong: It would not be technically consistent, but that said, this should compile and it would be a bit more explanatory.  Note sure

src/components/plots/areaPlot.ts
```diff
@@ -141,6 +168,22 @@ export module Plots {
...
       return { data: allPlotData.data, pixelPoints: allPlotData.pixelPoints, selection: d3.selectAll(allElements) };
     }
+
+    protected _constructAreaProjector(xProjector: _Projector, yProjector: _Projector, y0Projector: _Projector) {
+      var definedProjector = (d: any, i: number, dataset: Dataset) => {
+        var positionX = Plot._scaledAccessor(this.x())(d, i, dataset);
+        var positionY = Plot._scaledAccessor(this.y())(d, i, dataset);
+        return Utils.Methods.isValidNumber(positionX) && Utils.Methods.isValidNumber(positionY);
+      };
+      return (datum: any[], index: number, dataset: Dataset) => {
```
jtlan: Now the naming of the argument seems a bit weird, since it should be something like "`arrayDatum`"...

Issue #1984, opened 2015-05-07
[Drawer] - Should be bound to a dataset, not a key


jtlan:
`Drawer`s should take their `Dataset` directly -- a number of `HACKHACK`s have been created in tests by #1977.

bluong:
API Refactor since `Drawer`s are now considered part of the API

bluong:
Done!

PR #2158, opened 2015-05-26
[Drawer] - Constructor takes a Dataset instead of a key

bluong:
The key idea on the `Dataset` we found is not useful.  The key idea is only useful if there's a way to turn that key into something useful but that transformation is all stored within the `Plot`...  Instead, we will have the idea that `Drawer`s will be binded to a `Dataset` that it will draw.  This ties nicely to the fact that the input `key` was intended to be the dataset key anyways in the previous implementation.

Changes:
* The constructor now takes in a Dataset instead of a pretty arbitrary string key
* The `draw()` API no longer takes in a `Dataset` argument as the last argument
* A protected `dataset()` API has been added to `Drawer`

After this makes it to develop, #1984 can be closed


src/drawers/drawer.ts
```diff
@@ -77,12 +81,11 @@ export module Drawers {
       return data.length;
     }
 
-    private _applyMetadata(attrToProjector: AttributeToProjector,
-                          dataset: Dataset): AttributeToAppliedProjector {
+    private _applyMetadata(attrToProjector: AttributeToProjector): AttributeToAppliedProjector {
```
jtlan: Probably should be renamed to `_applyDataset()`?

src/drawers/rectDrawer.ts
```diff
@@ -84,16 +84,16 @@ export module Drawers {
       this._labelsTooWide = labelTooWide.some((d: boolean) => d);
     }
 
-    public draw(data: any[], drawSteps: DrawStep[], userMetadata: any) {
+    public draw(data: any[], drawSteps: DrawStep[]) {
```
jtlan: Uh, was this just *broken* before?

test/components/plots/plotTests.ts
```diff
@@ -155,9 +155,12 @@ describe("Plots", () => {
       var svg = TestMethods.generateSVG(400, 400);
       var plot = new Plottable.Plot();
 
+      var dataset1 = new Plottable.Dataset([{value: 0}, {value: 1}, {value: 2}]);
+      var dataset2 = new Plottable.Dataset([{value: 1}, {value: 2}, {value: 3}]);
+
       // Create mock drawers with already drawn items
       // HACKHACK #1984: Dataset keys are being removed, so this is the internal key
-      var mockDrawer1 = new Plottable.Drawers.AbstractDrawer("_0");
+      var mockDrawer1 = new Plottable.Drawers.AbstractDrawer(dataset1);
```
jtlan: Remove the `HACKHACK` comment since you're fixing the problem. You actually get to close #1984 in fact.

src/drawers/rectDrawer.ts
bluong: I wouldn't say broken... just majorly incorrect...
Yeah...
(The main of the problem is that we don't have a hard type on `metadata`, thus delegating to `any`)

src/drawers/rectDrawer.ts
```diff
@@ -11,8 +11,8 @@ export module Drawers {
     private _measurer: SVGTypewriter.Measurers.CacheCharacterMeasurer;
     private _writer: SVGTypewriter.Writers.Writer;
 
-    constructor(key: string, isVertical: boolean) {
-      super(key);
+    constructor(dataset: Dataset, isVertical: boolean) {
```
jtlan: Since the smarts are moving to the `Plot`, do we still need `isVertical`?

src/drawers/drawer.ts
```diff
@@ -36,16 +36,20 @@ export module Drawers {
...
-        this.key = key;
+    constructor(dataset: Dataset) {
+        this._dataset = dataset;
     }
 
     public setup(area: D3.Selection) {
       this._renderArea = area;
     }
 
+    protected dataset() {
```
jtlan: Why not just have the field be `protected`? Also the method should be underscore-named right?

src/drawers/rectDrawer.ts
bluong: There is some weirdness in which the text needs to know information about the plot's verticalness to write the text.  I will highly consider removing this constructor argument when I do remove it.

src/drawers/drawer.ts
bluong: There was a comment that @aicioara made that at his previous workplace, there was actually no concept of `protected` instance variables and that the way to get at the values would be through `protected` getters/setters.  After hearing about this, I think I am a huge fan of this model.

The biggest thing for me personally is that this will guard against setting variables in a parent class when they should have been only set at initialization.

Another benefit is that this will guard against API breaks since it will essentially allow us to put architecture surround these instance variables instead of setting them directly / indirectly

jtlan: Okay, makes sense to me.

src/drawers/drawer.ts
```diff
@@ -77,12 +81,11 @@ export module Drawers {
...
     }
 
-    private _applyMetadata(attrToProjector: AttributeToProjector,
-                          dataset: Dataset): AttributeToAppliedProjector {
+    private _appliedProjectors(attrToProjector: AttributeToProjector): AttributeToAppliedProjector {
       var modifiedAttrToProjector: AttributeToAppliedProjector = {};
       d3.keys(attrToProjector).forEach((attr: string) => {
         modifiedAttrToProjector[attr] =
-          (datum: any, index: number) => attrToProjector[attr](datum, index, dataset);
+          (datum: any, index: number) => attrToProjector[attr](datum, index, this.__dataset);
```
jtlan: We should use the getter here right?

bluong: This is inside the class where the instance variable is, and we're currently undecided on if we want to stick with getters or setters.

That said, I'm ok with using a getter here and I think using getters is more consistent even though it's slightly less performant

PR #2159, opened 2015-05-27
[Plots.StackedArea] - removing redundant class line

bluong:
It's already on `Plots.Area`.  Shouldn't need to be redeclared in `Plots.StackedArea`


aicioara:
Not sure what is the intended purpose of these CSS classes in the wild, but I am thinking we should distinguish between this and AreaPlot, maybe add a `stacked` class or something? Same for StackedBarPlot. I don't have a strong preference though, just makes more sense to me

PR #2160, opened 2015-05-27
[Plottable.css] - Cleaning useless declarations

bluong:
Quite a few declarations that we don't really document anywhere.

We should think about a way to document this better if possible

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -84,10 +84,6 @@ svg.plottable {
   font-weight: bold;
 }
 
-.plottable .text-label-vertical {
-  writing-mode: tb;
-}
-
 .plottable .axis line.baseline {
   stroke: #CCC;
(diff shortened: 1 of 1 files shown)


aicioara:
Can't really check the correctness of these, since they are not documented. The code itself looks good, but I'd give it QE as well to make sure no funny business happening

crmorford:
how are we sure no one is dependent on these styles?

bluong:
@crmorford As discussed with @jtlan , we can't be sure on how no one is dependent.  We should just be sure that our own stylings work nicely with what we have declared ourselves (I am mainly removing these rules since there are 0 references to these rules in our codebase) and then provide the obsolete rules in the upgrade guide

crmorford:
Cool, just note it in the upgrade instructions

PR #2161, opened 2015-05-27
[Drawers.Element] - x/y/defined attributes removed

bluong:
There was filtering done on the Drawer level.  This is now done on the appropriate `XYPlot` level.

### DIFF
--- a/src/components/plots/linePlot.ts
+++ b/src/components/plots/linePlot.ts
@@ -54,20 +54,14 @@ export module Plots {
 
     protected _generateAttrToProjector() {
       var attrToProjector = super._generateAttrToProjector();
-      var wholeDatumAttributes = this._wholeDatumAttributes();
-      var isSingleDatumAttr = (attr: string) => wholeDatumAttributes.indexOf(attr) === -1;
-      var singleDatumAttributes = d3.keys(attrToProjector).filter(isSingleDatumAttr);
-      singleDatumAttributes.forEach((attribute: string) => {
+      d3.keys(attrToProjector).forEach((attribute: string) => {
+        if (attribute === "d") { return; }
         var projector = attrToProjector[attribute];
         attrToProjector[attribute] = (data: any[], i: number, dataset: Dataset) =>
           data.length > 0 ? projector(data[0], i, dataset) : null;
       });
 
       return attrToProjector;
-    }
-
-    protected _wholeDatumAttributes() {
-      return ["x", "y", "defined", "d"];
     }
 
     public getAllPlotData(datasets = this.datasets()): Plots.PlotData {
--- a/src/components/plots/linePlot.ts
+++ b/src/components/plots/linePlot.ts
@@ -172,6 +172,14 @@ export module Plots {
                      .defined((innerDatum, innerIndex) => definedProjector(innerDatum, innerIndex, dataset))(datum, index);
       };
     }
+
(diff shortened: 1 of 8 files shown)


src/components/plots/xyPlot.ts
```diff
@@ -229,8 +216,26 @@ module Plottable {
...
+    }
+
+    protected _getDataToDraw() {
+      var datasets: D3.Map<any[]> = super._getDataToDraw();
+
+      var definedFunction = (d: any, i: number, dataset: Dataset) => {
+        var positionX = Plot._scaledAccessor(this.x())(d, i, dataset);
+        var positionY = Plot._scaledAccessor(this.y())(d, i, dataset);
+        return positionX != null && positionX === positionX &&
+               positionY != null && positionY === positionY;
```
jtlan: Should we use `Utils.isValidNumber()` here?

src/components/plots/xyPlot.ts
```diff
@@ -154,19 +154,6 @@ module Plottable {
       return this;
     }
 
-    protected _propertyProjectors(): AttributeToProjector {
```
jtlan: ??? was this not being used at all?

src/components/plots/linePlot.ts
```diff
@@ -54,10 +54,8 @@ export module Plots {
 
     protected _generateAttrToProjector() {
       var attrToProjector = super._generateAttrToProjector();
-      var wholeDatumAttributes = this._wholeDatumAttributes();
-      var isSingleDatumAttr = (attr: string) => wholeDatumAttributes.indexOf(attr) === -1;
-      var singleDatumAttributes = d3.keys(attrToProjector).filter(isSingleDatumAttr);
-      singleDatumAttributes.forEach((attribute: string) => {
+      d3.keys(attrToProjector).forEach((attribute: string) => {
+        if (attribute === "d") { return; }
```
jtlan: Why this change?

src/components/plots/linePlot.ts
```diff
@@ -172,6 +166,14 @@ export module Plots {
                      .defined((innerDatum, innerIndex) => definedProjector(innerDatum, innerIndex, dataset))(datum, index);
       };
     }
+
+    protected _getDataToDraw() {
```
jtlan: Why does `Line` have a different implementation for this method?

bluong: Because `Line` and `Area` have their own separate way of dealing with undefined data within their appropriate path generators

src/components/plots/linePlot.ts
bluong: Before the code would loop over anything that is not a `wholeDatumAttribute` which is defined to be x/y/defined/d from `generateAttrToProjector()`

Now that x/y/defined are all gone from `generateAttrToProjector()`, the only one left is d

src/components/plots/xyPlot.ts
bluong: Probably.  It'd make the code much more readable

src/components/plots/xyPlot.ts
bluong: It was used to set up the xProjector and the yProjector for pieces like `Plots.Bar` and `Drawers.Element`.  Now that the projectors are actually set up correctly, this override can finally go away.

PR #2162, opened 2015-05-27
[RenderController] Stringified policies

### LINKED ISSUE #2061: Remove RenderPolicy type?
Shold `RenderPolicy` exist as a distinct type, or is it sufficient to allow the user to set the rendering policy using a `string`y `enum`?

rcchen:
Decided not to remove the `RenderPolicy` type and instead just do some stringy enum cleanup in `RenderController.ts`

Fixes #2061

### DIFF
--- a/src/core/renderController.ts
+++ b/src/core/renderController.ts
@@ -31,19 +31,19 @@ module Plottable {
     }
     export var _renderPolicy: RenderPolicies.RenderPolicy = new RenderPolicies.AnimationFrame();
 
-    export function setRenderPolicy(policy: string): void {
-      switch ((<string> policy).toLowerCase()) {
+    export function setRenderPolicy(policy: string) {
+      switch (policy.toLowerCase()) {
         case Policy.IMMEDIATE:
-        _renderPolicy = new RenderPolicies.Immediate();
-        break;
+          _renderPolicy = new RenderPolicies.Immediate();
+          break;
         case Policy.ANIMATION_FRAME:
-        _renderPolicy = new RenderPolicies.AnimationFrame();
-        break;
+          _renderPolicy = new RenderPolicies.AnimationFrame();
+          break;
         case Policy.TIMEOUT:
-        _renderPolicy = new RenderPolicies.Timeout();
-        break;
+          _renderPolicy = new RenderPolicies.Timeout();
+          break;
         default:
-        Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
+          Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
         return;
       }
     }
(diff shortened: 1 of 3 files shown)


src/core/renderController.ts
```diff
@@ -24,26 +24,28 @@ module Plottable {
...
-          break;
-          case "timeout":
-          policy = new RenderPolicies.Timeout();
-          break;
-          default:
-          Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
-          return;
-        }
+    export function setRenderPolicy(policy: string): void {
+      switch ((<string> policy).toLowerCase()) {
```
aicioara: The cast is not needed, policy is of type string

src/core/renderController.ts
```diff
@@ -24,26 +24,28 @@ module Plottable {
...
-          policy = new RenderPolicies.AnimationFrame();
-          break;
-          case "timeout":
-          policy = new RenderPolicies.Timeout();
-          break;
-          default:
-          Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
-          return;
-        }
+    export function setRenderPolicy(policy: string): void {
```
aicioara: Return type is not needed, it is infered

src/core/renderController.ts
```diff
@@ -24,26 +24,28 @@ module Plottable {
...
+        break;
+        case Policy.ANIMATION_FRAME:
+        _renderPolicy = new RenderPolicies.AnimationFrame();
+        break;
+        case Policy.TIMEOUT:
+        _renderPolicy = new RenderPolicies.Timeout();
+        break;
+        default:
+        Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
+        return;
```
aicioara: we don't have any tslint rule to enforce this, but try to stick to some consistency. In general the statements after the `case`, including the `break` are on indented line

aicioara:
On a side note, I thought branch names should not have dashes in them?

rcchen:
@aicioara if they shouldn't, this is the first I've heard of it. That being said, it seems like both you and @bluong are using the [issue]fix convention without the dash, so I'll switch to that once my branches all clear out.

src/core/renderController.ts
```diff
@@ -24,26 +24,28 @@ module Plottable {
...
-          policy = new RenderPolicies.Timeout();
+        case Policy.TIMEOUT:
+          _renderPolicy = new RenderPolicies.Timeout();
           break;
-          default:
+        default:
           Utils.Methods.warn("Unrecognized renderPolicy: " + policy);
-          return;
-        }
+        return;
```
aicioara: No need to return, as that's the default at the end of the function

rcchen:
This was a straightforward renaming operation so I'll merge it in after Travis goes through.

PR #2166, opened 2015-05-27
[Drawers.Rect] - Text drawing moved to Plots.Bar

bluong:
The drawing of text for `Plots.Bar` labels used to be done on `Drawers.Rect`.  Now it is done on the `Plots.Bar` itself.  In the future, we may move this to a possible `Drawers.Text` but since moving to that idea is not an API break, we can do that at a later time (I could also do it here... but that could be timely)

Changes:
* `rectDrawer.removeLabels()` has been deleted
* `rectDrawer._getIfLabelsTooWide()` has been deleted
* `rectDrawer.drawText()` has been deleted
* `Drawers.Rect` no longer takes in an `isVertical` flag as the second argument to the constructor


test/components/plots/barPlotTests.ts
```diff
@@ -197,7 +197,7 @@ describe("Plots", () => {
         var d0Px: Plottable.Point, d1Px: Plottable.Point;
 
         beforeEach(() => {
-          bars = d3.selectAll(".bar-area rect");
+          bars = barPlot.getAllSelections();
```
jtlan: :/

src/components/plots/barPlot.ts
```diff
@@ -142,6 +159,31 @@ export module Plots {
       }
     }
 
+    public addDataset(dataset: Dataset) {
+      if (this._isSetup) {
+        var labelArea = this._renderArea.append("g").classed(Bar._LABEL_AREA_CLASS, true);
```
jtlan: Is there a way to move this logic to whatever logic is normally used to create new `<g>`s for the `Drawer`s?

src/components/plots/barPlot.ts
```diff
@@ -16,6 +19,9 @@ export module Plots {
     private _labelFormatter: Formatter = Formatters.identity();
     private _labelsEnabled = false;
     private _hideBarsIfAnyAreTooWide = true;
+    private _labelAreas: Utils.Map<Dataset, D3.Selection>;
+    private _labelMeasurers: Utils.Map<Dataset, SVGTypewriter.Measurers.Measurer>;
+    private _labelWriters: Utils.Map<Dataset, SVGTypewriter.Writers.Writer>;
```
jtlan: It seems like there would be less trouble if you had a single `Map` to a composite object.

jtlan:
Questions about streamlining. In particular, why are there two identical declarations of `_getDataToDraw()`?

bluong:
There is some logic done in `Drawers.Rect` where it prefilters data that result in invalid width/height/x/y and thus I put the prefilter in the `Plot`  The reason why there are 2 identical declarations is because there is no hierarchical join between `Plots.Bar` and `Plots.Rectangle`

src/components/plots/barPlot.ts
bluong: From an instance variable point of view, yeah.  From an implementation point of view, it's 1 step more trouble unless we have a good semantic way to group area / measurer / writer.

src/components/plots/barPlot.ts
bluong: in `addDataset()` on the parent `Plot` level there's a statement that uses `if (this._isSetup)`.  I could put the setup in there, or leave it here.

jtlan: Maybe we should have some sort of `_setUpDrawer()` call that gets invoked in `Plot`'s `addDataset()`?

bluong: Well it would be more aligned to DOM nodes that are correlated to each `Dataset`

src/components/plots/barPlot.ts
jtlan: Just an unnamed object will do:
```Typescript
private datasetToLabel: Utils.Map<Dataset, {
  labelArea: D3.Selection;
  measurer: SVGTypewriter.Measurers.Measurer;
  writer: SVGTypewriter.Writers.Writer;
}>;
```

jtlan:
Would it make sense to have the prefilter inside the `Drawer`? The danger is that the "projected" version has bad values, rather than the raw data values being bad.

PR #2167, opened 2015-05-28
ban d3.keys

bluong:
We should be using `Object.keys()` instead


aicioara:
What is the reasoning behind this?

aicioara:
> Should be ban d3.keys?  it doesn't have the hasOwnProperty filtering whereas Object.keys does I believe (@bluong)

PR #2170, opened 2015-05-28
Drawer calculations moved to Plot

bluong:
Drawers like `Drawers.Line` and `Drawers.Arc` would normally be able to take in an `AttributeToProjector` that would convert property attributes like "x" and "y" and transform them into DOM attributes like "d".  Now, all of those calculations are done at the `Plot` level meaning that the `Drawer`s are very thin, taking only DOM attributes

QE - From the `Drawer`s perspective, a lot of calculation has been removed so that `Drawer` must be handed the raw attributes to the DOM element.  From the `Plot`s perspective, nothing should be changed.  Thus I think this is mainly a regression pass


dbt:
This should mostly be a regression pass -- no API changes, just code moved from one place to another.

crmorford:
All my quicktests still work +1

PR #2171, opened 2015-05-28
Retrieve internal Interaction of DragBoxLayer with dragInteraction()

jtlan:
Since this is just a getter for an existing variable, I think it's probably [NOQE].


bluong:
Let's have @crmorford play with it in case she can think of potential use cases / pit falls with this idea

jtlan:
We're pretty low on QE time, and this is more a poweruser API point.

PR #2172, opened 2015-05-28
removing showEndTickLabel

bluong:
The function actually did absolutely nothing.  If we need to add the concept in, we can add it in at a later date.


aicioara:
I believe related to https://github.com/palantir/plottable/pull/1946

aicioara:
RFC: How about this method on the `Axes.Axis`. It appears to be in the same situation. Do we want this endpoint at all?

bluong:
We have knowledge that a customer is using the feature on the base axis, so we shouldn't remove that endpoint unless we know they aren't using it.

Issue #1762, opened 2015-03-24
Scale.ModifiedLog produces weird ticks under reasonable circumstances


jtlan:
Setting `domain([0, 10])` will cause no ticks to appear:

http://jsfiddle.net/rtdxr3xv/

aicioara:
Progress so far (all on 1762fix branch)
- the linear ticks in `modifiedLogScale.tick()` did not produce any ticks unless `showIntermediateTicks(true)` was invoked on the scale. Removed that limitation.
- The labels (and ticks) are no longer evenly spaced for a log scale, hence `abstractAxis._hideOverflowingTickLabels()` had to change, because at the moment there is some hack in place which just evenly spreads the labels. Removed that hack, but now will break #1196 
- Looking into http://vis.stanford.edu/files/2010-TickLabels-InfoVis.pdf as suggested by @jtlan

aicioara:
Updated fiddle v

PR #2173, opened 2015-05-28
[Scales.ModifiedLog] - removing showIntermediateTicks

bluong:
In a recent API Discussion, there was talk about the use of this idea and talk about removing it.


jtlan:
Related: #1762.

crmorford:
Let's see if we can find out why we did this and who wanted it, and if they still need it.

jtlan:
This feature was originally added in #710 (commit 8d6e26e); I think it was built for Slate. No one has touched it since then, so I think it's a fairly safe default. A `TickGenerator` can be used to generate more specific behavior.

dbt:
I'm punting this to rc5 since we've already released rc4.  Is this ready to merge?

jtlan:
@dbt @christinaforney was looking to see if people still cared about this feature. I think it's safe to remove it, since `TickGenerator`s can be used to fill the gap in functionality, if it was important to anyone.

jtlan:
[+1], since `TickGenerator`s are a better way to accomplish this. For that matter, can we set the default `TickGenerator` on `ModifiedLog` to only produce powers of the base?

PR #2174, opened 2015-05-28
[Plots.Bar] - removing barAlignment

bluong:
Preface - This change will be likely met with high friction but we should highly think about doing this / not doing this.

The idea of `barAlignment()` on `Plots.Bar` is to essentially account for creating Histograms.  However, with the introduction of `Plots.Rectangle` and knowledge that simulating Histograms using `Plots.Bar` is just... not really workable, `barAlignment()` never really worked correctly.

We should actually build Histogram as one of the more first things after v1.0 and delete `barAlignment()` as this is the best way for the user and us to have access to that feature.


crmorford:
Agreed

PR #2175, opened 2015-05-28
Goodbye dataset keys

bluong:
Internally within `Plot`s, `Dataset`s are referenced using an arbitrary key.  This is now no more!

QE: Regression pass.  Focus on any API regarding `Dataset`s.


src/components/plots/plot.ts
```diff
@@ -66,12 +55,10 @@ module Plottable {
       super();
       this._clipPathEnabled = true;
       this.classed("plot", true);
-      this._key2PlotDatasetKey = d3.map();
+      this._datasetToDrawer = new Utils.Map<Dataset, Drawers.AbstractDrawer>();
```
jtlan: `AbstractDrawer` was renamed to `Drawer` right?

src/components/plots/plot.ts
```diff
@@ -124,7 +108,7 @@ module Plottable {
     }
 
     protected _setupDatasetNodes(dataset: Dataset) {
-      var drawer = this._key2PlotDatasetKey.get(this._keyForDataset(dataset)).drawer;
+      var drawer = this._datasetToDrawer.get(dataset);
```
jtlan: Would it make sense to move the `Drawer` creation logic to this method?

src/components/plots/stackedAreaPlot.ts
```diff
@@ -18,7 +18,7 @@ export module Plots {
      */
     constructor(xScale: QuantitativeScale<X>, yScale: QuantitativeScale<number>) {
       super(xScale, yScale);
-      this.classed("area-plot", true);
+      this.classed("stacked-area-plot", true);
```
jtlan: ???

src/components/plots/plot.ts
bluong: The intention I had for this function was essentially
"Create the DOM nodes in accordance with a drawer" not really
"Create the drawer and the DOM nodes in accordance with it"

I would personally like to keep things simple and go wiht the former.  What do you think?

jtlan: Let's leave it how it currently is, since the separation of functionality is good.

src/components/plots/barPlot.ts
```diff
@@ -560,17 +558,16 @@ export module Plots {
     }
 
     protected _getDataToDraw() {
-      var datasets: D3.Map<any[]> = d3.map();
+      var dataToDraw = new Utils.Map<Dataset, any[]>();
```
jtlan: Looking at this, it seems like a better implementation might be to return a filter method that's applied to the `Dataset.data()` wherever that winds up being used.

src/components/plots/linePlot.ts
```diff
@@ -107,11 +107,9 @@ export module Plots {
...
 
     protected _getDataToDraw() {
-      var datasets: D3.Map<any[]> = d3.map();
-      this._datasetKeysInOrder.forEach((key: string) => {
-        datasets.set(key, this._key2PlotDatasetKey.get(key).dataset.data());
-      });
-      return datasets;
+      var dataToDraw = new Utils.Map<Dataset, any[]> ();
+      this.datasets().forEach((dataset) => dataToDraw.set(dataset, dataset.data()));
+      return dataToDraw;
```
jtlan: Isn't this the same implementation in `Plot`?

src/components/plots/plot.ts
```diff
@@ -434,10 +398,8 @@ module Plottable {
     }
 
     protected _getDataToDraw() {
-      var datasets: D3.Map<any[]> = d3.map();
-      this._datasetKeysInOrder.forEach((key: string) => {
-        datasets.set(key, this._key2PlotDatasetKey.get(key).dataset.data());
-      });
+      var datasets: Utils.Map<Dataset, any[]> = new Utils.Map<Dataset, any[]>();
```
jtlan: This variable should be named `dataToDraw`.

src/components/plots/linePlot.ts
bluong: Yes, but `Plots.Line` works really differently from `XYPlot`

jtlan:
@bluong, I'm going to merge this to avoid it getting stale.

PR #2179, opened 2015-05-28
ES6 Map and Set 

### LINKED ISSUE #2169: ES6 Map refactor
According to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), our `.keys()` and `.values()` methods return an array, instead of an iterator. Adapt this such that we are up to the standards.

aicioara:
Closes #2169 

Adapted the `Plottable.Utils.Map` and `Plottable.Utils.Set` to the ES6. [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)

In particular: 
### `Plottable.Utils.Map`
* `.values()` no longer existent
* `.keys()` no longer existent
* `.forEach()` added as per spec
* `.set()` returns `this`

### `Plottable.Utils.Set`
* `.values()` no longer existent
* `.forEach()` added as per spec
* `size` added as per spec

**QE notes:**
Regression test only. I checked myself with quicktests and nothing seems to be affected.
Basically what happened is that we changed the way we iterate over data structures which are literally used everywhere, especially on the interaction callbacks


src/scales/quantitativeScale.ts
```diff
@@ -123,7 +123,7 @@ module Plottable {
     }
 
     private _includeValues(domain: D[]) {
-      this._includedValues.values().forEach((value) => {
+      this._includedValues.forEach((value: D) => {
```
bluong: Is the D here necessary?  I think it should be implicitly defined?

src/scales/scale.ts
```diff
@@ -37,7 +37,11 @@ module Plottable {
     }
 
     protected _getAllExtents(): D[][] {
-      return d3.merge(this._extentsProviders.values().map((provider) => provider(this)));
+      var providerArray: D[][][] = [];
+      this._extentsProviders.forEach((provider: Scales.ExtentsProvider<D>) => {
+        providerArray.push(provider(this));
+      });
+      return d3.merge(providerArray);
```
bluong: I think the previous implementation was fine.

src/scales/quantitativeScale.ts
aicioara: Can't be implicitly defined because the function parameters are not defined. Let me check if I can change that

src/utils/callbackSet.ts
```diff
@@ -9,9 +9,9 @@ module Plottable {
      */
     export class CallbackSet<CB extends Function> extends Set<CB> {
       public callCallbacks(...args: any[]) {
-        this.values().forEach((callback) => {
+        this.forEach((callback: CB) => {
```
bluong: Can the type be implicit?

src/scales/quantitativeScale.ts
```diff
@@ -144,9 +144,18 @@ module Plottable {
...
+      this._paddingExceptions.forEach((value: D) => {
+        if (value === min) {
+          minExistsInExceptions = true;
+        }
+        if (value === max) {
+          maxExistsInExceptions = true;
+        }
+      });
+      var newMin = minExistsInExceptions ? min : this.invert(this.scale(min) - (this.scale(max) - this.scale(min)) * p);
+      var newMax = maxExistsInExceptions ? max : this.invert(this.scale(max) + (this.scale(max) - this.scale(min)) * p);
```
bluong: I think the pervious implementation is better?  Note that it is significantly shorter

src/utils/map.ts
```diff
@@ -60,21 +60,25 @@ export module Utils {
...
-    public values() {
-      return this._keyValuePairs.map((keyValuePair) => keyValuePair.value);
-    }
-
-    /**
-     * Return an array of keys in the Map.
+     * Callback is invoked with three arguments:
+     *   - the element value
+     *   - the element key
+     *   - the Map object being traversed
```
bluong: Can we find a way to put a small description and then reference MDN?

src/scales/quantitativeScale.ts
aicioara: Yup, fixed that

src/utils/map.ts
```diff
@@ -60,21 +60,25 @@ export module Utils {
...
-     * @return {K[]} The keys in the store
+     * @param {Function} callbackFn The callback to be invoked
+     * @param {any} thisArg The `this` context
      */
-    public keys() {
-      return this._keyValuePairs.map((keyValuePair) => keyValuePair.key);
+    public forEach(callbackFn: Function, thisArg?: any) {
+      if (thisArg == null) {
+        thisArg = this;
+      }
```
bluong: Let's use a default var?
```typescript
public forEach(callbackFn: Function, thisArg = this) {
```

src/utils/set.ts
```diff
@@ -7,6 +7,7 @@ module Plottable {
      * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
      */
     export class Set<T> {
+      public size = 0;
```
bluong: How do we have an instance variable be read only?

src/utils/map.ts
```diff
@@ -60,21 +60,25 @@ export module Utils {
...
+     *   - the element value
+     *   - the element key
+     *   - the Map object being traversed
      *
-     * @return {K[]} The keys in the store
+     * @param {Function} callbackFn The callback to be invoked
+     * @param {any} thisArg The `this` context
      */
-    public keys() {
-      return this._keyValuePairs.map((keyValuePair) => keyValuePair.key);
```
bluong: I personally like keys/values / I find them to be useful.  Can we keep this?

src/utils/set.ts
```diff
@@ -33,8 +36,26 @@ module Plottable {
...
+       *   - the element value
+       *   - the Set object being traversed
+       *
+       * @param {Function} callback The callback to be invoked
+       * @param {any} thisArg The `this` context
+       */
+      public forEach(callback: Function, thisArg?: any) {
+        if (thisArg == null) {
+          thisArg = this;
+        }
```
bluong: Same as above for default variables

src/scales/scale.ts
aicioara: `.values()` no longer exists, hence can't `.map()` it

src/utils/set.ts
bluong: Also it seems undefined is passed in according to MDN

src/utils/map.ts
bluong: MDN actually says that this should be `undefined`?

src/scales/quantitativeScale.ts
aicioara: `.values()` no longer exists, hence can't `indexOf` it. Notice that we are just using built in JS methods to do the same thing

src/utils/map.ts
aicioara: This is the MDN description. Reference will be added at the top

bluong: Oh I mean, let's add a small snippet and then just link to MDN for the full description

src/scales/scale.ts
bluong: I looked over the MDN idea for `.values()` and it seems that usage of `.values()` is not too clean?  See if the idea of `values()` / `keys()` is too annoying and if it's too annoying keep as is.

src/utils/map.ts
aicioara: That is because `foo.call(undefined)` is the same as `foo.call(this)` apparently. Simplified now.

src/utils/map.ts
aicioara: No, it is not part of the ES6 spec. The implementation of `.keys()` and `.values()` is different.

src/utils/set.ts
aicioara: https://github.com/Microsoft/TypeScript/issues/12

plottable.d.ts
```diff
@@ -204,17 +208,14 @@ declare module Plottable {
...
-             * Return an array of the values in the Map
-             *
-             * @return {V[]} The values in the store
-             */
-            values(): V[];
-            /**
-             * Return an array of keys in the Map.
+             * The forEach method executes the provided callback once for each key of the map which
+             * actually exist. It is not invoked for keys which have been deleted.
+             * However, it is executed for values which are present but have the value undefined.
```
bluong: Can we simplify the doc here to be
"Shim for ES6 Map.forEach" and then put the line to `forEach` on a new line

aicioara: But it is not consistent with the rest of the class. All methods in this class have a description of what the method actually does. The entire class is a `shim`. Also when reading the documentation, I believe it makes more sense to have all information required there and not a redirect. Redirects are annoying and should only be used when the reader voluntarily needs more information.

bluong: I guess my concern is mainly synchronization issues.  If MDN decides to change their documentation, we should change our documentation accordingly which... means we have to update consistently.  I think the better solution is say that this is a shim, provide our own description and then link.

src/utils/set.ts
jtlan: We could try to add some safety by doing this every time we want to set `size`:
```Typescript
Object.defineProperty(this, "size", { configurable: true, writable: false, value: newSize });
```
... although people could still mutate `size` by doing the same thing. The above code keeps people from doing something like `set.size++` or `set.size = 9001`, though.

test/tests.js
```diff
@@ -8012,6 +8012,35 @@ describe("Utils", function () {
...
+                assert.strictEqual(value1, values[index], "Value " + index + " is the expected one");
+                assert.strictEqual(passedSet, set, "The correct Set is passed as the third argument");
+                index++;
+            });
+            assert.strictEqual(index, values.length, "The expected number of iterations executed in the forEach");
+        });
+        it("forEach() not called on empty set", function () {
+            var set = new Plottable.Utils.Set();
+            set.forEach(function (value, value2, mp) {
+                assert.notOk(true, "forEach should not be called because the set is empty");
```
jtlan: Is this just an alias for `assert.isFalse()`?

src/utils/set.ts
aicioara: why configurable?

test/tests.js
aicioara: Yes, but it makes better semantics.

aicioara:
PhantomJS does not like it...

aicioara:
Opened this https://github.com/ariya/phantomjs/issues/13280 and will add a hackhack for testing

src/utils/set.ts
```diff
@@ -24,17 +28,38 @@ module Plottable {
...
           return true;
         }
         return false;
       }
 
+      private _setSize(size: number) {
+        Object.defineProperty(this, "size", {
+          value: this._values.length,
+          configurable: true,
+          writable: false
```
bluong: According to this
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

`Object.defineProperty` defaults to set immutable values and thus the `writable` parameter is default to false and this explicit setting can be removed

src/utils/set.ts
```diff
@@ -24,17 +28,38 @@ module Plottable {
...
+        });
+        if (this.size !== size) {
```
bluong: Can we more direct and put a line like
`if (isPhantomJS()) { this.size = size }`

We can even use can if-else to make it more clear that we only use this idea if we are in phantomJS

src/utils/set.ts
```diff
@@ -24,17 +28,38 @@ module Plottable {
         var index = this._values.indexOf(value);
         if (index !== -1) {
           this._values.splice(index, 1);
+          this._setSize(this._values.length);
           return true;
         }
         return false;
       }
 
+      private _setSize(size: number) {
```
bluong: Will `_setSize` ever be used for anything other than setting to `this._values.length`?  If not, we can remove the parameter here and just use `this._values.length` inside the code

aicioara: yes, look in the hackhack

src/utils/set.ts
aicioara: This is outdated, sorry

src/utils/set.ts
bluong: Even in the hackhack, it looks like this._values.length is only used there

bluong:
I'm unsure if this is NOQE.  Verify and assign accordingly?

aicioara:
QE Notes on top

PR #2186, opened 2015-05-29
Auto domain revamp

### LINKED ISSUE #2176: XYPlot.autorange() should function as a getter

bluong:
`xyPlot.autoRange()` has been renamed to be `xyPlot.adjustScaleType()` so that it can be turned into a getter/setter combo

Closes #2176


jtlan:
I think `autoranging()` would be a better name for this method. Originally we chose "`autorange()`" because we though that was the most likely name that users would look for.

bluong:
I'm thinking about the name of the function and I'm wondering if `autoranging` makes sense.  We're not automatically adjusting the range of the scales, we're adjusting the domain of one of the scales in accordance with the user input

jtlan:
The functionality adjusts the range of the *`Plot`* though. I'd like to get some user feedback before we proceed with this.

bluong:
Completed by #2201

PR #2190, opened 2015-05-29
[Drawer] - Remove setClass API

bluong:


src/drawers/arcDrawer.ts
```diff
@@ -6,6 +6,7 @@ export module Drawers {
 
     constructor(dataset: Dataset) {
       super(dataset);
+      this._className = "arc";
```
aicioara: Is this ever needed?

aicioara:
Rest looks fine

jtlan:
Hold on, does this remove the "arc" class?

src/drawers/arcDrawer.ts
bluong: The class name is added here to specify what types of paths an Arc Drawer will draw.  Note that Symbol Drawer labels with a "symbol" class

aicioara: Is there a reason why `Drawers.Rect` does not have a class associated with it?

bluong: The types of "rect"s that are drawn in a `Drawers.Rect` are... "rect"s?  It is really redundant I think in this case, whereas we have many different types of "path"s with no good way of differentiation

PR #2191, opened 2015-05-29
[Drawer] - AbstractDrawer -> Drawer / moved Drawer to Plottable module

bluong:


bluong:
This is mainly code movement more than anything else.  I think NOQE

PR #2194, opened 2015-05-29
[Drawers.Rect] - Drawers.Rect -> Drawers.Rectangle

bluong:
Abbreviation was bad... Let's stop abbreviating


aicioara:
Change **hugely** appreciated.

jtlan:
However, it does draw `<rect>` nodes, so there'd be some consistency in leaving it as "rect"...

PR #2202, opened 2015-06-02
Upgrade to Typescript 1.5 beta.

jtlan:
http://blogs.msdn.com/b/typescript/archive/2015/04/30/announcing-typescript-1-5-beta.aspx

Mostly, adds touch event typings so we don't have to force people to pull an extra set of typings.

You'll need to run `npm install` after this.


aicioara:
I restarted the failing build. Why are the typings not needed anymore?

PR #2203, opened 2015-06-02
[QuantitativeScale] padding exceptions and included values no longer take keys.

### LINKED ISSUE #2182: [QuantitativeScale] padding exceptions and included values should not take keys.
```Typescript
public addPaddingException(key: any, exception: D);
public removePaddingException(key: any);
public addIncludedValue(key: any, value: D);
public removeIncludedValue(key: any);
```

These methods should not take keys.

aicioara:
Closes #2182
Instead they now use providers.
> From an offline discussion, @aicioara and I realized that an "extent" (from ExtentsProvider) is just a pair of included values, because the order of values in the "extent" shouldn't matter.

As part of this change 
- refactored `addPaddingException(key: any, exception: D)` -> `addPaddingExceptionsProvider(provider: Scales.PaddingExceptionsProvider<D>)`
- refactored `removePaddingException(key: any)` -> `removePaddingExceptionsProvider(provider: Scales.PaddingExceptionsProvider<D>)`
- removed `addIncludedValue(key: any, value: D) `
- removed `removeIncludedValue(key: any, value: D) `
- renamed `addExtentsProvider(provider: Scales.ExtentsProvider<D>)` -> `addIncludedValuesProvider(provider: Scales.IncludedValuesProvider<D>)`
- renamed `removeExtentsProvider(provider: Scales.ExtentsProvider<D>)` -> `removeIncludedValuesProvider(provider: Scales.IncludedValuesProvider<D>)`
- Documentation for methods and interfaces
- Unit tests for `addPaddingExceptionsProvider`

One thing I still don't like is the mixture of extents and included values observed in `plot.ts`

**QE Notes**
Api break points:

Before:
```typescript
var scale = new Plottable.Scales.Linear();
var referenceKey = "referenceKey"
scale.addIncludedValues(referenceKey, 10);
scale.addIncludedValues(referenceKey, 13);
// assert.deepEqual(scale.domain(), [9.5, 13.5], "The domain is padded");

scale.addPaddingException(referenceKey, 13);
// assert.deepEqual(scale.domain(), [9.5, 13], "The right side of the domain is no longer padded");

scale.addPaddingException(referenceKey, 10);
// assert.deepEqual(scale.domain(), [10, 13], "The domain is no longer padded");
```
After:
```typescript
var scale = new Plottable.Scales.Linear();
scale.addIncludedValuesProvider(function() { return [10, 13]; });
// assert.deepEqual(scale.domain(), [9.5, 13.5], "The domain is padded");

scale.addPaddingExceptionsProvider(function() { return [13]; });
// assert.deepEqual(scale.domain(), [9.5, 13], "The right side of the domain is no longer padded");

scale.addPaddingExceptionsProvider(function() { return [10]; });
// assert.deepEqual(scale.domain(), [10, 13], "The domain is no longer padded");
```

Another change is that this has been renamed
Before:
```typescript  
var scale = new Plottable.Scales.Linear();
scale.addExtentsProvider(function() { return [10, 13]; });
```
After
```typescript
var scale = new Plottable.Scales.Linear();
scale.addIncludedValuesProvider(function() { return [10, 13]; });
```


src/components/plots/areaPlot.ts
```diff
@@ -146,10 +147,14 @@ export module Plots {
         return;
       }
 
+      if (this._constantBaselineValueProvider != null) {
+        yScale.removePaddingExceptionsProvider(this._constantBaselineValueProvider);
+        this._constantBaselineValueProvider = null;
```
jtlan: This logic doesn't seem correct. The previous logic removed the provider when `constantBaseline` was null, rather than when the provider was null.

With the new logic, if I set a constant baseline, then change `y0()` so the baseline isn't constant, the padding exception isn't removed.

src/components/plots/barPlot.ts
```diff
@@ -44,6 +45,7 @@ export module Plots {
       this.attr("fill", new Scales.Color().range()[0]);
       this.attr("width", () => this._getBarPixelWidth());
       this._labelConfig = new Utils.Map<Dataset, LabelConfig>();
+      this._baselineValueProvider = () => [this._baselineValue];
```
jtlan: Probably should write it as
```
() => [this.baseline()]
```

src/components/plots/barPlot.ts
```diff
@@ -301,13 +303,8 @@ export module Plots {
...
         var qscale = <QuantitativeScale<any>> valueScale;
-        if (this._baselineValue != null) {
-          qscale.addPaddingException(this, this._baselineValue);
-          qscale.addIncludedValue(this, this._baselineValue);
-        } else {
-          qscale.removePaddingException(this);
-          qscale.removeIncludedValue(this);
-        }
+        qscale.addPaddingExceptionsProvider(this._baselineValueProvider);
+        qscale.addIncludedValuesProvider(this._baselineValueProvider);
```
jtlan: Same complaint here: The logic has changed, because the Providers are never removed.

src/components/plots/stackedAreaPlot.ts
```diff
@@ -85,8 +87,8 @@ export module Plots {
       if (scale == null) {
         return;
       }
-      scale.addPaddingException(this, 0);
-      scale.addIncludedValue(this, 0);
+      scale.addPaddingExceptionsProvider(this._baselineValueProvider);
+      scale.addIncludedValuesProvider(this._baselineValueProvider);
```
jtlan: Looks like this was a bug, although I don't think the baseline can be set on `StackedArea`? At least it doesn't seem like it would work correctly.

src/components/plots/areaPlot.ts
aicioara: Can you give more details of the scenario?

jtlan: Never mind, I misread the logic. Looks good.

src/components/plots/barPlot.ts
jtlan: Let's fix this later, when we rename `baseline()` to `baselineValue()` (#2206).

aicioara:
There are failing overlaying tests, sorry... Fixing...

quicktests/overlaying/tests/functional/titleLegend_change.js
```diff
@@ -27,13 +27,26 @@ function run(svg, data, Plottable) {
   var xScale = new Plottable.Scales.Linear();
   var yScale = new Plottable.Scales.Linear();
   if (Plottable.Domainer != null) {
+try {
+    var domainer_X = new Plottable.Domainer().addPaddingExceptionsProvider(function() { return [0]; });
+    var domainer_Y = new Plottable.Domainer().addPaddingExceptionsProvider(function() { return [0]; });
```
jtlan: We can actually get rid of the `Domainer`s block entirely.

quicktests/overlaying/tests/functional/titleLegend_change.js
```diff
@@ -27,13 +27,26 @@ function run(svg, data, Plottable) {
...
     var domainer_X = new Plottable.Domainer().addPaddingException("overlaying", 0);
     var domainer_Y = new Plottable.Domainer().addPaddingException("overlaying", 0);
     xScale.domainer(domainer_X);
     yScale.domainer(domainer_Y);
+}
   } else {
+try {
+    xScale.padProportion(0).addPaddingExceptionsProvider(function() { return [0]; });
+    yScale.padProportion(0).addPaddingExceptionsProvider(function() { return [0]; });
+} catch (error) {
```
jtlan: This should be done with an `if`-block to avoid catching legitimate errors:
```Javascript
xScale.padProportion(0);
yScale.padProportion(0);
if (typeof xScale.addPaddingExceptionsProvider === "function") {
  xScale.addPaddingExceptionsProvider(function() { return [0]; });
  yScale.addPaddingExceptionsProvider(function() { return [0]; });
} else {
  xScale.addPaddingException("overlaying", 0);
  yScale.addPaddingException("overlaying", 0);
}
```

crmorford:
Hey @aicioara. I just pushed a test, padding.js.
The scale will redraw when you add a value, but not when you remove one, and that feels inconsistent. Let me know if I'm doing something wrong?

crmorford:
On that same test, with the addPaddingExceptionProviders uncommented, I get the error "component.detach is not a function"

crmorford:
on the last push, the extent function seems to work to expand the domain, but not to shrink it.
look at lines 47-51. extentFcnA vs extentFcnB

crmorford:
Looks like this is in an OK place now, I'm comfortable merging whenever

PR #2209, opened 2015-06-02
[2206rf] renamed baseline() to baselineValue()

### LINKED ISSUE #2206: [Plots.Bar] Rename baseline() to baselineValue()
`baseline()` actually gets/sets the *value* of the baseline, as opposed to the baseline selection itself, so it should be renamed to reflect that.

aicioara:
Closes #2206


jtlan:
[+1]

[NOQE] because straight-up rename.

PR #2212, opened 2015-06-03
Update D3 typings

jtlan:
From 837a1a0:

>Caught some typing errors in `Animators`.
No more `D3.QuantitativeScale`, so had to type some things more explicitly.
* `Scales.InterpolatedColor` was the most affected in that regard.

>Most other changes are to parameterize `Selections` -- Basically always `<void>`, or `<any>` if a datum is bound to it.

Also updated SVG-Typewriter, which had to be pulled along to the new version too.


src/animators/baseAnimator.ts
```diff
@@ -63,7 +63,7 @@ export module Animators {
       return time;
     }
 
-    public animate(selection: any, attrToProjector: AttributeToProjector): D3.Transition.Transition {
+    public animate(selection: d3.Selection<any>, attrToAppliedProjector: AttributeToAppliedProjector) {
```
aicioara: Can you tell me a bit about the difference between `AttributeToProjector` and `AttributeToAppliedProjector`

src/animators/rectAnimator.ts
```diff
@@ -19,25 +19,25 @@ export module Animators {
...
+    protected _startMovingProjector(attrToAppliedProjector: AttributeToAppliedProjector) {
       if (this.isVertical === this.isReverse) {
-        return attrToProjector[this._getMovingAttr()];
+        return attrToAppliedProjector[this._getMovingAttr()];
       }
-      var movingAttrProjector = attrToProjector[this._getMovingAttr()];
-      var growingAttrProjector = attrToProjector[this._getGrowingAttr()];
-      return (d: any, i: number, dataset: Dataset) => {
-        return movingAttrProjector(d, i, dataset) + growingAttrProjector(d, i, dataset);
+      var movingAttrAppliedProjector = attrToAppliedProjector[this._getMovingAttr()];
```
aicioara: `movingAttrToAppliedProjector`? Missing the `to`

src/components/axes/axis.ts
```diff
@@ -188,7 +188,7 @@ module Plottable {
     }
 
     protected _generateTickMarkAttrHash(isEndTickMark = false) {
-      var tickMarkAttrHash = {
+      var tickMarkAttrHash: { [key: string]: number | ((d: any) => number) } = {
```
aicioara: Is there a way to avoid this union? Shouldn't the second argument be `[key: string]: (d:any) => number`?

src/components/axes/numericAxis.ts
```diff
@@ -89,7 +89,7 @@ export module Axes {
     public renderImmediately() {
       super.renderImmediately();
 
-      var tickLabelAttrHash = {
+      var tickLabelAttrHash: { [key: string]: number | string | ((d: any) => number) } = {
```
aicioara: Same as above

src/components/axes/timeAxis.ts
```diff
@@ -281,7 +281,7 @@ export module Axes {
 
     private _getIntervalLength(config: TimeAxisTierConfiguration) {
       var startDate = this._scale.domain()[0];
-      var d3Interval = Formatters.timeIntervalToD3Time(config.interval);
+      var d3Interval = Formatters.timeIntervalTod3Time(config.interval);
```
aicioara: Capitalize D here for CamelCase

src/components/axes/timeAxis.ts
```diff
@@ -385,7 +385,7 @@ export module Axes {
         labelPos = tickPos;
       }
 
-      var tickLabels = container.selectAll("." + Axis.TICK_LABEL_CLASS).data(labelPos, (d) => d.valueOf());
+      var tickLabels = container.selectAll("." + Axis.TICK_LABEL_CLASS).data(labelPos, (d) => String(d.valueOf()));
```
aicioara: Why was this needed?

src/components/axes/timeAxis.ts
```diff
@@ -503,7 +503,7 @@ export module Axes {
     }
 
     private _hideOverlappingAndCutOffLabels(index: number) {
-      var boundingBox = this._element.select(".bounding-box")[0][0].getBoundingClientRect();
+      var boundingBox = (<Element> this._element.select(".bounding-box")[0][0]).getBoundingClientRect();
```
aicioara: Just a thought: we should consider if we want to enforce a whitespace between casting and variable or not. We have a mixture in the codebase

src/components/interpolatedColorLegend.ts
```diff
@@ -202,11 +202,11 @@ export module Components {
...
-      var boundingBoxAttr = {
-        x: 0,
-        y: padding,
-        width: 0,
-        height: 0
+      var boundingBoxAttr: { [key: string]: number} = {
+        "x": 0,
+        "y": padding,
+        "width": 0,
+        "height": 0
```
aicioara: We should be consistent, in general we don't stringify them (at least it did not happen above).

src/components/interpolatedColorLegend.ts
```diff
@@ -202,11 +202,11 @@ export module Components {
       var swatchX: (d: any, i: number) => number;
       var swatchY: (d: any, i: number) => number;
 
-      var boundingBoxAttr = {
-        x: 0,
-        y: padding,
-        width: 0,
-        height: 0
+      var boundingBoxAttr: { [key: string]: number} = {
```
aicioara: Space between `number` and `}`

src/components/plots/plot.ts
```diff
@@ -422,15 +422,16 @@ module Plottable {
      *
      * @param {Dataset[]} [datasets] The Datasets to retrieve the Selections for.
      *   If not provided, Selections will be retrieved for all Datasets on the Plot.
-     * @returns {D3.Selection}
+     * @returns {d3.Selection}
      */
-    public getAllSelections(datasets = this.datasets()): D3.Selection {
-      var allSelections: EventTarget[] = [];
+
```
aicioara: Extra whitespace

src/scales/linearScale.ts
```diff
@@ -53,7 +53,7 @@ export module Scales {
     }
 
     protected _niceDomain(domain: number[], count?: number): number[] {
-      return Utils.D3Scale.niceDomain(this._d3Scale, domain, count);
+      return this._d3Scale.copy().domain(domain).nice(count).domain();
```
aicioara: Why remove the D3Scale method?

src/scales/timeScale.ts
```diff
@@ -24,10 +24,10 @@ export module Scales {
     public tickInterval(interval: string, step?: number): Date[] {
       // temporarily creats a time scale from our linear scale into a time scale so we can get access to its api
       var tempScale = d3.time.scale();
-      var d3Interval = Formatters.timeIntervalToD3Time(interval);
+      var d3Interval: d3.time.Interval = Formatters.timeIntervalTod3Time(interval);
```
aicioara: Capitalize D for CamelCase. Is the explicit type d3.time.Interval needed?

src/scales/timeScale.ts
```diff
@@ -80,8 +80,8 @@ export module Scales {
       return this._d3Scale.ticks(QuantitativeScale._DEFAULT_NUM_TICKS);
     }
 
-    protected _niceDomain(domain: Date[], count?: number) {
-      return Utils.D3Scale.niceDomain(this._d3Scale, domain, count);
+    protected _niceDomain(domain: Date[]) {
+      return this._d3Scale.copy().domain(domain).nice().domain();
```
aicioara: Same as above

src/utils/formatters.ts
```diff
@@ -193,7 +193,7 @@ module Plottable {
      * Transforms the Plottable TimeInterval string into a d3 time interval equivalent.
      * If the provided TimeInterval is incorrect, the default is d3.time.year
      */
-    export function timeIntervalToD3Time(timeInterval: string) {
+    export function timeIntervalTod3Time(timeInterval: string) {
```
aicioara: Capitalize D for CamelCase

test/components/plots/rectanglePlotTests.ts
```diff
@@ -14,9 +14,9 @@ describe("Plots", () => {
       { x: 4, y: 4, x2: 5, y2: 5 }
     ];
 
-    var VERIFY_CELLS = (cells: D3.Selection) => {
+    var VERIFY_CELLS = (cells: d3.Selection<any>) => {
       assert.strictEqual(cells[0].length, 5);
-      cells.each(function(d: D3.Selection, i: number) {
+      cells.each(function(d: any, i: number) {
```
aicioara: Is this an `any`?

test/core/componentGroupTests.ts
```diff
@@ -65,7 +65,8 @@ describe("ComponentGroups", () => {
     componentGroup.remove(c1);
     assert.deepEqual(componentGroup.components(), [c0, c2],
     "removing a Component not in the Group does not remove Components from the Group");
-    assert.strictEqual(svg.node().childNodes[0], (<D3.Selection> (<any> c1)._element).node(), "The Component not in the Group stayed put");
+    assert.strictEqual((<Node> svg.node()).childNodes[0], (<d3.Selection<void>> (<any> c1)._element).node(),
```
aicioara: is the cast to `d3.Selection<void>` necessary?

src/animators/baseAnimator.ts
jtlan: an `AppliedProjector` has the `Dataset` rolled into it, so the signature is different:
```Typescript
/**
 */
export type _Projector = (datum: any, index: number, dataset: Dataset) => any;

export type AppliedProjector = (datum: any, index: number) => any;
```

src/animators/rectAnimator.ts
jtlan: It's not a dictionary, the type of that object is `AppliedProjector`.

src/components/axes/axis.ts
jtlan: I could convert the hash values to a constant
```Typescript
() => 0
```
instead of just using `number`, but that complicates the later code, and d3 does accept the raw `number`s. I wanted to modify as little of the actual code logic as possible in this PR.

I should remove the `<any>`s on the following 4 lines though.

src/components/axes/timeAxis.ts
jtlan: Good catch, bad find-replace on my part.

src/components/axes/timeAxis.ts
jtlan: The keying function (the second argument to `.data()`) [has to return `string`s](https://github.com/mbostock/d3/wiki/Selections#data); previously the return value was implicitly being cast by Javascript.

src/components/axes/timeAxis.ts
jtlan: I prefer the whitespace, personally.

src/scales/linearScale.ts
jtlan: It depended on the D3 type `QuantitativeScale` existing, but that type no longer exists. Rather than have it take some sort of union type, I thought it made more sense just to clone the implementation, in particular because the signature of the `nice()` method is different on different scales.

test/components/plots/rectanglePlotTests.ts
jtlan: Yes, because the `d` there is the data value, not the `Selection`. The previous author made a mistake.

test/core/componentGroupTests.ts
jtlan: No, but it provides better typing safety than just using `<any>`.

src/components/interpolatedColorLegend.ts
jtlan: Good catch, thanks.

src/scales/timeScale.ts
jtlan: Nope, not needed.

src/animators/rectAnimator.ts
aicioara: aw, my bad. However, in that case the `Attr` bit is extra

jtlan: I'll change it so it's clearer.

src/components/axes/axis.ts
aicioara: Let's keep the union, but isn't the second argument `[key: string]: (d:any) => number`? Like, it's a map from key to functions? I'm missing something here. How does this even compile?

test/core/componentGroupTests.ts
aicioara: But it's not an `<any>`. You cast the `_element`, which is of type `d3.Selection<void>` anyway

src/components/axes/axis.ts
jtlan: It's a map from strings to `number | (d: any) => number` (number or function).

test/core/componentGroupTests.ts
jtlan: The test code doesn't know that though, because `_element` isn't public. After we `<any>`-cast c1, any fields on it are also `<any>`.

That said, it's probably better just to rewrite this `assert`.

src/components/axes/axis.ts
aicioara: my bad, ty

PR #2215, opened 2015-06-03
[2214rf] cleaned the interfaces of public underscored methods

### LINKED ISSUE #2214: public-underscore types being exported
In **interfaces.ts**:
```Typescript
  /**
   */
  export type _Projector = (datum: any, index: number, dataset: Dataset) => any;

  /**
   * Defines a way how specific attribute needs be retrieved before rendering.
   */
  export type _Projection = {
    accessor: Accessor<any>;
    scale?: Scale<any, any>;
    attribute: string;
  }
```

There's a number of types here that seem extraneous and should probably be cleaned up anyway.

aicioara:
Closes #2214


src/core/interfaces.ts
```diff
@@ -9,36 +9,27 @@ module Plottable {
...
-  export type AttributeToProjector = { [attrToSet: string]: _Projector; };
+  export type AttributeToProjector = { [attrToSet: string]: Projector; };
 
   export type AttributeToAppliedProjector = { [attrToSet: string]: AppliedProjector; };
 
   export type SpaceRequest = {
     minWidth: number;
     minHeight: number;
-  }
+  };
```
jtlan: ??? How did this compile before?

jtlan:
[NOQE] because no JS changes.

src/core/interfaces.ts
aicioara: There is no TSLINT rule for semicolons after `export`. My `Sublime` complains all the time, so I just fix them as I see them

PR #2222, opened 2015-06-03
PaddingExceptionProvider can only take a quantitative scale now

aicioara:
Overlooked in #2203

### DIFF
--- a/src/scales/scale.ts
+++ b/src/scales/scale.ts
@@ -22,11 +22,11 @@ module Plottable {
      * If one end of the domain is set to an excepted value as a result of autoDomain()-ing,
      * that end of the domain will not be padded.
      *
-     * @param {Scale} scale
-     * @returns {D[]} An array of extents.
+     * @param {QuantitativeScale} scale
+     * @returns {D[]} An array of values to be added as exception.
      */
     export interface PaddingExceptionsProvider<D> {
-      (scale: Scale<D, any>): D[];
+      (scale: QuantitativeScale<D>): D[];
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/scales/scale.ts
```diff
@@ -22,11 +22,11 @@ module Plottable {
      * If one end of the domain is set to an excepted value as a result of autoDomain()-ing,
      * that end of the domain will not be padded.
      *
-     * @param {Scale} scale
-     * @returns {D[]} An array of extents.
+     * @param {QuantitativeScale} scale
+     * @returns {D[]} An array of values to be added as exception.
```
jtlan: "An array of padding exceptions"? They don't get added anywhere.

aicioara: Good point

src/scales/scale.ts
```diff
@@ -22,11 +22,11 @@ module Plottable {
      * If one end of the domain is set to an excepted value as a result of autoDomain()-ing,
      * that end of the domain will not be padded.
      *
-     * @param {Scale} scale
-     * @returns {D[]} An array of extents.
+     * @param {QuantitativeScale} scale
+     * @returns {D[]} An array of values to become padding exceptions.
```
jtlan: They don't "become" padding exceptions, they *are* the padding exceptions.

I'd actually argue that the commentary isn't necessary, since the free-form comment block at the start of the documentation already explains what they are.

aicioara: What do you mean? They *are* not exceptions, unless the providers provide them, hence they *become* exceptions

jtlan: What about "An array of values that should not be padded"?

aicioara:
[NOQE] No JS changes

PR #2223, opened 2015-06-03
[Plots.Bar] Fixed type for baselineValue()

### LINKED ISSUE #2208: [Plots.Bar] basevalue() is hardcoded to 0
Even though the scale might be non-numeric in the domain (time / category)

aicioara:
Closes #2208.

Well, this is one way of doing it. I did not really like the proper way, where 8 classes (scales) get polluted with a `getBaseValue` public method

**QE Notes**
Nothing should be affected. Tested overlaying tests and all works fine. Regression pass only. Bar Plots affected by this commit.

### DIFF
--- a/src/components/plots/barPlot.ts
+++ b/src/components/plots/barPlot.ts
@@ -123,20 +123,15 @@ export module Plots {
           return this._baselineValue;
         }
         if (!this._projectorsReady()) {
-          Utils.Methods.warn("Asking for the baseline value when no scale is attached is not safe");
           return 0;
         }
         var valueScale = this._isVertical ? this.y().scale : this.x().scale;
         if (!valueScale) {
-          Utils.Methods.warn("Asking for the baseline value when no value scale is attached is not safe");
           return 0;
         }
 
         if (valueScale instanceof Scales.Time) {
           return new Date(0);
-        }
-        if (valueScale instanceof Scales.Category) {
-          throw new Error("The value scale cannot be a Category Scale");
         }
 
         return 0;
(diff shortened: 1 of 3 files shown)


src/components/plots/barPlot.ts
```diff
@@ -107,20 +106,40 @@ export module Plots {
...
+        if (this._baselineValue != null) {
+          return this._baselineValue;
+        }
+        if (!this._projectorsReady()) {
+          Utils.Methods.warn("Asking for the baseline value when no scale is attached is not safe");
+          return 0;
+        }
+        var valueScale = this._isVertical ? this.y().scale : this.x().scale;
+        if (!valueScale) {
+          Utils.Methods.warn("Asking for the baseline value when no value scale is attached is not safe");
```
jtlan: Not strictly true, since it might just be a pixel value.

src/components/plots/barPlot.ts
```diff
@@ -107,20 +106,40 @@ export module Plots {
...
+          return 0;
+        }
+
+        if (valueScale instanceof Scales.Time) {
+          return new Date(0);
+        }
+        if (valueScale instanceof Scales.Category) {
+          throw new Error("The value scale cannot be a Category Scale");
```
jtlan: A little weird that the error message would occur when calling `baseline()`, right?

src/components/plots/barPlot.ts
```diff
@@ -107,20 +106,40 @@ export module Plots {
...
-    public baselineValue(value?: number): any {
+    public baselineValue(value: X|Y): Bar<X, Y>;
+    public baselineValue(value?: X|Y): any {
       if (value == null) {
-        return this._baselineValue;
+        if (this._baselineValue != null) {
+          return this._baselineValue;
+        }
+        if (!this._projectorsReady()) {
+          Utils.Methods.warn("Asking for the baseline value when no scale is attached is not safe");
```
jtlan: This warning seems redundant with the one below?

jtlan:
While the warnings are well-intentioned, I don't think they actually zero in on the correct problem. Let's leave them out for now, and wait until we have a better way of handling the generics here.

aicioara:
I agree with the points above. Although I believe we sacrificed safety for flexibility.

PR #2224, opened 2015-06-04
Utils Refactor

### LINKED ISSUE #2024: Utils Refactor
There is some inconsistency in the Utils package.

For Example, `Formatters` is under the `Utils` package, but it is not invoked as `Utils.Formatters`. -> put under module

Also the `utils.ts` file is actually a file of `Methods`, so all methods here are invoked as `Utils.Methods.method` -> filename change

aicioara:
Closes #2024

Splitting `utils.ts` (`Plottable.Utils.Methods`) into:

```
Utils.Methods.inRange()           -> Utils.Math.inRange();
Utils.Methods.clamp()             -> Utils.Math.clamp();
Utils.Methods.max()               -> Utils.Math.max();
Utils.Methods.min()               -> Utils.Math.min();
Utils.Methods.isNaN()             -> Utils.Math.isNaN();
Utils.Methods.isValidNumber()     -> Utils.Math.isValidNumber();
Utils.Methods.range()             -> Utils.Math.range();
Utils.Methods.distanceSquared()   -> Utils.Math.distanceSquared();

Utils.Methods.addArrays()         -> Utils.Array.add();
Utils.Methods.uniq()              -> Utils.Array.uniq();
Utils.Methods.createFilledArray() -> Utils.Array.createFilledArray();
Utils.Methods.flatten()           -> Utils.Array.flatten();
Utils.Methods.arrayEq()           -> DELETED

Utils.Methods.copyMap()           -> Utils.Window.copyObject();
Utils.Methods.populateMap()       -> DELETED

Utils.Methods.warn()              -> Utils.Window.warn();
Utils.Methods.setTimeout()        -> Utils.Window.setTimeout();
Utils.Methods.objEq()             -> DELETED
Utils.Methods.isIE()              -> DELETED
Utils.Methods.parseRange()        -> DELETED

Utils.Methods.colorTest()         -> Utils.Color.colorTest();
Utils.Methods.lightenColor()      -> Utils.Color.lightenColor();

Utils.Methods.intersectsBBox()    -> Utils.DOM.intersectsBBox();
```
For implementation details and description, read [here](https://github.com/palantir/plottable/blob/develop/src/utils/utils.ts)

---

Also:
- `formatters.ts` file moved under `core` (no JS change)
- `symbolFactories.ts` file moved under `core` (no JS change)

API Breaks:
- `Utils.Colors.contrast()` -> `Utils.Color.contrast()`


jtlan:
Maybe just name the file **methods.ts**?

aicioara:
Hmm. I somehow agree. Maybe they should not be `methods`. Since the `Utils.DOM` are also methods. Why would these be more special? Maybe we should call them **Utils.General**.

Another option is to bring them one level up. And have everything on Utils. Basically `Utils.min()` instead of `Utils.Methods.min()`

aicioara:
In other words
> I'm trying to figure out why we have a Methods module at all

aicioara:
RFC

jtlan:
> Another option is to bring them one level up. And have everything on Utils. Basically `Utils.min()` instead of `Utils.Methods.min()`

I like this. We used to have the methods directly under `Utils`, but #512 added the `Methods` module.

aicioara:
From offline conversation, just putting them in some category makes more sense. (say `Utils.Math.*`)

aicioara:
Hit a wall... We cannot have `Utils.Math.*` Because of the way JS resolves scopes, by doing this we just overridden `window.Math` for the entire project

aicioara:
Let's see if we can split them nicely into modules

```typescript export function inRange();
export function clamp();
export function warn();
export function addArrays();
export function populateMap<T>();
export function uniq<T>();
export function createFilledArray<T>();
export function flatten<T>();
export function arrayEq<T>();
export function objEq();
export function max<C>();
export function max<T, C>();
export function max();
export function min<C>();
export function min<T, C>();
export function min();
export function isNaN();
export function isValidNumber();
export function copyMap<T>();
export function range();
export function setTimeout();
export function colorTest();
export function lightenColor();
export function distanceSquared();
export function isIE();
export function intersectsBBox();
export function parseRange();
```

jtlan:
`populateMap()` and `copyMap()` don't return `Array`s, so putting them in the `Array` module seems incorrect. In fact, they don't even return the same type of "Map" ... !

What does "emm" mean?

Also, some of these methods, such as `objEq()`, aren't used. You can go ahead and delete them.

aicioara:
emm, means I have no idea

aicioara:
also, we can't call the module `Math`?

aicioara:
Will do a merge manually see what's up

aicioara:
Regression test only. Tested overlaying myself.

PR #2225, opened 2015-06-04
Missing type semicolon

aicioara:
I've just been hit by this so many times. Declaring type without semicolon ruins the syntax highlight on `Sublime` and it is not consistent.

Also, as part of this removed trailing whitespace across the entire project.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -833,7 +833,6 @@ var Plottable;
     Plottable.MILLISECONDS_IN_ONE_DAY = 24 * 60 * 60 * 1000;
     var Formatters;
     (function (Formatters) {
-        ;
         /**
          * Creates a formatter for currency values.
          *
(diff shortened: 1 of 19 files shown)


jtlan:
Can we add a tslint rule for this?

aicioara:
I believe there is no tslint rule for this. Let me double check

aicioara:
Yeah, there is no rule for it. I can either 
- Create a custom rule
- Submit a PR to tslint
- Ignore for now

jtlan:
How hard would it be to write a custom rule? I think this is something worth submitting to TSLint.

If it's difficult, we can save solving it for later. We should probably file an issue in that case.

aicioara:
I opened this https://github.com/palantir/plottable/issues/2226

jtlan:
[NOQE] because no JS changes.

Issue #2230, opened 2015-06-04
Legend.getEntry() should return Entity?


jtlan:
Currently, `Legend`'s `getEntry()` call returns a `d3.selection()`:

```Typescript
/**
 * Gets the entry under at given pixel position.
 * Returns an empty Selection if no entry exists at that pixel position.
 *
 * @param {Point} position
 * @returns {d3.Selection}
 */
getEntry(position: Point): d3.Selection<void>;
```

Should it return an `Entity`, the way `entityAt()` does on `Plot`? If so, `Entity` needs to be modified to have a `component` field instead of a `plot` field.

jtlan:
Closed in #2263.

Issue #2199, opened 2015-05-29
[Plot] - `getAllSelections()` may be redundant


bluong:
`getAllSelections()` is exactly the same as the reduction of the selection parameter on `entities()`.  We should consider if we want this convenience method and if we do, we will want to rename it to like `selections()` and also have the implementation of `getAllSelections` be just the modified output of `entities()` to keep behavior consistent

aicioara:
Need more context on this.

PR #2231, opened 2015-06-05
Remove get from public method names

### LINKED ISSUE #2198: [Drawer] - underscored endpoints should be removed or publicized
There are three public endpoints that are underscored.  They should be removed or actually publicized.

Note that their main usage is in `getAllSelections()` `entities()` and `closestEntity()`

jtlan:
Except for:
* `Map`'s `get()`,
* `getAllSelections()` (#2199)
* `getEntry()` (#2230)

Here's the renames:
#### QuantitativeScale
`getDefaultTicks()` --> `defaultTicks()`

#### Animator
`getTiming()` --> `totalTime()`

#### Dispatchers.Mouse
`getLastMousePosition()` --> `lastMousePosition()`

#### Drawers
`_getRenderArea()` --> `renderArea()`
`setup()` -> `renderArea()`
`_getSelector()` --> `selector()`
`_getSelection()` --> `selectionForIndex()`

Close #2198.


src/drawers/drawer.ts
```diff
@@ -111,16 +111,22 @@ export module Drawers {
      *
      * @returns {d3.Selection} the renderArea selection
      */
-    public _getRenderArea() {
+    public renderArea() {
```
aicioara: I can't comment on that LOC, but there is a `setup` method on the drawer, which just sets the `renderArea`. I suggest remove the setup call, and make this act as a setter as well

src/drawers/drawer.ts
```diff
@@ -111,16 +111,22 @@ export module Drawers {
...
       return this._renderArea;
     }
 
-    public _getSelector(): string {
+    /**
+     * Returns the selector for this Drawer's visual elements.
+     */
+    public selector() {
```
aicioara: Is this ever possible? If not `Utils.Methods.warn()`

src/drawers/drawer.ts
```diff
@@ -111,16 +111,22 @@ export module Drawers {
...
      * @returns {d3.Selection} the renderArea selection
      */
-    public _getRenderArea() {
+    public renderArea() {
       return this._renderArea;
     }
 
-    public _getSelector(): string {
+    /**
+     * Returns the selector for this Drawer's visual elements.
```
aicioara: What is a selector? I think the comment should be more precise, or the method should be renamed

test/components/plots/plotTests.ts
```diff
@@ -217,13 +217,13 @@ describe("Plots", () => {
       var renderArea1 = svg.append("g");
       var renderArea1Selection = renderArea1.append("circle").attr("cx", 100).attr("cy", 100).attr("r", 10);
       (<any> mockDrawer1).setup = () => (<any> mockDrawer1)._renderArea = renderArea1;
```
aicioara: You may decide to remove the `setup`, but there is no need for the `<any>` cast here. Feel free to ignore this comment as it is not part of the PR

src/drawers/drawer.ts
jtlan: ? Do you mean if "" is possible?

src/drawers/drawer.ts
jtlan: https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_started/Selectors

So maybe "CSS selector".

test/components/plots/plotTests.ts
jtlan: Probably better. I didn't realize setup() was public too.

src/drawers/drawer.ts
aicioara: I mean having a plain drawer.

jtlan: No, I don't think so.

aicioara: Then, I'd prefer to warn about the behavior

jtlan: In fact trying to use the selector `""` already throws an error, but a really vague one, so throwing our own would be better (or returning `null`). I'll do that.

src/drawers/areaDrawer.ts
```diff
@@ -11,9 +11,15 @@ export module Drawers {
...
 
-    public setup(area: d3.Selection<void>) {
-      Drawer.prototype.setup.call(this, area);
+    public renderArea(): d3.Selection<void>;
+    public renderArea(area: d3.Selection<void>): Drawer;
+    public renderArea(area?: d3.Selection<void>): any {
+      if (area == null) {
+        return super.renderArea();
+      }
+      Drawer.prototype.renderArea.call(this, area);
```
aicioara: `super(area)`

test/drawers/drawerTests.ts
```diff
@@ -89,15 +90,15 @@ describe("Drawers", () => {
       assert.deepEqual(timings, [0, 20, 30], "setTimeout called with appropriate times");
     });
 
-    it("_getSelection", () => {
+    it("selectionForIndex", () => {
```
aicioara: `selectionForIndex()`

src/drawers/areaDrawer.ts
jtlan: Calling `super()` would attach a line since `Drawers.Area` extends `Drawers.Line`, so I assume the `.call()` structure was added on purpose.

This weirdness suggests that something is off about the inheritance (which switching to `Drawers.Path` would probably fix...).

aicioara: Oh... Well, in that case `Drawers.Area` should not inherit from `Drawers.Line`. I see no point in keeping consistency with plot inheritance

jtlan: I think there are some methods that would be duplicated between them, but I agree that the inheritance is odd.

aicioara: Opened https://github.com/palantir/plottable/issues/2232

PR #2235, opened 2015-06-05
Drawer.draw() endpoint returns `this`

### LINKED ISSUE #2197: [Drawer] - the `draw()` endpoint should return this
The `draw()` endpoint is inconsistent in that it draws stuff to the DOM and returns the time it takes to draw.  `draw()` maybe should be more consistent and return `this` but then the time idea is important.  There might be a need to get the time it takes to draw in another endpoint

### LINKED ISSUE #2232: Fix Drawer inheritance structure
`Drawers.Area` should not inherit from `Drawers.Line`. I see no point in keeping consistency with plot inheritance hierarchy. 

Also, looking at the `Drawers.Area`, there is no `super` call which uses the `Drawers.Line`, hence the inheritance is useless.

There might be more issues, this is just what arised from the thread [here](https://github.com/palantir/plottable/pull/2231#discussion_r31837732)

aicioara:
Closes #2197

This is one option. Is there a reason why we draw the labels (`Plots.Bar` only) at the end? That was the only place where the drawer time to draw was used. If we want that behavior, we can: 
- add another public method on the drawer to just give us this value after a `draw()` call. Let's call it `.animationTime()`. It will return 0 if the animation has already finished, warn if it did not yet start (`.draw()` not yet triggered) or the amount in milliseconds until the animation stops (which requires carying around a new instance variable) 
- leave it like this (discard this PR)
- Something else? I feel like there should be a 3rd option

The issue right now can be observed in the `base_animator.ts` test in the overlaying tests


jtlan:
>Is there a reason why we draw the labels (Plots.Bar only) at the end?

It's so that the bars finish animating before the labels show up.

aicioara:
Opened #2247 for a second attempt

PR #2236, opened 2015-06-05
Fix extent calculation on QuantitativeScale to account for non-numbers.

jtlan:
`d3.extent()` behaves differently if the first element in the array isn't a `number`, so I filtered the values before passing them in.

Test fiddle showing the bug (on develop): http://jsfiddle.net/b48jk5yx/

One thing we should consider is if we want to have `Scales.Time` reject non-`Date` objects -- see the skipped test in **timeScaleTests.ts**.

One more thing to fix here is that `entities()` returns corrupted data if there are datums that are skipped. going to get to that next.

Test fiddle showing the bug (on `numericExtent`): http://jsfiddle.net/njqnu1vf/ (open the console)

### DIFF
--- a/src/scales/quantitativeScale.ts
+++ b/src/scales/quantitativeScale.ts
@@ -223,8 +223,8 @@ module Plottable {
     }
 
     public extentOfValues(values: D[]): D[] {
-      // HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
-      var extent = d3.extent(<any[]> values);
+//       HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
+      var extent = d3.extent(<any[]> values.filter((value) => Utils.Math.isValidNumber(+value)));
       if (extent[0] == null || extent[1] == null) {
         return [];
       } else {
--- a/src/scales/quantitativeScale.ts
+++ b/src/scales/quantitativeScale.ts
@@ -223,7 +223,7 @@ module Plottable {
     }
 
     public extentOfValues(values: D[]): D[] {
-//       HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
+      // HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
       var extent = d3.extent(<any[]> values.filter((value) => Utils.Math.isValidNumber(+value)));
       if (extent[0] == null || extent[1] == null) {
         return [];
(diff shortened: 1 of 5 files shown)


aicioara:
Does this close any issue? How does this PR fit in the bigger picture?

src/scales/quantitativeScale.ts
```diff
@@ -223,8 +223,8 @@ module Plottable {
     }
 
     public extentOfValues(values: D[]): D[] {
-      // HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
-      var extent = d3.extent(<any[]> values);
+//       HACKHACK: TS1.4 doesn't consider numbers to be Number-like (valueOf() returning number), so D can't be typed correctly
```
aicioara: Indentation

test/scales/linearScaleTests.ts
```diff
@@ -4,6 +4,14 @@ var assert = chai.assert;
 
 describe("Scales", () => {
   describe("Linear Scales", () => {
+    it("extentOfValues() filters out invalid numbers", () => {
+      var scale = new Plottable.Scales.Linear();
+      var expectedExtent = [0, 1];
+      var arrayWithBadValues: any[] = [null, NaN, undefined, Infinity, -Infinity, "a string", 0, 1];
```
aicioara: Do you need explicit typing?

test/scales/timeScaleTests.ts
```diff
@@ -3,6 +3,16 @@
 var assert = chai.assert;
 
 describe("TimeScale tests", () => {
+    it.skip("extentOfValues() filters out invalid Dates", () => {
+      var scale = new Plottable.Scales.Time();
+      var expectedExtent = [new Date("2015-06-05"), new Date("2015-06-04")];
+      var arrayWithBadValues: any[] = [null, NaN, undefined, Infinity, -Infinity, "a string",
+        0, new Date("2015-06-05"), new Date("2015-06-04")];
+      var extent = scale.extentOfValues(arrayWithBadValues);
+      assert.strictEqual(extent[0].getTime(), expectedExtent[0].getTime(), "returned correct min");
```
aicioara: They are not going to be equal because they are 2 different objects (reference wise). Not if `mocha` goes around that for dates, but this will fail when you unskip

test/scales/linearScaleTests.ts
jtlan: Yes, otherwise it becomes an array of `string | number`, which can't be passed in to `extentOfValues()`.

test/scales/timeScaleTests.ts
jtlan: The error is actually that `0` makes it through, since it's a valid number.

aicioara: oh, sorry, my bad

aicioara:
We'll want to refer to this when addressing this https://github.com/palantir/plottable/issues/1866

crmorford:
so confused by this function. But your fix looks good!
screen shot 2015-06-05 at 3 17 06 pm

jtlan:
Was going to do the 2nd fix on this branch, but decided to move it to a different PR.

PR #2237, opened 2015-06-05
Fix entity bug

jtlan:
The secondary bug mentioned on #2236.

Test fiddle showing the bug: http://jsfiddle.net/njqnu1vf/1/ (open the console)
**expected**: circles are drawn around the position of each `Entity`
**actual**: no circles drawn, error in console

#### The fix
Keep a different index that only counts the valid datums, and use that index to retrieve the `Selection`. Invalid datums are never drawn, so they have no `Selection`.

A better fix might be to have a method `selectionForDatum()` rather than `selectionForIndex()` on `Drawer`.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -1711,8 +1711,8 @@ var Plottable;
             return this;
         };
         QuantitativeScale.prototype.extentOfValues = function (values) {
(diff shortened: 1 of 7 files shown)


jtlan:
Going to recreate this since there seems to be some sort of bug in the display of the diff.

PR #2240, opened 2015-06-05
Use loopIndex when searching for selection.

jtlan:
The secondary bug mentioned on #2236, remake of #2237.

Test fiddle showing the bug: http://jsfiddle.net/njqnu1vf/1/ (open the console)
**expected**: circles are drawn around the position of each `Entity`
**actual**: no circles drawn, error in console

#### The fix
Keep a different index that only counts the valid datums, and use that index to retrieve the `Selection`. Invalid datums are never drawn, so they have no `Selection`.

A better fix might be to have a method `selectionForDatum()` rather than `selectionForIndex()` on `Drawer`.

Fiddle as fixed on this branch: http://jsfiddle.net/ap0ujhu0/

### DIFF
--- a/src/components/plots/plot.ts
+++ b/src/components/plots/plot.ts
@@ -459,19 +459,21 @@ module Plottable {
       var entities: Plots.Entity[] = [];
       datasets.forEach((dataset) => {
         var drawer = this._datasetToDrawer.get(dataset);
-        dataset.data().forEach((datum: any, index: number) => {
-          var position = this._pixelPoint(datum, index, dataset);
+        var loopIndex = 0;
+        dataset.data().forEach((datum: any, datasetIndex: number) => {
+          var position = this._pixelPoint(datum, datasetIndex, dataset);
           if (position.x !== position.x || position.y !== position.y) {
             return;
           }
           entities.push({
             datum: datum,
-            index: index,
+            index: datasetIndex,
             dataset: dataset,
             position: position,
-            selection: drawer.selectionForIndex(index),
+            selection: drawer.selectionForIndex(loopIndex),
             plot: this
           });
+          loopIndex++;
         });
       });
       return entities;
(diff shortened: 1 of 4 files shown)


src/components/plots/plot.ts
```diff
@@ -459,19 +459,21 @@ module Plottable {
...
             datum: datum,
-            index: index,
+            index: datasetIndex,
             dataset: dataset,
             position: position,
-            selection: drawer.selectionForIndex(index),
+            selection: drawer.selectionForIndex(loopIndex),
             plot: this
           });
+          loopIndex++;
```
aicioara: How can `loopIndex` and `datasetIndex` ever be different?

src/components/plots/plot.ts
```diff
@@ -459,19 +459,21 @@ module Plottable {
       var entities: Plots.Entity[] = [];
       datasets.forEach((dataset) => {
         var drawer = this._datasetToDrawer.get(dataset);
-        dataset.data().forEach((datum: any, index: number) => {
-          var position = this._pixelPoint(datum, index, dataset);
+        var loopIndex = 0;
```
aicioara: the var name is misleading. `drawerIndex`, `entitiesIndex` `selectionIndex`, `drawerSelectionIndex` could be more appropriate

jtlan: `validDatumIndex`?

aicioara: sounds good

aicioara:
We should try to iterate open PRs faster. I totally forgot I did this in the past already and it was last commit only.

jtlan:
@crmorford I think the unit tests covered this case decently, although you might want to check what happens with `undefined` and `Infinity` instead of `NaN`. When checking it on my own it seemed fine though.

`null` gets coerced to `0` instead of behaving like `undefined`, but I think that's a separate problem.

PR #2242, opened 2015-06-05
Update Plottable to use UMD.

### LINKED ISSUE #2241: Convert plottable to use UMD
Necessary so users can upgrade slowly.

dbt:
Closes #2241.


jtlan:
Assuming control of this pull request.

crmorford:
regression pass looks okay

test/drawers/drawerTests.ts
```diff
@@ -20,10 +20,12 @@ class MockAnimator implements Plottable.Animators.Plot {
   }
 }
 
-class MockDrawer extends Plottable.Drawer {
-  public _drawStep(step: Plottable.Drawers.AppliedDrawStep) {
-    step.animator.animate(this.renderArea(), step.attrToAppliedProjector);
-  }
+function getMockDrawer(dataset: Plottable.Dataset) {
```
bluong: Rename function to `constructMockDrawer()` or `createMockDrawer()`?

Also, unless we are having this be the endpoint for having a mock drawer, we should probably name this function to be more aligned with what you are mocking.

bluong:
Do developers need to `npm install` / `npm update` after this?  What's the update process for us?

Gruntfile.js
```diff
@@ -147,6 +147,13 @@ module.exports = function(grunt) {
   var configJSON = {
     pkg: grunt.file.readJSON("package.json"),
     bump: bumpJSON,
+    umd: {
+      all: {
+        src: "plottable.js", 
+        objectToExport: "Plottable",
+        amdModuleId: "Plottable",
```
bluong: What happens if we don't set `objectToExport` and `amdModuleId`?  I'm reading the documentation on https://github.com/bebraw/grunt-umd and I'm unsure what's the advantages / disadvantages of setting this directly

bluong:
This doesn't seem like something we can unit test.  Can some test or so be pushed up so that we can prove that this works?

test/drawers/drawerTests.ts
jtlan: `createMockDrawer()` seems fine to me. What do you mean by "more aligned with what you are mocking"?

Gruntfile.js
jtlan: I think we need to set `objectToExport`, since an object (named "`Plottable`") is being exported.

`amdModuleId` seems like it shouldn't be left out since I don't think we want the module to be anonymous.

jtlan:
> Do developers need to `npm install` / `npm update` after this?  What's the update process for us?

You'll need to `npm install` after this to get the new module.

test/drawers/drawerTests.ts
bluong: Note that this mock drawer specifically mocks out the `drawStep` function.  Does that mean that if other tests wanted to use a mock drawer, should they always retrieve this one that has its `drawStep` mocked out? or should this function be renamed to more hint at the fact that the `drawStep` function is mocked out?

jtlan:
>This doesn't seem like something we can unit test. Can some test or so be pushed up so that we can prove that this works?

I'll make a small thing now; eventually we might want to convert our overlaying tests or something to use UMD.

quicktests/umd/app.js
```diff
@@ -0,0 +1,36 @@
...
+    plot.x(function(d) { return d.name; }, xScale);
+    plot.y(function(d) { return d.value; }, yScale);
+    
+    var chart = new Plottable.Components.Table([
+      [yAxis, plot],
+      [null, xAxis]
+    ]);
+    chart.renderTo(d3.select("#chart"));
+  });
+});
```
bluong: Add a newline here please

test/drawers/drawerTests.ts
bluong: Please change this to createMockDrawer

bluong:
Please provide QE with context on this one.

crmorford:
My understanding is that I'm just checking to make sure this doesn't break anything, since I can't really simulate how the customer is going to use this. Sound correct?

jtlan:
The UMD quicktest is supposed to demonstrate using requirejs to load Plottable. I think checking that loads up correctly is the extent of what we can do for now.

PR #2243, opened 2015-06-07
[Legend] Return Entities instead of Selection when getting entries

jtlan:
`Entity` is now generic on the `Component` type:
```Typescript
export interface Entity<C extends Component> {
  datum: any;
  index: number;
  position: Point;
  selection: d3.Selection<any>;
  component: C;
}
```

And `Plots.PlotEntity` extends `Entity`:
```Typescript
interface PlotEntity extends Entity<Plot> {
  dataset: Dataset;
  component: Plot;
}
```

On `Legend`, `entitiesAt()` replaces `getEntry()`, returns `Entity<Legend>[]`. Usually, only one `Entity` is ever returned. The method is still named `entitiesAt()` for consistency with the `Plot` methods. Multiple `Entries` might be returned if the query point is exactly on the border between them.

Example usage: http://jsfiddle.net/0d5tk6v2/1/


src/components/legend.ts
```diff
@@ -206,36 +206,49 @@ export module Components {
...
+              highY >= p.y && lowY <= p.y) {
+            var entrySelection = d3.select(this);
+            var datum = entrySelection.datum();
+            var index = domain.indexOf(datum);
+            entities.push({
+              datum: datum,
+              index: index,
+              position: { x: symbolX, y: symbolY },
+              selection: entrySelection,
+              component: legend
```
aicioara: I don't yet have the full picture of how is this going to be consumed later in the pipeline, but it seems like some of the entities can be derived from the others? Do we want to do something about it?

src/components/plots/barPlot.ts
```diff
@@ -206,16 +206,16 @@ export module Plots {
...
+     * Returns the PlotEntity nearest to the query point according to the following algorithm:
+     *   - If the query point is inside a bar, returns the PlotEntity for that bar.
+     *   - Otherwise, gets the nearest PlotEntity by the primary direction (X for vertical, Y for horizontal),
      *     breaking ties with the secondary direction.
-     * Returns undefined if no Entity can be found.
+     * Returns undefined if no PlotEntity can be found.
      *
      * @param {Point} queryPoint
-     * @returns {Plots.Entity} The nearest Entity, or undefined if no Entity can be found.
+     * @returns {Plots.PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
```
aicioara: `Plots.PlotEntity` not in sync with the bellow where we say `PlotEntity` (not prefixed)

src/components/plots/barPlot.ts
```diff
@@ -519,7 +519,7 @@ export module Plots {
       return barPixelWidth;
     }
 
-    public entities(datasets = this.datasets()): Plots.Entity[] {
+    public entities(datasets = this.datasets()): Plots.PlotEntity[] {
```
aicioara: Not in sync with the above where we just say `PlotEntity`

src/core/interfaces.ts
```diff
@@ -46,4 +46,12 @@ module Plottable {
     topLeft: Point;
     bottomRight: Point;
   };
+
+  export interface Entity<C extends Component> {
+    datum: any;
+    index: number;
+    position: Point;
+    selection: d3.Selection<any>;
+    component: C;
```
aicioara: what is the difference between index and position?

src/core/interfaces.ts
```diff
@@ -46,4 +46,12 @@ module Plottable {
     topLeft: Point;
     bottomRight: Point;
   };
+
+  export interface Entity<C extends Component> {
```
aicioara: Some documentation would be nice. At least a hint of how is this going to be used

test/testMethods.ts
```diff
@@ -91,18 +91,27 @@ module TestMethods {
     assert.strictEqual(height, String(heightExpected), "height: " + message);
   }
 
-  export function assertEntitiesEqual(actual: Plottable.Plots.Entity, expected: Plottable.Plots.Entity,
-      msg: string) {
+  export function assertEntitiesEqual(
+    actual: Plottable.Entity<Plottable.Component>,
+    expected: Plottable.Entity<Plottable.Component>,
+    msg: string) {
```
aicioara: I prefer double indentation here (add 2 more spaces) to distinguish from the body of the method in some way)

test/testMethods.ts
```diff
@@ -91,18 +91,27 @@ module TestMethods {
...
       assert.strictEqual(element, expected.selection[0][index], msg + " (selection contents)");
     });
-    assert.strictEqual(actual.plot, expected.plot, msg + " (plot)");
+    assert.strictEqual(actual.component, expected.component, msg + " (component)");
+  }
+
+  export function assertPlotEntitiesEqual(
+    actual: Plottable.Plots.PlotEntity,
+    expected: Plottable.Plots.PlotEntity,
+    msg: string) {
```
aicioara: Same

aicioara:
Just minor things like consistency in naming between `Plots.PlotEntity` and `PlotEntity`. Also make tests pass on IE

src/components/legend.ts
jtlan: What do you mean by "derived from the others"?

src/core/interfaces.ts
jtlan: Previously `index` would have been the position in the `Dataset` array, and `position` is a representative location on the screen. Possibly one or both of them doesn't belong in the base `Entity`?

test/testMethods.ts
jtlan: Sounds good.

src/components/plots/barPlot.ts
jtlan: Some of this confusion is because on `Plot` we have to say `Plots.PlotEntity` (since `Plot` is outside of the `Plots` module), but not on any of the subclasses.

src/components/legend.ts
aicioara: For example you push the `entitySelection`, but then the datum is just calling `.datum()` on that selection, which seems redundant. Then `legend` is just `this`, but `entitySelection` is just `d3.select(this)`, so `entitySelection` in itself seems redundant. It also exposes a d3 entity to the consumer of a public method, which I thought we avoid at all costs

src/core/interfaces.ts
aicioara: That was just a first read impression. Maybe it makes sense. Maybe `position` should be renamed to `coordinate`. I just think the two sound too similar to be able to tell the difference at a first glance

src/components/legend.ts
jtlan: > Then `legend` is just `this`, but `entitySelection` is just `d3.select(this)`, so `entitySelection` in itself seems redundant.

Actually the `this` being selected is the `<g>` for the entry, not the `Legend` itself (the `this`-context is different inside the closure). So the `Selection` and the `Component` are quite different.

I thought about returning the `Element` instead of a `Selection`. Thoughts on the matter?

src/core/interfaces.ts
jtlan: The type of `Position` is `Point`, so that should be enough indication for Typescript consumers. I did consider the name `Location` though.

`index` doesn't make much sense without `Dataset`, so I think I'll move that to `PlotEntity`.

src/components/legend.ts
aicioara: you mean the `<g>` element? That sounds better than exposing `d3`. Also, how about the other redundancy?

src/core/interfaces.ts
aicioara: Location is better than position, so I am okay with that change alone. 

Other than that, I have the feeling that this object is too verbose (exposing too much). Is everything above strictly necessary? This actually now brings us back to the other thread (above)

src/components/legend.ts
jtlan: It's not redundant in all cases, since for the `Entity` returned by `Plots.Line`, `selection.datum()` and `datum` are different.

PR #2245, opened 2015-06-08
Update d3 definitions to latest hash.

jtlan:
No actual diff in d3.d.ts since the PR I sent to DefinitelyTyped
was the latest version...


jtlan:
[NOQE] because only change was to **tsd.json**

PR #2248, opened 2015-06-08
[Animator.Rect] - Deleting this

bluong:
It seems that only `StackedBar Plot` is the only user of this animator... and the animation logic is basically the same as `Bar Plot`.  Thus, deleting all references to `Rect Animator`.

Also, since there was a talk a long long time ago about what the role of the animator should be, I would prefer to let the plot handle plot-specific animation until we reach a conclusion.  We can make a firm decision after we v1.0

API changes:
* `Animators.Rect` has been deleted.
* `Animators.MovingRect` has been deleted


aicioara:
Code wise, there's not much to review. Are we sure no one was using those end points? Is everything still working? If so I am okay with that change. Maybe @jtlan has more context on the matter, as I haven't played much with the animators.

jtlan:
I think the `MovingRect` `Animator` was intended for some sort of "one series pops up at a time" animation, I'm surprised that `Animators.Base` works on `Plots.StackedBar` though?

aicioara:
Sounds more like a QE pass to me? If everything works as before, less is better, hence let's go with it.

bluong:
@jtlan A great long time ago, I think I switched the animation logic so that the bars all expand at once.  That said, the logic used in the current develop logic is unless I'm mistaken, the same exact logic used for `Plots.Bar`.

Possibly irrelevant aside, a great long time ago, I tried to pursue the animator idea that plots should not need to define that weird reset animator and have the reset idea stored inside the animator logic, but as that is inconsistent with our other animator usages, I'm removing the inconsistency.

crmorford:
So there's just no animation for rectangle plots now?

bluong:
@crmorford We can check out develop to be sure, but I believe that rectangle plots had no animation to begin with.  I think it would make sense as one cannot really make any good assumptions for animation on rectangles (maybe the fade in @jtlan suggested?)

PR #2251, opened 2015-06-09
[Animator.Base] Refactor

### LINKED ISSUE #2246: [Animator.Base] methods names have to be refactored
Playing with the methods inside the Base Animator, lots of things did not work as expected. That is because poor documentation and non descriptive method names. Fix for `1.0.0`.

In particular
- `maxIterativeDelay` -> `iterativeDelay`
- `maxTotalDuration` should win over `duration`
- `duration` sounds too much like a total Duration, when it is in fact animation per bar (in bar plot) (`stepDuration`)
- `delay` is actually the starting delay
- the `maxTotalDuration` has a non `Infinity` default
- All the numbers should be tested more thorroughly
- The overlaying tests should be repaired (for example `Plot.animator("bars", animator) ->       Plot.animator("main", animator)`)

aicioara:
Closes #2246 

Changes:
- Added tests for the base animator timing
- Extracted common code inside the `Animator.Base`
- Rename `delay()` -> `startDelay()`
- Rename `duration()` -> `stepDuration()`
- Rename `maxIterativeDelay` -> `iterativeDelay()`
- `DEFAULT_MAX_TOTAL_DURATION_MILLISECONDS` defaults to `Infinity`
- Rename `DEFAULT_DELAY_MILLISECONDS` -> `DEFAULT_START_DELAY_MILLISECONDS`
- Rename `DEFAULT_DURATION_MILLISECONDS` -> `DEFAULT_STEP_DURATION_MILLISECONDS`
- Rename `DEFAULT_MAX_ITERATIVE_DELAY_MILLISECONDS` -> `DEFAULT_ITERATIVE_DELAY_MILLISECONDS`
- Fixed off by one errors
- Fixed documentation
- Added `ANIMATION_MAX_DURATION` as a public static field on the `Plot` to suggest that 600 ms is the ideal max total duration for a plot animation

**QE Notes:** Mainly regression pass. It is merely just a method rename as seen above. Everything should work as before.


src/components/plots/scatterPlot.ts
```diff
@@ -16,7 +16,11 @@ export module Plots {
     constructor() {
       super();
       this.classed("scatter-plot", true);
-      this.animator(Plots.Animator.MAIN, new Animators.Base().duration(250).delay(5));
+      var animator = new Animators.Base();
+      animator.startDelay(5);
+      animator.stepDuration(250);
+      animator.maxTotalDuration(600);
```
bluong: I see this 600 all over the place.  Let's standardize this to a variable?

test/animators/baseAnimatorTests.ts
```diff
@@ -0,0 +1,156 @@
...
+        animator.stepDuration(stepDuration);
+        animator.iterativeDelay(iterativeDelay);
+        animator.maxTotalDuration(maxTotalDuration);
+
+        var expectedIterativeDelay = 500;
+        // 1 |  ###########
+        // 2 |       ###########
+        //   +------------------------
+        //   |  |    |    |    |    |
+        //     0.0  0.5  1.0  1.5  2.0
```
bluong: Is there a way we can get this idea through without pictures.  Pictures take up space :/

src/components/plots/scatterPlot.ts
bluong: Can we instead of making a new animator, just take the existing animator and then configure that?

test/animators/baseAnimatorTests.ts
aicioara: But it's pretty. And more expressive than a 1 paragraph explanation.

src/components/plots/scatterPlot.ts
aicioara: I can create a new variable on the base animator called `RECOMMENDED_MAX_TOTAL_DURATION`. I would want to avoid creating a new animator for just one variable. It does not really make sense.

aicioara: However, this is something very plot specific. 5 and 250 are also there by themselves

bluong: Yes and thus, I believe the Plot parent should hold a static variable for the maxTotalDuration for its base animator

aicioara: Sounds good

src/components/plots/linePlot.ts
```diff
@@ -12,7 +12,11 @@ export module Plots {
     constructor() {
       super();
       this.classed("line-plot", true);
-      this.animator(Plots.Animator.MAIN, new Animators.Base().duration(600).easing("exp-in-out"));
+      var animator = new Animators.Base();
+      animator.stepDuration(600);
```
bluong: Do you think the 600 was intentioned to be the ANIMATION_MAX_DURATION?

aicioara: Hmm, that's a good point. I think you're right

aicioara: Changing...

PR #2252, opened 2015-06-09
Using Enum instead of string for Plot.Animator.MAIN

aicioara:
Same for Plots.Animator.RESET


jtlan:
Did you check the animations on the Overlaying Tests? If so it's probably safe to [NOQE].

aicioara:
Checked overlaying tests myself, [NOQE]. Nothing has been changed anyway, as we are just getting the exact same value from a constant

PR #2254, opened 2015-06-09
Privitizing

bluong:
There are several instance variables that are protected that should be instead private.  Few cleanups that we would want to do before 1.0

QE: Regression pass please.  The JS got affected functionally in very small ways, specifically with how `Axis.Time` sets up its DOM elements and how `Plots.ClusteredBar` lays out the bars for the cluster, but these changes should hopefully not affect you.

### DIFF
--- a/src/components/plots/stackedAreaPlot.ts
+++ b/src/components/plots/stackedAreaPlot.ts
@@ -97,10 +97,6 @@ export module Plots {
       return this;
     }
 
-    protected _wholeDatumAttributes() {
-      return ["x", "y", "defined", "d"];
-    }
-
     protected _updateExtentsForProperty(property: string) {
       super._updateExtentsForProperty(property);
       if ((property === "x" || property === "y") && this._projectorsReady()) {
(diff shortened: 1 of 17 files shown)


src/components/plots/stackedAreaPlot.ts
```diff
@@ -97,10 +97,6 @@ export module Plots {
       return this;
     }
 
-    protected _wholeDatumAttributes() {
```
jtlan: Was this just unused? How did that happen?

jtlan:
[+1] Probably a regression pass is sufficient, if that's even necessary.

src/components/plots/stackedAreaPlot.ts
bluong: I likely forgot this when i was cleaning up areaPlot

bluong:
Let's be safe unless bandwidth gets clogged

crmorford:
Axis.Time looks good in the regression pass, wanna check out clustered bar a little bit.

PR #2255, opened 2015-06-09
Underscore fixes

bluong:
Some variables were underscored incorrectly and thus are fixed here

### DIFF
--- a/src/core/renderController.ts
+++ b/src/core/renderController.ts
@@ -18,10 +18,10 @@ module Plottable {
    * ```
    */
   export module RenderController {
-    var componentsNeedingRender = new Utils.Set<Component>();
-    var componentsNeedingComputeLayout = new Utils.Set<Component>();
-    var animationRequested = false;
-    var isCurrentlyFlushing = false;
+    var _componentsNeedingRender = new Utils.Set<Component>();
+    var _componentsNeedingComputeLayout = new Utils.Set<Component>();
+    var _animationRequested = false;
+    var _isCurrentlyFlushing = false;
     export module Policy {
       export var IMMEDIATE = "immediate";
       export var ANIMATION_FRAME = "animationframe";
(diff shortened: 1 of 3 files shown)


src/core/renderController.ts
```diff
@@ -18,10 +18,10 @@ module Plottable {
...
    */
   export module RenderController {
-    var _componentsNeedingRender = new Utils.Set<Component>();
-    var _componentsNeedingComputeLayout = new Utils.Set<Component>();
-    var _animationRequested = false;
-    var _isCurrentlyFlushing = false;
+    var componentsNeedingRender = new Utils.Set<Component>();
+    var componentsNeedingComputeLayout = new Utils.Set<Component>();
+    var animationRequested = false;
+    var isCurrentlyFlushing = false;
```
aicioara: Why this change? These are all private variables.

bluong: They are not exposed to the user.  There's no way for them to access these vars.

aicioara: What is the issue with the underscore than?

bluong: Having the underscore convention here is similar to having underscores for our local variables within function definitions.  The intention of the underscore was so if a user were to call like object._someFunction, they would be signaled that the function usage as at your own risk.  However, since these variables are not even exported, there is no RenderController._componentsNeedingRender being shown to the user.

aicioara: Okay then

aicioara: I think it is more consistent with the rest of the rest of our codebase. Underscored always means not exported, why are these more special?

jtlan: Agree with @aicioara here -- while it's not necessary to underscore them, it will result in less confusion.

aicioara:
NOQE because it is just private method rename

PR #2256, opened 2015-06-09
[Plot] - animateOnNextRender turned into a function

bluong:
Mainly so descendents have a consistent way to check if animations should happen.  Also, this helps so that the variable does not have to be manually updated

API Breaks:
plot._animate is private
plot._dataChanged is private
plot._animateOnNextRender turned into a function

QE: I would say regression pass on mainly animations if you have enough bandwidth


jtlan:
[+1] Nice.

crmorford:
On the stocks.js quicktest, there is an animation every time you zoom or pan.

PR #2257, opened 2015-06-09
[Legend] - legend.scale -> legend.colorScale

bluong:
By definition, a Legend is a visualization of a colorScale.  It should be named such

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1929,7 +1929,7 @@ declare module Plottable {
              * @constructor
              * @param {Scale.Color} scale
              */
-            constructor(scale: Scales.Color);
+            constructor(colorScale: Scales.Color);
             protected _setup(): void;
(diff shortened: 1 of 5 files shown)


jtlan:
If it's implied that the `Scale` is a `Scales.Color`, why would we need to name the variable as such? On `Axes.Time` we don't call the `Scale` parameter "`timeScale`", for example.

aicioara:
But the `Axes.Time` does not have a method called `scale()`. This one does. Argument name is not as important as method name, and I believe the change makes sense. 

However, we have to watch out for inconsistencies (if there are any I did not spot)

jtlan:
Makes sense. Also might be helpful if we want to add a `radiusScale()` API point later or something.

bluong:
Simple rename.  NOQE

PR #2258, opened 2015-06-09
[Plots.Bar] - labelFormatter -> labelsFormatter

bluong:
Being consistent since a parallel endpoint `labelsEnabled` is also plural

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2738,14 +2738,14 @@ declare module Plottable {
             /**
              * Gets the Formatter for the labels.
              */
-            labelFormatter(): Formatter;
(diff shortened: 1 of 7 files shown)


jtlan:
The function of the word "label" in the two names is different, since "`labelsEnabled`" answers a question about the labels, but in "`labelFormatter`" the word "label" is describing "Formatter". I think the rename is fine though.

Be sure to fix the Overlaying Test for this.

aicioara:
It is also now in sync with the old comment
```typescript
    /**
     * Gets the Formatter for the labels.
     */
```

bluong:
Tis a simple rename.  Decently sure it's NOQE

PR #2263, opened 2015-06-09
Move index out of Entity into PlotEntity

jtlan:
`index` doesn't have a well-defined meaning for a general `Component`, so I've removed it from the base `Entity` type.


bluong:
I Think this is NOQE

Issue #2268, opened 2015-06-09
Dispatcher and Interaction refactor


aicioara:
- [x] Dispatcher has an instance variable called _event2Callback.  We should change the “2" to a “To"
- [x] Dispatchers.Mouse.lastMousePosition should be explicitly typed to Point
- [x] Interactions.Key has “onKey” endpoints that should probably be renamed to something like “onKeyPress”
- [x] PanZoomInteraction has an exposed PIXELS_PER_LINE that should be looked at.
- [x] Interactions.Drag has a constrainToComponent() endpoint that should be renamed so that dragInteraction.constrainToComponent() more looks like a getter

bluong:
Got all the checkmarks.

PR #2270, opened 2015-06-09
[Interactions.Key] - onKey -> onKeyPress

bluong:


aicioara:
Do you want to take the entire ticket here as a PR https://github.com/palantir/plottable/issues/2268 ?

bluong:
I would rather not.  Unitize as much as possible

aicioara:
Don't forget to revert quicktests guards at the end

bluong:
Should be done after the QE pass at the end

PR #2275, opened 2015-06-10
Axes Refactor

### LINKED ISSUE #2266: Axes Refactor
Axis.gutter maybe should be worded to margin. The documentation surrounding should point to the endpoint more clearly.

The formatter constructor argument on Axes can probably be removed.

aicioara:
Closes #2266

- Documentation
- `Axis.gutter()` -> `Axis.margin()`
- `new Plottable.Axis` no longer takes a `Formatter` as the third argument. Instead the `.formatter()` method should be used after instantiation

**QE Notes:** The formatter bit has changed for the axes. In particular, the numeric axis should behave just like before (not show too many / too little decimals). On quicktests it seemed alright. The other half of this PR is just method rename


aicioara:
Fixing documentation

src/components/axes/axis.ts
```diff
@@ -358,26 +357,26 @@ module Plottable {
     }
 
     /**
-     * Gets the size of the gutter in pixels.
-     * The gutter is the extra space between the tick labels and the outer edge of the Axis.
+     * Gets the size of the margin in pixels.
+     * The margin is the extra space between the tick labels and the outer edge of the Axis.
```
bluong: Can you rewrite the documentation so that it more hints that the margin is indeed the size?

So like
"Gets the margin in pixels."

aicioara:
Needs QE for the formatter bit. In particular, the numeric axis should behave just like before (not show too many too little decimals). On quicktests it seemed alright.

jtlan:
> The formatter bit has changed for the axes. In particular, the numeric axis should behave just like before (not show too many / too little decimals). 

The whole thing is just renames / moving calls, so nothing should have changed, right?

aicioara:
Correct

crmorford:
Formatters look good. 
In overlaying tests (comparing develop vs #local) symbols.js from this branch doesn't work on develop; the pointer and key interactions are broken (the callbacks never happen). I notice you didn't change the test, so do you want to see if that happens on your computer? Not sure if tests are flaking or if it's a sign of a bug.

aicioara:
It might be because of lots of breaking changes on develop. I just merged develop in, does it still happen? Note I could not spot anything wrong in my version of `symbols.js`

PR #2276, opened 2015-06-10
[Dispatcher.Mouse] - lastMousePosition explicitly typed as Point

bluong:


src/dispatchers/mouseDispatcher.ts
```diff
@@ -191,7 +191,7 @@ export module Dispatchers {
      *
      * @return {Point}
      */
-    public lastMousePosition() {
+    public lastMousePosition(): Point {
```
aicioara: That is so weird. Looks like a bug from typescript side.

Oh, well

aicioara:
Links here https://github.com/palantir/plottable/issues/2268

src/dispatchers/mouseDispatcher.ts
bluong: It's not really a bug from Typescript side, because how would Typescript know that this is a `Point` rather than an x,y object?

That said, I would much rather have Typescript better support implicit return types on types aliases

Issue #2272, opened 2015-06-09
Small Refactor issues


aicioara:
- [x] There is a POLYFILL_TIMEOUT_MSEC variable exposed in Utils.DOM.  We should probably hide this variable. (PR: #2293)
- [x] There is a function in Plottable.Formatters that is not a Formatter. (PR: #2278)
- [x] The parameter name to XYPlot.autoRange does not reflect the endpoint name (PR: #2295).
- [x] The RenderController has a setRenderPolicy function.  We might want to split this to a getter/setter combo. (PR: #2280)

**EDIT**
<del>Plottable has a Configs module used for currently shutting off the warnings thrown to the console.  However, we just learned recently that this endpoint may not be used.  We may want to remove this. (PR: #2279)</del>

jtlan:
@bluong has the last checkbox on his PR, so I'm reassigning to him.

bluong:
Got all the checkboxes

PR #2278, opened 2015-06-10
[Formatters] - moving intervalToTime function to the Scales.Time

bluong:
The calculation is specific to `Axes.Time` and `Scales.Time`.  Since `Scales.Time` is more of the brains of the two, I thought that it should go on the `Scale` class

API Changes:
* `Formatters.timeIntervalTod3Time()` has been moved to `Scales.Time.timeIntervalTod3Time()`


aicioara:

PR #2279, opened 2015-06-10
[Config] - delete config.ts

bluong:
We found that I think the requester for this feature actually did not use this feature.  We should probably take this chance to remove this idea from the codebase so that we can possibly restart and think more heavily about a "developer-mode"/"production-mode" usage of Plottable, which is what `SHOW_WARNINGS` was essentially trying to accomplish


aicioara:
Links here https://github.com/palantir/plottable/pull/2279

jtlan:
Just because they haven't turned it on doesn't mean it's not valuable...?

PR #2280, opened 2015-06-10
[RenderController] - Adding getter/setter renderPolicy endpoint

bluong:
API Breaks:
* `RenderController._renderPolicy` is no longer exposed
* `RenderController.setRenderPolicy()` -> `RenderController.renderPolicy()`

QE: Regression pass should be enough.  You can try setting the renderPolicy if you're so curious, albeit I don't think you'll see much functional changes and mainly performance changes.


aicioara:
Can you link to the issue?

aicioara:

PR #2281, opened 2015-06-10
[Interaction.PanZoom] - privitizing PIXELS_PER_LINE

bluong:
Doing this mainly so that we can answer the question for how to deal with this variable later.

The idea of PIXELS_PER_LINE is so the user has an understanding of the conversion for browsers like Firefox that report their wheel events differently than others.  This should in my opinion be configurable so that users can set it to a more correct value if they should so choose but currently as it is a static variable, it has the same on all plots which may or may not be correct.

What also is concerning is that I actually do not think that this belongs in `Interaction.PanZoom`.  It should instead be in something like `Interactions.Scroll` (which doesn't exist) or maybe even in `Dispatchers.Wheel`.

If we are certain that it should be a property of `Dispatchers.Wheel` then I can make it thus, but if we aren't sure, let's hide it for now until we are.

API Breaks:
* Interactions.PanZoom no longer exposes a `PIXELS_PER_LINE` static variable

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3527,10 +3527,6 @@ declare module Plottable {
     module Interactions {
         class PanZoom extends Interaction {
             /**
-             * The number of pixels occupied in a line.
-             */
-            static PIXELS_PER_LINE: number;
-            /**
              * A PanZoom Interaction updates the domains of an x-scale and/or a y-scale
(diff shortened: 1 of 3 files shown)


aicioara:
Links here https://github.com/palantir/plottable/issues/2268

Issue #2267, opened 2015-06-09
Legend Refactor


aicioara:
- [x] Legend.maxEntriesPerRow should have a more reflected parameter name (PR: #2283)
- [x] Legend.symbolFactoryAccessor does not take in a Plottable Accessor (PR: #2309).
- [x] InterpolatedColorLegend has an orientation constructor argument that is optional, which is inconsistent. (PR: #2284)

bluong:
All checks.  CLose!

PR #2284, opened 2015-06-10
[InterpolatedColorLegend] - removing optional arguments from the constructor

bluong:
API Breaks:
* `InterpolatedColorLegend` no longer takes in an `orientation` parameter + `formatter` parameter as arguments to the constructor

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2024,10 +2024,8 @@ declare module Plottable {
              *
              * @constructor
              * @param {Scales.InterpolatedColor} interpolatedColorScale
-             * @param {string} [orientation="horizontal"] One of "horizontal"/"left"/"right".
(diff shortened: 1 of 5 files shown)


aicioara:
Again, can you link to the issues so we keep better track of what's been done and what's left to be done

aicioara:
Links here https://github.com/palantir/plottable/issues/2267

crmorford:
So now can InterpolatedColorLegends only be horizontal? Can I call orientation later?

bluong:
The constructor to `InterpolatedColorLegend` now only takes in an `InterpolatedColorScale`.  There are already existing endpoints for changing the formatter and the orientation, so just use those endpoints as needed to change the configuration of the legend

PR #2285, opened 2015-06-10
Scale Refactor

### LINKED ISSUE #2271: Scale Refactor
- [ ] QuantitativeScale._DEFAULT_NUM_TICKS seems like it is not used and should be deleted.
- [ ] Possibly remove the colorRange constructor parameter from InterpolatedColorScale.

aicioara:
Closes #2271

Interpolated Color scale no longer takes the colorRange as a constructor argument. Instead, it is now set via the `range()` method

**Reason:** We want InterpolatedColor to be consistent with Color, which takes as a first argument the scale type. In the same time, the argument was optional, and could have been set by the `colorRange()` method

Changelog:
- `Scales.InterpolatedColor.colorRange()` -> `Scales.InterpolatedColor.range()`
- `new Scales.InterpolatedColor(colorRange, scaleType)` -> `new Scales.InterpolatedColor(scaleType)`

**QE Notes** Regression pass on interpolated color scale. Checked the overlaying tests myself

### DIFF
--- a/src/scales/interpolatedColorScale.ts
+++ b/src/scales/interpolatedColorScale.ts
@@ -2,7 +2,7 @@
 
 module Plottable {
 export module Scales {
-  type supportedScale = d3.scale.Linear<number, string> | d3.scale.Log<number, string> | d3.scale.Pow<number, string>
+  type supportedScale = d3.scale.Linear<number, string> | d3.scale.Log<number, string> | d3.scale.Pow<number, string>;
 
   export class InterpolatedColor extends Scale<number, string> {
     public static REDS = [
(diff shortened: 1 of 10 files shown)


jtlan:
Should `colorRange()` just be called `range()`?

aicioara:
I am okay with the change. This makes it more consistent with `_getRange()` which just returns `this.colorRange()`

src/scales/interpolatedColorScale.ts
```diff
@@ -53,14 +53,11 @@ export module Scales {
...
-     * @constructor
-     * @param {string[]} [colors=InterpolatedColor.REDS] an array of strings representing color hex values
-     *   ("#FFFFFF") or keywords ("white").
      * @param {string} [scaleType="linear"] One of "linear"/"log"/"sqrt"/"pow".
      */
-    constructor(colorRange = InterpolatedColor.REDS, scaleType = "linear") {
+    constructor(scaleType = "linear") {
       super();
-      this._colorRange = colorRange;
+      this._colorRange = InterpolatedColor.REDS;
```
jtlan: Shouldn't we use `this.range(InterpolatedColor.REDS)`?

jtlan:
[-1] minor.

PR #2286, opened 2015-06-10
Switch the template used for UMD.

jtlan:
Avoids exporting `Plottable` to the global namespace if a module loader is used.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -150,8 +150,8 @@ module.exports = function(grunt) {
     umd: {
       all: {
         src: "plottable.js", 
+        template: "unit",
         objectToExport: "Plottable",
-//        amdModuleId: "Plottable",
       }
     },
     concat: {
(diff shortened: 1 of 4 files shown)


bluong:
Can you update your umd example or provide an example that checks that this indeed works?

jtlan:
Sure, basically `window.Plottable` should be null if require is used.

jtlan:
@cmslewis pulled the version off this branch and has confirmed that it works.

crmorford:
QE passing, since Chris has already confirmed it

cmslewis:
I think it'd be nice to add some kind of test of this form:

```html
<!-- Import an old version of Plottable, globally. -->
<script type="text/javascript" src="plottable-v0.32.js"></script>
<script type="text/javascript">
  // Verify that this old version of Plottable was loaded as a global variable.
  expect( window.Plottable.version ).to.deep.equal( '0.32.0' );
  
  // Now include the latest version of Plottable within a require module. 
  require(['plottable-v1.0.0-rc6.js'], function(Plottable) {
    // The old version should still be accessible via the global variable.
    expect( window.Plottable.version ).to.deep.equal( '0.32.0' );
    
    // But the locally scoped 'Plottable' variable should be the new version.
    expect( Plottable.version ).to.deep.equal( '1.0.0-rc6' );
  });
</script>

```

PR #2289, opened 2015-06-10
[InterpolatedColorLegend] - InterpolatedColorLegend -> InterpolatedLegend

bluong:
Similar to how `Legend` is not called `ColorLegend`, the "Color" part in `InterpolatedColorLegend` is redundant I feel.

API Changes:
* `InterpolatedColorLegend` has been renamed to `InterpolatedLegend`

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2007,7 +2007,7 @@ declare module Plottable {
 
 declare module Plottable {
     module Components {
-        class InterpolatedColorLegend extends Component {
+        class InterpolatedLegend extends Component {
             /**
              * The css class applied to the legend labels.
              */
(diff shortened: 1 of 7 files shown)


jtlan:
I'd argue that the "color" in the default `Legend` is implied. If I encounter "'InterpolatedLegend`" having never seen the library before, am I going to know what it is?

bluong:
A `Legend` by our definition is a representation of a `Scales.Color` by listing domain values next to some colored symbol.

With this, the parallel for `InterpolatedLegend` would be
An `InterpolatedLegend` by our definition is a representation of a `Scales.InterpolatedColor` by listing domain values next to some interpolated colored symbols

The parallel seems clean enough such that if people know what legends are, then they'd understand what an interpolated legend is imo

jtlan:
>The parallel seems clean enough such that if people know what legends are, then they'd understand what an interpolated legend is imo

They'd have to know specifically what our `Legend` is, instead of just the general concept of a "legend".

I'd say the primary strike against the name "`InterpolatedLegend`" is that the *colors* are being interpolated, not the `Legend`.

bluong:
Valid enough.  I'll close the PR

PR #2291, opened 2015-06-10
[WindowUtils] - delete copyObject

bluong:
It was used in `Drawers.Line` / `Drawers.Area` for not really much good reason.  Deleting its usage

API Changes:
* `Utils.Window.copyObject()` has been removed.

QE: Regression pass only for `Plots.Area` / `Plots.Line` and their animating


aicioara:
If nothing broke, I am fine with the change. I thought we make a copy of the object because we don't want it to change inside the method (i.e. guarantee a c-like `const`)

bluong:
I believe it was there for before I did the `Drawer` calculation movement.

PR #2292, opened 2015-06-10
[Utils.Math] - actually use isNaN()

bluong:
Utilizing `Utils.Math.isNaN()` so that we have a reason to expose the function

### DIFF
--- a/src/utils/domUtils.ts
+++ b/src/utils/domUtils.ts
@@ -151,10 +151,7 @@ export module Utils {
     function _parseStyleValue(style: CSSStyleDeclaration, property: string): number {
       var value = style.getPropertyValue(property);
       var parsedValue = parseFloat(value);
-      if (parsedValue !== parsedValue) {
-          return 0;
-      }
-      return parsedValue;
+      return Utils.Math.isNaN(parsedValue) ? 0 : parsedValue;
     }
   }
 }
--- a/src/utils/domUtils.ts
+++ b/src/utils/domUtils.ts
@@ -151,7 +151,7 @@ export module Utils {
     function _parseStyleValue(style: CSSStyleDeclaration, property: string): number {
       var value = style.getPropertyValue(property);
       var parsedValue = parseFloat(value);
-      return Utils.Math.isNaN(parsedValue) ? 0 : parsedValue;
+      return parsedValue || 0;
     }
   }
 }
--- a/src/utils/mathUtils.ts
+++ b/src/utils/mathUtils.ts
@@ -61,18 +61,11 @@ export module Utils {
     }
 
     /**
-     * Returns true **only** if x is NaN
-     */
-    export function isNaN(n: any) {
-      return n !== n;
-    }
-
-    /**
      * Returns true if the argument is a number, which is not NaN
      * Numbers represented as strings do not pass this function
      */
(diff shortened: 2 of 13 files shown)


src/utils/mathUtils.ts
```diff
@@ -61,18 +61,11 @@ export module Utils {
...
-      return n !== n;
-    }
-
-    /**
      * Returns true if the argument is a number, which is not NaN
      * Numbers represented as strings do not pass this function
      */
     export function isValidNumber(n: any) {
-      return typeof n === "number" && !Plottable.Utils.Math.isNaN(n) && isFinite(n);
+      return typeof n === "number" && n === n && isFinite(n);
```
aicioara: Leave isNaN as a private method. I don't like `n === n`. It is not expressive at all.

jtlan:
Does this change affect anything if later we want `isValidNumber()` to return true for `Infinity`/`-Infinity`?

src/utils/mathUtils.ts
jtlan: Also there are a number of places in the codebase where people were using `x !== x` instead of the method. While the strict-equal check is concise, I think it can be confusing for people who haven't seen it before.

aicioara:
Why would `+- Infinity` ever be a valid number?

bluong:
`Infinity` as a number is useful under some circumstances. (Take `legend.maxEntriesPerRow` as a good example)

src/utils/mathUtils.ts
bluong: Are you suggesting that we should use `isNaN()` in those situations rather than removing this function?  Note that I saw 0 usage of this function in the source code.  It is only used in the tests, meaning that at best it should be a function in the utils, but since it's used in this one place and `isNaN` is so short, I decided to just inline the definition

jtlan:
Or maybe we figure out a way to display it properly on `Plot`s without distorting the domain of the associated `Scale`.

src/utils/mathUtils.ts
aicioara: I suggest leave `isNaN()` in and instead use it where it should be used. The following RegEx'es might help

```regex
(\w+) !== \1
```

aicioara:
I suggest leave `isNaN()` in and instead use it where it should be used. The following RegEx'es might help

```regex
(\w+) !== \1
```
```regex
\s(\w+) === \1
```
```regex
([^\s]+) !== \1
```

src/utils/domUtils.ts
```diff
@@ -151,10 +151,7 @@ export module Utils {
     function _parseStyleValue(style: CSSStyleDeclaration, property: string): number {
       var value = style.getPropertyValue(property);
       var parsedValue = parseFloat(value);
-      if (parsedValue !== parsedValue) {
-          return 0;
-      }
-      return parsedValue;
+      return Utils.Math.isNaN(parsedValue) ? 0 : parsedValue;
```
aicioara: I would have preferred `return parsedValue || 0`.

bluong: I think I like that better as well.  Let's make the small adjustment

bluong:
I think NOQE.  There is essentially no functionality changes, just some rerouting to functionality changes

PR #2293, opened 2015-06-10
[Utils.DOM] - Clean dom utils

bluong:
Last minute cleanup of the DOM utils module.

API Changes:
* `Utils.DOM.POLYFILL_TIMEOUT_MILLISECONDS` has been renamed to `Utils.DOM.SCREEN_REFRESH_RATE_MILLISECONDS`
* `Utils.DOM.boxesOverlap()` has been renamed to `Utils.DOM.clientRectsOverlap()`
* `Utils.DOM.boxIsInside()` has been renamed to `Utils.DOM.clientRectInside()`

Link to #2272


src/utils/domUtils.ts
```diff
@@ -20,13 +20,13 @@ export module Utils {
       return bbox;
     }
 
-    export var POLYFILL_TIMEOUT_MILLISECONDS = 1000 / 60; // 60 fps
+    export var REQUEST_ANIMATION_FRAME_TIMEOUT_MILLISECONDS = 1000 / 60; // 60 fps
```
aicioara: Can we shorten it in any way?

aicioara: maybe drop MILLISECONDS?

src/utils/domUtils.ts
```diff
@@ -20,13 +20,13 @@ export module Utils {
...
 
     export function requestAnimationFramePolyfill(fn: () => void) {
```
aicioara: What is a `Polyfill` do you think this is an appropriate name now that the constant is named differently?

src/utils/domUtils.ts
aicioara: We were arguing this should not be exported? I could not find a way not to export it, unless actually using the `requestAnimationFramePolyfill` itself

aicioara:

src/utils/domUtils.ts
jtlan: I think keeping the old name is better, since the *actual* `requestAnimationFrame` timeout could be different.

src/utils/domUtils.ts
```diff
@@ -64,7 +64,7 @@ export module Utils {
       return selection;
     }
 
-    export function boxesOverlap(boxA: ClientRect, boxB: ClientRect) {
+    export function clientRectOverlap(boxA: ClientRect, boxB: ClientRect) {
```
jtlan: `clientRectsOverlap`?

src/utils/domUtils.ts
bluong: Based on the first google result of "polyfill" you get...

"A polyfill, or polyfiller, is a piece of code (or plugin) that provides the technology that you, the developer, expect the browser to provide natively." - https://remysharp.com/2010/10/08/what-is-a-polyfill

Which basically says that the word polyfill here means almost nothing.  Our `Map`s by this definition are I think polyfills and thus having the variable name `POLYFILL_TIMEOUT_MILLISECONDS` I think is really not useful

src/utils/domUtils.ts
bluong: `RenderPolicies.Timeout` uses this value as the default value, and that mainly is the thing that's blocking from hiding this variable.  After thinking about it, I think it's ok to expose this variable

src/utils/domUtils.ts
bluong: I went for this approach since I was thinking "does clientRect 1 overlap clientRect 2?".  That said, your proposal works just as well.

What do you think?

jtlan: `clientRectsOverlap()` makes more sense to me. Also avoids suggesting that it returns the overlap.

src/utils/domUtils.ts
jtlan: >Which basically says that the word polyfill here means almost nothing.  Our `Map`s by this definition are I think polyfills and thus having the variable name `POLYFILL_TIMEOUT_MILLISECONDS` I think is really not useful

I don't see how the first part follows from the second part here.

For that matter, is there a reason this needs to be exported?

bluong: `POLYFILL_TIMEOUT_MILLISECONDS` can be read as the "timeout for a polyfill in milliseconds".  However, this isn't the timeout for just a polyfill, it is the timeout for the requestAnimationFrame polyfill.

bluong: It needs to be exported since the `Timeout` RenderPolicy uses it as the default timeout

aicioara:
Suggestion `REQUEST_ANIMATION_FRAME_TIMEOUT_MILLISECONDS` -> `ANIMATION_FRAME_TIMEOUT_MILLISECONDS`. No verb.

bluong:
Renames. NOQE

jtlan:
Docs?

src/utils/domUtils.ts
```diff
@@ -64,30 +100,50 @@ export module Utils {
...
 
-    export function boxIsInside(inner: ClientRect, outer: ClientRect) {
+    /**
+     * Checks if the first ClientRect is inside the second.
+     * 
+     * @param {ClientRect} clientRectA The first ClientRect
+     * @param {ClientRect} clientRectB The second ClientRect
+     * @returns {boolean} If the first ClientRect is inside the second.
+     */
+    export function clientRectInside(clientRectA: ClientRect, clientRectB: ClientRect) {
```
jtlan: Why did you rename these from `inner` and `outer`? I think the old names were clearer.

jtlan:
[-1] param names seem less clear now.

src/utils/domUtils.ts
bluong: Mainly to reflect the docs.  The `inner` one is not really confirmed to be in the second one unless the function returns true.

Do you think it should be `innerClientRect`?

jtlan: No, because people using Typescript should have tooling available to help them out there.

Javascript consumers would probably be looking at the docs or the d.ts file, otherwise the parameter name wouldn't be available to help them anyway.

PR #2294, opened 2015-06-10
[animatorRefactor] Forgoten animator quicktests fix

aicioara:


bluong:
We may not want to do this until we merge the code into master.  @jtlan When do you think is the best timing to remove the guards?

aicioara:
Hmm, I always removed the guards before committing to `develop` but I can see your point

aicioara:
@crmorford thoughts?

jtlan:
Wait for merge to **`master`**, so we can diff against it during the regression pass.

PR #2296, opened 2015-06-11
Fix destroy() method on Group

jtlan:
The list of `Component`s in the implementation of `_forEach()` should be a shallow copy.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -3399,7 +3399,7 @@ var Plottable;
                 components.forEach(function (c) { return _this.append(c); });
             }
             Group.prototype._forEach = function (callback) {
-                this._components.forEach(callback);
+                this.components().forEach(callback);
             };
             /**
              * Checks whether the specified Component is in the Group.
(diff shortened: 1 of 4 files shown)


bluong:
Provide QE with how to test this

jtlan:
The QE test would pretty much be identical to the unit test: `destroy()` should destroy the sub-`Component`s of `Group.

Fiddle: http://jsfiddle.net/2kqcyed2/

crmorford:
Why did this fiddle destroy c1 (http://jsfiddle.net/2kqcyed2/1/) but not c2?

jtlan:
To explain:
If `destroy()` didn't operate on a shallow copy of the array, the list of `Component`s would mutate during the `forEach()` loop, similar to the following code:
```Javascript
arr = [0,1,2,3,4,5,6,7,8,9];
arr.forEach(function(d, i) { console.log(d); arr.splice(i,1); })
```

Using a shallow copy fixes the problem:
```Javascript
arr.slice().forEach(function(d, i) { console.log(d); arr.splice(i,1); })
```

PR #2297, opened 2015-06-11
[Plot] - getAllSelections -> selections

### LINKED ISSUE #2199: [Plot] - `getAllSelections()` may be redundant

bluong:
API Changes:
* `plot.getAllSelections()` has been renamed to `plot.selections()`

Closes #2199


jtlan:
Simple find-replace. [NOQE]?

PR #2298, opened 2015-06-11
[Drawer] - Massive cleanup

### LINKED ISSUE #2232: Fix Drawer inheritance structure

bluong:
Many of the protected functions have been made private after finding that most of the logic in the subclasses can be folded into the parent class.

API Changes:
* `drawer._svgElementName` has been added to specify what type of svg element should be appended per datum onto the DOM
* `drawer._applyDefaultAttributes()` has been added if subclasses want to apply their own default stylings / defaults to the selection as it's being binded to data
* `drawer._enterData()` has been privitized and renamed since we found that all `Drawer`s have a consistent way to binding data.  They take the selection, bind the data, create new nodes as needed, and remove old nodes as needed.
* `drawer._drawStep()` has been privitized since we found that all `Drawer`s have a consistent way of drawing.  They pre-apply things that should not be animated (namely color), use the animator to animate the step, and then apply the enforced class name if one exists.
* `drawer._numberOfAnimationIterations()` has been removed since we found that it was essentially `data.length` for everything except `Drawers.Line` and `Drawers.Area`.  With a bit of implementation tinkering, this is no longer the case.
* `Drawers.Element` has been removed since we found that after the cleanup, there were no overrides in the class and semantically there really was no difference.  In accordance, subclasses of `Drawers.Element` now subclass from `Drawer`.  The hierarchy now is almost perfectly flat.
* `Drawers.Line` and `Drawers.Area` no longer expose their class strings.  Mainly to keep consistent with other `Drawer`s like `Drawers.Arc`

QE: This is mainly a regression check, unless you want to delve into `Drawer`s since I basically cleaned up a bunch of the protected functions, so extensibility is a lot cleaner.  Focus the attention on essentially anything dealing with drawing onto the DOM, including animations.

Fixes #2232


src/drawers/drawer.ts
```diff
@@ -76,12 +87,16 @@ export module Drawers {
...
-    protected _numberOfAnimationIterations(data: any[]): number {
-      return data.length;
+    private _drawStep(step: Drawers.AppliedDrawStep) {
+      var selection = this._selection();
+      var colorAttributes = ["fill", "stroke"];
+      colorAttributes.forEach((colorAttribute) => {
+        if (step.attrToAppliedProjector[colorAttribute] != null) {
+          selection.attr(colorAttribute, step.attrToAppliedProjector[colorAttribute]);
+        }
+      });
```
jtlan: Why these changes? What if someone wants the colors to animate?

src/drawers/drawer.ts
```diff
@@ -67,8 +69,17 @@ export module Drawers {
...
-    protected _enterData(data: any[]) {
-      // no-op
+    private _setSelectionData(data: any[]) {
+      var dataElements = this._selection().data(data);
+      dataElements.enter().append(this._svgElementName);
+      dataElements.exit().remove();
+      this._setDefaultAttributes(dataElements);
+    }
+
+    protected _setDefaultAttributes(selection: d3.Selection<any>) {
```
jtlan: Should be `Selection<void>`, since we're not doing anything with the data on the `Selection` (compare a callback that returns `void` to signal that we won't be using the return value).

src/drawers/drawer.ts
```diff
@@ -76,12 +87,16 @@ export module Drawers {
...
-
-    protected _numberOfAnimationIterations(data: any[]): number {
-      return data.length;
+    private _drawStep(step: Drawers.AppliedDrawStep) {
+      var selection = this._selection();
+      var colorAttributes = ["fill", "stroke"];
+      colorAttributes.forEach((colorAttribute) => {
+        if (step.attrToAppliedProjector[colorAttribute] != null) {
+          selection.attr(colorAttribute, step.attrToAppliedProjector[colorAttribute]);
+        }
```
jtlan: We might consider (later) moving this logic to the `Plot`, in case users want the `stroke` or `fill` to animate.

src/drawers/drawer.ts
```diff
@@ -67,8 +69,17 @@ export module Drawers {
      *
      * @param{any[]} data The data to be drawn
      */
-    protected _enterData(data: any[]) {
-      // no-op
+    private _setSelectionData(data: any[]) {
```
jtlan: maybe `bindSelectionData()`? "set" seems like it should be reserved for simple setters.

src/drawers/drawer.ts
jtlan: Similarly, "`applyDefaultAttributes()`" might be more appropriate than using "set", which is probably better reserved for simple setters.

jtlan:
So the base `Drawer` produces `<g>`s? Why?

bluong:
Since there was a lot of logic movement upstairs into the `Drawer` level, there are some defaults that I needed to place at the `Drawer` level.

One example is the _svgElementName.  We could have `Drawer` be perfectly abstract and have the `_svgElementName` be empty string or so, but then we have a lot of tests that test on the base `Drawer` class directly.  So, instead I tried to go for what's possibly the most generic svg element name and I got "g".

Thus, I'm faced with two options:  Change the tests and override `_svgElementName` every time, or guess at a generic _svgElementName which hopefully is valid like "g"

src/drawers/drawer.ts
bluong: `Selection<void>` seems incorrect since the data bound to the selection is actually typed any

bluong: Imagine the scenario where an extender of `_setDefaultAttributes` wanted to set attributes based on the data

aicioara:
Regarding the `<g>` issue. How big is the change for overriding `_svgElementName` all the time? It sounds like a better idea to me. We should not bend functionality to accommodate the tests. As far as I see, the drawer test are 3 `it()`s. It sounds more correct to me for those classes that extend `Drawer` and forget to override the method.

bluong:
The default `selector()` returning the `svgElementName` makes a lot of sense to me as it applies to almost all drawers, unless a drawer were to draw a group inside group or something odd like that.

bluong:
I do agree withthe fact that the implementation of `Drawer` should not be bound to the fact that tests have to do more to get a valid drawer.

However, the fact that we use the abstract `Drawer`s begs the question to me if we can actually find a generic element to use.  If there wasn't a generic element to use, I would wholly agree with your sentiment, but since "g" is so generic / flexible, what is wrong with using "g"?

jtlan:
Is there a significant difference between the `Line` and `Area` `Drawer`s any more?

bluong:
@jtlan Only the class that's applied to both of them and the default stylings placed on their nodes

jtlan:
Is it possible to combine `Line` and `Area` then?

bluong:
The default stylings is what separates the two

jtlan:
What if we made the styling configurable, and the `Plot`s set the default stylings?

bluong:
How would this look in practice?

From the `Plots.Area` perspective, if it had a single `Drawer` to work for both area and line, it would have to take the drawer, set it to the area config, then draw the area.  Then, it would grab the same drawer again, set it to the line config, then draw the line.  The problem here is that the configuration at the end would be set to the line configuration which seems wrong to me.

crmorford:
Gonna wait for travis to pass before QEing

crmorford:
Didn't wait on travis. Animations all look fine, and I did a regression pass.

PR #2305, opened 2015-06-11
Clean up AppliedProjector types.

jtlan:
These were super confusing (particularly with `Accessor` and `Projector` already running around), so I'm clearing them out.

### DIFF
--- a/src/animators/animator.ts
+++ b/src/animators/animator.ts
@@ -7,13 +7,15 @@ module Plottable {
      * Applies the supplied attributes to a d3.Selection with some animation.
      *
      * @param {d3.Selection} selection The update selection or transition selection that we wish to animate.
-     * @param {AttributeToAppliedProjector} attrToAppliedProjector The set of
-     *     AppliedProjectors that we will use to set attributes on the selection.
-     * @return {any} Animators should return the selection or
-     *     transition object so that plots may chain the transitions between
-     *     animators.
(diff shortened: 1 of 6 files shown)


src/animators/animator.ts
```diff
@@ -7,13 +7,15 @@ module Plottable {
...
+     * @param {{ [attr: string]: (datum: any, index: number) => any; })} attrToAppliedProjector
+     *     An object that maps attributes to functions that generate attribute values.
+     * @return {any} Animators should return the Selection or
+     *     Transition object so that plots may chain the transitions between
+     *     Animators.
      */
-    animate(selection: d3.Selection<any>, attrToAppliedProjector: AttributeToAppliedProjector): d3.Selection<any> | d3.Transition<any>;
+    animate(selection: d3.Selection<any>,
+            attrToAppliedProjector: { [attr: string]: (datum: any, index: number) => any; }
+           ): d3.Selection<any> | d3.Transition<any>;
```
aicioara: Is this line needed?

aicioara: (the explicit return type)

bluong: Which line?  The output types?

aicioara:
I personally don't think this is clearer. I would have preferred the original version

jtlan:
We then have to explain what an `AppliedProjector` is though...

PR #2307, opened 2015-06-11
Component docs

jtlan:


src/components/component.ts
```diff
@@ -470,7 +480,14 @@ module Plottable {
       return this;
     }
 
+    /**
+     * Gets the parent ComponentContainer for this Component.
+     */
     public parent(): ComponentContainer;
+    /**
+     * Sets the parent ComponentContainer for this Component.
+     * The call will fail if the parent does not contain this Component.
```
aicioara: Fail how?

src/components/group.ts
```diff
@@ -69,6 +69,10 @@ export module Components {
       return this._components.slice();
     }
 
+    /**
+     * Adds a Component to this Group.
+     * The added Component will be rendered on top of Components already in the Group.
```
aicioara: Not sure if the concept of `top` is proper here. I believe `above` (if that is the case) is mor appropriate

src/components/component.ts
jtlan: An `Error` will be thrown. I can specify that.

PR #2309, opened 2015-06-12
[Legend] - rename symbolFactoryAccessor -> symbol

bluong:
API Changes:
* `legend.symbolFactoryAccessor()` has been renamed to `legend.symbol()`

QE: Simple rename.  Should not need QE.


aicioara:
What is the reasoning behind this? The method still requires an accessor

bluong:
Similar reasoning to why `Plots.Scatter` has a `symbol` endpoint.  Also, the word accessor here is not correct.  It doesn't contain the `Dataset` parameter

PR #2311, opened 2015-06-12
[Plot] - _isVisibleOnPlot -> _visibleOnPlot

bluong:
We're currently consistent on trying to shorten our variable/function names where applicable (removing "get" / "set" / "is" in front where we can).  Since I think `_visibleOnPlot()` conveys the meaning of `_isVisibleOnPlot()` and it's more consistent with our vocabulary in our codebase, we should change the name.

API Changes:
* `plot._isVisibleOnPlot()` -> `plot._visibleOnPlot()`

QE: Protected function rename.  Should not require QE

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2350,7 +2350,7 @@ declare module Plottable {
          * @returns {Plots.PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
          */
         entityNearest(queryPoint: Point): Plots.PlotEntity;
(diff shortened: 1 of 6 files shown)


aicioara:
I disagree, the new name looks like it could act like a setter as well.

`is` is important to infer the fact that it is returning a boolean

aicioara:
Do we do this anywhere else in the codebase?

aicioara:
I'd prefer 2 characters + clarity over ambiguity

PR #2312, opened 2015-06-12
[Plot] - split animated() to enableAnimation()/disableAnimation()/willAnimate()

bluong:
Functions that take in boolean parameters usually can be reworked so that it is significantly more readable / easier to use.

API Changes:
* `plot.animated()` has been removed in favor of the following API endpoints.
* `plot.enableAnimation()` has been added to enable animation on plots.
* `plot.disableAnimation()` has been added to disable animation on plots.
* `plot.willAnimate()` has been added to allow checking if the plot will animate.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2302,13 +2302,23 @@ declare module Plottable {
         protected _generateAttrToProjector(): AttributeToProjector;
         renderImmediately(): Plot;
         /**
-         * Returns whether the plot will be animated.
+         * Enables animation on the plot.
+         *
+         * @returns {Plot} The calling Plot.
          */
-        animated(): boolean;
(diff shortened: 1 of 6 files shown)


jtlan:
I don't think this is a change we should make. Plottable uses combined getter-setters for methods that set flags, even if they're booleans. Compare:
```Typescript
Axis.showEndTickLabels();
Plots.Bar.labelsEnabled();
Interactions.Drag.constrainedToComponent();
SelectionBoxLayer.boxVisible();
DragBoxLayer.resizable();
```

bluong:
I personally think that we should change the boolean parameter functions but this is way too much of a change to make right now.  Let's have the discussion about boolean parameter functions later

PR #2314, opened 2015-06-12
Compress computeStackOffsets() down into a single method.

jtlan:
A whole bunch of shimming was being done to satisfy d3's stack layout,
but we didn't even need all its power.

**[-2]**-ing myself since this is a demo for @aicioara. @aicioara, you can merge this if you want to pick up my changes.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -292,6 +292,7 @@ declare module Plottable {
                 value: number;
                 offset?: number;
             };
+            function normalizeKey(key: any): string;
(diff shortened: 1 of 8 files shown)


jtlan:
One thing to note is that after this change, the `key` property in `StackedDatum` is no longer needed.

aicioara:
This looks neat. I believe I'll integrate.

aicioara:
I just fast-forwarded my branch to this as I think it looks neater. I would like to somehow not need a keyAccessor for the `computeStackExtent()`

PR #2316, opened 2015-06-12
fixed overlaying tests

aicioara:


quicktests/overlaying/tests/realistic/baseball.js
```diff
@@ -56,16 +56,16 @@ function run(svg, data, Plottable) {
   var squareFactory = Plottable.SymbolFactories.square();
   var circleFactory = Plottable.SymbolFactories.circle();
 
-  legend.symbolFactoryAccessor(function (d, i) {
+  legend.symbol(function (d, i) {
```
jtlan: We should be using the `typeof` pattern so we can diff against **`master`**.

Issue #2265, opened 2015-06-09
Documentation


aicioara:
We will need to do a documentation sweep as a large number of our exposed stuff is not well documented.

For now, we should aim for basic documentation on all of our public methods and once we have a basic sweep, we can refine this to have more nicely worded documentation.

Modules that need to be documented:

**please do not edit description, just check things on/off and comment as a separate comment**

- [x] Animators
- [x] Axes
- [x] Plots
- [x] Components
- [x] Utils (@aicioara)
- [x] Scales
- [x] Drawers
- [x] Interactions (@jtlan)
- [x] Dispatchers (@jtlan)
- [x] Misc (The stuff in the core directory containing like Formatters/Dataset/RenderController)

aicioara:
the branch `docs` is aimed for this

aicioara:
I'll take `Utils`

bluong:
I got `Plots`

jtlan:
`Dispatcher`s and `Interaction`s check out.

aicioara:
Taking Animators

aicioara:
Taking `Scales`

aicioara:
`Scales` are fine

jtlan:
Checking out `Axes`.

aicioara:
Taking `Components`

jtlan:
`Axes` check out.

bluong:
Getting `Drawer`s

aicioara:
`Components` are fine

jtlan:
Going after `misc`.

aicioara:
Ok, so that should be all

jtlan:
`misc` checks out. Done?

aicioara:
Waiting for my build to pass

aicioara:
It merged

jtlan:
We still need to merge `docs` into `develop`.

PR #2319, opened 2015-06-12
Utils docs

aicioara:
Part of #2265 
Did not know what to say for `colorUtils.colorTest()`

### DIFF
--- a/src/utils/clientToSVGTranslator.ts
+++ b/src/utils/clientToSVGTranslator.ts
@@ -8,9 +8,8 @@ export module Utils {
     private _measureRect: SVGElement;
 
     /**
-     * Returns the singleton (per document) ClientToSVGTranslator.
-     *
-     * It is recommended to use this method instead of the explicit constructor to ensure singularity.
+     * Returns the ClientToSVGTranslator for the <svg> containing elem.
+     * If one already exists on that <svg>, it will be returned; otherwise, a new one will be created.
      */
     public static getTranslator(elem: SVGElement): ClientToSVGTranslator {
       var svg = Utils.DOM.boundingSVG(elem);
--- a/src/utils/clientToSVGTranslator.ts
+++ b/src/utils/clientToSVGTranslator.ts
@@ -7,6 +7,11 @@ export module Utils {
     private _svg: SVGElement;
     private _measureRect: SVGElement;
 
+    /**
+     * Returns the singleton (per document) ClientToSVGTranslator.
+     *
+     * It is recommended to use this method instead of the explicit constructor to ensure singularity.
+     */
     public static getTranslator(elem: SVGElement): ClientToSVGTranslator {
       var svg = Utils.DOM.boundingSVG(elem);
 
--- a/src/utils/colorUtils.ts
+++ b/src/utils/colorUtils.ts
@@ -27,6 +27,14 @@ export module Utils {
       return hsl.rgb().toString();
     }
 
+    /**
+     * Gets the Hex Code of the color resulting by applying the className CSS class to the
+     * colorTester selection. Returns null if the tester is transparent.
+     *
+     * @param {d3.Selection<void>} colorTester The d3 selection to apply the CSS class to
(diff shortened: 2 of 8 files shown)


src/utils/clientToSVGTranslator.ts
```diff
@@ -7,6 +7,11 @@ export module Utils {
     private _svg: SVGElement;
     private _measureRect: SVGElement;
 
+    /**
+     * Returns the singleton (per document) ClientToSVGTranslator.
```
jtlan: Not quite right, since it's one per `<svg>`. Better might be
```
Returns the ClientToSVGTranslator for the <svg> containing elem.
If one already exists on that <svg>, it will be returned; otherwise, a new one will be created.
```

See `Dispatcher.Mouse`'s documentation.

src/utils/colorUtils.ts
```diff
@@ -18,6 +18,10 @@ export module Utils {
       return l1 > l2 ? l1 / l2 : l2 / l1;
     }
 
+    /**
+     * Returns a brighter copy of this color. Each channel is multiplied by 0.7 ^ -factor.
+     * Channel values are capped at the maximum value of 255, and the minimum value of 30.
```
jtlan: I'm not sure it's worth describing implementation details, since that makes the docs fragile to functionality changes.

src/utils/mathUtils.ts
```diff
@@ -75,6 +75,10 @@ export module Utils {
       return typeof n === "number" && !Plottable.Utils.Math.isNaN(n) && isFinite(n);
     }
 
+    /**
+     * Generates an array of consecutive, strictly increasing numbers
+     * in the range [start, stop) at a step distance of each other
```
jtlan: "separated by step" rather than "at step distance of each other"

src/utils/mathUtils.ts
```diff
@@ -89,6 +93,13 @@ export module Utils {
       return range;
     }
 
+    /**
+     * Returns the square of the distance between two points
+     *
+     * @param p1
+     * @param p2
+     * @return dist(p1, p2)^2
```
jtlan: Incorrect docs here. It should be:
```
/**
 * @param {number} p1
 * @param {number} p2
 * @returns {number}
 */
```

However, I think this method's name is very self-explanatory, so it'd be better just not to have any docs on this method at all.

jtlan:
[-1] Some clarification, and removing docs on self-documenting methods. In particular, `Set` shouldn't need any docs at all -- users should know how a set works, and if not there's an ES6 spec for it.

src/utils/colorUtils.ts
aicioara: This is a copy + paste from the d3 description

src/utils/mathUtils.ts
aicioara: Yes, although if you didn't get it means I was too subtle. [ ] generally denote closed interval, ( ) generally denote open interval

aicioara:
Well, we wrote them already. And it is consistent with the rest of the documentation. I suggest leave them in.

aicioara:
> jtlan added a note 11 minutes ago
> "separated by step" rather than "at step distance of each other". Is the close-paren intentional?

> aicioara added a note 3 minutes ago
> Yes, although if you didn't get it means I was too subtle. [ ] generally denote closed interval, ( ) generally denote open interval

src/utils/mathUtils.ts
```diff
@@ -75,6 +75,10 @@ export module Utils {
       return typeof n === "number" && !Plottable.Utils.Math.isNaN(n) && isFinite(n);
     }
 
+    /**
+     * Generates an array of consecutive, strictly increasing numbers
+     * in the range [start, stop) step by step
```
jtlan: ", separated by step"

jtlan:
`colorTest` is only used in `Scales.Color`. We could consider inlining it.

aicioara:
Opened https://github.com/palantir/plottable/issues/2320

PR #2322, opened 2015-06-12
[Plots] - Adding docs

bluong:
Mainly peeling the scales from the constructor docs and adding constructor docs to inform what the class is

### DIFF
--- a/src/components/plots/piePlot.ts
+++ b/src/components/plots/piePlot.ts
@@ -11,8 +11,6 @@ export module Plots {
     private _endAngles: number[];
 
     /**
-     * A Pie Plot draws a circle separated into sectors in accordance with the sector values in the data.
-     * 
      * @constructor
      */
     constructor() {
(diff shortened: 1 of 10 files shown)


src/components/plots/stackedAreaPlot.ts
```diff
@@ -11,9 +11,9 @@ export module Plots {
     private _baselineValueProvider: () => number[];
 
     /**
+     * A StackedArea Plot draws Areas stacked on top of each other.
```
jtlan: Seems obvious enough that we probably don't need this.

src/components/plots/piePlot.ts
```diff
@@ -11,6 +11,8 @@ export module Plots {
     private _endAngles: number[];
 
     /**
+     * A Pie Plot draws a circle separated into sectors in accordance with the sector values in the data.
```
jtlan: Probably not necessary to document this one either.

PR #2328, opened 2015-06-12
Style final touches

### LINKED ISSUE #1942: [Style] Final touches - indentation, folder structure and method order
Final touches:
- [x] indentation screen shot 2015-04-29 at 5 29 05 pm
- [x] Plots should not be inside the components folder - as per module hierarchy
- [ ] Public first, protected after. They are so mixed up in the plottable.d.ts file. I think it makes sense to move protected after public. Private is not so much a big deal and it might make sense to keep them.

aicioara:
Closes #1942

### DIFF
--- a/src/animators/easingAnimator.ts
+++ b/src/animators/easingAnimator.ts
@@ -1,195 +1,195 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export module Animators {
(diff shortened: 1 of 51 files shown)


aicioara:
Dropping this as per a Friday discussion with @bluong

PR #2329, opened 2015-06-12
File shuffle

aicioara:
Part of #1942

Basically, mimic the module structure as much as possible:
- Plots no longer part of the components folder
- Axes no longer part of the components folder
- Test folders to match the avoce
- `ComponentGroupTests.ts` -> `groupTests.ts`
- some tests were in the wrong folders.


bluong:
Not sure about this change.  `Plot`s and `Axes` are indeed components and the main reason why the module hierarchy would suggest not is mainly because of convenience.

rcchen:
+1 to @bluong, it seems more correct to have them under the `Components` module

aicioara:
I disagree with that. The folder structure is mainly helping to find files. If the user knows what class he/she is looking for, they most definitely know the full module name, and having the folder structure imitate that will aid the process

PR #2330, opened 2015-06-15
Indentation

aicioara:
Part of #1942

As per a conversation with @bluong, we have two options
- Indent everything properly
- Consider `module Plottable {` to be an exception

The advantage of the former is that of consistency. The advantage of the latter is horizontal space saving. Also, as we can see, every single file begins with `module Plottable {`. This suggests that probably the same behavior can be achieved through tooling. I looked into it for a while and could not figure out how to wrap everything in a module, but if we manage to do that, then the change will be super easy, just delete 2 lines.

Either way we go, we need to be consistent across files.

### DIFF
--- a/src/animators/animator.ts
+++ b/src/animators/animator.ts
@@ -2,26 +2,26 @@
 
 module Plottable {
 
-  export interface Animator {
-    /**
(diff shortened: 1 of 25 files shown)


aicioara:
[NOQE] as no JS changes

PR #2331, opened 2015-06-15
Clean overlaying tests

aicioara:
There were a bunch of overlaying tests (in particular the `html` and the `js` folder) which were not used anymore. I assume the logic has been ported to the `overlaying` tests folder. These tests were outdated and using pre 1.0.0 API. 

Were we planning to restore these? It appeared to me that we are only using the overlaying anyway.

**Also:** I was careful not to affect the existing overlaying tests. Tested and they all still work.

**Reasoning behind the change:** Name clashes with existing overlaying tests. So many times I was editing the wrong file.


aicioara:
@crmorford Were you using these tests anymore?

jtlan:
These were the non-overlaying tests. We had meant to delete them a while ago and just completely forgot about it.

bluong:
Github I didn't press to remove +1 ....

PR #2332, opened 2015-06-15
Gridlines on Scales.Category

### LINKED ISSUE #1289: Gridlines on Scale.Category

aicioara:
Closes #1289

Consider the scenario of scatter plot + gridlines on category scale. It should be valid IMO. 

**Fiddle**: http://jsfiddle.net/haku8pmk/1/
**QE Notes:** Before, using gridlines with a Category Scale, just threw an error (watch console). Now it should work as expected.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2120,10 +2120,10 @@ declare module Plottable {
         class Gridlines extends Component {
             /**
              * @constructor
-             * @param {QuantitativeScale} xScale The scale to base the x gridlines on. Pass null if no gridlines are desired.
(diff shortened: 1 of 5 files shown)


jtlan:
We probably want to get some details about who this was for and why they needed it. Part of the issue here is that for `Bar` `Plot`s, people might want the lines to come out between the ticks or something.

aicioara:
I really doubt we can remember something from 2014. Especially since the issue does not seem well documented. My suggestion is won't fix until someone actually requests this feature

bluong:
In retrospect, we should have documented that issue better...

aicioara:
Closing... Issue is too old and poorly documented. If the feature is needed, it will be requested again and we can refer to this.

PR #2337, opened 2015-06-15
Map and Set performance improvements

aicioara:
As part of the API refactor, we introduced the ES6 shims for Map and Set. Also, we decided not to expose the d3 map/set anymore. The problem is that our Map and Set were not efficient.

However, Chrome and FF support ES6 Map / Set for quite a while, hence we can leverage their power and cut down O(n) to O(log n) / O(1).

[Browser Support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map):
```
Chrome: 38
Firefox: 25
IE: 11
Opera: 25
Safari: 7.1
```
For reference, [latest version](http://browsehappy.com/) of each browser at the time of writing is:
```
Chrome: 43
Firefox: 38
IE: 11
Opera: 30
Safari: 8
```

**Notice** The `size` field on the Set is no longer read-only. The reason for that is 25% time reduction (from 90-100 ms down to 15-25 ms) on the following test: http://jsfiddle.net/jto18oyh/2/. It is not excellent, but since a set is a core component and performance is important, I am in favor for this change

**Also:** Backward compatible 

---
#### Numbers
On the stacked plot test, the improvement is substantial. (if looking into the JSfiddle, I suggest open the console before clicking on the link)

Before http://jsfiddle.net/r44qLe06/2/
```
t: 11793.249ms
t: 12763.007ms
t: 13495.781ms
t: 14244.572ms
t: 14458.213ms
```
After http://jsfiddle.net/r44qLe06/1/
```
t: 1377.659ms
t: 1260.547ms
t: 1304.650ms
t: 1322.013ms
t: 1646.604ms
```

**QE Notes** Regression pass. IE should behave exactly the same. Firefox and Chrome should be faster everywhere as we improved the speed of our core Set and Map data structures. As a side note, stacking for stacked bar and stacked line should be improved by up to 10x

### DIFF
--- a/src/utils/map.ts
+++ b/src/utils/map.ts
@@ -38,9 +38,9 @@ export module Utils {
       return this;
     }
 
-    public get(key: K): V {
+    public get(key: K) {
       if (this._es6Map != null) {
-        return this._es6Map.get(key);
+        return <V>this._es6Map.get(key);
       }
 
       for (var i = 0; i < this._keyValuePairs.length; i++) {
@@ -51,9 +51,9 @@ export module Utils {
       return undefined;
     }
 
-    public has(key: K): boolean {
+    public has(key: K) {
       if (this._es6Map != null) {
-        return this._es6Map.has(key);
+        return <boolean>this._es6Map.has(key);
       }
 
       for (var i = 0; i < this._keyValuePairs.length; i++) {
@@ -75,9 +75,9 @@ export module Utils {
       });
     }
 
-    public delete(key: K): boolean {
+    public delete(key: K) {
       if (this._es6Map != null) {
-        return this._es6Map.delete(key);
+        return <boolean>this._es6Map.delete(key);
       }
 
       for (var i = 0; i < this._keyValuePairs.length; i++) {
--- a/src/utils/map.ts
+++ b/src/utils/map.ts
@@ -7,12 +7,27 @@ export module Utils {
    * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
    */
   export class Map<K, V> {
-    private _keyValuePairs: { key: K; value: V; }[] = [];
+    private _keyValuePairs: { key: K; value: V; }[];
+    private _es6Map: any;
+
+    public constructor() {
+      if (typeof (<any>window).Map === "function") {
+        this._es6Map = new (<any>window).Map();
+      } else {
+        this._keyValuePairs = [];
+      }
+    }
 
     public set(key: K, value: V) {
       if (Utils.Math.isNaN(key)) {
         throw new Error("NaN may not be used as a key to the Map");
       }
+
+      if (this._es6Map != null) {
+        this._es6Map.set(key, value);
+        return this;
+      }
+
       for (var i = 0; i < this._keyValuePairs.length; i++) {
         if (this._keyValuePairs[i].key === key) {
           this._keyValuePairs[i].value = value;
@@ -23,7 +38,11 @@ export module Utils {
       return this;
     }
 
-    public get(key: K) {
+    public get(key: K): V {
+      if (this._es6Map != null) {
+        return this._es6Map.get(key);
+      }
+
       for (var i = 0; i < this._keyValuePairs.length; i++) {
         if (this._keyValuePairs[i].key === key) {
           return this._keyValuePairs[i].value;
@@ -32,7 +51,11 @@ export module Utils {
       return undefined;
     }
 
-    public has(key: K) {
+    public has(key: K): boolean {
+      if (this._es6Map != null) {
+        return this._es6Map.has(key);
+      }
+
       for (var i = 0; i < this._keyValuePairs.length; i++) {
         if (this._keyValuePairs[i].key === key) {
           return true;
@@ -42,12 +65,21 @@ export module Utils {
     }
 
     public forEach(callbackFn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any) {
+      if (this._es6Map != null) {
+        this._es6Map.forEach(callbackFn, thisArg);
+        return;
+      }
+
       this._keyValuePairs.forEach((keyValuePair) => {
         callbackFn.call(thisArg, keyValuePair.value, keyValuePair.key, this);
       });
     }
 
-    public delete(key: K) {
+    public delete(key: K): boolean {
+      if (this._es6Map != null) {
+        return this._es6Map.delete(key);
+      }
+
       for (var i = 0; i < this._keyValuePairs.length; i++) {
         if (this._keyValuePairs[i].key === key) {
           this._keyValuePairs.splice(i, 1);
--- a/src/utils/map.ts
+++ b/src/utils/map.ts
@@ -66,7 +66,8 @@ export module Utils {
 
     public forEach(callbackFn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any) {
       if (this._es6Map != null) {
-        this._es6Map.forEach(callbackFn, thisArg);
+        var callbackWrapper = (value: V, key: K) => callbackFn.call(thisArg, value, key, this);
+        this._es6Map.forEach(callbackWrapper, thisArg);
         return;
       }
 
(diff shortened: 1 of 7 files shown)


bluong:
Having the `size` field not be readonly is incredibly tricky... as now we have a `Set` that can lie...

src/utils/map.ts
```diff
@@ -7,12 +7,27 @@ export module Utils {
    * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
    */
   export class Map<K, V> {
-    private _keyValuePairs: { key: K; value: V; }[] = [];
+    private _keyValuePairs: { key: K; value: V; }[];
+    private _es6Map: any;
```
bluong: Can we type the `_es6Map`?

aicioara:
I prefer that and performance, to a 4x slowdown tbh. Enforcing it to be read-only is trickier because a new object is created every time, hence garbage collection pressure

src/utils/map.ts
```diff
@@ -7,12 +7,27 @@ export module Utils {
...
+
+    public constructor() {
+      if (typeof (<any>window).Map === "function") {
+        this._es6Map = new (<any>window).Map();
```
bluong: Why is anycasting necessary here?

src/utils/map.ts
aicioara: Looked into `lib.d.ts`. There doesn't seem to be a type for it

src/utils/map.ts
```diff
@@ -7,12 +7,27 @@ export module Utils {
...
+      } else {
+        this._keyValuePairs = [];
+      }
+    }
 
     public set(key: K, value: V) {
       if (Utils.Math.isNaN(key)) {
         throw new Error("NaN may not be used as a key to the Map");
```
bluong: Can we modify the message here to mirror the es6 version?

src/utils/map.ts
```diff
@@ -7,12 +7,27 @@ export module Utils {
...
       }
+
+      if (this._es6Map != null) {
+        this._es6Map.set(key, value);
+        return this;
+      }
```
bluong: I think we should move this statement higher?

bluong:
We should almost certainly submit a pull request to DefinitelyTyped as now we have a bunch of cruft in our codebase regarding any types.  I don't want to block this going through on that pull request, but having strong, clean is important enough that I would put that on high priority after this is done.

src/utils/map.ts
aicioara: Because Map does not exist on type window. Also I can't just call Map() here, because it will resolve scope to Utils.Map

src/utils/map.ts
aicioara: This is ugly. ES6 Map accepts NaN as a valid key. However, we cannot do that because `NaN` is never `===` with `NaN` for the time we do reference equality in the array, hence if we want to fully support the behavior we need to put a hack there to check for NaN

src/utils/map.ts
aicioara: However, if we leave the line there, the two maps will behave the same, and there will be no differences between browsers (IE failing in unexpected ways)

aicioara:
Will do

src/utils/map.ts
bluong: Scope resolution can be fixed by specifying the path to the right symbol.  Where is the ES6 Map located?

aicioara: It is top level. The best way I could find to access it was through `window.Map`. You see any workaround?

src/utils/map.ts
jtlan: Right, we'd have to have:
```Typescript
private static _keysEqual(key1: K, key2: K) {
  return key1 === key2 || (key1 !== key1 && key2 !== key2);
}
```

jtlan:
There's probably definitions for `Set` and `Map` already.

test/utils/mapTests.ts
```diff
@@ -44,7 +44,11 @@ describe("Map", () => {
     map.forEach((value: number, key: string, mp: Plottable.Utils.Map<string, number>) => {
       assert.strictEqual(value, values[index], "Value " + index + " is the expected one");
       assert.strictEqual(key, keys[index], "Key " + index + " is the expected one");
-      assert.strictEqual(mp, map, "The correct map is passed as the third argument");
+      if ((<any>map)._es6Map != null) {
+        assert.strictEqual(mp, (<any>map)._es6Map, "The correct map is passed as the third argument (ES6)");
```
jtlan: Is there a reason we don't just always pass our own `Map`?

jtlan:
It seems like most of our large data-scale cases involve `number`s or `string`s. Can we update our shim to use a simple `Object` as a dictionary for those cases, with a `string`ified key? That would improve perf even in cases without the ES6 `Map` / `Set`.

bluong:
@jtlan I wasn't able to find the `Map` / `Set` definitions in DefinitelyTyped, unfortunately :cry:

aicioara:
@jtlan The IE issue is out of the scope of this PR, hence opened https://github.com/palantir/plottable/issues/2342

Addressed the other issue

src/utils/map.ts
bluong: We should probably do the above, but this is a side issue that already has been issued up

test/utils/setTests.ts
```diff
@@ -10,17 +10,23 @@ describe("Utils", () => {
       var value1 = { value: "one" };
       set.add(value1);
       assert.strictEqual(set.size, 1, "set contains one value");
-      assert.strictEqual((<any>set)._values[0], value1, "the value was added to the set");
+      if ((<any>set)._values != null) {
+        assert.strictEqual((<any>set)._values[0], value1, "the value was added to the set");
+      }
```
bluong: Let's use the public API where possible.  Please use a `forEach` here.

test/utils/setTests.ts
```diff
@@ -10,17 +10,23 @@ describe("Utils", () => {
...
 
       set.add(value1);
       assert.strictEqual(set.size, 1, "same value is not added twice");
-      assert.strictEqual((<any>set)._values[0], value1, "list still contains the value");
+      if ((<any>set)._values != null) {
+        assert.strictEqual((<any>set)._values[0], value1, "list still contains the value");
+      }
```
bluong: Let's use the public API where possible.  Please use a `forEach` here.

test/utils/setTests.ts
```diff
@@ -10,17 +10,23 @@ describe("Utils", () => {
...
 
       var value2 = { value: "two" };
       set.add(value2);
       assert.strictEqual(set.size, 2, "set now contains two values");
-      assert.strictEqual((<any>set)._values[0], value1, "set contains value 1");
-      assert.strictEqual((<any>set)._values[1], value2, "set contains value 2");
+      if ((<any>set)._values != null) {
+        assert.strictEqual((<any>set)._values[0], value1, "set contains value 1");
+        assert.strictEqual((<any>set)._values[1], value2, "set contains value 2");
+      }
```
bluong: Let's use the public API where possible.  Please use a `forEach` here.

aicioara:
Actually removed those calls altogether as .forEach() is tested later

jtlan:
@aicioara, can you explain what sort of testing might need to be done for this? It seems like the unit tests cover this pretty well right?

crmorford:
Before this PR (on chrome) http://jsfiddle.net/v5p1d7vs/3/

aicioara:
Performance testing: Stacking should be affected the most.
Regression pass: everything can be affected as we use the datastructures in lots of places

crmorford:
Bar labels are in the wrong spots
screen shot 2015-06-16 at 5 55 17 pm

crmorford:
Regression looks good!

crmorford:
Will do FF and IE tomorrow morning

crmorford:
FF before http://jsfiddle.net/v5p1d7vs/8/
FF afer http://jsfiddle.net/v5p1d7vs/9/

crmorford:
IE before (not as many data points) http://jsfiddle.net/v5p1d7vs/10/
IE after (about the same) http://jsfiddle.net/v5p1d7vs/11/

PR #2338, opened 2015-06-16
[QuantitativeScale] - Adding `zoomLevel()` endpoint

bluong:
Something I probably should have done a while back...

`Interactions.PanZoom` currently taps into the `QuantitativeScale`'s API in order to determine the new domain of the scale.  Because of this, the zooming idea is not enforced at the `Scale` level.

In other words, unless users wanted to do tricky math, there was no way to zoom into a scale domain without using `Interactions.PanZoom`.  Now, `zoomLevel()` has been exposed such that users can, through code, set the zoom factor of the scale's domain.

EDIT: New functionality - Setting the domain of a `QuantitativeScale` will set the `zoomLevel` back to 1.0.

This is mainly relevant as a precommit for #1639 

API Changes:
* `quantitativeScale.zoomLevel()` has been exposed to get the zoom factor for the scale's domain.
* `quantitativeScale.zoomLevel(zoomLevel: number)` has been exposed to set the zoom factor for the scale's domain.  The scale will then dispatch this update accordingly


jtlan:
Why "`domainZoomFactor`" and not just "`zoomFactor`", or "`zoomLevel`"

jtlan:
Also, why is this API point on the scale? When I change the domain, what happens to the zoom level?

Are there unit tests that lay out how we expect this to work?

bluong:
I was unsure if `zoomFactor` was enough to signal that only the domain'll be affected.  That said, let me try removing the `domain` prefix.

This API point is on the scale, because it should be the scale that is managing the zoom idea, this allows flexibility so that panZoomInteraction is not the only thing that can use the zoom concept.  In close accordance, I am trying to follow a model close to `SelectionBoxLayer` / `DragBoxLayer` where there is a non-interactable object but satisfiable endpoints and an interactable object that just uses the above endpoints to perform interactions.

src/interactions/panZoomInteraction.ts
```diff
@@ -93,12 +93,14 @@ export module Interactions {
       var newCornerDistance = this._cornerDistance();
 
       if (this._xScale != null && newCornerDistance !== 0 && oldCornerDistance !== 0) {
-        PanZoom._magnifyScale(this._xScale, oldCornerDistance / newCornerDistance, oldCenterPoint.x);
-        PanZoom._translateScale(this._xScale, oldCenterPoint.x - newCenterPoint.x);
+        var invertedPX = this._xScale.invert(oldCenterPoint.x);
+        this._xScale.zoomLevel(this._xScale.zoomLevel() * oldCornerDistance / newCornerDistance);
+        PanZoom._translateScale(this._xScale, this._xScale.scale(invertedPX) - newCenterPoint.x);
```
jtlan: Why this change? Do `_xScale.zoomLevel(...)` and `_magnifyScale(...)` not have the same functionality?

bluong: `_magnifyScale(...)` was a private static method to pull out the calculations for calculating the new domain of the scale.  Now that we actually have an endpoint on the scale that will do the calculations inherently, I didn't see the need of `_magnifyScale(...)`.

jtlan: But the call to `_translateScale()` has changed. What's the difference in functionality between  `_xScale.zoomLevel(...)` and `_magnifyScale(...)` that necessitates this change?

bluong: `_magnifyScale(...)` used to take in the center of where the magnification should be.  With the center idea removed from the signature, the center is assumed to be the center of the scale range, meaning that a translation has to be done afterwards to get the scale position to be in the right spot.

Decently important note, based on the usage above + a bunch of searching online, the above is essentially the common solution for this type of problem

jtlan: Right, I understand why a translation has to be done afterwards. But why is it important to save off `invertedPX` before? The previous implementation didn't do that.

bluong: One of the implicit goals in this pull request as you see is to remove the `centerValue` idea from the scale magnification.  With this idea in mind, I needed to somehow remove this `oldCenterPoint.x` parameter.

To do this, first I need to magnify / affect the zoom level first.

Then, I need to translate to correct for the centerValue.

This translation is (you can see a similar calculation in the mousewheeling scenario)
`this._xScale.scale(invertedPX) - oldCenterPoint.x`

In short, the last two lines in this description should probably be...
```typescript
this._xScale.zoomLevel(this._xScale.zoomLevel() * oldCornerDistance / newCornerDistance);
PanZoom._translateScale(this._xScale, this._xScale.scale(invertedPX) - oldCenterPoint.x);
PanZoom._translateScale(this._xScale, oldCenterPoint.x - newCenterPoint.x);
```

However, since translation is a linear transform, I combined the last two translations into one, ending in the result you see here.

jtlan: So basically because the implicit translate in `_magnifyScale()` was removed, you moved the translation to `_translateScale()`? That makes sense.

bluong: Yup

jtlan:
This feels a bit off to me, since now both `domain()` and `zoomLevel()` can be used to set the apparent domain, so you can have the same domain state with multiple model states. For instance, `scale.domain(scale.domain())` now produces a state change, which seems counterintuitive.

It also feels like the API point is in the wrong spot, since it's not immediately obvious to me why a `QuantitativeScale` has a zoom API unless I know there's an `Interaction` that's going to use it.

Finally, what happens when the user calls `autoDomain()`? Does the zoom level get reset?

bluong:
Found that this was probably the incorrect way to go about this feature.  Closing

PR #2340, opened 2015-06-16
[Plots.Pie] Add .entitiesAt() to Plots.Pie

rcchen:
Adds `.entitiesAt(point)` to `Plots.Pie`

Fiddle: http://jsfiddle.net/ogj7qLxw/2/

### DIFF
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -157,6 +157,42 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Returns the PlotEntity at the query point.
+     * 
+     * @param {Point} queryPoint
+     * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found. 
+     */
+    public entityAt(queryPoint: Point): PlotEntity {
+      var bbox = d3.select(this._boundingBox[0][0]);
+      var center = { x: +bbox.attr("width") / 2, y: +bbox.attr("height") / 2 };
+
+      var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
+      var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
+      var angle = Math.acos(-adjustedQueryPoint.y / (1 + radius));
+      if (adjustedQueryPoint.x < 0) {
+        angle = Math.PI * 2 - angle;
+      }
+
+      for (var i = 0; i < this.entities().length; i++) {
+        var entity = this.entities()[i];       
+        var innerRadius = this.innerRadius().accessor(entity.datum, entity.index, entity.dataset);
+        if (this.innerRadius().scale) {
+          innerRadius = this.innerRadius().scale.scale(innerRadius);
+        }
+        var outerRadius = this.outerRadius().accessor(entity.datum, entity.index, entity.dataset);
+        if (this.outerRadius().scale) {
+          outerRadius = this.outerRadius().scale.scale(outerRadius);
+        }        
+        if (this._startAngles[i] <= angle && this._endAngles[i] > angle &&
+            innerRadius < radius && outerRadius > radius) {
+          return this.entities()[i];
+        }
+      }
+
+      return null;
+    }
+
     protected _propertyProjectors(): AttributeToProjector {
       var attrToProjector = super._propertyProjectors();
       var innerRadiusAccessor = Plot._scaledAccessor(this.innerRadius());
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -163,10 +163,8 @@ export module Plots {
      * @param {Point} queryPoint
      * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found. 
      */
-    public entityAt(queryPoint: Point): PlotEntity {
-      var bbox = d3.select(this._boundingBox[0][0]);
-      var center = { x: +bbox.attr("width") / 2, y: +bbox.attr("height") / 2 };
-
+    public entitiesAt(queryPoint: Point): PlotEntity[] {
+      var center = { x: this.width() / 2, y: this.height() / 2 };
       var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
       var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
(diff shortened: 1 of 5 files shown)


jtlan:
Modification to the fiddle to make things a bit easier to see: http://jsfiddle.net/ogj7qLxw/1/

src/plots/piePlot.ts
```diff
@@ -157,6 +157,42 @@ export module Plots {
...
 
+    /**
+     * Returns the PlotEntity at the query point.
+     * 
+     * @param {Point} queryPoint
+     * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found. 
+     */
+    public entityAt(queryPoint: Point): PlotEntity {
+      var bbox = d3.select(this._boundingBox[0][0]);
+      var center = { x: +bbox.attr("width") / 2, y: +bbox.attr("height") / 2 };
```
jtlan: Just use `this.height()/2` and `this.width()/2`.

We should probably have a private `_center()` method.

test/plots/piePlotTests.ts
```diff
@@ -165,6 +165,31 @@ describe("Plots", () => {
         svg.remove();
       });
 
+      it("entityAt with defaults", () => {
+        var click1 = { x: 300, y: 200 };
+        var entity1 = piePlot.entityAt(click1);
+        assert.strictEqual(entity1.datum.value, 5);
```
jtlan: Please use `TestMethods.assertPlotEntitiesEqual()` to check the return value.

src/plots/piePlot.ts
```diff
@@ -157,6 +157,42 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Returns the PlotEntity at the query point.
+     * 
+     * @param {Point} queryPoint
+     * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found. 
+     */
+    public entityAt(queryPoint: Point): PlotEntity {
```
jtlan: For consistency the method should be named `entitiesAt()`, even though we won't ever return more than one slice.

rcchen:
Fiddle updated for the changed API point http://jsfiddle.net/ogj7qLxw/2/

src/plots/piePlot.ts
```diff
@@ -157,6 +157,40 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Returns the PlotEntity at the query point.
+     * 
+     * @param {Point} queryPoint
+     * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found.
```
jtlan: Wrong docs. See `Plots.Bar` for the correct docs.

src/plots/piePlot.ts
```diff
@@ -157,6 +157,40 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Returns the PlotEntity at the query point.
+     * 
+     * @param {Point} queryPoint
+     * @param {PlotEntity} The PlotEntity at the point, or null if no PlotEntity can be found. 
+     */
+    public entitiesAt(queryPoint: Point): PlotEntity[] {
```
jtlan: Better to let the compiler infer the return type.

test/plots/piePlotTests.ts
```diff
@@ -165,6 +165,31 @@ describe("Plots", () => {
...
+        svg.remove();
+      });
+
+      it("entityAt with radius modifications", () => {
+        piePlot.innerRadius(100).render();
+        var click1 = { x: 250, y: 250 };
+        var entity1 = piePlot.entitiesAt(click1);
+        assert.strictEqual(entity1.length, 0, "no entities returned");
+        piePlot.outerRadius(150).render();
+        var click2 = { x: 300, y: 100 };
```
jtlan: It seems like this test would read more logically if this was a click in the same direction as the first one, but closer to the center so it misses the doughnut slice.

test/plots/piePlotTests.ts
```diff
@@ -165,6 +165,31 @@ describe("Plots", () => {
         svg.remove();
       });
 
+      it("entityAt with defaults", () => {
```
jtlan: "`entitiesAt()`". Also what are the defaults? A more descriptive name for the test would be better.

test/plots/piePlotTests.ts
```diff
@@ -165,6 +165,31 @@ describe("Plots", () => {
...
+        var click2 = { x: 200, y: 300 };
+        var entity2 = piePlot.entitiesAt(click2);
+        TestMethods.assertPlotEntitiesEqual(entity2[0], piePlot.entities()[1], "entities are equal");
+        var click3 = { x: 0, y: 0 };
+        var entity3 = piePlot.entitiesAt(click3);
+        assert.strictEqual(entity3.length, 0, "no entities returned");
+        svg.remove();
+      });
+
+      it("entityAt with radius modifications", () => {
```
jtlan: Same naming stuff here. Maybe something like "points within innerRadius() or outside of outerRadius() don't return any entities"?

jtlan:
When updating API points be careful to update the docs and test names too...

src/plots/piePlot.ts
```diff
@@ -157,6 +157,40 @@ export module Plots {
...
+     * Gets the Entities at a particular Point.
+     * 
+     * @param {Point} p
+     * @param {PlotEntity[]} 
+     */
+    public entitiesAt(queryPoint: Point) {
+      var center = { x: this.width() / 2, y: this.height() / 2 };
+      var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
+      var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
+      var angle = Math.acos(-adjustedQueryPoint.y / (1 + radius));
```
jtlan: Why `1+radius` here?

jtlan:
Re-updated for correct signature: http://jsfiddle.net/ogj7qLxw/3/

src/plots/piePlot.ts
rcchen: Since 0 is defined to be the vector 1j, in order to find the angle from that vector to any arbitrary vector, we use the dot product defined as a∙b = ||a|| ||b|| cos(theta). 

Magnitude of the first vector is just 1, magnitude of the second is radius, which gives us (1+r)

Dot product of the two vectors resolves to the negative y value of the adjusted query point, since the first vector only has a non-zero y value.

Other techniques for finding the angle would involve less math and a lot more code, as I would have to set conditional statements for figuring out what quadrant the second vector was in, and adjusting formulas to match.

Issue #1284, opened 2014-11-03
Panzoom Interaction Performance for StackedBar with Auto Bar Width


terekim:
In this repro, I have dataset length of 50, and here you can definitely see performance changes in PanZooming. Compare this with modified auto barwidth plot in v0.34.0.

current -> http://jsfiddle.net/tkim/pboc0st2/36/
old, hacky auto-barwidth -> http://jsfiddle.net/tkim/pboc0st2/37/

aicioara:
Slow version: http://jsfiddle.net/aycxz2ap/1/
Fast version: http://jsfiddle.net/aycxz2ap/2/

aicioara:
I was hoping it will be caught by this https://github.com/palantir/plottable/pull/2337 but it seems like it is not the case. Will investigate further.

aicioara:
It seems like the flush() method has to be improved. In the fast version it takes 13 ms, in the current version it takes 121 ms. Working on that...

aicioara:
`_getBarPixelWidth()` is the bottleneck. In particular it is called about 80+ times during a `flush()` procedure, which does not seem right, especially considering its complexity

aicioara:
Improved this to `10 ms` by caching the `barPixelWidth` value and updating it only where necessary

bluong:
Progress looks good so far :smiley:

PR #2344, opened 2015-06-16
Bar plot performance

### LINKED ISSUE #1289: Gridlines on Scale.Category

aicioara:
Closes https://github.com/palantir/plottable/issues/1284

Improved the call of `flush()` from `132 ms` down to `10 ms` by caching the result of `_getBarPixelWidth()`

Historic Before: http://jsfiddle.net/aycxz2ap/4/
Before: http://jsfiddle.net/aycxz2ap/1/
After: http://jsfiddle.net/aycxz2ap/3/

**QE Notes:** Bar Plot should behave the same. In particular the width of the bars should not behave weirdly. Also The main performance boost is in the PanZoom animation effect (as can be seen in the fiddle)

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -149,14 +149,16 @@ export module Plots {
 
     public addDataset(dataset: Dataset) {
       dataset.onUpdate(this._getBarPixelWidthCallback);
+      super.addDataset(dataset);
       this._getBarPixelWidth();
-      return super.addDataset(dataset);
+      return this;
     }
 
     public removeDataset(dataset: Dataset) {
       dataset.offUpdate(this._getBarPixelWidthCallback);
+      super.removeDataset(dataset);
       this._getBarPixelWidth();
-      return super.removeDataset(dataset);
+      return this;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/plots/barPlot.ts
```diff
@@ -27,6 +27,9 @@ export module Plots {
     private _labelConfig: Utils.Map<Dataset, LabelConfig>;
     private _baselineValueProvider: () => (X|Y)[];
 
+    private _barPixelWidth = 0;
+    private _updateBarPixelWidthCallback: (() => void);
```
bluong: I think? the parenthese for `() => void` is unnecessary?

src/plots/barPlot.ts
```diff
@@ -479,6 +505,7 @@ export module Plots {
      * If the position scale of the plot is a QuantitativeScale, then _getMinimumDataWidth is scaled to compute the barPixelWidth
      */
     protected _getBarPixelWidth(): number {
+
```
bluong: I don't think the whitespace here is necessary

src/plots/barPlot.ts
```diff
@@ -495,14 +522,14 @@ export module Plots {
 
         numberBarAccessorData.sort((a, b) => a - b);
 
-        var barAccessorDataPairs = d3.pairs(numberBarAccessorData);
+        var scaledData = numberBarAccessorData.map((datum: number) => barScale.scale(datum));
```
bluong: Can you remove the `number` type here or no?

src/plots/barPlot.ts
```diff
@@ -518,6 +545,10 @@ export module Plots {
       return barPixelWidth;
     }
 
+    private _updateBarPixelWidth() {
+      this._barPixelWidth = this._getBarPixelWidth();
```
bluong: Why do we need the function wrapping around this line?

src/plots/barPlot.ts
```diff
@@ -553,6 +584,11 @@ export module Plots {
       return { x: x, y: y };
     }
 
+  protected _uninstallScaleForKey(scale: Scale<any, number>, key: string) {
+    scale.offUpdate(this._updateBarPixelWidthCallback);
+    super._uninstallScaleForKey(scale, key);
+  }
```
bluong: Indentation is messed up

src/plots/barPlot.ts
aicioara: Backward compatibility

aicioara: if I add that line in the `_getBarPixelWidth()` it does not really make much sense from a reading stand point. Also, can't rename or remove old function

PR #2345, opened 2015-06-17
[Interactions.PanZoom] - Usage on multiple scales

bluong:
Allowing usage of `Interactions.PanZoom` on multiple scales.  Through the usage of either of the `addXScale()` / `addYScale()` / `xScales()` / `yScales()`, users can install multiple scales to use for the interaction.  In consequence, a panning or zooming will occur on all installed scales at once.

API Changes:
* `panZoomInteraction.addXScale()` has been added to add an x scale to the interaction.
* `panZoomInteraction.addYScale()` has been added to add a y scale to the interaction.
* `panZoomInteraction.removeXScale()` has been added to from an x scale from the interaction.
* `panZoomInteraction.removeYScale()` has been added to from an y scale from the interaction.
* `pandZoomInteraction.xScales()` has been added to get/set the x scales for the interaction.
* `pandZoomInteraction.yScales()` has been added to get/set the y scales for the interaction.

Example JSFiddle:
http://jsfiddle.net/bluong63/j8fgxw0o/2/

### DIFF
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -185,7 +185,15 @@ export module Interactions {
       });
     }
 
+    /**
+     * Gets the x scales for this PanZoom Interaction.
+     */
     public xScales(): QuantitativeScale<any>[];
+    /**
+     * Sets the x scales for this PanZoom Interaction.
+     * 
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
     public xScales(xScales: QuantitativeScale<any>[]): Interactions.PanZoom;
     public xScales(xScales?: QuantitativeScale<any>[]): any {
       if (xScales == null) {
@@ -195,7 +203,15 @@ export module Interactions {
       return this;
     }
 
+    /**
+     * Gets the y scales for this PanZoom Interaction.
+     */
     public yScales(): QuantitativeScale<any>[];
+    /**
+     * Sets the y scales for this PanZoom Interaction.
+     * 
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
     public yScales(yScales: QuantitativeScale<any>[]): Interactions.PanZoom;
     public yScales(yScales?: QuantitativeScale<any>[]): any {
       if (yScales == null) {
@@ -205,6 +221,12 @@ export module Interactions {
       return this;
     }
 
+    /**
+     * Adds an x scale to use for this PanZoom Interaction
+     * 
+     * @param {QuantitativeScale<any>} An x scale to add
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
     public addXScale(xScale: QuantitativeScale<any>) {
       if (this.xScales().indexOf(xScale) === -1) {
         return this;
@@ -213,6 +235,12 @@ export module Interactions {
       return this;
     }
 
+    /**
+     * Removes an x scale that would have been used for this PanZoom Interaction
+     * 
+     * @param {QuantitativeScale<any>} An x scale to add
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
     public removeXScale(xScale: QuantitativeScale<any>) {
       var xScaleIndex = this.xScales().indexOf(xScale);
       if (xScaleIndex === -1) {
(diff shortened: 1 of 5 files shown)


src/interactions/panZoomInteraction.ts
```diff
@@ -169,18 +173,111 @@ export module Interactions {
...
+    public addXScale(xScale: QuantitativeScale<any>) {
+      if (this.xScales().indexOf(xScale) !== -1) {
+        return this;
+      }
+      this._xScales.push(xScale);
+      return this;
+    }
+
+    /**
+     * Removes an x scale that would have been used for this PanZoom Interaction
```
jtlan: "Removes an X scale from this PanZoom Interaction"?

src/interactions/panZoomInteraction.ts
```diff
@@ -8,8 +8,8 @@ export module Interactions {
      */
     private static _PIXELS_PER_LINE = 120;
 
-    private _xScale: QuantitativeScale<any>;
-    private _yScale: QuantitativeScale<any>;
+    private _xScales: QuantitativeScale<any>[];
+    private _yScales: QuantitativeScale<any>[];
```
jtlan: Use a `Set` for both of these...

jtlan:
Switch to using a `Set`. Also, can you make an example fiddle?

src/interactions/panZoomInteraction.ts
bluong: I was highly thinking of using a `Set` but then I run into the situation where I have to turn the `Set` into an array for the getters.  Are we ok with this?

jtlan: Yeah. It pushes the weirdness into the setter/getters, but avoids duplicating the `Set` logic. It also reduces the chance of array-mutation bugs when users operate on the return values of `xScales()`/`yScales()`/=.

crmorford:
When you try to remove a scale that hasn't been added, I'd expect an error. Look at lines 30 and 32 in this fiddle http://jsfiddle.net/j8fgxw0o/17/

crmorford:
How come you can add scales with an array, but only remove them one at a time? 

I tried to use an array in the remove call (line 32), and there was no error, AND the scales didn't get removed. I'd expect one or the other.
 http://jsfiddle.net/j8fgxw0o/18/

crmorford:
In lines 29-34, I tried to make scales to add to the PanZoomInteraction (line 38), but now Pan Zooming in the x direction doesn't work at all.
Not sure if this is a bug in my fiddle, or if I shouldn't be trying this for some reason, or if it's a bug in the PR.
http://jsfiddle.net/j8fgxw0o/19/

bluong:
@crmorford Currently, `remove*Scale()` has no effect if the scale is not on the interaction.  Do you prefer an error or a warning instead of a no-op?

bluong:
@crmorford the `*Scale()` setter is not the only way to set the scales; it is meant for specifically setting the scales to be the input array.  There is also an `add*Scale()` endpoint that is meant to be the parallel to `remove*Scale()`.  Both should take a single scale as an argument.

bluong:
@crmorford The scales that you made on-the-fly are not attached to any component, so at the very least, those scales can't be interacted with physically through the PanZoomInteraction.  In accordance, in line 37, you set the scales to be initially xScale/xScale2, but then you immediately overwrite what the xScales are in the following line.  Thus, xScale/xScale2 are no longer on the PanZoomInteraction.

jtlan:
> The scales that you made on-the-fly are not attached to any component, so at the very least, those scales can't be interacted with physically through the PanZoomInteraction. 

`PanZoom` should be able to manipulate `Scale`s even if they're not powering anything in the DOM, right?

bluong:
> `PanZoom` should be able to manipulate `Scale`s even if they're not powering anything in the DOM, right?

As long as the `Scale` has a domain and a range, attaching it to a `Interactions.PanZoom` will have the desired effect.  It's just that if it's not powering anything in the DOM, then there's nothing visually that will change.

crmorford:
I'd prefer an error when you try to remove a scale that was never added, if it's consistent with the rest of our code. @jtlan, opinions?

crmorford:
Oh okay. I didn't read the documentation, because this seemed really self explanatory to me (so partially my fault), but I think some of this is unintuitive. 

* I didn't expect .xScales to override other calls to .xScales, I thought it'd be additive (most of our stuff feels that way)
* Probably related: it feels weird to be able to add an array but not have a call to remove the same array.

bluong:
The closest example I can find is in `plot.datasets()`.  For that function, calls to further `datasets()` setters would replace the prior calls.

This seems to make sense to me as the semantics of it reads "I'm setting the datasets to be the input datasets".  What do you think?

Also, `plot.removeDataset()` does not report any warning or error if the input dataset is not part of the plot.  Removing the dataset having a warning is indeed helpful if you want to be informed if the dataset was indeed removed, but if the end goal was to ensure that the dataset is not on the plot anymore, maybe the warning is unnecessary.  Not sure on this one.

crmorford:
Okay I'm on board now :)

PR #2346, opened 2015-06-17
[Plots.Pie] Add labels to slices

### LINKED ISSUE #1970: Static Labels on Pie Plots

rcchen:
Turns out that I can't use `arc.centroid()` directly here from d3, but I've implemented an identical technique as a substitute.

image

Fixes #1970 

Fiddle: http://jsfiddle.net/2z4pckh8/1/

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -63,6 +63,11 @@ svg.plottable {
   font-size: 14px;
 }
 
+.plottable .label-area text {
+  font-family: "Helvetica Neue", sans-serif;
+  fill: #32313F;
+}
+
 .plottable .light-label text {
   fill: white;
 }
--- a/plottable.css
+++ b/plottable.css
@@ -64,8 +64,9 @@ svg.plottable {
 }
 
 .plottable .label-area text {
-  font-family: "Helvetica Neue", sans-serif;
   fill: #32313F;
+  font-family: "Helvetica Neue", sans-serif;
+  font-size: 14px;
 }
 
 .plottable .light-label text {
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -9,6 +9,7 @@ export module Plots {
     private static _SECTOR_VALUE_KEY = "sector-value";
     private _startAngles: number[];
     private _endAngles: number[];
+    private _labelsEnabled = false;
 
     /**
      * @constructor
@@ -157,6 +158,29 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Get whether slice labels are enabled.
+     *
+     * @returns {boolean} Whether slices should display labels or not.
+     */
+    public labelsEnabled(): boolean;
+    /**
+     * Sets whether labels are enabled.
+     *
+     * @param {boolean} labelsEnabled
+     * @returns {Pie} The calling Pie Plot.
+     */
+    public labelsEnabled(enabled: boolean): Pie;
+    public labelsEnabled(enabled?: boolean): any {
+      if (enabled === undefined) {
+        return this._labelsEnabled;
+      } else {
+        this._labelsEnabled = enabled;
+        this.render();
+        return this;
+      }
+    }
+
     protected _propertyProjectors(): AttributeToProjector {
       var attrToProjector = super._propertyProjectors();
       var innerRadiusAccessor = Plot._scaledAccessor(this.innerRadius());
@@ -208,6 +232,55 @@ export module Plots {
       var endAngle = pie[index].endAngle;
       var avgAngle = (startAngle + endAngle) / 2;
       return { x: avgRadius * Math.sin(avgAngle), y: -avgRadius * Math.cos(avgAngle) };
+    }
+
+    protected _additionalPaint(time: number) {
+      if (this._labelsEnabled) {
+        Utils.Window.setTimeout(() => this._drawLabels(), time);
+      }
+    }
+
+    private _drawLabels() {
+      var attrToProjector = this._generateAttrToProjector();
+      var labelArea = this._renderArea.append("g").classed("label-area", true);
+      var measurer = new SVGTypewriter.Measurers.CacheCharacterMeasurer(labelArea);
+      var writer = new SVGTypewriter.Writers.Writer(measurer);
+      this.entities().forEach((entity) => {
+        var value = "" + this.sectorValue().accessor(entity.datum, entity.index, entity.dataset);
+        var measurement = measurer.measure(value);
+
+        var center = this._center();
+        var theta = (this._endAngles[entity.index] + this._startAngles[entity.index]) / 2;
+        var outerRadius = this.outerRadius().accessor(entity.datum, entity.index, entity.dataset);
+        if (this.outerRadius().scale) {
+          outerRadius = this.outerRadius().scale.scale(outerRadius);
+        }
+        var innerRadius = this.innerRadius().accessor(entity.datum, entity.index, entity.dataset);
+        if (this.innerRadius().scale) {
+          innerRadius = this.innerRadius().scale.scale(innerRadius);
+        }
+        var labelRadius = (outerRadius + innerRadius) / 2;
+
+        var x = Math.sin(theta) * labelRadius - measurement.width / 2;
+        var y = -Math.cos(theta) * labelRadius - measurement.height / 2;
+
+        var color = attrToProjector["fill"](entity.datum, entity.index, entity.dataset);
+        var dark = Utils.Color.contrast("white", color) * 1.6 < Utils.Color.contrast("black", color);
+        var g = labelArea.append("g").attr("transform", "translate(" + x + "," + y + ")");
+        var className = dark ? "dark-label" : "light-label";
+        g.classed(className, true);
+
+        writer.write(value, measurement.width, measurement.height, {
+          selection: g,
+          xAlign: "center",
+          yAlign: "center",
+          textRotation: 0
+        });
+      });
+    }
+
+    private _center() {
+      return { x: this.width() / 2, y: this.height() / 2 };
     }
   }
 }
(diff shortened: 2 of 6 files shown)


jtlan:
Can you give us some examples on how this degrades if there's not enough room for the text, etc?

plottable.css
```diff
@@ -63,6 +63,11 @@ svg.plottable {
   font-size: 14px;
 }
 
+.plottable .label-area text {
+  font-family: "Helvetica Neue", sans-serif;
+  fill: #32313F;
```
jtlan: Where'd you get this CSS?

src/plots/piePlot.ts
```diff
@@ -209,6 +233,50 @@ export module Plots {
...
+    protected _additionalPaint(time: number) {
+      if (this._labelsEnabled) {
+        Utils.Window.setTimeout(() => this._drawLabels(), time);
+      }
+    }
+
+    private _drawLabels() {
+      var attrToProjector = this._generateAttrToProjector();
+      var labelArea = this._renderArea.append("g").classed("label-area", true);
+      var measurer = new SVGTypewriter.Measurers.CacheCharacterMeasurer(labelArea);
```
jtlan: Creating a new cacheing `Measurer` sort of defeats the purpose :(

src/plots/piePlot.ts
```diff
@@ -209,6 +233,50 @@ export module Plots {
...
+      var writer = new SVGTypewriter.Writers.Writer(measurer);
+      this.entities().forEach((entity) => {
```
jtlan: Why use `entities()` inside here? It seems like the `Plot` should already be able to get the necessary values by other means right?

jtlan:
Tests?

plottable.css
```diff
@@ -63,6 +63,12 @@ svg.plottable {
   font-size: 14px;
 }
 
+.plottable .label-area text {
+  fill: #32313F;
+  font-family: "Helvetica Neue", sans-serif;
+  font-size: 14px;
+}
+
```
rcchen: Didn't respond to this last time; it was pulled from a couple of lines above. I think that we should standardize the `.label-area` class across all plots that can show labels since they more or less will end up sharing styles.

src/plots/piePlot.ts
```diff
@@ -243,6 +267,73 @@ export module Plots {
...
+          { x: svgX, y: svgY + measurement.height}
+        ];
+
+        var showLabel = true;
+        corners.forEach((corner) => {
+          var entities = this.entitiesAt(corner);
+          if (entities.length < 1) {
+            showLabel = false;
+          } else {
+            if (entities[0].index != entity.index) {
```
rcchen: @jtlan I'm curious if there's a good way to do object equality between `Entities`. This works because there's only a single Entity/Dataset to ever be concerned with in this plot but this is definitely not generalizable.

src/plots/piePlot.ts
```diff
@@ -243,6 +267,76 @@ export module Plots {
...
+        var corners = [
+          { x: svgX, y: svgY },
+          { x: svgX + measurement.width, y: svgY },
+          { x: svgX + measurement.width, y: svgY + measurement.height },
+          { x: svgX, y: svgY + measurement.height}
+        ];
+
+        var showLabel = true;
+        corners.forEach((corner) => {
+          var entities = this.entitiesAt(corner);
```
jtlan: Is there a way to do the math without using `entitiesAt()`?

src/plots/piePlot.ts
```diff
@@ -243,6 +267,76 @@ export module Plots {
...
+          if (entities.length < 1) {
+            showLabel = false;
```
jtlan: Should `break` here.

src/plots/piePlot.ts
```diff
@@ -243,6 +267,97 @@ export module Plots {
...
       return { x: avgRadius * Math.sin(avgAngle), y: -avgRadius * Math.cos(avgAngle) };
     }
+
+    protected _additionalPaint(time: number) {
+      if (this._labelsEnabled) {
+        Utils.Window.setTimeout(() => this._drawLabels(), time);
+      }
+    }
+
+    private _getSliceIndexForPoint(p: Point) {
```
jtlan: No "get"...!

src/plots/piePlot.ts
```diff
@@ -243,6 +267,97 @@ export module Plots {
...
+
+        // Hide the label if it is outside of the slice area
+        var corners = [
+          { x: x, y: y },
+          { x: x, y: y + measurement.height},
+          { x: x + measurement.width, y: y },
+          { x: x + measurement.width, y: y + measurement.height }
+        ];
+
+        // We use the first entity as a point of comparison
```
jtlan: Straggling comment

src/plots/piePlot.ts
```diff
@@ -243,6 +267,97 @@ export module Plots {
...
+        var innerRadius = this.innerRadius().accessor(datum, i, dataset);
+        if (this.innerRadius().scale) {
+          innerRadius = this.innerRadius().scale.scale(innerRadius);
+        }
+        var labelRadius = (outerRadius + innerRadius) / 2;
+
+        var x = Math.sin(theta) * labelRadius - measurement.width / 2;
+        var y = -Math.cos(theta) * labelRadius - measurement.height / 2;
+
+        // Hide the label if it is outside of the slice area
```
jtlan: Don't need this comment, the logic should speak for itself.

src/plots/piePlot.ts
```diff
@@ -243,6 +267,95 @@ export module Plots {
...
+      var index: number;
+      for (var i = 0; i < this._startAngles.length; i++) {
+        if (this._startAngles[i] < pointAngle && this._endAngles[i] > pointAngle) {
+          index = i;
+          break;
+        }
+      }
+      if (index !== undefined) {
+        var dataset = this.datasets()[0];
+        var datum = dataset.data()[i];
```
jtlan: `i`?

src/plots/piePlot.ts
```diff
@@ -243,6 +267,95 @@ export module Plots {
...
+          { x: x + measurement.width, y: y },
+          { x: x + measurement.width, y: y + measurement.height }
+        ];
+
+        var showLabel = true;
+        var index = this._sliceIndexForPoint(corners[0]);
+        for (var j = 1; j < corners.length; j++) {
+          var sliceIndex = this._sliceIndexForPoint(corners[j]);
+          if (sliceIndex == null || sliceIndex !== index || sliceIndex !== i) {
+            showLabel = false;
```
jtlan: Can we replace the loop with the more compact:
```Typescript
```

test/plots/piePlotTests.ts
```diff
@@ -244,6 +244,27 @@ describe("Plots", () => {
...
+      it("labels are shown and hidden appropriately", () => {
+        piePlot.removeDataset(simpleDataset);
+        var data = [
+          { key: "A", value: 1 }, { key: "B", value: 50 },
+          { key: "C", value: 1 }, { key: "D", value: 50 },
+          { key: "E", value: 1 }, { key: "F", value: 50 }
+        ];
+        var dataset = new Plottable.Dataset(data);
+        piePlot.addDataset(dataset).labelsEnabled(true);
+        $(".label-area").children("g").each(function(i) {
```
jtlan: JQuery instead of D3?

rcchen: Only need first level children for this, the children wrap <g> elements and it can get wonky.

src/plots/piePlot.ts
```diff
@@ -243,6 +267,89 @@ export module Plots {
...
+
+        var sliceIndices = corners.map((corner) => this._sliceIndexForPoint(corner));
+        var showLabel = sliceIndices.every((index) => (index != null && index === sliceIndices[0]));
+
+        var color = attrToProjector["fill"](datum, i, dataset);
+        var dark = Utils.Color.contrast("white", color) * 1.6 < Utils.Color.contrast("black", color);
+        var g = labelArea.append("g").attr("transform", "translate(" + x + "," + y + ")");
+        var className = dark ? "dark-label" : "light-label";
+        g.classed(className, true);
+        g.style("display", showLabel ? "block" : "none");
```
jtlan: Usually we use `visibility` instead:
```Typescript
g.style("visibility", showLabel ? "inherit" : "hidden");
```

src/plots/piePlot.ts
```diff
@@ -243,6 +267,89 @@ export module Plots {
...
+
+        writer.write(value, measurement.width, measurement.height, {
```
jtlan: Also some question as to whether we want to write the text at all if it's not visible. It's useful for debugging though, so I think we should keep doing it.

crmorford:
I made a bug :( middle pie plot http://jsfiddle.net/2z4pckh8/2/

src/plots/piePlot.ts
```diff
@@ -243,6 +267,89 @@ export module Plots {
...
+
+        var corners = [
+          { x: x, y: y },
+          { x: x, y: y + measurement.height},
+          { x: x + measurement.width, y: y },
+          { x: x + measurement.width, y: y + measurement.height }
+        ];
+
+        var sliceIndices = corners.map((corner) => this._sliceIndexForPoint(corner));
+        var showLabel = sliceIndices.every((index) => (index != null && index === i));
```
jtlan: Just `index === i` would be enough right?

Also it's worth renaming `i` to something like `datumIndex` so the logic is clearer.

crmorford:
Ok this is admittedly a kinda stupid way to use the plot, but if the labels only show up halfway because of the bounds of the svg, they don't get removed. http://jsfiddle.net/2z4pckh8/3/

rcchen:
Bar plots also suffer in this degenerate case (http://jsfiddle.net/dk7La3hn/1/), going to go with won't fix for now.

crmorford:
Okay, works for me :) Gonna look around one more time.

PR #2347, opened 2015-06-17
[Typescript] - upgrade to use 1.5-beta

bluong:
Developer actions:
* Please run `npm install` to upgrade to the latest compiler after this branch merges to develop.

QE:
* Regression pass please on the overall code.  We updated a tool that generates the entire JS.


bluong:
We updated a tool that affects our JS.  We should have a QE pass.

crmorford:
my tests look ok

aicioara:
Do we want to upgrade to the beta? After recent discussions about backwards compatibility with 1.4 it seems more like an overhead (i.e. figure out if a feature we use has been introduced in 1.5 or not)

bluong:
Based on prior discussion and the fact that 1.5 is hopefully due to be out any day now, let's hold off on the upgrade

aicioara:
This was second attempt. 3rd time lucky

PR #2348, opened 2015-06-17
[Tooling] - Adding grunt-jscs

bluong:
We currently have jshint helping us in our quicktests to make sure that our code is correct from a language perspective, however it does not cover aspects about code style (indentation + newline usage).

Since we are highly considering converting quicktest examples into website examples + so the javascript files are much readable at a later period of time, we should highly consider putting code style rules.

@crmorford Please weigh in if you think that using and maintaining this tool is not that helpful.  I personally think it would be nice if our code style in our quicktests was more consistent.


crmorford:
definitely think it's useful.

hzoo:
Maybe look into using a jscs [preset](http://jscs.info/overview.html#presets) to enforce a consistent style or make your own config - the [autofix feature](http://jscs.info/overview.html#cli) with `--fix` is pretty useful too!

PR #2349, opened 2015-06-17
Closest point performance improvement

### LINKED ISSUE #2195: getClosestPlotData() should use an efficient data structure
We're using getClosestPlotData() for interactive tooltips and the performance just isn't cutting it. The implementation should use an efficient space-partitioning data structure (i.e., kd-tree) to reduce query time to O(log n).

aicioara:
Closes #2195

Before: `660ms` http://jsfiddle.net/9emes3ud/2/ 
After: `1.5ms` http://jsfiddle.net/9emes3ud/3/
`600x` improvement

Stretching:
- 10 000 datapoints - `10ms` - http://jsfiddle.net/9emes3ud/4/ 
- 100 000 datapoints - `100ms` - http://jsfiddle.net/9emes3ud/5/ (mainly limited by the core rendering)

Bar Plot: 
- smooth: 100 bars - `1ms` - http://jsfiddle.net/9emes3ud/9/
- stretch: 5000 bars - `100ms` -  http://jsfiddle.net/9emes3ud/10/

---

Also as part of this PR: 
- Added `Utils.Window.deprecated()`
- deprecated `Plot._visibleOnPlot()`
- added `protected Plot._entityVisibleOnPlot(pixelPoint, datum, index, dataset)`

---

QE Notes: 
- Performance testing: closest points on Scatter Plot, Bar Plot, Line Plot should be improved 
- Regression pass: I changed the logic for retrieving closest points and for computing which data points are visible on the screen.


src/utils/windowUtils.ts
```diff
@@ -41,6 +41,30 @@ export module Utils {
...
+          .trim()
+          .split(/\s|@/)
+          .filter((keyword: string) => !keyword.match(/at/))[0];
+      } catch (err) {
+        callingMethod = "called";
+      }
+
+      Utils.Window.warn("Method " + callingMethod + " has been deprecated in version " + version +
+        ". Please refer to the release notes.");
+    }
```
bluong: How annoying would it be to separate this function into a separate PR?

aicioara: Not a lot, but since this introduces a deprecated function for the first time, we need the method anyway. I can alter the title of the PR if you want

bluong: In the future, you probably wnat to side branch this feature and then make a pull request for just `deprecated()`.  It'll make it substantially easier on the reviewer

plottable.d.ts
```diff
@@ -1251,7 +1258,7 @@ declare module Plottable {
         class Line extends Drawer {
             constructor(dataset: Dataset);
             protected _applyDefaultAttributes(selection: d3.Selection<any>): void;
-            selectionForIndex(index: number): d3.Selection<void>;
```
bluong: And... we have an API break?

src/drawers/areaDrawer.ts
```diff
@@ -16,7 +16,7 @@ export module Drawers {
     }
 
     public selectionForIndex(index: number) {
-      return this.renderArea().select(this.selector());
```
bluong: The more that I use `d3.select` the more that I get scared.  Can we instead use `this.renderArea()` instead of `d3`?

src/drawers/drawer.ts
```diff
@@ -170,7 +179,7 @@ export class Drawer {
    * Returns the D3 selection corresponding to the datum with the specified index.
    */
   public selectionForIndex(index: number) {
-    return d3.select(this._selection()[0][index]);
```
bluong: Use `this.renderArea()` instead of `d3`?

src/drawers/lineDrawer.ts
```diff
@@ -16,7 +16,7 @@ export module Drawers {
     }
 
     public selectionForIndex(index: number) {
-      return this.renderArea().select(this.selector());
```
bluong: Same as above

plottable.d.ts
aicioara: Yey...
I guess I can force cast it. If we look above at the drawer, the `selectionForIndex` actually returns a `d3.Selection<any>`. It's weird that this one diverged.

bluong:
This pull request is certainly doing a lot at once.

It is:
* Improvements at the plot level through `_datumVisibleOnPlot()`
* Improvements at the drawer level through `_cachedSelection`
* `deprecated()` addition

How hard would it be to split this into a side branch and then having 3 separate prs?

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,30 @@ export module Utils {
       }
     }
 
+    /**
+     * To be used in the first line of a deprecated method
+     *
+     * @param {string} version The version when the tagged method became obsolete
+     */
+    export function deprecated(version: string) {
```
bluong: It may be useful to have an optional `message` parameter so we can say "Please use xMethod instead"

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,30 @@ export module Utils {
       }
     }
 
+    /**
+     * To be used in the first line of a deprecated method
```
bluong: What does `deprecated()` do?  The action of the function should be the first sentence.  This sentence should probably go second.

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,30 @@ export module Utils {
...
+
+      var callingMethod = "";
+
+      try {
+        callingMethod = (<any> new Error).stack
+          .split("\n")
+          .filter((step: string) => step.match(/http/))[1]
+          .trim()
+          .split(/\s|@/)
+          .filter((keyword: string) => !keyword.match(/at/))[0];
```
bluong: Errr... I have no idea what is happening here.  Please explain?  Probably through comments or some better organization

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,30 @@ export module Utils {
...
+      } catch (err) {
+        callingMethod = "called";
```
bluong: Why would be have an error case?

src/plots/barPlot.ts
```diff
@@ -298,6 +299,24 @@ export module Plots {
       return Plottable.Utils.DOM.intersectsBBox(xRange, yRange, barBBox);
     }
 
+    protected _datumVisibleOnPlot(pixelPoint: Point, datum: any, index: number, dataset: Dataset) {
```
bluong: Why is the order like so?  Right now we have an order of
`datum`, `pixelPoint`, `index`, `dataset` consistently

src/drawers/areaDrawer.ts
aicioara: The newer version is faster because selection() is cached. The `d3.select` only puts the wrapper around the item. The old version is doing a DOM lookup for the `this.selector()` which is slower

src/plots/scatterPlot.ts
```diff
@@ -109,6 +110,21 @@ export module Plots {
       return Utils.DOM.intersectsBBox(xRange, yRange, translatedBbox);
     }
 
+    protected _datumVisibleOnPlot(pixelPoint: Point, datum: any, index: number, dataset: Dataset) {
+      var xRange = { min: 0, max: this.width() };
+      var yRange = { min: 0, max: this.height() };
+
+      var diameter = this.size().accessor(datum, index, dataset);
```
bluong: Please use `Plot._scaledAccessor()` in case the `size` is under a scale

src/plots/scatterPlot.ts
```diff
@@ -109,6 +110,21 @@ export module Plots {
...
+      var translatedBbox: SVGRect = {
```
bluong: Does this need to be explicitly typed `SVGRect`?

src/drawers/areaDrawer.ts
aicioara: On top of that, it is not consistent with the super implementation of selectionForIndex(). I was having different expectations from this method when I was playing with it and wasted some time figuring out that it is accessing the same data as the parent class but in a different way.

aicioara: Wait, maybe I misunderstood. I wait for you to finish the CR actually

src/plots/plot.ts
```diff
@@ -20,6 +20,11 @@ export module Plots {
   }
 }
 
+interface LightweightPlotEntity extends Plots.PlotEntity {
```
bluong: The name is a misnomer.  By having the prefix `Lightweight`, this implies that there are less items in `LightweightPlotEntity` than `Plots.PlotEntity` which is not the case.

src/plots/plot.ts
```diff
@@ -451,10 +456,24 @@ export class Plot extends Component {
    * @return {Plots.PlotEntity[]}
    */
   public entities(datasets = this.datasets()): Plots.PlotEntity[] {
-    var entities: Plots.PlotEntity[] = [];
+
```
bluong: Don't really care for the newline here.

src/plots/plot.ts
```diff
@@ -465,7 +484,9 @@ export class Plot extends Component {
           index: datasetIndex,
           dataset: dataset,
           position: position,
-          selection: drawer.selectionForIndex(validDatumIndex),
+          drawer: drawer,
+          validDatumIndex: validDatumIndex,
+          selection: null,
```
bluong: The fact that `selection` is null here sounds really mysterious.  Why can't this attribute be just defined instead?

src/plots/plot.ts
```diff
@@ -482,9 +503,10 @@ export class Plot extends Component {
    */
   public entityNearest(queryPoint: Point): Plots.PlotEntity {
     var closestDistanceSquared = Infinity;
-    var closest: Plots.PlotEntity;
-    this.entities().forEach((entity) => {
-      if (!this._visibleOnPlot(entity.datum, entity.position, entity.selection)) {
+    var closest: LightweightPlotEntity;
+    var entities = this._lightweightEntities();
```
bluong: Name variable to `lightweightEntities`

src/plots/plot.ts
```diff
@@ -482,9 +503,10 @@ export class Plot extends Component {
    */
   public entityNearest(queryPoint: Point): Plots.PlotEntity {
     var closestDistanceSquared = Infinity;
-    var closest: Plots.PlotEntity;
-    this.entities().forEach((entity) => {
-      if (!this._visibleOnPlot(entity.datum, entity.position, entity.selection)) {
+    var closest: LightweightPlotEntity;
```
bluong: Name variable to `closestLightweightPlotEntity`?

bluong:
Most of the PR looks fine aside from the idea of `LightweightPlotEntity`  I'm a bit unsure of the purpose of its existence

src/drawers/drawer.ts
aicioara: same

src/drawers/lineDrawer.ts
aicioara: same

plottable.d.ts
bluong: I would recommend force casting it I think in this case, unless our type here is just wrong

src/drawers/areaDrawer.ts
bluong: I essentially mean `this.renderArea().select(this.selection()[0][0])`

Whenever I see `d3.select`, I personally get scared that it'll look throughout the entire page for the selector, which can lead to bugs sometimes

src/plots/barPlot.ts
aicioara: The new ordering makes more sense because the `datum`, `index`, `dataset` are required for the accessor, so they belong to the same logical unit. I am not sure where else we have `datum`, `pixelPoint`, `index`, `dataset` in this order?

bluong: Would `_entityVisibleOnPlot` be a bad idea?

src/plots/plot.ts
aicioara: defined as?

aicioara:
`LightweightPlotEntity` is the core of the improvement. Basically, computing the `selection` for each entity was really time consuming. Therefore, we lazily compute it. That is why selection is set as null and only computed for the closest point. 

Now, I can see the source of confusion. The `LightweightPlotEntity` extents from `PlotEntity`, but that is not totally correct, as the `selection` field should be missing, and instead the `drawer` and `validDatumIndex` fields will be present (for later computation of `selector`). In the same time, I wanted this to inherit from `PlotEntity`, in case we want to add extra fields in the future.

Not really sure how to handle this.

src/drawers/areaDrawer.ts
aicioara: But this was the main source of improvement. If I do what you suggested, I have no more caching. Plus, the `d3.select()` and `whateverD3SelectionHere.select()` are essentially the same thing. By looking at the argument passed (an Element, not a string), you can tell that it will not look throughout the entire page

src/plots/barPlot.ts
aicioara: you mean I should rename it to that?

aicioara:
Writing some tests for the `deprecated()` method

src/drawers/areaDrawer.ts
bluong: Using the `renderArea()` getter does not reset the caching state (as it should).  So, I don't see why we shouldn't reduce the scope for d3 to search for the right selection

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
       }
     }
 
+    /**
+     * Sends a warning to the console. The warning includes the version number of the deprecation,
+     * the name of the function which was deprecated and an optional message.
+     *
+     * Tu be used in the first line of a deprecated method.
```
bluong: "To be used"...

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
       }
     }
 
+    /**
+     * Sends a warning to the console. The warning includes the version number of the deprecation,
```
bluong: Probably "Sends a deprecation warning to the console."?

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
...
+     * the name of the function which was deprecated and an optional message.
+     *
+     * Tu be used in the first line of a deprecated method.
+     *
+     * @param {string} version The version when the tagged method became obsolete
+     * @param {string?} message Optional message to be shown with the warning
+     */
+    export function deprecated(version: string, message = "") {
+
```
bluong: I personally don't like the newline after the function declaration.  Not a big deal though

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
...
+      var callingMethod = "";
+
+      try {
+        // Getting the name of the calling method through the stack trace.
+        callingMethod = (<any> new Error).stack
+          .split("\n")
+          .filter((step: string) => step.match(/http/))[1] // Just method names
+          .trim()
+          .split(/\s|@/) // Accounthing for both Chrome and Firefox
```
bluong: Accounting*

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
...
+          .filter((keyword: string) => !keyword.match(/at/))[0]; // Dropping extra keywords in Chrome
+      } catch (err) { // IE9 does not give a stack trace
```
bluong: IE9 -____-

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,35 @@ export module Utils {
...
+     */
+    export function deprecated(version: string, message = "") {
+
+      var callingMethod = "";
+
+      try {
+        // Getting the name of the calling method through the stack trace.
+        callingMethod = (<any> new Error).stack
+          .split("\n")
+          .filter((step: string) => step.match(/http/))[1] // Just method names
```
bluong: Curious, how does this get method names?

src/plots/plot.ts
```diff
@@ -482,23 +502,34 @@ export class Plot extends Component {
...
+    return closestPointEntity;
   }
 
   protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
+    Utils.Window.deprecated("v1.1.0");
+    return !(pixelPoint.x < 0 || pixelPoint.y < 0 ||
+      pixelPoint.x > this.width() || pixelPoint.y > this.height());
+  }
+
+  protected _entityVisibleOnPlot(pixelPoint: Point, datum: any, index: number, dataset: Dataset) {
```
bluong: Can the input to this function just be an `entity` now?

src/plots/plot.ts
bluong: `entity.drawer.selectionForIndex(entity.validDatumIndex)`?  This one call has caching built in, so I don't see why `_lightweightEntities` doesn't

src/utils/windowUtils.ts
aicioara: lol

src/utils/windowUtils.ts
aicioara: There is a reference to the JS file, which always come from a http URL. This way we avoid the mocka injections and the extra lines (for Example Chrome also says Error on a line by itself).

Now thinking of it, I believe the deprecated method is so hacky, maybe I don't want to be so smart about it. Also using reflection is always a bad thing

src/plots/plot.ts
aicioara: Good point

plottable.d.ts
aicioara: There is a RFC about this. d3.Selection<any> is the right type to return here

src/drawers/areaDrawer.ts
aicioara: I actually have a discussion about this, because `renderArea()` actually resets the caching state (as I thought it should)

jtlan: @bluong wrote:
> Whenever I see `d3.select`, I personally get scared that it'll look throughout the entire page for the selector, which can lead to bugs sometimes

In this case though, we're selecting an existing element, not using a selector string. Previous bugs caused by `d3.select()` were due to passing in a selector string.

src/plots/plot.ts
```diff
@@ -451,27 +456,42 @@ export class Plot extends Component {
    * @return {Plots.PlotEntity[]}
    */
   public entities(datasets = this.datasets()): Plots.PlotEntity[] {
-    var entities: Plots.PlotEntity[] = [];
+    var entities = this._lightweightEntities(datasets);
+
+    entities.forEach((entity) => {
+      entity.selection = entity.drawer.selectionForIndex(entity.validDatumIndex);
+      delete(entity.drawer);
+      delete(entity.validDatumIndex);
```
jtlan: Wouldn't `map()` be more semantic than using `delete`?

aicioara: `map()` is harder to implement because we don't know what are the properties on the `entity`. If we hardcode them, we have to maintain this bit of code, which leads to fragility

bluong: Why does entity.selection have to be evaluated here?  Since `selectionForIndex` caches, why can't we just fold `_lightweightEntities` into `entities`?

aicioara: What do you mean?

bluong: I guess I mainly don't understand why `_lightweightEntities()` sets the `selection` to `null`.  This can be incredibly confusing later if developers down the line call `_lightweightEntities` expecting to have a `selection` parameter

bluong:
The idea behind `_lightweightEntities()` seems really odd to me.  Judging from looking at the raw type definitions, `LightweightEntitiy` are actually heavier than the `Entity` type.  Also there is a lot of mysterious logic with the `selection` parameter being set to null first and then being set to a valid value later.

aicioara:
It was weird that lightweight entities were inheriting from entities. 

The main reason behind doing so was such that the conversion back to a normal PlotEntity would be done by blacklisting extra fields instead of whitelisting the needed fields. The issue was that this bit of code would become fragile, but since the compiler is doing the type check, I believe it is a good compromise. Will do that

aicioara:
I am in favor of making the `LightweightPlotEntity` interface public. That is because:
`entityNearest()` in BarPlot and LinePlot can also leverage a 10x extra increase by not calling `entities()` and instead calling the protected `lightweightEntities` (just like the other plots do). However, protected methods cannot use unexported types

src/plots/barPlot.ts
```diff
@@ -291,6 +291,7 @@ export module Plots {
     }
 
     protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
+      Utils.Window.deprecated("v1.1.0");
```
bluong: Change all the deprecated functions to aim at v1.2.0?

As a side note, it'd be nice if we can find a way to automate this deprecated idea.

src/plots/plot.ts
```diff
@@ -20,6 +20,20 @@ export module Plots {
   }
 }
 
+/**
+ * Lightweight version of Plots.PlotEntity. The field selection is missing, but can be deduced
+ * using the drawer and the validDatumIndex.
```
bluong: Possibly give a description on how this is Lightweight in comparison to the original Plots.PlotEntity?  Also, consider making it a `type`

src/plots/plot.ts
```diff
@@ -451,27 +465,45 @@ export class Plot extends Component {
    * @return {Plots.PlotEntity[]}
    */
   public entities(datasets = this.datasets()): Plots.PlotEntity[] {
-    var entities: Plots.PlotEntity[] = [];
+    return this._lightweightEntities(datasets).map((entity) => this._lightweightEntityToPlotEntity(entity));
```
bluong: What happened to the `selection`?

src/plots/plot.ts
```diff
@@ -482,23 +514,30 @@ export class Plot extends Component {
...
+        closestPointEntity = entity;
       }
     });
 
-    return closest;
+    return this._lightweightEntityToPlotEntity(closestPointEntity);
   }
 
   protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
+    Utils.Window.deprecated("v1.1.0");
```
bluong: Probably change to v1.2.0

src/plots/scatterPlot.ts
```diff
@@ -94,6 +94,7 @@ export module Plots {
     }
 
     protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
+      Utils.Window.deprecated("v1.1.0");
```
bluong: Probably change to v1.2.0

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,34 @@ export module Utils {
...
+    /**
+     * Sends a deprecation warning to the console. The warning includes the version number of the deprecation,
+     * the name of the function which was deprecated and an optional message.
+     *
+     * To be used in the first line of a deprecated method.
+     *
+     * @param {string} version The version when the tagged method became obsolete
+     * @param {string?} message Optional message to be shown with the warning
+     */
+    export function deprecated(version: string, message = "") {
```
bluong: Do we thinkg that the name `deprecated` is a good name?

This essentially marks that a function is deprecated in the input version with a message if given, so I personally think it makes sense, but we should probably survey the name.

src/utils/windowUtils.ts
```diff
@@ -41,6 +41,34 @@ export module Utils {
...
+      var callingMethod = "";
+
+      try {
+        // Getting the name of the calling method through the stack trace.
+        callingMethod = (<any> new Error).stack
```
bluong: Why anycasting

bluong: Why are we using `new Error` here?

src/utils/windowUtils.ts
bluong: Unless there is a determined way to get the method name that isn't hacky (do a bit of research on this), we should probably go with a dumber approach (sending the methodName through to the function)

bluong: Consider this? https://stackoverflow.com/questions/280389/how-do-you-find-out-the-caller-function-in-javascript

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,58 @@
+///<reference path="../testReference.ts" />
+
+var assert = chai.assert;
+
+describe("Utils.Window", () => {
+  it("deprecated()", () => {
```
bluong: Probably move `deprecated()` into it's own `describe()` block

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,58 @@
...
+    var oldWarn = Plottable.Utils.Window.warn;
+    Plottable.Utils.Window.warn = (msg: string) => {
+      warningTriggered = true;
+    };
+
+    Plottable.Utils.Window.deprecated("v0.77.2");
+
+    assert.isTrue(warningTriggered, "the warning has been triggered");
+
+    Plottable.Utils.Window.warn = oldWarn;
```
bluong: Use a `before()` and `after()` block to set / reset the warning

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,58 @@
...
+  it("deprecated()", () => {
+    var warningTriggered = false;
+    var oldWarn = Plottable.Utils.Window.warn;
+    Plottable.Utils.Window.warn = (msg: string) => {
+      warningTriggered = true;
+    };
+
+    Plottable.Utils.Window.deprecated("v0.77.2");
+
+    assert.isTrue(warningTriggered, "the warning has been triggered");
```
bluong: Can we get better information from the `deprecated` call?  We want the warning to trigger, but we also want it to trigger with the methodName in mind right?

src/plots/barPlot.ts
aicioara: What do you mean with automate? Also, I am aiming to release this in v1.1, i.e. demo it on Friday

bluong: Possibly put a symbol here that allows Grunt to find/replace the symbol with current version number + 1.  That said, that's outside scope of this PR, but something to keep in mind especially as we are forced to deprecate more things

src/utils/windowUtils.ts
bluong: Weird it seems that my comment got nuked somewhere...  Oh well.

Consider using this post to get the method name information instead of how you're doing it ccurrently https://stackoverflow.com/questions/280389/how-do-you-find-out-the-caller-function-in-javascript

src/plots/plot.ts
aicioara: Why not keep it consistent with the PlotEntity (which is an interface)?

src/utils/windowUtils.ts
aicioara: Survey where? Is this an important thing? It is supposed to only be used by developers.

src/utils/windowUtils.ts
aicioara: The idea here is that I wanted to automate this method as much as possible. This bit of code tries to get the calling method name by getting the stack trace at runtime. Any casting is because for some reason the error does not have a stack field in Typescript

test/utils/windowUtilsTests.ts
aicioara: There is a special test case for that

aicioara:
callee / caller are deprecated https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee

src/utils/windowUtils.ts
bluong: It will be exposed to users so we should just make sure that the spec of the function is right.  Probably survey it within our group and see if people have objections to the spec.

jtlan: I'd vouch for passing in the method name rather than trying to infer it.

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    it("deprecated()", () => {
+      var warningTriggered = false;
+      Plottable.Utils.Window.warn = (msg: string) => {
+        warningTriggered = true;
+      };
+
+      Plottable.Utils.Window.deprecated("deprecatedMethod", "v0.77.2");
+      assert.isTrue(warningTriggered, "the warning has been triggered");
+    });
```
bluong: I think this test is not useful now

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,62 @@
...
+      Plottable.Utils.Window.warn = (msg: string) => {
+        warningTriggered = true;
+        if (!TestMethods.isIE() && !window.PHANTOMJS) {
+          assert.isNotNull(msg.match(new RegExp(methodName)), "The method name exists in the message " + msg);
+        }
+      };
+
+      var reallyOutdatedCallerMethod = () => Plottable.Utils.Window.deprecated(methodName, "v0.1.2");
+      reallyOutdatedCallerMethod();
+    });
```
bluong: I think this test is not useful now.

test/utils/windowUtilsTests.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      var warningTriggered = false;
+      Plottable.Utils.Window.warn = (msg: string) => {
+        assert.isNotNull(msg.match(/v\d\.\d\d\.\d/), "There exists a version number " + msg);
+        assert.strictEqual(msg.match(/v\d\.\d\d\.\d/)[0], version, "The version number has been correctly passed in " + msg);
+        assert.isNotNull(msg.match(message)[0], "The message exists in the warning message " + msg);
+        var regEx = new RegExp(message + "$");
+        assert.strictEqual(msg.match(regEx)[0], message, "The message appears at the end of the warning message " + msg);
+        warningTriggered = true;
+      };
```
bluong: We should probably check the method name is in the message.  The fact that the warning got called and that the content of the warning says that `methodName` got deprecated are the 2 defining things of this function

src/plots/plot.ts
```diff
@@ -20,6 +20,21 @@ export module Plots {
   }
 }
 
+/**
+ * Computing the selection of an entity is an expensive operation. This object aims to
+ * reproduce the bahavior of the Plots.PlotEntity, excluding the selection, but including
```
bluong: behavior*

src/plots/plot.ts
```diff
@@ -451,27 +466,45 @@ export class Plot extends Component {
...
+          validDatumIndex: validDatumIndex
         });
         validDatumIndex++;
       });
     });
-    return entities;
+    return lightweightEntities;
+  }
+
+  private _lightweightEntityToPlotEntity(entity: LightweightPlotEntity) {
```
bluong: `_lightweightPlotEntityToPlotEntity`?

test/utils/windowUtilsTests.ts
aicioara: Disagree. This is the base test. I will remove the third one

test/utils/windowUtilsTests.ts
aicioara: Fixed

PR #2351, opened 2015-06-18
[Tslint] - adding variable-name rule

bluong:
Putting more enforcement through tslint.  The rule enforces that variables are either camelCased or UPPER_CASED.  Mainly to keep to a consistent style that Typescript variables are normally under the above heuristic.

Invalid:
```typescript
var one_two_three;
```

Valid:
```typescript
var oneTwoThree;
```


jtlan:
Can we also add something to enforce that `private` and `protected` vars start with an underscore?

bluong:
That rule does not yet currently exist, but we should definitely look to make it a rule.

bluong:
Just variable renames NOQE

PR #2358, opened 2015-06-19
adding eslint + disabling default rules

bluong:
Disabling the default rules.  I'll enable them as needed gradually


aicioara:
Unless I'm mistaking, this will replace JSHint. If that is the case, can we create a new ticket for removing jshint, and then link it here? I want to avoid leaving trash behind (`Gruntfile` already has some tasks which I believe are not used anymore)

bluong:
Done https://github.com/palantir/plottable/issues/2361

bluong:
NOQE since no js change

PR #2365, opened 2015-06-20
Safari test fix. ES6 Set forEach() badly implemented in Safari

### LINKED ISSUE #2353: Tests failing in Safari
There is 1 set test failing in Safari

aicioara:
Closes #2353

As [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/forEach) suggests, the second argument for the `forEach()` method is the same as the first argument. However, Safari is not following the spec, so had to add a HackHack for the test to pass.

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -8341,7 +8341,7 @@ describe("Utils", function () {
             set.add(values[1]);
             var index = 0;
             set.forEach(function (value1, value2, passedSet) {
-                // HACKHACK: Safari does not pass a duplicate value as the second argument, as the spec suggests.
+                // HACKHACK: Safari bug #21489317: Safari passes undefined instead of a duplicate value for value2.
                 if (value2 !== undefined) {
                     assert.strictEqual(value1, value2, "The two value arguments passed to the callback are the same");
                 }
--- a/test/tests.js
+++ b/test/tests.js
@@ -8341,7 +8341,10 @@ describe("Utils", function () {
             set.add(values[1]);
             var index = 0;
             set.forEach(function (value1, value2, passedSet) {
-                assert.strictEqual(value1, value2, "The two value arguments passed to the callback are the same");
+                // HACKHACK: Safari does not pass a duplicate value as the second argument, as the spec suggests.
+                if (value2 !== undefined) {
+                    assert.strictEqual(value1, value2, "The two value arguments passed to the callback are the same");
+                }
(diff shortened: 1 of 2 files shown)


bluong:
How about for Mobile Safari?  How do `Set`s work there?  Do we need to change our implementation for our code to work there?

aicioara:
It should work on mobile Safari as well, but testing now

aicioara:
Tested in Safari mobile and it works

test/tests.js
```diff
@@ -8341,7 +8341,10 @@ describe("Utils", function () {
             set.add(values[1]);
             var index = 0;
             set.forEach(function (value1, value2, passedSet) {
-                assert.strictEqual(value1, value2, "The two value arguments passed to the callback are the same");
+                // HACKHACK: Safari does not pass a duplicate value as the second argument, as the spec suggests.
+                if (value2 !== undefined) {
```
jtlan: Can we have an `else` that tests the Safari implementation separately?

jtlan:
Does this mean that our `forEach()` just doesn't work correctly on Safari?

test/tests.js
aicioara: It is covered. Safari has `value2 === undefined`. the else would look like
`typescript
else {
  assert.strictEqual(value2, undefined, "Safari")
}
```
which looks a bit silly

aicioara:
It does not really matter because we don't use value 2 anyway. It was designed like that such that it is consistent with the Map API (argument1, argument2, callback). 

However, if we are to be pedantic then yes, `forEach` does not work the same on Safari as it does on FF and Chrome

test/tests.js
aicioara: Maybe I should amend the message of the hackhack, since it was not clear to you that safari passes undefined?

jtlan: Does Safari Pass the `Set` as the third argument?

If there's a bug # for it on Safari, I would link that.

aicioara: You are right, it does not pass the `Set` as the third argument either

aicioara: 3rd argument does not matter. We have a wrapper for it anyway because we pass our own `Set`

aicioara:
I opened bug #21489317 on Apple platform, but I have no idea how to see it from an external source

jtlan:
[NOQE], test-only changes.

PR #2366, opened 2015-06-20
[ESLint] - allowing for default usage of comma-dangle

bluong:
Default usage is where commas cannot be dangling at the end.

Invalid:
[ {foo: "bar"},
  {foo: "baz"}, ]

Valid:
[ {foo: "bar"},
  {foo: "baz"} ]


aicioara:
I don't really like this feature. But sure

PR #2367, opened 2015-06-20
[ESLint] - applying default usage of no-mixed-spaces-and-tabs

bluong:
Tab characters in general should not be used.  Only spaces for alignment.

Using {\t} as a tab character

Invalid:
```javascript
{\t}{\t}var foo = "bar;
```

Valid:
```javascript
  var foo = "bar";
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,7 +14,6 @@
     "comma-spacing": 0,
     "key-spacing": 0,
     "space-infix-ops": 0,
-    "no-mixed-spaces-and-tabs": 0,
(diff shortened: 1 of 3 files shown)


aicioara:
I think we should completely forbid the usage of the tab character (I can't think of a usage for it in our codebase). I believe this rule does not do that. If you have a file all-tabs-indentation, it will pass?

bluong:
I'll look for a rule very soon that will ban tab characters.  This is mainly to get to the defaults as a starting point

PR #2368, opened 2015-06-20
[ESLint] - applying modified default usage of no-unused-vars

bluong:
Default usage means that if a variable is defined, it should be used somewhere.

Invalid:
```javascript
function foo() {
  var a = "test";
  return;
}
```

Valid:
```javascript
function foo() {
  var a = "test";
  doStuff(a);
  return;
}
```

A modification has been added so that global variables are exceptions to the rule.


aicioara:
Why not `vars: all`?

.eslintrc
```diff
@@ -8,7 +8,7 @@
     "quotes": 0,
     "consistent-return": 0,
     "no-console": 0,
-    "no-unused-vars": 0,
+    "no-unused-vars": [2, {"vars": "local"}],
```
aicioara: Why not `vars: all`?

Gruntfile.js
```diff
@@ -388,7 +385,7 @@ module.exports = function(grunt) {
   var updateQuickTestsJSON = function() {
     var qtJSON = [];
     var rawtests = grunt.file.expand("quicktests/overlaying/tests/**/*.js");
-    rawtests.forEach(function(value, index, array){
+    rawtests.forEach(function(value){
```
aicioara: Was this change needed?
The default setting in eslint is `"args": "after-used"`, so it should not be the case to flag this one. Also, I personally prefer leaving in all arguments for a function, in case future development needs them

bluong:

.eslintrc
bluong: `vars: all` enforces that functions that are exposed also need to be used, which doesn't make much sense.  Especially if they are utility functions.

Gruntfile.js
bluong: `"args": "after-used"` means that if `array` is used in the above, then the rule ignores the `value` and the `index` parameter usage.  However, in the example above, `index` and `array` are just not used.

I personally prefer the slim approach, only exposing what one needs, but I can see the validity in your statement.  We can do a quick survey to see what we prefer

aicioara:
Oh, I see

PR #2369, opened 2015-06-20
[ESLint] - applying default usage of no-return-assign

bluong:
Mainly by disabling the linter on require.js

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,6 @@
 {
   "rules": {
     "no-eval": 0,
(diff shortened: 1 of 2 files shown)


aicioara:
Can we configure Grunt to omit that folder? It seems odd to disable jslint rules in a 3rd party library. I'd also want this to run as fast as possible, by omitting needless folders.

bluong:

aicioara:
Rule is fine, I hoping for it + no change in the JS

PR #2370, opened 2015-06-20
[ESLint] - applying default usage of no-undef

bluong:
The rule disallows the usage of variables before they are defined

Invalid:
```javascript
function foo() {
  doStuff(a);
}
```

Valid:
```javascript
function foo() {
  var a = ...;
  doStuff(a);
}
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,18 @@
 {
+  "env": {
+    "browser": true,
+    "jquery": true,
+    "node": true
+  },
+  "globals": {
+    "d3": true,
+    "Plottable": true,
+    "makeRandomData": true,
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -1,8 +1,18 @@
...
+  "env": {
+    "browser": true,
+    "jquery": true,
+    "node": true
+  },
+  "globals": {
+    "d3": true,
+    "Plottable": true,
+    "makeRandomData": true,
+    "deep_copy": true
```
aicioara: What method is this?

aicioara: However, not part of this PR

PR #2371, opened 2015-06-20
[ESLint] - applying default usage of no-shadow

bluong:
Rule disallows shadowing of variables.

Invalid:
```javascript
var a = 10;
function foo(a) {
  doStuff(a);
}
```

Valid:
```javascript
var a = 10;
function foo(aTest) {
  doStuff(aTest);
}
```

### DIFF
--- a/quicktests/umd/lib/require.js
+++ b/quicktests/umd/lib/require.js
@@ -7,6 +7,7 @@
 //problems with requirejs.exec()/transpiler plugins that may not be strict.
 /*jslint regexp: true, nomen: true, sloppy: true */
 /*global window, navigator, document, importScripts, setTimeout, opera */
+/*eslint-disable */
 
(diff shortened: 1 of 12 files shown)


quicktests/umd/lib/require.js
```diff
@@ -7,6 +7,7 @@
 //problems with requirejs.exec()/transpiler plugins that may not be strict.
 /*jslint regexp: true, nomen: true, sloppy: true */
 /*global window, navigator, document, importScripts, setTimeout, opera */
+/*eslint-disable */
```
jtlan: Modifying the library? Is there a way to tell ESLint to ignore the directory instead?

bluong: Refer to this: https://github.com/palantir/plottable/pull/2373

bluong:

PR #2372, opened 2015-06-20
[ESLint] - allowing for default usage of no-trailing-spaces

bluong:
Rules disallows ending with trailing whitespace.

Using {\space} as a whitespace character.

Invalid:
```javascript
var foo = 3;{\space}{\space}{\space}
```

Valid:
```javascript
var foo = 3;
```


aicioara:
Nice

PR #2373, opened 2015-06-20
[ESLint] - applying default usage of consistent-return

bluong:
Rules disallows functions being inconsistent with returning something or returning nothing.

Invalid:
```javascript
function foo() {
  if (boolean) {
    return bar;
  } else {
    return;
  }
}
```

Valid:
```javascript
function foo() {
  if (boolean) {
    return bar;
  } else {
    return baz;
  }
}
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,7 +6,6 @@
     "no-shadow": 0,
     "no-trailing-spaces": 0,
     "quotes": 0,
-    "consistent-return": 0,
     "no-console": 0,
     "no-unused-vars": 0,
     "no-multi-spaces": 0,
--- a/quicktests/umd/lib/require.js
+++ b/quicktests/umd/lib/require.js
@@ -7,6 +7,7 @@
 //problems with requirejs.exec()/transpiler plugins that may not be strict.
 /*jslint regexp: true, nomen: true, sloppy: true */
 /*global window, navigator, document, importScripts, setTimeout, opera */
+/*eslint-disable */
 
 var requirejs, require, define;
 (function (global) {


aicioara:
But we do this in our TypeScript (at least returning different types). Why would this be forbidden in JS? Again, it looks weird that we have to disable the rule for a 3rd party library. Maybe we should exclude the file in Grunt somehow?

bluong:
We return different types in Typescript, but we at least return something at all times.  This is to mainly check that we don't have a function that returns nothing or something.

Also, note the above example.  From my personal perspective, I suspect that the Invalid case is more likely to be something that a developer would write as a mistake and that the Valid case is more likely what that developer intended.

bluong:
In accordance with the ignoring, so far the easiest way I found to ignore the file is by...
* Create a ".eslintignore" file that contains paths to ignored files (more files to maintain -.-; but not a big deal)
* Running "grunt eslint" now shows the files ignored as a warning to the console (This is pretty annoying.  I want the fact that we are ignoring this file to be hidden, not be shot as a warning on every run.  Having warnings be shot every run is just bad practice.)

There are a few solutions that I can think of:
* Modify the targets parameter to route around this file (this sounds unmaintainable if we have more files under the glob that we want to ignore)
* Move the according file to a different location that is outside the glob scope.
* Configure "grunt-eslint" so that we get what we want.  We essentially work with the API to hopefully ignore warnings that come from ignoring files.  (This "might" be the best way but it is also the most difficult).

aicioara:
Can we extract all 3rd party libraries to a separate folder? Then configure eslint to either ignore that folder or glob around it. I still do not agree in changing the code of a library we don't own (this will make updating a pain if we ever do it), I prefer either amend somehow or drop the PR entirely. In other words I doubt this rule is as useful as the downside of modifying the `require.js` file

bluong:
Ok, in the near future, I'll likely drop this PR and then make a new one that will essentially just relocate the require.js file

aicioara:
okay

bluong:

PR #2374, opened 2015-06-20
[ESLint] - applying default usage of semi-spacing

bluong:
The rule enforces spacing around semicolons.  There can be no whitespace before and there must be some after.

Invalid:
```javascript
var foo = function (a) { return a ;};
```

Valid:
```javascript
var foo = function (a) { return a; };
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,7 +15,6 @@
     "key-spacing": 0,
     "space-infix-ops": 0,
     "no-mixed-spaces-and-tabs": 0,
-    "semi-spacing": 0,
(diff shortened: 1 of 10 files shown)


aicioara:
Modify the example, there is a spacing difference before `(a)`, which is not part of this PR

aicioara:
If it's not to hard, rewrite the history a bit before merging. `git commit --amend -m "Better message"`, to make the git history cleaner. `default rule` seems like a theme throughout the PRs. I like it better than seeing it in the merge information, which is harder to follow.

PR #2376, opened 2015-06-22
Add enable/disable functionality to Interaction

### LINKED ISSUE #696: Interactions should have disable() and enable() calls

jtlan:
For temporarily disabling an `Interaction` without permanently disconnecting it from the `Component` it was attached to.

API:
```Typescript
/**
 * Gets whether this Interaction is enabled.
 */
public enabled(): boolean;
/**
 * Enables or disables this Interaction.
 */
public enabled(enabled: boolean): Interaction;
```

Example fiddle: http://jsfiddle.net/j0fnbmvr/

Open questions: Should interactive `Component`s (like `DragBoxLayer`) implement this functionality as well? Or is giving access to the internal `Interaction` sufficient?

Close #696.

### DIFF
--- a/src/interactions/interaction.ts
+++ b/src/interactions/interaction.ts
@@ -7,6 +7,7 @@ export class Interaction {
   private _anchorCallback = (component: Component) => this._anchor(component);
 
   private _isAnchored: boolean;
+  private _enabled = true;
 
   protected _anchor(component: Component) {
     this._isAnchored = true;
@@ -24,14 +25,19 @@ export class Interaction {
    * @returns {Interaction} The calling Interaction.
    */
   public attachTo(component: Component) {
-    if (this._componentAttachedTo) {
-      this.detachFrom(this._componentAttachedTo);
-    }
-
+    this._detach();
     this._componentAttachedTo = component;
-    component.onAnchor(this._anchorCallback);
-
+    this._attach();
     return this;
+  }
+
+  /**
+   * Attaches to existing _componentAttachedTo if enabled.
+   */
+  private _attach() {
+    if (this.enabled() && this._componentAttachedTo != null && !this._isAnchored) {
+      this._componentAttachedTo.onAnchor(this._anchorCallback);
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/interactions/interaction.ts
```diff
@@ -42,12 +48,42 @@ export class Interaction {
...
+      this._componentAttachedTo.offAnchor(this._anchorCallback);
+    }
+  }
 
+  /**
+   * Gets whether this Interaction is enabled.
+   */
+  public enabled(): boolean;
+  /**
+   * Enables or disables this Interaction.
```
bluong: Probably document the fact that this method can chain

src/interactions/interaction.ts
```diff
@@ -24,17 +25,22 @@ export class Interaction {
...
+    this._detach();
     this._componentAttachedTo = component;
-    component.onAnchor(this._anchorCallback);
-
+    this._attach();
     return this;
   }
 
   /**
+   * Attaches to existing _componentAttachedTo if enabled.
```
bluong: The docs point to a private instance variable which sounds.. a bit incorrect.  Javascript users now might wonder what this `_componentAttachedTo` is and then run into trouble interacting with it.

Let's just use english here.  This also helps us if we decide to change the name of the variable later down the road

bluong:
Documentation issues.

src/interactions/interaction.ts
jtlan: The variable is `protected` so we can't change the name without causing a break.

Technically it's attached to the Component, just not listening...

bluong: Oh I just realized that this is documentation on a private function...  Never mind, I think? this is fine.  Not sure, but I think this is ok.

jtlan:
Probably `DragBoxLayer` should have its own implementation -- right now if you set `resizable(true)` and `enabled(false)`, you still get the "`<-->`" cursor.

src/interactions/interaction.ts
jtlan: done.

src/interactions/interaction.ts
```diff
@@ -24,16 +25,18 @@ export class Interaction {
...
-
+    this._disconnect();
     this._componentAttachedTo = component;
-    component.onAnchor(this._anchorCallback);
-
+    this._connect();
     return this;
   }
 
+  private _connect() {
```
bluong: Unsure if I like the name.  My first response is ok `interaction._connect()` will connect to...?

That said, this is a private function and it alludes to what it's doing sorta somewhat ok.  Not a huge deal.

jtlan:
Secondary issue filed at #2386.

PR #2379, opened 2015-06-23
[Interactions.PanZoom] - Scale extent endpoints

bluong:
Endpoints have been added to `Interactions.PanZoom` to set the maximum/minimum allowable domain extents for the input scale.  The interaction upon zooming will then enforce these extents (stopping a zoom out if the resultant zoom out is too much, etc.).

Note that these calculations will not work well if the input scale is a `Scales.ModifiedLog` or to be more specific, any `Scale` that is not linear in the mathematical sense.  There is then an appropriate warning thrown to the console if the input scale to the setter is not a `Scales.Linear` or a `Scales.Time`.

Example fiddle: http://jsfiddle.net/bluong63/bzbbaxcy/

`Scales.Time` example: http://jsfiddle.net/00hrdqsf/

API Changes:
*  The following `minDomainExtent` API has been added to `Interactions.PanZoom` to get/set the minimum allowable extent for the domain of the input scale.
```typescript
minDomainExtent<D>(qScale: QuantitativeScale): D;
minDomainExtent<D>(qScale: QuantitativeScale, minDomainExtent: D): Interactions.PanZoom;
```

*  The following `maxDomainExtent` API has been added to `Interactions.PanZoom` to get/set the maximum allowable extent for the domain of the input scale.
```typescript
maxDomainExtent<D>(qScale: QuantitativeScale): D;
maxDomainExtent<D>(qScale: QuantitativeScale, minDomainExtent: D): Interactions.PanZoom;
```


jtlan:
Example fiddle?

src/interactions/panZoomInteraction.ts
```diff
@@ -144,14 +166,41 @@ export module Interactions {
...
+        constrainedDomainValues[1] = domainCenter + minDomainExtent / 2;
+      }
+
+      var maxDomainExtent = this._maxDomainExtents.get(scale);
+      if (domainExtent > maxDomainExtent) {
+        constrainedDomainValues[0] = domainCenter - maxDomainExtent / 2;
+        constrainedDomainValues[1] = domainCenter + maxDomainExtent / 2;
+      }
+
+      var constrainedDomain = constrainedDomainValues.map((value) => scale instanceof Scales.Time ? new Date(value) : value);
```
jtlan: I don't think we should have an `instanceof` conditional here, which suggests that this might not be quite the right implementation. Should the constraint logic live on the `Scale`, with the actual constraint values stored on the `PanZoom`?

src/interactions/panZoomInteraction.ts
```diff
@@ -271,9 +329,21 @@ export module Interactions {
...
       return this;
     }
 
+    private _ensureDefaultDomainExtents(scale: QuantitativeScale<any>) {
+      if (this._minDomainExtents.get(scale) == null) {
+        this._minDomainExtents.set(scale, 0);
+      }
+      if (this._maxDomainExtents.get(scale) == null) {
+        var maxDateValue = 8640000000000000;
+        var defaultMaxDomainExtent = scale instanceof Scales.Time ? maxDateValue : Infinity;
```
jtlan: Same issue here -- the `instanceof` suggests that the `Scale`s should be implementing some of this functionality.

jtlan:
[-1] I think some of the logic needs to be shifted to the `Scale`s.

src/interactions/panZoomInteraction.ts
bluong: @istarnx can pop in since he has a better grasp on this perspective, but his guess is that the constraining logic should not be on the `Scale`

bluong:
Which piece of logic?  I originally had the logic on the `Scale`s until that turned out to be incorrect.

src/interactions/panZoomInteraction.ts
bluong: `Scale`s probably do need a way to convert from a number to their Domain type however.

jtlan:
I guess the question now is, is it fair to assume that all `QuantitativeScale`s have an extent that can be expressed as a `number`? One constraint that we have implicitly is that the type `D` has to be `{ valueOf(): number } | number`, which suggests that users should be able to pass in extents as `number`s.

bluong:
I believe that is a fair assumption

jtlan:
I notice that repeatedly zooming in and out on the fiddle can cause the scales to get out of sync, which doesn't seem right to me.

bluong:
What do you mean by out of sync?

jtlan:
If you zoom in and out a whole bunch the y-scale continues to zoom in further, even though the x-scale is blocked. This gradually changes the relative "scaling" of the two scales, affecting the appearance of the data (on the Fiddle, the angle of the line will appear to change)

src/interactions/panZoomInteraction.ts
```diff
@@ -161,11 +205,19 @@ export module Interactions {
 
         var deltaPixelAmount = e.deltaY * (e.deltaMode ? PanZoom._PIXELS_PER_LINE : 1);
         var zoomAmount = Math.pow(2, deltaPixelAmount * .002);
+
+        if (zoomAmount < 1 && this._scalesAtMinExtent()) {
+          return;
+        }
+
+        if (zoomAmount > 1 && this._scalesAtMaxExtent()) {
```
jtlan: Shouldn't this logic apply to non-wheel zoom events as well?

src/interactions/panZoomInteraction.ts
```diff
@@ -161,13 +164,57 @@ export module Interactions {
...
+      var upperBound = extentIncreasing ? Infinity : constrainedZoomAmount;
+      var iterations = 20;
+      var magnifyTransform = (rangeValue: number) => scale.invert(centerValue - (centerValue - rangeValue) * constrainedZoomAmount);
+      var scaleRange = scale.range();
+      for (var i = 0; i < iterations; i++) {
+        var transformedDomain = scaleRange.map(magnifyTransform);
+        var transformedDomainExtent = Math.abs(transformedDomain[1] - transformedDomain[0]);
+        if (transformedDomainExtent === boundingDomainExtent) { return constrainedZoomAmount; }
+
+        if (extentIncreasing === transformedDomainExtent < boundingDomainExtent) {
```
jtlan: Can you add parens to this, for readers who don't know the [operator precedence rules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence) off the top of their head?

jtlan:
Fiddle produces odd behavior with `ModifiedLog`: http://jsfiddle.net/8Lkdu7dn/

bluong:
Define odd please.  Note that the starting requirements for the fiddle are currently infeasible

jtlan:
> Define odd please.
* Can't zoom in after zooming out.
* Seem to be unable to pan, even though I'm panning in such a way to keep the domain within the extent limits.

src/interactions/panZoomInteraction.ts
```diff
@@ -248,6 +366,9 @@ export module Interactions {
      * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
      */
     public addXScale(xScale: QuantitativeScale<any>) {
+      if (this._nonLinearScaleWithExtents(xScale)) {
```
jtlan: This isn't going to trigger properly. New `Scale`s won't have any extent limits attached, so there won't be any warnings. At the same time, attempting to put an extent limit on the `Scale` won't produce warnings.

jtlan:
[-1] Warnings won't get triggered at the right time.

In addition, removing a `Scale` from the `Interaction` should remove all extents associated with it -- this prevents memory leaks.

jtlan:
`Scales.Time` example: http://jsfiddle.net/00hrdqsf/

src/interactions/panZoomInteraction.ts
```diff
@@ -284,6 +336,32 @@ export module Interactions {
       this._yScales.delete(yScale);
       return this;
     }
+
+    public minDomainExtent<D>(quantitativeScale: QuantitativeScale<D>): D;
+    public minDomainExtent<D>(quantitativeScale: QuantitativeScale<D>, minDomainExtent: D): Interactions.PanZoom;
+    public minDomainExtent<D>(quantitativeScale: QuantitativeScale<D>, minDomainExtent?: D): any {
```
jtlan: Docs?

src/interactions/panZoomInteraction.ts
```diff
@@ -284,6 +336,32 @@ export module Interactions {
       this._yScales.delete(yScale);
       return this;
     }
+
+    public minDomainExtent<D>(quantitativeScale: QuantitativeScale<D>): D;
+    public minDomainExtent<D>(quantitativeScale: QuantitativeScale<D>, minDomainExtent: D): Interactions.PanZoom;
```
jtlan: Docs for these two methods.

src/interactions/panZoomInteraction.ts
```diff
@@ -284,6 +336,32 @@ export module Interactions {
...
+      }
+      if (this._nonLinearScaleWithExtents(quantitativeScale)) {
+        Utils.Window.warn("Panning and zooming with extents on a nonlinear scale may have unintended behavior.");
+      }
+      this._minDomainExtents.set(quantitativeScale, minDomainExtent);
+      return this;
+    }
+
+    public maxDomainExtent<D>(quantitativeScale: QuantitativeScale<D>): D;
+    public maxDomainExtent<D>(quantitativeScale: QuantitativeScale<D>, maxDomainExtent: D): Interactions.PanZoom;
```
jtlan: Docs here too.

jtlan:
[-1] Docs, and unit tests.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -229,5 +229,101 @@ describe("Interactions", () => {
...
+        var domainExtent = Math.abs(xScale.domain()[1] - xScale.domain()[0]);
+        assert.strictEqual(domainExtent, minimumDomainExtent, "xScale zooms to the correct domain via scroll");
+        svg.remove();
+      });
+
+      it("Pinching in cannot go beyond the specified domainExtent", () => {
+        // HACKHACK PhantomJS doesn't implement fake creation of WheelEvents
+        // https://github.com/ariya/phantomjs/issues/11289
+        if ( window.PHANTOMJS ) {
+          svg.remove();
```
jtlan: PhantomJS should work for touch events right?

crmorford:
Since on reversed domains, it still works with positive numbers (http://jsfiddle.net/bzbbaxcy/3/), maybe we should take absolute value of the number given to minDomainExtent/maxDomainExtent?  There doesn't seem to be a reason to allow negative numbers, and they kinda break things.

crmorford:
Once we get errors for negative numbers, looks good :)

crmorford:
One more concern actually :/
It works when max < min which I'm okay with http://jsfiddle.net/bzbbaxcy/5/
....except when max is 0 and things go bad fast http://jsfiddle.net/bzbbaxcy/6/
Ideas
* enforce min < max
* don't allow max == 0

PR #2382, opened 2015-06-23
[Bower] - Adding requirejs as a dev dependency

bluong:
Our "umd" quicktest currently holds a copy of require.js.  We should instead move this to be a dependency, only to be used for development.


aicioara:
Travis.. I restarted the build

aicioara:
Nice

PR #2383, opened 2015-06-23
[Plots.ErrorBar] Implement ErrorBarPlots

### LINKED ISSUE #1379: Error Bars
Implement error bars:

screen shot 2014-11-21 at 2 43 42 pm

Things to watch out for:
- Upper and lower bound might not be the same distance away from the data point
- Error bars might not be the same size between data points
- How does this work for stacked plots?

rcchen:
Initial implementation.
image

- [x] Add API point for top/bottom bars and their lengths
- [ ] Write tests

Fixes #1379

### DIFF
--- a/src/drawers/errorBarDrawer.ts
+++ b/src/drawers/errorBarDrawer.ts
@@ -5,7 +5,6 @@ export module Drawers {
   export class ErrorBar extends Drawer {
 
     private _ERROR_BAR_CLASS = "error-bar";
-    private _ERROR_BAR_WIDTH = 10;
 
     constructor(dataset: Dataset) {
       super(dataset);
--- a/src/drawers/errorBarDrawer.ts
+++ b/src/drawers/errorBarDrawer.ts
@@ -1,0 +1,17 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Drawers {
+  export class ErrorBar extends Drawer {
+
+    private _ERROR_BAR_CLASS = "error-bar";
+    private _ERROR_BAR_WIDTH = 10;
+
+    constructor(dataset: Dataset) {
+      super(dataset);
+      this._className = this._ERROR_BAR_CLASS;
+      this._svgElementName = "line";
+    }
+  }
+}
+}
--- a/src/plots/errorBarPlot.ts
+++ b/src/plots/errorBarPlot.ts
@@ -1,0 +1,98 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Plots {
+  export class ErrorBar<X, Y> extends XYPlot<X, Y> {
+
+    constructor() {
+      super();
+      this.addClass("error-bar-plot");
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for X.
+     */
+    public x(): Plots.AccessorScaleBinding<X, number>;
+    public x(x: number | Accessor<number>): ErrorBar<X, Y>;
+    public x(x: X | Accessor<X>, xScale: Scale<X, number>): ErrorBar<X, Y>;
+    public x(x?: number | Accessor<number> | X | Accessor<X>, xScale?: Scale<X, number>): any {
+      if (x == null) {
+        return super.x();
+      }
+
+      if (xScale == null) {
+        super.x(<number | Accessor<number>>x);
+      } else {
+        super.x(< X | Accessor<X>>x, xScale);
+      }
+
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for X2.
+     */
+    public x2(): Plots.AccessorScaleBinding<X, number>;
+    public x2(x2: number | Accessor<number>): ErrorBar<X, Y>;
+    public x2(x2: X | Accessor<X>, xScale: Scale<X, number>): ErrorBar<X, Y>;
+    public x2(x2?: number | Accessor<number> | X | Accessor<X>, xScale?: Scale<X, number>): any {
+      if (x2 == null) {
+        return this._propertyBindings.get("x2");
+      }
+      var xBinding = this.x();
+      var xScale = xBinding && xBinding.scale;
+      this._bindProperty("x2", x2, xScale);
+      this.render();
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for Y.
+     */
+    public y(): Plots.AccessorScaleBinding<Y, number>;
+    public y(y: number | Accessor<number>): ErrorBar<X, Y>;
+    public y(y: Y | Accessor<Y>, yScale: Scale<Y, number>): ErrorBar<X, Y>;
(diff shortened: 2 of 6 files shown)


src/drawers/errorBarDrawer.ts
```diff
@@ -0,0 +1,16 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Drawers {
+  export class ErrorBar extends Drawer {
```
bluong: Oof this is gonna be tricky.  This is not really an `ErrorBar` Drawer but instead a ... `Line` Drawer...

Try to name it to something that is not specific to `ErrorBar`s and more aligned with `Line` that doesn't collide with `LineDrawer`

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      if (xScale == null) {
+        super.x(<number | Accessor<number>>x);
+      } else {
+        super.x(< X | Accessor<X>>x, xScale);
+      }
+
+      return this;
+    }
```
bluong: This is just to get the chaining?

src/drawers/errorBarDrawer.ts
rcchen: `PrimitiveLineDrawer`?

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,115 @@
...
+    protected _createDrawer(dataset: Dataset): Drawer {
+      return new Plottable.Drawers.ErrorBar(dataset);
+    }
+
+    protected _propertyProjectors(): AttributeToProjector {
+      var attrToProjector = super._propertyProjectors();
+      attrToProjector["x1"] = Plot._scaledAccessor(this.x());
+      attrToProjector["y1"] = Plot._scaledAccessor(this.y());
+      attrToProjector["x2"] = this.x2() != null ? Plot._scaledAccessor(this.x2()) : Plot._scaledAccessor(this.x());
+      attrToProjector["y2"] = this.y2() != null ? Plot._scaledAccessor(this.y2()) : Plot._scaledAccessor(this.y());
```
bluong: Instead of this... you could trying setting x2 to be the same thing as x1.  This makes sense as is though.  What I'm suggesting just gives a good transparency through the getter

src/drawers/errorBarDrawer.ts
bluong: What differentiates a line `path` and a `line` svg element...?  Not sure if `Primitive` is the right term?

jtlan: `Drawers.LineSegment`? This confusion is part of why I wanted a `Drawers.Path`...

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,115 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Plots {
+  export class ErrorBar<X, Y> extends XYPlot<X, Y> {
```
jtlan: Is there a name that doesn't carry as much intent?

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,115 @@
...
+      super();
+      this.addClass("error-bar-plot");
+    }
+
+    /**
+     * Sets the tick length
+     */
+    public tickLength(): number;
+    public tickLength(tickLength: number): ErrorBar<X, Y>;
+    public tickLength(tickLength?: number): any {
```
jtlan: What if I want to set the top and bottom tick lengths independently? Is that something we should support?

src/plots/errorBarPlot.ts
rcchen: I think we went through an RFC the first time around re: the naming, the conclusion was that since these things are so commonly called Error Bars, we should just stick to this naming.

src/plots/errorBarPlot.ts
bluong: Can we query to see the usecase of these `tickLength()` endpoints?  We can add `topTickLength()`/`bottomTickLength()` endpoints instead of / after this one and I think it should work nicely additively, but we should just choose the one more aligned with our usecase for now.

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.render();
+      return this;
+    }
+
+    protected _additionalPaint(time: number) {
+      super._additionalPaint(time);
+      console.log(this._tickLength);
+      if (this._tickLength != null) {
+        var lowerBars = this._renderArea.append("g").classed("error-lower-bars", true);
+        var upperBars = this._renderArea.append("g").classed("error-upper-bars", true);
```
jtlan: The logic here doesn't seem quite right. For example, repeated calls to `_additionalPaint` will generate more `<g>`s every time.

In addition, we now have the problem that the `Entity` for a given point contains only the center part of the error bar and not the two ends.

rcchen: Agreed, and this puts us back into the land of composite drawers, where the drawers will need to be slightly more intelligent than they are right now. @bluong thoughts?

src/drawers/errorBarDrawer.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    private _ERROR_BAR_CLASS = "error-bar";
```
bluong: This class should be applied from the plot side

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,117 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Plots {
+  export class ErrorBar<X, Y> extends XYPlot<X, Y> {
+
+    private _tickLength = 10;
+    private _LOWER_BARS_CLASS = "error-lower-bars";
+    private _UPPER_BARS_CLASS = "error-upper-bars";
```
bluong: private static

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,117 @@
...
+      var yBinding = this.y();
+      var yScale = yBinding && yBinding.scale;
+      this._bindProperty("y2", y2, yScale);
+      this.render();
+      return this;
+    }
+
+    protected _additionalPaint(time: number) {
+      super._additionalPaint(time);
+      if (this._tickLength != null) {
```
bluong: `tickLength` of 0 seems more apt.  When would this be `null`?

src/plots/errorBarPlot.ts
```diff
@@ -0,0 +1,117 @@
...
+        var lowerBars = this._renderArea.append("g").classed(this._LOWER_BARS_CLASS, true);
+        var upperBars = this._renderArea.append("g").classed(this._UPPER_BARS_CLASS, true);
```
bluong: This should be done in `_setup`

bluong:
Let's see if we actually need this plot now that we've gotten feedback

rcchen:
Temporarily closing this one in favor of upcoming Plots.Segment

PR #2384, opened 2015-06-23
PanZoom performance improvement

### LINKED ISSUE #2352: Scatter Plot pan/zoom is slow
Maybe it was improved by https://github.com/palantir/plottable/pull/2337 , but it needs double checking.

Spec:
` [...] we are having to limit to 5,000 dots for scatterplots so that they stay usable, we’d like to get upwards around 20,000 – 30,000 dots per plot eventually.
`

aicioara:
Closes #2352

**Description**: If we are having a ScatterPlot with a lot of data points, every time we update the scale (which happens on every single pan or zoom), all the `DOM` elements corresponding to the points get modified (in particular translated). This is a costly operation and can be improved by translating the parent `<g>` and letting the drawing engine do the heavy lifting. 

**Problem1**: We cannot account nicely for scaling (when zooming) here. That is because the scale on the top element also modifies the size of the points (and that is not acceptable). Therefore, we ran for a compromise: Scale the points (i.e. zoom badly and lose resolution) during the zoom and recompute all the points after no more interaction exists.

**Problem2**: Because of Problem1, we no longer have the same default behavior. Therefore, there is a switch `XYPlot.deferredRendering()`. 

**Unimplemented ideas**:
- <del>Maybe call performanceEnabled() something else, such that we can have finer grained performance enabling in the future </del>
- <del>Document everything, especially the new public method </del>
- <del>Extract the 500 ms in the setTimeout in some variable </del>
- Add tests?

---

#### Fiddles
Before: http://jsfiddle.net/rk0sx69o/
After: http://jsfiddle.net/rk0sx69o/7/

---

#### Numbers
Scatter Plot with 8000 points `6x increase`:
- Before: `175 ms` per update
- After: `3 ms` per update (capped by browser)

Old Scatter Plot with 8000 points has same performance as new Scatter Plot with 50 000

---

#### QE Notes: 
- Regression test: With the option `deferredRendering` turned off (that is the default). There should be absolutely no change.
- Feature test: With the option `deferredRendering` turned on, panning should behave the same. Zooming however, should behave weirdly (points / lines on plot appear to grow or shrink), but it should default to better quality after 0.5 seconds. The behavior should be the same as when zooming in/out on google map, i.e. low quality during zoom and get back to normal once we have data.
- Performance test: With the option `deferredRendering` turned on, panning and zooming should be 10x or more faster
- Additional notes: All plots except Pie Plot should be affected

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2541,8 +2541,8 @@ declare module Plottable {
          * @param {Scale} yScale The y scale to use.
          */
         constructor();
-        performanceEnabled(): boolean;
-        performanceEnabled(performanceEnabled: boolean): XYPlot<X, Y>;
+        lazyDomainChange(): boolean;
+        lazyDomainChange(lazyDomainChange: boolean): XYPlot<X, Y>;
         /**
          * Gets the AccessorScaleBinding for X.
          */
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2533,7 +2533,6 @@ declare module Plottable {
     class XYPlot<X, Y> extends Plot {
         protected static _X_KEY: string;
         protected static _Y_KEY: string;
-        tempScrollTimeout: number;
         /**
          * An XYPlot is a Plot that displays data along two primary directions, X and Y.
          *
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2307,6 +2307,7 @@ declare module Plottable {
     class Plot extends Component {
         protected static _ANIMATION_MAX_DURATION: number;
         protected _renderArea: d3.Selection<void>;
+        protected _renderCallback: ScaleCallback<Scale<any, any>>;
         protected _propertyExtents: d3.Map<any[]>;
         protected _propertyBindings: d3.Map<Plots.AccessorScaleBinding<any, any>>;
         /**
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2541,6 +2541,8 @@ declare module Plottable {
          * @param {Scale} yScale The y scale to use.
          */
         constructor();
+        performanceEnabled(): boolean;
+        performanceEnabled(performanceEnabled: boolean): XYPlot<X, Y>;
         /**
          * Gets the AccessorScaleBinding for X.
          */
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2542,12 +2542,12 @@ declare module Plottable {
          */
         constructor();
         /**
-         * Returns the lazy domain change setting on the plot
-         * @return {boolean} The lazy domain change setting
+         * Returns the whether or not the rendering is deferred for performance boost.
+         * @return {boolean} The deferred rendering option
          */
         deferredRendering(): boolean;
         /**
-         * Sets / unsets the lazy domain change on the plot.
+         * Sets / unsets the deferred rendering option
          * Activating this option improves the performance of plot interaction (pan / zoom) by
          * performing lazy renders, only after the interaction has stopped. Because re-rendering
          * is no longer performed during the interaction, the zooming might experience a small
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2545,7 +2545,7 @@ declare module Plottable {
          * Returns the lazy domain change setting on the plot
          * @return {boolean} The lazy domain change setting
          */
-        lazyDomainChange(): boolean;
+        deferredRendering(): boolean;
         /**
          * Sets / unsets the lazy domain change on the plot.
          * Activating this option improves the performance of plot interaction (pan / zoom) by
@@ -2555,7 +2555,7 @@ declare module Plottable {
          *
          * This option is intended for cases where performance is an issue.
          */
-        lazyDomainChange(lazyDomainChange: boolean): XYPlot<X, Y>;
+        deferredRendering(deferredRendering: boolean): XYPlot<X, Y>;
         /**
          * Gets the AccessorScaleBinding for X.
          */
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2541,7 +2541,20 @@ declare module Plottable {
          * @param {Scale} yScale The y scale to use.
          */
         constructor();
+        /**
+         * Returns the lazy domain change setting on the plot
+         * @return {boolean} The lazy domain change setting
+         */
         lazyDomainChange(): boolean;
+        /**
+         * Sets / unsets the lazy domain change on the plot.
+         * Activating this option improves the performance of plot interaction (pan / zoom) by
+         * performing lazy renders, only after the interaction has stopped. Because re-rendering
+         * is no longer performed during the interaction, the zooming might experience a small
+         * resolution degradation, before the lazy re-render is performed.
+         *
+         * This option is intended for cases where performance is an issue.
+         */
         lazyDomainChange(lazyDomainChange: boolean): XYPlot<X, Y>;
         /**
          * Gets the AccessorScaleBinding for X.
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2533,6 +2533,7 @@ declare module Plottable {
     class XYPlot<X, Y> extends Plot {
         protected static _X_KEY: string;
         protected static _Y_KEY: string;
+        tempScrollTimeout: number;
         /**
          * An XYPlot is a Plot that displays data along two primary directions, X and Y.
          *
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -60,8 +60,8 @@ export class XYPlot<X, Y> extends Plot {
       }
       _lastSeenDomainX = scale.domain();
       _scalingX = (scale.scale(this._cachedDomainX[1]) - scale.scale(this._cachedDomainX[0])) /
-        (scale.scale(_lastSeenDomainX[1]) - scale.scale(_lastSeenDomainX[0]));
-      _deltaX = scale.scale(this._cachedDomainX[0]) - scale.scale(_lastSeenDomainX[0]);
+        (scale.scale(_lastSeenDomainX[1]) - scale.scale(_lastSeenDomainX[0])) || 1;
+      _deltaX = scale.scale(this._cachedDomainX[0]) - scale.scale(_lastSeenDomainX[0]) || 0;
 
       _registerDeferredRendering();
     };
@@ -72,9 +72,8 @@ export class XYPlot<X, Y> extends Plot {
       }
       _lastSeenDomainY = scale.domain();
       _scalingY = (scale.scale(this._cachedDomainY[1]) - scale.scale(this._cachedDomainY[0])) /
-        (scale.scale(_lastSeenDomainY[1]) - scale.scale(_lastSeenDomainY[0]));
-      _deltaY = scale.scale(this._cachedDomainY[0]) -
-        scale.scale(_lastSeenDomainY[0]) * _scalingY;
+        (scale.scale(_lastSeenDomainY[1]) - scale.scale(_lastSeenDomainY[0])) || 1;
+      _deltaY = scale.scale(this._cachedDomainY[0]) - scale.scale(_lastSeenDomainY[0]) * _scalingY || 0;
 
       _registerDeferredRendering();
     };
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -10,11 +10,11 @@ export class XYPlot<X, Y> extends Plot {
   private _adjustXDomainOnChangeFromYCallback: ScaleCallback<Scale<any, any>>;
 
   private _lazyDomainChange = false;
-  private _fastPanZoomOnXCallback: ScaleCallback<Scale<X, any>>;
-  private _fastPanZoomOnYCallback: ScaleCallback<Scale<Y, any>>;
+  private _lazyDomainChangeCallbackX: ScaleCallback<Scale<X, any>>;
+  private _lazyDomainChangeCallbackY: ScaleCallback<Scale<Y, any>>;
 
-  private _lazyDomainingKnownDomainX: X[] = [null, null];
-  private _lazyDomainingKnownDomainY: Y[] = [null, null];
+  private _lazyDomainChangeCachedDomainX: X[] = [null, null];
+  private _lazyDomainChangeCachedDomainY: Y[] = [null, null];
 
   /**
    * An XYPlot is a Plot that displays data along two primary directions, X and Y.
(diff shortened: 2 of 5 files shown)


plottable.d.ts
```diff
@@ -2540,6 +2541,8 @@ declare module Plottable {
          * @param {Scale} yScale The y scale to use.
          */
         constructor();
+        performanceEnabled(): boolean;
+        performanceEnabled(performanceEnabled: boolean): XYPlot<X, Y>;
```
bluong: What's the point of having `performanceEnabled()` be off?

bluong:
One thing that I'm highly concerned with is that setting `performanceEnabled()` will give as a loss in features if it is set to true.  It makes sense to me that other features get slower at the cost of boosting the performance of panZoom but it doesn't make too much sense to me that features get axed due to the performance boost.

plottable.d.ts
jtlan: So the default behavior is the same as before.

bluong:
In terms of user usage, let's say I find this performanceEnabled endpoint.  Aside from looking at the documentation (which is a step we generally want to somewhat avoid), there is nothing that says that the user shouldn't leave this at false.

plottable.d.ts
aicioara: Documented above

jtlan:
@bluong wrote:
>One thing that I'm highly concerned with is that setting performanceEnabled() will give as a loss in features if it is set to true. It makes sense to me that other features get slower at the cost of boosting the performance of panZoom but it doesn't make too much sense to me that features get axed due to the performance boost.

The high-performance mode is there to offer a tradeoff: better speed at the cost of some precision / beauty during `Scale` updates. I think it makes sense to offer users the option to make that tradeoff.

@aicioara wrote:
>Maybe call performanceEnabled() something else, such that we can have finer grained performance enabling in the future

What about "`performanceMode()`"? Or something else where a `string` is passed in.

bluong:
According to @aicioara however, it seems that the issue isn't some precision, it seems the plot will straight up not update nicely when zooming happens.

aicioara:
@bluong 
Have you checked the fiddle?

src/plots/xyPlot.ts
```diff
@@ -9,6 +9,19 @@ export class XYPlot<X, Y> extends Plot {
...
+  private _fastPanZoomOnYCallback: ScaleCallback<Scale<any, any>>;
+
+  private _fastPanZoomTimeoutReferenceX = 0;
+  private _fastPanZoomTimeoutReferenceY = 0;
+  private _fastPanZoomDeltaX = 0;
+  private _fastPanZoomDeltaY = 0;
+  private _fastPanZoomScaleX = 1;
+  private _fastPanZoomScaleY = 1;
+  private _fastPanZoomKnownDomainX: X[] = [null, null];
+  private _fastPanZoomKnownDomainY: Y[] = [null, null];
```
bluong: A bunch of the above variables allude to some `_fastPanZoom` idea but nowhere in the below code references panning or zooming.  All it seems to be is scale domain updating.

aicioara:
The issue is the following: In order to keep the points to the same radius on a scatter plot (or the width of the lines on a line plot for what matters), we need to translate and not scale them. This means the translation is not the same for all points -> we cannot apply it to the parent DOM element -> we have to apply a different translate to every single point on the plot -> modify lots of DOM elements -> slow

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
...
+        "translate(" + this._fastPanZoomDeltaX + ", " + this._fastPanZoomDeltaY + ")" +
+        "scale(" + this._fastPanZoomScaleX + ", " + this._fastPanZoomScaleY + ")");
+      clearTimeout(this._fastPanZoomTimeoutReferenceX);
+      this._fastPanZoomTimeoutReferenceX = setTimeout(() => {
+        this._fastPanZoomKnownDomainX = domain;
+        this._fastPanZoomDeltaX = 0;
+        this._fastPanZoomDeltaY = 0;
+        this.render();
+        this._renderArea.attr("transform", "translate(0, 0) scale(1, 1)");
+      }, 500);
```
bluong: Why is a timeout idea here?

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
...
+
+    var domain = scale.domain();
+    this._fastPanZoomScaleX = (scale.scale(this._fastPanZoomKnownDomainX[1]) - scale.scale(this._fastPanZoomKnownDomainX[0])) /
+                              (scale.scale(domain[1]) - scale.scale(domain[0]));
+    this._fastPanZoomDeltaX = scale.scale(this._fastPanZoomKnownDomainX[0]) - scale.scale(domain[0]);
+
+    if (this._renderArea != null) {
+      this._renderArea.attr("transform",
+        "translate(" + this._fastPanZoomDeltaX + ", " + this._fastPanZoomDeltaY + ")" +
+        "scale(" + this._fastPanZoomScaleX + ", " + this._fastPanZoomScaleY + ")");
```
bluong: Where else are these instance variables use?  Why not wrap this function into a `_fastPanZoom` function()?

Also, this seems to be working with any type of scale updating.

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
...
+      clearTimeout(this._fastPanZoomTimeoutReferenceY);
+      this._fastPanZoomTimeoutReferenceY = setTimeout(() => {
+        this._fastPanZoomKnownDomainY = domain;
+        this._fastPanZoomDeltaX = 0;
+        this._fastPanZoomDeltaY = 0;
+        this.render();
+        this._renderArea.attr("transform", "translate(0, 0) scale(1, 1)");
+      }, 500);
+    }
+  }
```
bluong: Lot's of duplicated logic it seems.  Can we factor out the logic.  Also, try to reduce the number of instance variables

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
...
+
+  public performanceEnabled(): boolean;
+  public performanceEnabled(performanceEnabled: boolean): XYPlot<X, Y>;
```
bluong: If we want this endpoint, the documentation surrounding this will be ridiculously important (as well as the function name and the parameter name)

src/plots/xyPlot.ts
aicioara: if no more PanZoom happens for 500 ms (will be checked with a designer), then do a render, basically restore the dimension of the points in the plot

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
...
+      if (this.x() && this.x().scale) {
+        this._fastPanZoomKnownDomainX = this.x().scale.domain();
+        this.x().scale.onUpdate(this._fastPanZoomOnXCallback);
+        this.x().scale.offUpdate(this._renderCallback);
+      }
+      if (this.y() && this.y().scale) {
+        this._fastPanZoomKnownDomainY = this.y().scale.domain();
+        this.y().scale.onUpdate(this._fastPanZoomOnYCallback);
+        this.y().scale.offUpdate(this._renderCallback);
+      }
```
bluong: What happens if the scale is replaced after `performanceEnabled()` is set?

src/plots/xyPlot.ts
jtlan: You get a similar effect in most map programs if you try to zoom in and out too fast; you'll initially see a lower-resolution version of what you zoomed in on, then a higher-res version will load.

src/plots/xyPlot.ts
jtlan: Sounds like RFC material.

src/plots/xyPlot.ts
```diff
@@ -22,6 +35,95 @@ export class XYPlot<X, Y> extends Plot {
 
     this._adjustYDomainOnChangeFromXCallback = (scale) => this._adjustYDomainOnChangeFromX();
     this._adjustXDomainOnChangeFromYCallback = (scale) => this._adjustXDomainOnChangeFromY();
+
+    this._fastPanZoomOnXCallback = (scale) => this._fastPanZoomOnX(scale);
+    this._fastPanZoomOnYCallback = (scale) => this._fastPanZoomOnY(scale);
```
bluong: Maybe `_fastRenderCallback`?

src/plots/xyPlot.ts
bluong: We should maybe have this 500 be configurable for tests

bluong:
Based on the demo you showed, you might want to do a survey to see if this should actually be the default / only behavior

src/plots/xyPlot.ts
bluong: If you decide to keep this endpoint, maybe rename it to like `smoothScalingEnabled()`?

src/plots/xyPlot.ts
aicioara: Good catch. It fails horribly for some reason

src/plots/xyPlot.ts
aicioara: Fixed. Renamed to `lazyDomainChange[...]`

src/plots/xyPlot.ts
aicioara: done

src/plots/xyPlot.ts
aicioara: Renamed to `lazyDomainChange` and added documentation

src/plots/xyPlot.ts
aicioara: Fixed. It failed because other (now) known bugs https://github.com/palantir/plottable/issues/2396

src/plots/xyPlot.ts
aicioara: it's not a render, hence called it `lazyDomainChangeCallback`

aicioara:
Used the constructor as a closure. Could not really think of another way to do it nicely. This just means the constructor code is bigger

aicioara:
Writing an RFC for this. Code-wise, I think it is okay now.

src/plots/xyPlot.ts
```diff
@@ -135,15 +237,28 @@ export class XYPlot<X, Y> extends Plot {
     var adjustCallback = key === XYPlot._X_KEY ? this._adjustYDomainOnChangeFromXCallback
                                                : this._adjustXDomainOnChangeFromYCallback;
     scale.onUpdate(adjustCallback);
+
+    if (this._lazyDomainChange) {
+      scale.offUpdate(this._renderCallback);
```
jtlan: Instead of doing this on/off switch, wouldn't it make more sense if `_renderCallback`'s internal behavior switched on `_lazyDomainChange`?

aicioara: The issue is that we need to not change the reference of the `this._renderCallback`. This means that we cannot plug and play a new `_rendererCallback`. This means that some sort of switch has to go inside the _rendererCallback. This means that the `Plot` needs to know about the fast version (since the `rendererCallback` is defined there)

aicioara: Alternative is to do this switch inside the `lazyDomainChange`, and fetch the scales from inside there

jtlan: Could we have a separate function that switches between them only for the X and Y `Scale`s?
```Typescript
this._xScaleRenderCallback = (scale) => {
  if (this.lazyDomainChange()) {
    this._lazyDomainChangeCallbackX(scale);
  } else {
    this._renderCallback(scale);
  }
};

...

protected _installScaleForKey(scale: Scale<any, any>, key: string) {
  if (key === XYPlot._X_KEY) {
    scale.onUpdate(this._xScaleRenderCallback);
  } else {
    super._installScaleForKey(scale, key);
  }
}
```

Not sure this is better, though.

aicioara: We'll miss important stuff from the parent class (like `scale.addIncludedValuesProvider(this._includedValuesProvider);`

jtlan: Alternatively:
```Typescript
this._renderCallback = (scale) => {
  if (scale === this.x().scale && this.lazyDomainChange()) {
    this._lazyDomainChangeCallbackX(scale);
  } else if (scale === this.y().scale && this.lazyDomainChange()) {
    this._lazyDomainChangeCallbackY(scale);
  } else {
    this._render();
  }
};
```

aicioara: That looks great!

src/plots/xyPlot.ts
```diff
@@ -53,10 +152,6 @@ export class XYPlot<X, Y> extends Plot {
       this._updateYExtentsAndAutodomain();
     }
 
-    if (xScale != null) {
-      xScale.onUpdate(this._adjustYDomainOnChangeFromXCallback);
```
jtlan: ?? Why was this removed

aicioara: Redundant. `this.bindProperty()` -> `_bind()`-> `_installScaleForKey()` -> `scale.onUpdate(adjustDomainCallback)`

jtlan: Neat.

jtlan:
Pending rename of API point for enabling this.

aicioara:
Added the performance configuration concept here If we decide not to go for it, I can always reset a couple of commits back, once we figure out how to name things

aicioara:
Blocking until we find better naming

crmorford:
I'm seeing very strange behavior from horizontal bar plots, it's hard to figure out what's happening. On this test, when you pan, the size of the bars is changing unpredictably. http://jsfiddle.net/2ctcjx83/

On the quicktests (about to push) categoryAxis_timeAxis has a panning lag, which I didn't expect. The bars are 'jumping' to their end position when you stop dragging.

aicioara:
This is the same issue as here https://github.com/palantir/plottable/pull/2344
Since this is the third time this feature is signaled as a flag (twice by @crmorford and once by Slate), I believe we should eliminate the feature.
The issue was that the test was containing a 0-length bar which made it harder to see what is going on. Amended the test here http://jsfiddle.net/2ctcjx83/4/ 

For completeness, the feature that can be observed in the fiddle is the following: `Plots.Bar` tries to fit all its bars in the given space if possible. As you approach upper or lower margins the behavior can be observed. This feature cannot exist in the fast version of panzoom. 

@crmorford Let me know if the test makes sense now

aicioara:
The categoryAxis_timeAxis is an issue. Solving now. It probably has to do with grouping

aicioara:
Fixed

crmorford:
Oh yeah, the 0-length bar made it way harder to understand. Now it's much more predictable.

crmorford:
And categoryAxis_timeAxis looks good now.

aicioara:
I had a followup on this issue.

Issue: Scaling on a SVG influences both the size of the symbols and the distance between them. What we want is to influence the distance between symbols while leaving the symbol sizes untouched. 

@dcervelli suggested that we can maybe apply a CSS3 transform on the symbol class, which will do an inverse scaling on the symbol itself and that would solve the problem. There were two issues with that:
- CSS3 transforms on SVG elements are not supported in IE [known issue](https://connect.microsoft.com/IE/feedbackdetail/view/1046003/css-transforms-feature-detection-not-working-on-svg-elements)
- Applying CSS3 transforms in Chrome overwrites the SVG transform attributes (so we lose the position of the points)

PR #2391, opened 2015-06-24
[ESLint] - Applying default no-shadow rule

bluong:
Rule disallows shadowing of variables.

Invalid:
```javascript
var a = 10;
function foo(a) {
  doStuff(a);
}
```

Valid:
```javascript
var a = 10;
function foo(aTest) {
  doStuff(aTest);
}
```


quicktests/overlaying/tests/realistic/cities.js
```diff
@@ -7,7 +7,7 @@ function run(svg, data, Plottable) {
   "use strict";
 
   d3.csv("/quicktests/overlaying/data/cities.csv").get(function(error, rows) {
-  var data = rows;
+  data = rows;
```
aicioara: This looks like a breaking change. I hope the test still works

PR #2392, opened 2015-06-24
[Quicktest] - Deleting 2Up + other files

bluong:
There were a bunch of files in our quicktests directory that are essentially obsolete at this point.  From my understanding the 2Up has been completely ignored in favor of the overlaying idea for quicktests.  Keeping it around may mean that we have to maintain it.  We might as well just delete this things.

There were also quite a few files that are essentially dead.  Deleting those as well.


jtlan:
Are you going to delete the directory of old tests too?

bluong:
What directory are you referring to?

PR #2395, opened 2015-06-24
[Plots.Segment] Implementation

### LINKED ISSUE #2394: Line Segment Plot
7fa1ff70-6389-47a2-9d55-023acd396d9f

rcchen:
image

http://jsfiddle.net/qLjusc6w/

Fixes #2394


jtlan:
Our modern art repertoire expands!

src/drawers/segmentDrawer.ts
```diff
@@ -0,0 +1,18 @@
...
+export module Drawers {
+  export class Segment extends Drawer {
+
+    constructor(dataset: Dataset) {
+      super(dataset);
+      this._svgElementName = "line";
+    }
+
+    protected _applyDefaultAttributes(selection: d3.Selection<any>) {
+      super._applyDefaultAttributes(selection);
```
jtlan: What's the point of this override?

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,79 @@
...
+      var attrToProjector = super._generateAttrToProjector();
+      return attrToProjector;
+    }
+
+    protected _generateDrawSteps(): Drawers.DrawStep[] {
+      return [{attrToProjector: this._generateAttrToProjector(), animator: this._getAnimator(Plots.Animator.MAIN)}];
+    }
+
+    public x2(): AccessorScaleBinding<X, number>;
+    public x2(x2: number | Accessor<number> | X | Accessor<X>): Plots.Segment<X, Y>;
```
jtlan: Docs.

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,79 @@
...
+      this.addClass("segment-plot");
+    }
+
+    protected _createDrawer(dataset: Dataset) {
+      return new Drawers.Segment(dataset);
+    }
+
+    protected _generateAttrToProjector() {
+      var attrToProjector = super._generateAttrToProjector();
+      return attrToProjector;
```
jtlan: What's the point of this override?

jtlan:
Example Fiddle?

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    protected _generateDrawSteps(): Drawers.DrawStep[] {
+      return [{attrToProjector: this._generateAttrToProjector(), animator: this._getAnimator(Plots.Animator.MAIN)}];
```
jtlan: Not sure this looks quite right. We should stick with an `Animators.Null` until we get a spec from design for what this should look like.

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,79 @@
...
+      this.render();
+      return this;
+    }
+
+    protected _propertyProjectors(): AttributeToProjector {
+      var attrToProjector = super._propertyProjectors();
+      attrToProjector["x1"] = Plot._scaledAccessor(this.x());
+      attrToProjector["x2"] = Plot._scaledAccessor(this.x2());
+      attrToProjector["y1"] = Plot._scaledAccessor(this.y());
+      attrToProjector["y2"] = Plot._scaledAccessor(this.y2());
```
jtlan: This implies that `x2` and `y2` are mandatory, unlike on `Plots.Rectangle`. Is that true?

rcchen:
In which I once again forget to remove stuff that was added when hacking out the initial impl...

src/plots/segmentPlot.ts
rcchen: I implemented it as mandatory, but now come to think of it, it is possible for either one to be optional (not both though, as you end up with a point).

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @param {number|Accessor<number>|Y|Accessor<Y>} y2
+     * @returns {Plots.Segment} The calling Segment Plot
+     */
+    public x2(x2: number | Accessor<number> | X | Accessor<X>): Plots.Segment<X, Y>;
+    public x2(x2?: number | Accessor<number> | X | Accessor<X>): any {
+      if (x2 == null) {
+        return this._propertyBindings.get(Segment._X2_KEY);
+      }
+      var xBinding = this.x();
+      var xScale = xBinding && xBinding.scale;
```
jtlan: If the user sets `x2()` before `x()`, they won't get the `Scale`, although I'll note that this is an open bug on `Plots.Rectangle` as well.

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @param {number|Accessor<number>|Y|Accessor<Y>} y2
+     * @returns {Plots.Segment} The calling Segment Plot.
+     */
+    public y2(y2: number | Accessor<number> | Y | Accessor<Y>): Plots.Segment<X, Y>;
+    public y2(y2?: number | Accessor<number> | Y | Accessor<Y>): any {
+      if (y2 == null) {
+        return this._propertyBindings.get(Segment._Y2_KEY);
+      }
+      var yBinding = this.y();
+      var yScale = yBinding && yBinding.scale;
```
jtlan: Same `Scale` problem here.

src/plots/segmentPlot.ts
```diff
@@ -0,0 +1,146 @@
...
+     * @returns {Plots.Segment} The calling Segment Plot
+     */
+    public x2(x2: number | Accessor<number> | X | Accessor<X>): Plots.Segment<X, Y>;
+    public x2(x2?: number | Accessor<number> | X | Accessor<X>): any {
+      if (x2 == null) {
+        return this._propertyBindings.get(Segment._X2_KEY);
+      }
+      var xBinding = this.x();
+      var xScale = xBinding && xBinding.scale;
+
```
jtlan: Whitespace here, but not in `y2()`?

plottable.css
```diff
@@ -158,3 +158,8 @@ svg.plottable {
   stroke: #CCC;
   stroke-width: 1px;
 }
+
+.plottable .segment-plot line {
+  stroke: #555;
+  stroke-width: 1px;
```
jtlan: Do we have design sign-off on this? I would instead default to the same color other `Plot`s do, using an `attr()` call in the `constructor`.

crmorford:
Is there anything we can do about the different width of different segments? Seems like an aliasing thing. Compare the first line (furthest left) to any other line.
screen shot 2015-06-25 at 3 43 00 pm

rcchen:
Dunno if it's avoidable. http://bl.ocks.org/mbostock/1322948

jtlan:
It'd be `grunt eslint` now right?

crmorford:
+1 pending on me...awkward

PR #2397, opened 2015-06-25
[Gitignore] - ignoring tests.js

bluong:
One thing that felt really incredibly odd to me is the fact that we had a bunch of built files that were checked into the repo.

Not only would this often cause merge conflicts, but it would make git management awkward when working around these files.  Note that our build directory is completely ignored and that at least in my experience with java projects, the build files are not checked in.  They are just, well, built before usage.

For plottable.js and plottable.d.ts, these are files that we export out to customers and these are also files that we test against when quality testing.  I don't currently have a solution for this, but I'll think about it as time goes on.  For now, tests.js is not exported to anyone, and it is mainly used for generating our unit test suite.  I think it should be safe to not have in the repo.


aicioara:
Sounds reasonable to me

PR #2399, opened 2015-06-25
[ESLint] - applying default usage of strict

bluong:
This forces javascript files to be in "strict mode".

Invalid:
```javascript
function start() {
//doStuff
}
```

Valid:
```javascript
function start() {
"use strict";
//doStuff
}
```


quicktests/umd/app.js
```diff
@@ -1,4 +1,5 @@
 /*jshint strict:false */
+/*eslint strict: 0*/
```
aicioara: Can we just `"use strict"` instead of disabling the rule?

bluong: Note that there is some message about jshint being set to false...

@jtlan Can you chime in on why this should not be in strict mode?

jtlan: I think it's because you have to use "`require`" before it's declared?

quicktests/umd/app.js
```diff
@@ -1,6 +1,4 @@
-/*jshint strict:false */
-
-var require;
+/*eslint-env node */
```
aicioara: `/*eslint-env amd*/` ?

bluong: Good catch

PR #2409, opened 2015-06-26
Deprecate Plottable.MILLISECONDS_IN_ONE_DAY & Formatters.relativeDate()

### LINKED ISSUE #2402: Deprecate MILLISECONDS_IN_ONE_DAY
Remove all usage from code and/or mark functions that use it as deprecated.

We can't remove it until v2 due to API breaking.

dbt:
Also remove from non-deprecated src.

Fixes #2402

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -596,7 +596,7 @@ declare module Plottable {
      * The number of milliseconds between midnight one day and the next is
      * not a fixed quantity.
      *
-     * use date.setDate(Date.getDate() + n) instead.
+     * Use date.setDate(date.getDate() + number_of_days) instead.
      *
      */
     var MILLISECONDS_IN_ONE_DAY: number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -596,7 +596,7 @@ declare module Plottable {
      * The number of milliseconds between midnight one day and the next is
      * not a fixed quantity.
      *
-     * use Date#setDate(Date#getDate() + n) instead.
+     * use date.setDate(Date.getDate() + n) instead.
      *
      */
     var MILLISECONDS_IN_ONE_DAY: number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -590,6 +590,15 @@ declare module Plottable {
 
 declare module Plottable {
     type Formatter = (d: any) => string;
+    /**
+     * This field is deprecated and will be removed in v2.0.0.
+     *
+     * The number of milliseconds between midnight one day and the next is
+     * not a fixed quantity.
+     *
+     * use Date#setDate(Date#getDate() + n) instead.
+     *
+     */
     var MILLISECONDS_IN_ONE_DAY: number;
     module Formatters {
         /**
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -4,6 +4,15 @@ module Plottable {
 
 export type Formatter = (d: any) => string;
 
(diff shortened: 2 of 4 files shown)


plottable.d.ts
```diff
@@ -590,6 +590,15 @@ declare module Plottable {
 
 declare module Plottable {
     type Formatter = (d: any) => string;
+    /**
+     * This field is deprecated and will be removed in v2.0.0.
```
bluong: We should put when the field has been deprecated.

plottable.d.ts
```diff
@@ -590,6 +590,15 @@ declare module Plottable {
...
+     *
+     * The number of milliseconds between midnight one day and the next is
+     * not a fixed quantity.
+     *
+     * use Date#setDate(Date#getDate() + n) instead.
+     *
+     */
     var MILLISECONDS_IN_ONE_DAY: number;
```
bluong: We should put when the field has been deprecated.

src/core/formatters.ts
```diff
@@ -199,6 +208,7 @@ export module Formatters {
    * @returns {Formatter} A formatter for time/date values.
    */
   export function relativeDate(baseValue: number = 0, increment: number = MILLISECONDS_IN_ONE_DAY, label: string = "") {
+    Plottable.Utils.Window.deprecated("relativeDate()", "1.2", "Not safe for use with time zones.");
```
bluong: Is there some form of satisfiable replacement that we can list in the message parameter?  One reason the message parameter exists is so we can put messages on what the user should use instead.

The lucky thing in this particular instance is that since Formatters are so lightweight, we probably can tell users to just "make their own".  Not sure if this is the right route to go.

src/scales/timeScale.ts
```diff
@@ -38,8 +38,10 @@ export module Scales {
...
 
     protected _defaultExtent(): Date[] {
-      var endTimeValue = new Date().valueOf();
-      var startTimeValue = endTimeValue - MILLISECONDS_IN_ONE_DAY;
+      var now = new Date();
+      var endTimeValue = now.valueOf();
+      now.setDate(now.getDate() - 1);
+      var startTimeValue = now.valueOf();
       return [new Date(startTimeValue), new Date(endTimeValue)];
     }
```
bluong: This looks like a bug fix.  Can we get a test up in regards to the bug fix?

src/core/formatters.ts
```diff
@@ -4,6 +4,15 @@ module Plottable {
 
 export type Formatter = (d: any) => string;
 
+/**
+ * This field is deprecated and will be removed in v2.0.0.
+ * 
+ * The number of milliseconds between midnight one day and the next is 
+ * not a fixed quantity.
+ * 
+ * use Date#setDate(Date.getDate() + n) instead.
```
jtlan: Shouldn't this be a dot instead of a "#"? As in: "`Use date.setDate(date.getDate() + numberOfDays) instead.`".

src/scales/timeScale.ts
jtlan: Looks like the test would probably just be the same implementation as this method, so I'm not sure it'd be helpful.

crmorford:
This is a regression pass, correct?

rcchen:
@crmorford looks like you just want to be checking if time-related formatters continue to work correctly, so regression pass.

crmorford:
Thanks @rcchen!

PR #2412, opened 2015-06-26
Ignore clicks not directed at Plottable graphs. Fixes #2142

### LINKED ISSUE #2142: Click Interactions work behind overlays
Interactions can be detected through overlays:
http://jsfiddle.net/ohk89gth/

**Actual**: Clicking anywhere in the `<svg>` triggers a message.
**Expected**: Clicking on the part of the `<svg>` covered by the overlay doesn't trigger anything.

raiju:
Fixes #2142

### DIFF
--- a/src/dispatchers/mouseDispatcher.ts
+++ b/src/dispatchers/mouseDispatcher.ts
@@ -54,12 +54,12 @@ export module Dispatchers {
       this._callbacks = [this._moveCallbacks, this._downCallbacks, this._upCallbacks, this._wheelCallbacks,
                          this._dblClickCallbacks];
 
-      var processMoveCallback = (e: MouseEvent) => this._measureAndDispatch(e, this._moveCallbacks);
+      var processMoveCallback = (e: MouseEvent) => this._measureAndDispatch(e, this._moveCallbacks, "page");
       this._eventToCallback["mouseover"] = processMoveCallback;
       this._eventToCallback["mousemove"] = processMoveCallback;
       this._eventToCallback["mouseout"] = processMoveCallback;
       this._eventToCallback["mousedown"] = (e: MouseEvent) => this._measureAndDispatch(e, this._downCallbacks);
-      this._eventToCallback["mouseup"] = (e: MouseEvent) => this._measureAndDispatch(e, this._upCallbacks);
+      this._eventToCallback["mouseup"] = (e: MouseEvent) => this._measureAndDispatch(e, this._upCallbacks, "page");
       this._eventToCallback["wheel"] = (e: WheelEvent) => this._measureAndDispatch(e, this._wheelCallbacks);
       this._eventToCallback["dblclick"] = (e: MouseEvent) => this._measureAndDispatch(e, this._dblClickCallbacks);
     }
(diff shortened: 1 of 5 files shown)


src/dispatchers/mouseDispatcher.ts
```diff
@@ -178,11 +178,13 @@ export module Dispatchers {
      * Computes the mouse position from the given event, and if successful
      * calls all the callbacks in the provided callbackSet.
      */
-    private _measureAndDispatch(event: MouseEvent, callbackSet: Utils.CallbackSet<MouseCallback>) {
-      var newMousePosition = this._translator.computePosition(event.clientX, event.clientY);
-      if (newMousePosition != null) {
-        this._lastMousePosition = newMousePosition;
-        callbackSet.callCallbacks(this.lastMousePosition(), event);
+    private _measureAndDispatch(event: MouseEvent, callbackSet: Utils.CallbackSet<MouseCallback>, captureAllEvents: boolean = false) {
```
dbt: Please don't use boolean parameters in methods.

http://blogs.msdn.com/b/oldnewthing/archive/2006/08/28/728349.aspx

raiju:
Would a scope: string with "PAGE"/"ELEMENT" (or "CAPTURE_PAGE" and "CAPTURE_ELEMENT") be preferable? Or an enum? (In that case, where do you want the enum to live?)

dbt:
In general we're using lowercase strings for these, and if it's a public method you should expose them as constants.  You don't need to do that here.  I'd lean towards "page"/"element" over capture_...

src/utils/clientToSVGTranslator.ts
```diff
@@ -70,6 +70,18 @@ export module Utils {
 
       return scaledPosition;
     }
+    public isWithinTarget(e: MouseEvent) {
+      return this._isChildOf(<Element>e.target, this._svg);
```
jtlan: You could probably use an existing `Utils` method here:
```Typescript
return Utils.DOM.boundingSVG(<SVGElement>e.target) === this._svg);
```

src/dispatchers/mouseDispatcher.ts
```diff
@@ -178,11 +178,13 @@ export module Dispatchers {
      * Computes the mouse position from the given event, and if successful
      * calls all the callbacks in the provided callbackSet.
      */
-    private _measureAndDispatch(event: MouseEvent, callbackSet: Utils.CallbackSet<MouseCallback>) {
-      var newMousePosition = this._translator.computePosition(event.clientX, event.clientY);
-      if (newMousePosition != null) {
-        this._lastMousePosition = newMousePosition;
-        callbackSet.callCallbacks(this.lastMousePosition(), event);
+    private _measureAndDispatch(event: MouseEvent, callbackSet: Utils.CallbackSet<MouseCallback>, scope: string = "element") {
```
dbt: One tiny nit.  We should validate the input is "element" or "page".

dbt:
lgtm

src/utils/clientToSVGTranslator.ts
```diff
@@ -70,6 +70,9 @@ export module Utils {
 
       return scaledPosition;
     }
+    public isWithinTarget(e: MouseEvent) {
```
jtlan: Probably should be renamed to "`insideSVG()`", and take a generic `Event`. (Renaming it also makes the method self-documenting).

src/dispatchers/mouseDispatcher.ts
jtlan: Looks like this has been addressed.

crmorford:
Tested:
- [x] click
- [x] panzoom
- [ ] double click
- [ ] drag

jtlan:
@crmorford 
Two NOTEs probably have the same root cause. I'd expect no `Interaction`s to trigger on top of the modal, so there's probably a few corrections to be made here.

crmorford:
Ok, I'm going to QE -1 then and pass this back to @raiju.

bluong:
@jtlan Note that both the NOTEs occur on mouseover/out rather than on clickDown/clickUp

jtlan:
Taking care of `Pointer` and `Key` probably requires some mechanism where they don't respond to `Dispatcher` updates if the dispatched event isn't targeted at the `<svg>`, which means that the `Dispatcher` probably needs to pass that information to its listeners somehow.

crmorford:
Can I get a summary of what happened?
Tests were added, yes? And pointer/key weren't fixed. Any other functionality changes?

raiju:
Oh, sorry. Yes, I added a test which tests whether events on an overlay above the SVG are captured by the underlying SVG. No other functionality changes were made though.

crmorford:
Wow, thanks for the quick response :)

PR #2423, opened 2015-07-01
[ESLint] - applying default usage of no-console

bluong:
Enforces that console statements are not used.  Especially for `console.log`, these statements are generally not helpful in production level code.  Albeit, they are useful at development level.  We can decide if we want to utilize this or not.

At the very least, we should probably use Plottable's warn so that we don't run into IE issues


quicktests/overlaying/overlaying.js
```diff
@@ -217,10 +217,6 @@ function loadAllQuickTests(quicktestsPaths, firstQTBranch, secondQTBranch){
   quicktestsPaths.forEach(function(path) { //for each quicktest
     var name = path.replace(/\w*\/|\.js/g, "");
     d3.text("http://localhost:9999/" + path, function(error, text) {
-      if (error !== null) {
-        console.warn("Tried to load nonexistant quicktest ");
```
aicioara: `throw new Error("Tried to load nonexistant quicktest")`. Being silent while testing is not a good thing.

PR #2424, opened 2015-07-01
Numeric Axis does not wrap the labels anymore.

aicioara:


jtlan:
Explanation: `Axis.Numeric` never wraps its labels anyway, so there's no purpose to the code that was removed.

PR #2425, opened 2015-07-01
[manyPlotsPerf] No more redraw on numeric axis rescale

aicioara:
It seemed like it was some sort of optimization, but in reality the `_computeWidth()` was an expensive operation in itself. All tests are still passing (including overlaying)

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -4338,13 +4338,6 @@ var Plottable;
                 if (!this._isSetup) {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


jtlan:
This change means that the `Axis` no longer redraws if the width would change (by virtue of zooming in or panning to a region with larger numbers). That seems like a regression in functionality?

aicioara:
Reopened this here https://github.com/palantir/plottable/pull/2432

PR #2426, opened 2015-07-01
Implement Plots.Waterfall

### LINKED ISSUE #1853: Support for waterfall charts

### LINKED ISSUE #1862: .project('class') on waterfall charts clobbers styling
The style classes for waterfall charts:
```
.waterfall-base {
    fill: #039BE5;
}
.waterfall-decline {
    fill: #D32F2F;
}
.waterfall-growth {
    fill: #43A047;
}
```
gets overwritten if the user uses `Plot.project('class', function(){...})` to add their own classes. Would it be possible to make a way to only add classes rather than determine them completely?

rcchen:
image

Based off of `Plots.Bar`, `Plots.Waterfall` has the following API points in addition to what is expected from `Plots.Bar`:
- `.connectorsEnabled()`/`.connectorsEnabled(enabled)` - gets/sets whether connector lines should be enabled or not.
- `.total()`/`.total(total)` - gets/sets the accessor for whether each bar represents a total or not. Must return a boolean.

Data model:

``` javascript
{
  x: "A", // xValue for each bar. Will be interpreted as a category
  y: 150, // yValue for each bar. Can be a total or a delta.
  t: "total" // the type of each bar. You can put anything here but its accessor should return a boolean
}
```

Fiddle: http://jsfiddle.net/a3mmsbdt/

Closes #1853 
Closes #1862


jtlan:
I think you could save yourself some trouble by extending `Plots.Bar` instead of `XYPlot`.

joyt:
Hey, I really like the idea of having the `total()` accessor (instead of my "group label" idea). I would suggest renaming `total()` to `isTotal()` as `total()` can be construed in a lot of ways completely unrelated to waterfall charts (and one would expect it to take/return a number, not a bool). Along that line of thinking, an even clearer name would be `isDelta()` IMO.

jtlan:
@joyt re: "`isTotal()`"/"`isDelta()`":

I think the names with "is" might be confusing in the case where we're using the method to get the current accessor:
```Typescript
var totalAccessor = waterfall.isTotal(); // does this return a boolean?
```
A cursory reading might suggest that "`isTotal()`" returns some information about the `Waterfall`, rather than returning an `AccessorScaleBinding`.

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+     * Gets the AccessorScaleBinding for whether a bar represents a total or a delta.
+     */
+    public total<T>(): Plots.AccessorScaleBinding<T, boolean>;
+    /**
+     * Sets total to a constant number or the result of an Accessor
+     * 
+     * @param {boolean|Accessor<number>}
+     * @returns {Waterfall} The calling Waterfall Plot.
+     */
+    public total(total: boolean | Accessor<boolean>): Waterfall<X, Y>;
```
jtlan: Is there any value in having this take a `boolean`? If `true` is passed in, `Waterfall` just becomes a `Bar` `Plot`, and if `false` is passed in, I'm not sure it even works.

Related: what happens if the `Accessor` evaluates to `false` on the first element?

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Sets total to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing
+     * 
+     * @param {T|Accessor<T>}
+     * @param {Scale<T, number>} scale
+     * @returns {Waterfall} The calling Waterfall Plot.
+     */
+    public total<T>(total: boolean | Accessor<T>, scale: Scale<T, boolean>): Waterfall<X, Y>;
```
jtlan: I don't think we need this signature either -- there aren't any `Scale`s in Plottable that return `boolean`s!

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+          return Math.abs(yScale.scale(<any> currentValue) - yScale.scale(<any> 0));
+        } else {
+          var currentSubtotal = this._subtotals[i];
+          var priorSubtotal = this._subtotals[i - 1];
+          var height = Math.abs(yScale.scale(<any> currentSubtotal) - yScale.scale(<any> priorSubtotal));
+          return height;
+        }
+      };
+
+      attrToProjector["class"] = (d, i, dataset) => {
```
jtlan: This will bomb any classes the user sets with `attr()`, right?

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+      }
+    }
+
+    protected _generateAttrToProjector() {
+      var attrToProjector = super._generateAttrToProjector();
+
+      var yScale = this.y().scale;
+      var totalAccessor = Plot._scaledAccessor(this.total());
+
+      attrToProjector["y"] = (d, i, dataset) => {
```
jtlan: This bombs a user-set `attr("y", ...).

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+          var priorSubtotal = this._subtotals[i - 1];
+          if (currentSubtotal > priorSubtotal) {
+            return yScale.scale(<any> currentSubtotal);
+          } else {
+            return yScale.scale(<any> priorSubtotal);
+          }
+        }
+      };
+
+      attrToProjector["height"] = (d, i, dataset) => {
```
jtlan: This bombs a user-set `attr("height", ...).

jtlan:
I'm not sure we should be setting colors in the CSS classes we apply -- they essentially override `attr()` without any clear way for the user to figure out what happened.

src/plots/waterfallPlot.ts
rcchen: image

(it's not happy)

@joyt what is the expected behavior if the first bar is not a total bar?

src/plots/waterfallPlot.ts
rcchen: No property binding sans scale though?

src/plots/waterfallPlot.ts
rcchen: Yeah (reported in #1862 from before), need to figure out a way to resolve this one..

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+
+module Plottable {
+export module Plots {
+  export class Waterfall<X, Y> extends Bar<X, Y> {
+    private static _BAR_DECLINE_CLASS = "waterfall-decline";
+    private static _BAR_GROWTH_CLASS = "waterfall-growth";
+    private static _BAR_TOTAL_CLASS = "waterfall-total";
+    private static _CONNECTOR_CLASS = "connector";
+    private static _CONNECTOR_AREA_CLASS = "connector-area";
+    private static _TOTAL_KEY = "total";
```
aicioara: For completeness add `_DELTA_KEY`, even though we do not use it?

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public total<T>(): Plots.AccessorScaleBinding<T, boolean>;
+    /**
+     * Sets total to a constant number or the result of an Accessor
+     * 
+     * @param {Accessor<boolean>}
+     * @returns {Waterfall} The calling Waterfall Plot.
+     */
+    public total(total: Accessor<boolean>): Waterfall<X, Y>;
+    public total<T>(total?: Accessor<boolean> | T | Accessor<T>): any {
```
aicioara: Why `|T | Accessor<T>`? I don't see that in any of the above signatures.

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+      if (total === undefined) {
```
aicioara: `total == null` to be consistent with the rest of the codebase

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    protected _generateAttrToProjector() {
+      var attrToProjector = super._generateAttrToProjector();
+
+      var yScale = this.y().scale;
+      var totalAccessor = Plot._scaledAccessor(this.total());
+
+      var yAttr = this.attr("y");
+      if (yAttr === undefined) {
```
aicioara: `== null`

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+              return yScale.scale(<any> currentSubtotal);
+            } else {
+              return yScale.scale(<any> priorSubtotal);
+            }
+          }
+        };
+      }
+
+      var heightAttr = this.attr("height");
+      if (heightAttr === undefined) {
```
aicioara: `== null`

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+        attrToProjector["height"] = (d, i, dataset) => {
+          var isTotal = totalAccessor(d, i, dataset);
+          var currentValue = this.y().accessor(d, i, dataset);
+          if (isTotal) {
+            return Math.abs(yScale.scale(<any> currentValue) - yScale.scale(<any> 0));
```
aicioara: Are the `<any>` casts necessary?

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+          } else {
+            var currentSubtotal = this._subtotals[i];
+            var priorSubtotal = this._subtotals[i - 1];
+            var height = Math.abs(yScale.scale(<any> currentSubtotal) - yScale.scale(<any> priorSubtotal));
```
aicioara: same for the `<any>` casts

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+            return height;
```
aicioara: Inline to be consistent with the above

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,195 @@
...
+          }
+        };
+      }
+
+      attrToProjector["class"] = (d, i, dataset) => {
+        var baseClass = "";
+        if (this.attr("class") !== undefined) {
```
aicioara: `!= null`

test/plots/waterfallPlotTests.ts
```diff
@@ -0,0 +1,122 @@
...
+  describe("Waterfall Plot", () => {
+    var svg: d3.Selection<void>;
+    var dataset: Plottable.Dataset;
+    var xScale: Plottable.Scales.Category;
+    var yScale: Plottable.Scales.Linear;
+    var plot: Plottable.Plots.Waterfall<string, number>;
+    var renderArea: d3.Selection<void>;
+    var SVG_WIDTH = 600;
+    var SVG_HEIGHT = 400;
+	  var data = [
```
aicioara: Indentation

test/plots/waterfallPlotTests.ts
```diff
@@ -0,0 +1,122 @@
...
+          if (currentTotal || currentValue < 0) {
+            assert.isTrue(+previousBar.attr("y") === +currentBar.attr("y"), "bars are top/top aligned");
+          } else {
+            assert.isTrue(+previousBar.attr("y") === +currentBar.attr("y") + +currentBar.attr("height"),
+              "bars are top/bottom aligned");
+          }
+        } else {
+          if (previousValue > 0) {
+            if (currentTotal || currentValue < 0) {
+              assert.isTrue(+previousBar.attr("y") === +currentBar.attr("y"), "bars are top/top aligned");
```
aicioara: Why is the coercion required?

test/testReference.ts
```diff
@@ -42,6 +42,7 @@
 ///<reference path="plots/stackedBarPlotTests.ts" />
 ///<reference path="plots/clusteredBarPlotTests.ts" />
 ///<reference path="plots/segmentPlotTests.ts" />
+///<reference path="plots/waterfallPlotTests.ts" />
```
aicioara: Why did this not get added to the `test.js` file?

test/plots/waterfallPlotTests.ts
```diff
@@ -0,0 +1,122 @@
...
+          if (yAccessor(data[index], index, dataset) > 0) {
+            assert.isTrue(selection.classed("waterfall-growth"));
+          } else {
+            assert.isTrue(selection.classed("waterfall-decline"));
+          }
+        }
+      });
+      svg.remove();
+    });
+    it("renders connector lines correctly", () => {
```
aicioara: can we have a \n between tests?

test/plots/waterfallPlotTests.ts
```diff
@@ -0,0 +1,122 @@
...
+      plot.connectorsEnabled(true);
+      plot.renderTo(svg);
+      var bars = renderArea.selectAll("rect")[0];
+      var connectors = renderArea.selectAll("line.connector")[0];
+      assert.isTrue(bars.length - 1 === connectors.length, "there is one more bar than num. connectors");
```
aicioara: no abbreviation.

aicioara:
Mainly style issues. The logic seems all fine. Fix IE9 issue.

src/plots/waterfallPlot.ts
rcchen: One way to address this is by extending Bar<X, number> instead of Bar<X, Y>. If Plots.Waterfall only support the vertical orientation this is OK.

test/plots/waterfallPlotTests.ts
rcchen: .attr returns a string value, not a number

src/plots/waterfallPlot.ts
aicioara: Extending what?

rcchen: Sorry forgot that \< and \> needed to be escaped in Markdown

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+    public connectorsEnabled(): boolean;
+    /**
+     * Sets whether connectors are enabled.
+     * 
+     * @param {boolean} enabled
+     * @returns {Waterfall} The calling Waterfall Plot.
+     */
+    public connectorsEnabled(enabled: boolean): Waterfall<X, Y>;
+    public connectorsEnabled(enabled?: boolean): any {
+      if (enabled === undefined) {
```
aicioara: Sorry, I missed this one `enabled == null`

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Gets the AccessorScaleBinding for whether a bar represents a total or a delta.
+     */
+    public total<T>(): Plots.AccessorScaleBinding<T, boolean>;
+    /**
+     * Sets total to a constant number or the result of an Accessor
+     * 
+     * @param {Accessor<boolean>}
+     * @returns {Waterfall} The calling Waterfall Plot.
```
aicioara: `@returns Plots.Waterfall`

crmorford:
@aicioara what was the ie9 issue?

rcchen:
@crmorford it wasn't doing number addition correctly, so I changed the tests to have a threshold of 1 as opposed to a+b === c exactly

crmorford:
on this: http://jsfiddle.net/a3mmsbdt/1/
i made the decline very large, so that the next  "total" would be negative. Is this how waterfall charts work? I'd expect it to look like a bar growing down from 0, but instead it grows down from the height of the last bar.
screen shot 2015-07-06 at 2 39 17 pm

crmorford:
It doesn't look like you can start out with deltas. http://jsfiddle.net/a3mmsbdt/3/

rcchen:
@crmorford starting off with deltas doesn't have any meaning because there's no baseline "total" value to work from.

crmorford:
@rcchen i'm not sure if I like that. There are a lot of images that come up where the chart starts with deltas when you google search waterfall chart.

Would it be possible to have a default start value of 0, and have some endpoint to change that, if necessary?

@dbt @christinaforney, what's your opinion on this?

crmorford:
Here's some examples I found online:
deltas before total: https://www.freakalytics.com/blog/2009/11/17/wc/ and https://en.wikipedia.org/wiki/Waterfall_chart

Deltas only: http://2.bp.blogspot.com/-rAVQnOBS2kI/UbnRWPbNS4I/AAAAAAAAAxA/Zw6xhcKj8_c/s1600/Chart.PNG

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (heightAttr == null) {
+        attrToProjector["height"] = (d, i, dataset) => {
+          var isTotal = totalAccessor(d, i, dataset);
+          var currentValue = this.y().accessor(d, i, dataset);
+          if (isTotal) {
+            return Math.abs(yScale.scale(currentValue) - yScale.scale(0));
+          } else {
+            var currentSubtotal = this._subtotals[i];
+            if (i === 0) {
+              return Math.abs(yScale.scale(Math.abs(currentValue)) - yScale.scale(0));
```
aicioara: I don't think you need the inner Math.abs

aicioara: Hmm, what happens when the first value is not 0-based (say the entire waterfall plot starts from value 100). 

Say the first delta is 100-200. Will this have height 200? 

To my understanding the height attr should be the height of the bar (so a delta from 100 to 200 should have height 100?

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,219 @@
...
+        }
+        this._subtotals.push(total);
+        if (total < min) {
+          min = total;
+        }
+        if (total > max) {
+          max = total;
+        }
+        if (startValue === -Infinity && isTotal) {
+          var adjustment = currentValue - total;
```
aicioara: `adjustment` is not very descriptive. I believe this is the offset of all deltas. Call it `startValue` or something more descriptive

src/plots/waterfallPlot.ts
```diff
@@ -0,0 +1,219 @@
...
+      var dataset = this.datasets()[0];
+      for (var datumIndex = 1; datumIndex < dataset.data().length; datumIndex++) {
+        var prevIndex = datumIndex - 1;
+        var datum = dataset.data()[datumIndex];
+        var prevDatum = dataset.data()[prevIndex];
+        var x = attrToProjector["x"](prevDatum, prevIndex, dataset);
+        var x2 = attrToProjector["x"](datum, datumIndex, dataset) + attrToProjector["width"](datum, datumIndex, dataset);
+        var y = attrToProjector["y"](datum, datumIndex, dataset);
+        if ((this._subtotals[datumIndex] > 0 && this._subtotals[datumIndex] > this._subtotals[prevIndex]) ||
+            (this._subtotals[datumIndex] < 0 && this._subtotals[datumIndex] >= this._subtotals[prevIndex])) {
```
aicioara: This looks dodgy. Why is this `>=` and before it was just `>`?

rcchen: This used to be a couple of nested if statements that was eventually consolidated down to this.

- If your current sum is positive and has a greater subtotal than the prior bar, you'll want to draw from the bottom of your current bar (first case)
- If your current sum is negative and has a greater or equal subtotal relative to the prior bar, you'll also want to draw from the bottom of your current bar (second case)

The reason the second one looks dodgy is because it's essentially the inverted case to the top one (<=) but ends up getting flipped in the other direction due to operating in negative space.

crmorford:
If all the data is "totals", it doesn't render any bars
http://jsfiddle.net/a3mmsbdt/4/
screen shot 2015-07-07 at 5 43 36 pm

crmorford:
Because this is derived from Plots.Bar, users might expect 
* plot.labelsEnabled
* new Plottable.Plots.Waterfall("horizontal")
to work.

Is there a way to remove these from documentation on the website when it goes up?

rcchen:
@crmorford the "totals" issue is because it adjusts the extent of the yScale to fit the range of deltas, which does not deviate from 0 in your example. Adding `yScale.domain([-30, 50]);` will manually address it.

crmorford:
All the issues I still have will be covered by documentation, and a statement that this is an 'experimental' plot.

Documentation should cover:
* Incorrect totals will not throw a warning
* API may change
* labelsEnabled is not implemented
* There is no horizontal orientation

I can write this if we want, but it should hit the website the same time as waterfall plot is released.

aicioara:
> - API may change

We should probably not state that. It sounds like the plot API can break at any point. All pre 2.0 releases will be backward compatible with the previous version and also we should not make any promises about the future of this plot as we are unsure of future priorities. 

Rest sounds sensible. Do we want to release this in 1.3 and update the website on Friday?

aicioara:
Let's create an experimental branch, always up to date with master, where we dump experimental features? Not really sure what is the best approach for this, such that it is scalable, and maintainable

rcchen:
I chatted with @christinaforney and we're using the feature branch idea for now, where people interested in this feature can pick this off of the specific branch. I merged 1.2.0 into it so it's master + this feature.

aicioara:
Sounds good

rcchen:
Checked off with @joyt, going to go ahead and merge it in now. Will be available in the release a week from now.

aicioara:
no more experimental branch?

rcchen:
We shipped a feature branch for testing, it was evaluated to work properly, so it merged back into develop.

freedomflyer:
Had to update fiddle to fix plottable.js/css links:
http://jsfiddle.net/rLwo8eyk/

erichiller:
As @crmorford said, I've a number of scenarios in which non-zero start values would apply. Should I put this in as a feature request or is it dead in the water at this point?

rcchen:
@erichiller please put it in as a new issue/FR, thanks!

PR #2427, opened 2015-07-02
add `callUpdateHandler` for improved performance when adding many datasets

### LINKED ISSUE #2418: Unnecessary quadratic slowdown when adding datasets to stacked plot
## Problem

If I have 150 datasets in `myDatasets`, and I call

``` javascript
new Plottable.Plots.StackedArea().datasets(myDatasets);
```

the resulting computation can take over 30 seconds on a powerful machine.
## Cause

The cause of this slow-down is that `.datasets` calls  `.addDataset` for each dataset, and every call to `.addDataset` ends up triggering a call to `Plottable.Utils.Stacking.stack`. Every call to `.stack` recomputes the stacking from scratch. It never takes more than ~0.3sec, but it's being called for every added dataset, so the time adds up.

_Stack trace:_

```
stack @ plottable.js:663
StackedArea._updateStackExtentsAndOffsets @ plottable.js:8124
StackedArea._updateExtentsForProperty @ plottable.js:8103
(anonymous function) @ plottable.js:6127
d3_class.forEach @ d3.js:341
Plot._updateExtents @ plottable.js:6127
Plot._onDatasetUpdate @ plottable.js:6035
Area._onDatasetUpdate @ plottable.js:7851
StackedArea._onDatasetUpdate @ plottable.js:8097
Plot.addDataset @ plottable.js:6015
Area.addDataset @ plottable.js:7860
(anonymous function) @ plottable.js:6228
Plot.datasets @ plottable.js:6228
```
## Proposed solution

The easiest way I can think of to fix this would be to add an additional argument to `.addDataset` called `suppressUpdates` (or something less silly). If provided, it would tell `.addDataset` not to call `._onDatasetUpdate()`. Then `.datasets` could choose to suppress updates and call `._onDatasetUpdate()` itself after all the datasets were added correctly.

Thoughts? I'd be happy to submit a PR myself, or to let someone else take care of this. :stuck_out_tongue:

jhorowitz-coursera:
Resolves #2418. I'm not sure if there are additional build steps I should perform in this PR, or what the process for that is. Please let me know if I missed something. Thanks!

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -100,9 +100,11 @@ export class Plot extends Component {
    * Adds a Dataset to the Plot.
    *
    * @param {Dataset} dataset
+   * @param {boolean} callUpdateHandler Whether the typical on-dataset-update handler should be called.
+   *   If not provided, defaults to true.
    * @returns {Plot} The calling Plot.
    */
-  public addDataset(dataset: Dataset) {
+  public addDataset(dataset: Dataset, callUpdateHandler: boolean = true) {
     if (this.datasets().indexOf(dataset) > -1) {
       this.removeDataset(dataset);
     };
@@ -114,7 +116,9 @@ export class Plot extends Component {
     }
 
     dataset.onUpdate(this._onDatasetUpdateCallback);
-    this._onDatasetUpdate();
+    if (callUpdateHandler) {
+      this._onDatasetUpdate();
+    }
     return this;
   }
 
@@ -403,7 +407,8 @@ export class Plot extends Component {
       return currentDatasets;
     }
     currentDatasets.forEach((dataset) => this.removeDataset(dataset));
-    datasets.forEach((dataset) => this.addDataset(dataset));
+    datasets.forEach((dataset) => this.addDataset(dataset, false));
+    this._onDatasetUpdate();
     return this;
   }
 
--- a/src/plots/areaPlot.ts
+++ b/src/plots/areaPlot.ts
@@ -80,13 +80,13 @@ export module Plots {
       this._updateYScale();
     }
 
-    public addDataset(dataset: Dataset) {
+    public addDataset(dataset: Dataset, callUpdateHandler: boolean = true) {
       var lineDrawer = new Drawers.Line(dataset);
       if (this._isSetup) {
         lineDrawer.renderArea(this._renderArea.append("g"));
       }
       this._lineDrawers.set(dataset, lineDrawer);
-      super.addDataset(dataset);
+      super.addDataset(dataset, callUpdateHandler);
       return this;
     }
 
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -153,9 +153,9 @@ export module Plots {
       return this;
     }
 
-    public addDataset(dataset: Dataset) {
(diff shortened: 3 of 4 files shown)


src/plots/plot.ts
```diff
@@ -100,9 +100,11 @@ export class Plot extends Component {
    * Adds a Dataset to the Plot.
    *
    * @param {Dataset} dataset
+   * @param {boolean} callUpdateHandler Whether the typical on-dataset-update handler should be called.
+   *   If not provided, defaults to true.
    * @returns {Plot} The calling Plot.
    */
-  public addDataset(dataset: Dataset) {
+  public addDataset(dataset: Dataset, callUpdateHandler: boolean = true) {
```
jtlan: The explicit typing of  `boolean` is unnecessary here, since the type can be inferred from the fact that the value is set to `true`.

jtlan:
>I'm not sure if there are additional build steps I should perform in this PR

You'll want to run `grunt` to generate the `plottable.js` file, and upload that too -- this allows us to test the code against **`develop`** on jsfiddle.

jtlan:
I think it would make more sense to leave `addDatasets()` as it is, and have `datasets()` use specialized logic to avoid triggering extra updates, by only triggering a single update at the end.

While the approach submitted in this PR improves the performance of `addDataset()` itself, I think the semantics are not as good. In particular, the presence of a `boolean` parameter can be confusing, as it's hard to tell what the method call does without going back to the docs:
```Javascript
plot.addDataset(dataset1, false);
plot.addDataset(dataset2, false);
plot.addDataset(dataset3); // ???
```

jhorowitz-coursera:
@jtlan I agree that boolean parameters are confusing. However, I think duplicating `addDataset()`'s logic in `datasets()` would make for fragile code.

How about I factor most of `addDataset()` into `_addDatasetWithoutCallingUpdateHandler()` (awful name) so that `addDataset()` looks like:
```typescript
public addDataset(dataset: Dataset) {
  this._addDatasetWithoutCallingUpdateHandler(dataset);
  this._onDatasetUpdate();
  return this;
}
```
while `datasets()` can end with
```typescript
datasets.forEach((dataset) => this._addDatasetWithoutCallingUpdateHandler(dataset));
this._onDatasetUpdate();
```
?

(I would appreciate naming advice there. :smile:)

jtlan:
The suggested refactor makes sense to me. What about just calling it "`_addDataset()`"?

jtlan:
In addition, @jhorowitz-coursera, could you fill out this [Contributor License Agreement](https://github.com/palantir/plottable/blob/develop/Plottable_Corporate_Contributor_License_Agreement.pdf?raw=true)? (I assumed you would need the Corporate version of the form based on your username; otherwise, please fill out [this one](https://github.com/palantir/plottable/blob/develop/Plottable_Personal_Contributer_License_Agreement.pdf?raw=true) instead). Thanks!

christinaforney:
@jhorowitz-coursera have you had a chance to fill out the CLA? We'd love to be able to bing in your PR!

aicioara:
@jhorowitz-coursera ping :)

aicioara:
Dropping as inactive and unamended

PR #2431, opened 2015-07-02
Cache the plottable colors on the Color Scale

PeterFaiman:
Color._getPlottableColors() hits the DOM every time, which in turn
occurs every time a Color Scale is constructed. Adding a static cache
saves a few ms on every new Color().

### DIFF
--- a/src/scales/colorScale.ts
+++ b/src/scales/colorScale.ts
@@ -8,6 +8,8 @@ export module Scales {
     // The maximum number of colors we are getting from CSS stylesheets
     private static _MAXIMUM_COLORS_FROM_CSS = 256;
 
+    private static _plottableColorCache: string[] = null;
+
     private _d3Scale: d3.scale.Ordinal<string, string>;
 
     /**
@@ -24,7 +26,10 @@ export module Scales {
       switch (scaleType) {
         case null:
         case undefined:
-          scale = d3.scale.ordinal<string, string>().range(Color._getPlottableColors());
+          if (Color._plottableColorCache == null) {
+            Color._plottableColorCache = Color._getPlottableColors();
+          }
+          scale = d3.scale.ordinal<string, string>().range(Color._plottableColorCache);
           break;
         case "Category10":
         case "category10":
@@ -59,6 +64,10 @@ export module Scales {
     // Duplicated from OrdinalScale._getExtent - should be removed in #388
     protected _getExtent(): string[] {
       return Utils.Array.uniq(this._getAllIncludedValues());
+    }
+
+    public static invalidateColorCache(): void {
+      Color._plottableColorCache = null;
     }
 
     private static _getPlottableColors(): string[] {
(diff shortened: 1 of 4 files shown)


jtlan:
Can you add a test for the invalidation mechanism? Essentially, a test that checks that the colors *aren't* recomputed unless the cache is invalidated.

src/scales/colorScale.ts
```diff
@@ -8,6 +8,8 @@ export module Scales {
     // The maximum number of colors we are getting from CSS stylesheets
     private static _MAXIMUM_COLORS_FROM_CSS = 256;
 
+    private static _plottableColorCache: string[] = null;
```
aicioara: Specifically setting the variable to null is not needed as that is the default
`private static _plottableColorCache: string[];` is the convention.

src/scales/colorScale.ts
```diff
@@ -61,6 +66,10 @@ export module Scales {
       return Utils.Array.uniq(this._getAllIncludedValues());
     }
 
+    public static invalidateColorCache(): void {
```
aicioara: `:void` can be inferred, hence it can be removed

test/scales/scaleTests.ts
```diff
@@ -140,6 +142,23 @@ describe("Scales", () => {
         "Unloading the CSS should cause color scales fallback to default colors");
     });
 
+     it("caches CSS specified colors unless the cache is invalidated", () => {
+      var style = d3.select("body").append("style");
+      style.html(".plottable-colors-0 {background-color: #ff0000 !important; }");
+      Plottable.Scales.Color.invalidateColorCache();
```
aicioara: `Plottable` bit is optional. 

`Scales.Color.InvalidateColorCache()`

test/scales/scaleTests.ts
```diff
@@ -140,6 +142,23 @@ describe("Scales", () => {
...
+      var scale = new Plottable.Scales.Color();
+      style.remove();
+
+      assert.strictEqual(scale.range()[0], "#ff0000", "User has specified red color for first color scale color");
+
+      var scaleCached = new Plottable.Scales.Color();
+      assert.strictEqual(scaleCached.range()[0], "#ff0000", "The red color should still be cached");
+
+      Plottable.Scales.Color.invalidateColorCache();
```
aicioara: same

aicioara:
rest I believe is fine

test/scales/scaleTests.ts
PeterFaiman: The rest of the tests already written use the full name with Plottable, so I stuck with that convention.

aicioara:
Sorry about that, you're right

crmorford:
Talked to dev, treating this as a regression pass around Scales.Color.

aicioara:
Part of #2433

PR #2432, opened 2015-07-02
Some optimisations for the axes

aicioara:
Not really much faster unfortunately. It could stack up with some future improvements.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -4338,13 +4338,6 @@ var Plottable;
                 if (!this._isSetup) {
                     return;
                 }
-                if (!this._isHorizontal()) {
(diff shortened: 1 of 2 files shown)


jtlan:
Wrong behavior, since now `Axes.Numeric` will hide the first and last tick labels even if they would fit:
screen shot 2015-07-02 at 1 20 09 pm

Repro: http://jsfiddle.net/539prsr0/

aicioara:
hmm. Maybe this is a problem with the default and my understanding of how `showEndTickLabels` should work

aicioara:
Dropping this as well...

Issue #2433, opened 2015-07-02
Multiple plots on the same page take significantly longer than D3 version


aicioara:

aicioara:
Working on this branch `manyPlotsPerf`

aicioara:
50% faster already

aicioara:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering

PR #2435, opened 2015-07-02
Many plots perf improvement

aicioara:
I was hoping I can get more stuff in this side branch, but this is literally just a QE review of https://github.com/palantir/plottable/pull/2424 .
Merging it into develop now so @PeterFaiman can get these changes as well.

QE Notes: 
- **Regression pass**: nothing should be changed
- **What has been modified**: The numeric axis had logic for wrapping tick labels, but that never happened anyway because the tick labels which were too big were just hidden.
- **Performance improvement**: Not big, but for 60 plots on a page the load time decreases `1700ms` -> `1100ms`


aicioara:
Part of #2433

Issue #1030, opened 2014-09-12
Plottable.renderTo() should trigger initial flush


teamdandelion:
Right now if you call Plottable.renderTo and then try to do any computation that depends on the generated objects, it will fail because it has been async queued for render rather than rendered directly. We should have renderTo call flush at the end of its invocation to sidestep this issue.

PR #2436, opened 2015-07-02
[flushPerf] improvement

aicioara:
Perf improvement:
60 Plots of 10 points each
- Before: `1700ms`
- After: `700ms`

All unit tests passing. Overlaying tests seem okay. Am I missing something?

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -49,6 +49,9 @@ export class Component {
   private _onAnchorCallbacks = new Utils.CallbackSet<ComponentCallback>();
   private _onDetachCallbacks = new Utils.CallbackSet<ComponentCallback>();
 
+  private _onRenderCallbacks = new Utils.CallbackSet<Function>();
+
+
   public constructor() {
     this._cssClasses.add("component");
   }
@@ -283,6 +286,14 @@ export class Component {
     this.computeLayout();
     this.render();
     return this;
+  }
+
+  public onRender(callback: Function) {
+    this._onRenderCallbacks.add(callback);
+  }
+
+  public dispatchRender() {
+    this._onRenderCallbacks.callCallbacks();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/components/component.ts
```diff
@@ -282,8 +282,6 @@ export class Component {
     }
     this.computeLayout();
     this.render();
-    // flush so that consumers can immediately attach to stuff we create in the DOM
-    RenderController.flush();
```
jtlan: Without this, the following code has a chance of failing:

```Typescript
plot.renderTo(svg);
var circle = plot.foreground().append("circle"); // might fail if the render hasn't yet occurred
```

jtlan:
If we add a place to attach callbacks that get called after rendering, I think this change should be safe.

aicioara:
Will do.

aicioara:
Actually cannot reproduce the above

src/components/component.ts
jtlan: Apparently was added in PR #1082? I don't see a repro of the specific problem in the ticket (#1030) though.

jtlan: Found a repro:

```Typescript
barplot.renderTo(svg);
console.log(barplot.content().selectAll("rect").size()); // gets 0 instead of the expected value
```

aicioara:
Will this be an API breaking change?

jtlan:
No, since any code written previously will still compile, but the functionality may change slightly.

aicioara:
Actually this is not an improvement :( . I just deferred the flush for later, which is no way better than the previous flush split

aicioara:
My test was not recording the async flush.

aicioara:
Dropping this as well I guess

rcchen:
@aicioara can we close this PR? Github preserves the branch in case you want to open a PR again

aicioara:
Agree

PR #2438, opened 2015-07-06
Rename tickLength() to innerTickLength()

### LINKED ISSUE #2227: tickLength call only changes inner ticks
I'd think that calling tickLength(0) should get rid of all ticks, but it leaves the outer ticks, so you also have to call endTickLength(0);
this feels incorrect to me.

see: http://jsfiddle.net/v20vw15v/; ticks super bold for emphasis

estas:
Fixes #2227

### DIFF
--- a/src/axes/axis.ts
+++ b/src/axes/axis.ts
@@ -23,7 +23,7 @@ export class Axis<D> extends Component {
   protected _computedWidth: number;
   protected _computedHeight: number;
   private _endTickLength = 5;
-  private _tickLength = 5;
+  private _innerTickLength = 5;
   private _tickLabelPadding = 10;
   private _margin = 15;
   private _showEndTickLabels = false;
@@ -203,7 +203,7 @@ export class Axis<D> extends Component {
       tickMarkAttrHash["y2"] = scalingFunction;
     }
 
-    var tickLength = isEndTickMark ? this._endTickLength : this._tickLength;
+    var tickLength = isEndTickMark ? this._endTickLength : this._innerTickLength;
 
     switch (this._orientation) {
       case "bottom":
@@ -279,22 +279,22 @@ export class Axis<D> extends Component {
   /**
    * Gets the tick mark length in pixels.
    */
-  public tickLength(): number;
+  public innerTickLength(): number;
   /**
    * Sets the tick mark length in pixels.
    *
    * @param {number} length
    * @returns {Axis} The calling Axis.
    */
-  public tickLength(length: number): Axis<D>;
-  public tickLength(length?: number): any {
+  public innerTickLength(length: number): Axis<D>;
+  public innerTickLength(length?: number): any {
     if (length == null) {
-      return this._tickLength;
+      return this._innerTickLength;
     } else {
       if (length < 0) {
-        throw new Error("tick length must be positive");
+        throw new Error("inner tick length must be positive");
       }
-      this._tickLength = length;
+      this._innerTickLength = length;
       this.redraw();
       return this;
     }
@@ -326,9 +326,9 @@ export class Axis<D> extends Component {
 
   protected _maxLabelTickLength() {
     if (this.showEndTickLabels()) {
-      return Math.max(this.tickLength(), this.endTickLength());
+      return Math.max(this.innerTickLength(), this.endTickLength());
     } else {
-      return this.tickLength();
+      return this.innerTickLength();
     }
   }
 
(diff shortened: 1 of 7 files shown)


aicioara:
Thank you for the PR and interest in Plottable.

Unfortunately we cannot only remove a public method. That would be an API break (if people using the library were calling tickLength(), now that method disappeared and their code would break). 

Instead, the procedure is to leave that method in and deprecate it (`Utils.Window.deprecated()`). See [`Plots._visibleOnPlot()`](https://github.com/palantir/plottable/blob/develop/src/plots/plot.ts#L535) for an example. I suggest you just create the `innerTickLength()` along with `tickLength()`, (of course extract duplicated code in a private method)

test/axes/baseAxisTests.ts
```diff
@@ -29,11 +29,11 @@ describe("BaseAxis", () => {
     var verticalAxis = new Plottable.Axis(scale, "right");
     verticalAxis.renderTo(svg);
 
-    var expectedWidth = verticalAxis.tickLength() + verticalAxis.margin(); // tick length and margin by default
+    var expectedWidth = verticalAxis.innerTickLength() + verticalAxis.margin(); // innerTickLength and margin by default
```
aicioara: the comment was fine as it was. The reader of this comment doesn't particularly care if it's the inner tick length (actually the reader would wonder why is the inner tick more special)

test/axes/baseAxisTests.ts
```diff
@@ -47,11 +47,11 @@ describe("BaseAxis", () => {
     var horizontalAxis = new Plottable.Axis(scale, "bottom");
     horizontalAxis.renderTo(svg);
 
-    var expectedHeight = horizontalAxis.tickLength() + horizontalAxis.margin(); // tick length and margin by default
+    var expectedHeight = horizontalAxis.innerTickLength() + horizontalAxis.margin(); // innerTickLength and margin by default
```
aicioara: Comment was fine

test/axes/baseAxisTests.ts
rcchen: Let's actually go ahead and remove this comment. Comments should not be necessary if the code is well-written and clear, and in this case, the comment basically says the same thing as the code preceding it almost verbatim.

test/axes/baseAxisTests.ts
rcchen: Same here.

test/axes/baseAxisTests.ts
aicioara: The only extra information is that this is the default width, hence let's rename the variable to `defaultWidth` and remove the comment.

aicioara:
Also, for more information into how we deprecate stuff, this PR is perfect. Just check the `formatters.ts` file https://github.com/palantir/plottable/pull/2409/files#diff-7309a7fc81ae78def5ae0dfc185a5020R211

src/axes/axis.ts
```diff
@@ -288,13 +290,33 @@ export class Axis<D> extends Component {
    */
   public tickLength(length: number): Axis<D>;
   public tickLength(length?: number): any {
+    Utils.Window.deprecated("tickLength()", "v1.2.0", "Renamed to innerTickLength()");
+    return this._tickLength(length);
```
aicioara: Since this is the same as saying `return this.innerTickLength(length)`, how about we transform `_tickLength()` into `innerTickLength()`. 

In other words, have just `tickLength()` and `innerTickLength`. The former will call `deprecated` and then `innerTickLength`.

src/axes/axis.ts
```diff
@@ -277,6 +277,8 @@ export class Axis<D> extends Component {
   }
 
   /**
+   * This method is deprecated. Use innerTickLength() instead.
```
aicioara: We are still considering what is the best way to comment on a deprecated method, but I think `@deprecated  As of release 1.3, replaced by innerTickLength()` is the way to go now.

aicioara:
Everything else looks good.

aicioara:
Looks great! Thank you. Sending it to QE.

@estas in addition, could you fill out this [Contributor License Agreement](https://github.com/palantir/plottable/blob/develop/Plottable_Personal_Contributer_License_Agreement.pdf?raw=true) and send it to opensource@palantir.com?

estas:
You're welcome! Of cource, I'll send it.

crmorford:
@estas looks good! ready to merge as soon as we have the CLA :)

aicioara:
Merged. Also @estas this is our first ever accepted pull request coming from outside Palantir. Congratulations and thank you! :balloon:

estas:
I'm very happy! This is my first commit to open source! :octocat:

PR #2441, opened 2015-07-07
Fixed animate line data parsing issue in FF

### LINKED ISSUE #2440: animate_line quicktest broken on firefox
Can't find any reason in the console.
I've run it a few times so it doesn't seem like a quicktest flake.
The line just isn't showing up.
None of the other line quicktests seem broken.
screen shot 2015-07-06 at 3 21 13 pm

aicioara:
Closes #2440
Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

### DIFF
--- a/quicktests/overlaying/tests/animations/animate_line.js
+++ b/quicktests/overlaying/tests/animations/animate_line.js
@@ -1,8 +1,8 @@
 
 function makeData() {
   "use strict";
-  var data = [{x: new Date("2014-04-01 00:24"), y: 4},
-              {x: new Date("2014-08-29 00:24"), y: 6}];
+  var data = [{x: new Date("4/1/2014 00:24"), y: 4},
+              {x: new Date("8/29/2014 00:24"), y: 6}];
   return data;
 }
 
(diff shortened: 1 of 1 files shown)


aicioara:
Similar solution for https://github.com/palantir/plottable-website/issues/105

quicktests/overlaying/tests/animations/animate_line.js
```diff
@@ -1,8 +1,8 @@
 
 function makeData() {
   "use strict";
-  var data = [{x: new Date("2014-04-01 00:24"), y: 4},
-              {x: new Date("2014-08-29 00:24"), y: 6}];
+  var data = [{x: new Date("4/1/2014 00:24"), y: 4},
+              {x: new Date("8/29/2014 00:24"), y: 6}];
```
rcchen: Could we use the [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) standard instead? Just for the sake of eliminating confusion.

rcchen:
Let's use the standard instead. I tested it in Firefox and new Date() will recognize it properly.

aicioara:
This will break the test. We need a localized version of the date. 
```
> new Date("2014-04-01T00:24Z")
< Mon Mar 31 2014 17:24:00 GMT-0700 (PDT)
```

crmorford:
wait, was the fix just in the test, not in Plottable? That's what it looks like in the diff.

aicioara:
Yes, it's the Time fun. Firefox does not know how to parse dates like 
```typescript 
new Date("2014-04-01 00:24")
```
(if you open console and write that, you will get `Invalid Date`, while in Chrome it still works)

crmorford:
ah okay, that's a relief :)

PR #2444, opened 2015-07-07
Hide labels that are not rendered within the renderArea for Bar/PiePlots

### LINKED ISSUE #1728: Bar Chart Labels are only partially rendered if cut off by min/max

rcchen:
Closes #1728

Bar labels that are not within the render area will no longer be visible.

Repros: 

Bar Plot
Before: http://jsfiddle.net/5mh3rdko/
After: http://jsfiddle.net/5mh3rdko/1/

Pie Plot:
Before: http://jsfiddle.net/th7qmqoa/5/
After: http://jsfiddle.net/th7qmqoa/6/


src/plots/barPlot.ts
```diff
@@ -441,9 +441,20 @@ export module Plots {
             x += offset;
           }
 
+          var showLabel = true;
+          var labelPosition = {
+            x: x + w / 2 - measurement.width / 2,
+            y: positive ? y : y + h
+          };
+          if (labelPosition.x < 0 || labelPosition.x + measurement.width > +this._boundingBox.attr("width") ||
+              labelPosition.y < 0 || labelPosition.y + measurement.height > +this._boundingBox.attr("height")) {
```
aicioara: Are you sure this is correct in the case we have multiple datasets / stacked bars (not sure if those accept labels)

src/plots/piePlot.ts
```diff
@@ -332,8 +332,18 @@ export module Plots {
           { x: x + measurement.width, y: y + measurement.height }
         ];
 
-        var sliceIndices = corners.map((corner) => this._sliceIndexForPoint(corner));
-        var showLabel = sliceIndices.every((index) => index === datumIndex);
+        var absoluteCenter = { x: this.width() / 2, y: this.height() / 2 };
+        var showLabel = true;
+        for (var i = 0; i < corners.length; i++) {
```
aicioara: I prefer `corners.forEach()`

src/plots/barPlot.ts
rcchen: This check is strictly for component render area boundaries via boundingBox so it shouldn't matter what sort of plot it is.

test/plots/barPlotTests.ts
```diff
@@ -694,6 +694,41 @@ describe("Plots", () => {
...
+        xScale.domain([1, 3]);
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.labelsEnabled(true);
+        plot.renderTo(svg);
+        var texts = svg.selectAll("text")[0];
+        texts.forEach((text, i) => {
+          var selection = d3.select(text);
+          var visibility = selection.style("visibility");
+          if (i === 1) {
+            assert.isTrue(visibility === "visible" || visibility === "inherit", "bar label is visible");
```
aicioara: Comment should be `"Second bar label is visible"`

test/plots/barPlotTests.ts
```diff
@@ -694,6 +694,41 @@ describe("Plots", () => {
...
+          } else {
+            assert.isTrue(visibility === "hidden", "bar label is hidden");
```
aicioara: Comment should be `"bar label " + (i + 1) + " should be hidden"`

test/plots/barPlotTests.ts
```diff
@@ -694,6 +694,41 @@ describe("Plots", () => {
...
+        svg = TestMethods.generateSVG();
+        xScale = new Plottable.Scales.Linear();
+        yScale = new Plottable.Scales.Linear();
+        plot = new Plottable.Plots.Bar<number, number>();
+        plot.x((d) => d.x, xScale);
+        plot.y((d) => d.y, yScale);
+      });
+
+      it("hides labels outside of the visible render area", () => {
+        var data = [{ x: 1, y: -10.1 }, { x: 2, y: -5.3 }, { x: 3, y: -2.8 }];
```
aicioara: From this data, why is the second label visible only? It seems to me like the first one should be visible as well.

test/plots/barPlotTests.ts
```diff
@@ -694,6 +694,41 @@ describe("Plots", () => {
...
+          if (i === 1) {
+            assert.isTrue(visibility === "visible" || visibility === "inherit", "bar label is visible");
+          } else {
+            assert.isTrue(visibility === "hidden", "bar label is hidden");
+          }
+        });
+        svg.remove();
+      });
+    });
+
```
aicioara: This tests only the height space. How about we get some test where we also check for the width? (render same plot on a large SVG and on a very small one)

test/plots/piePlotTests.ts
```diff
@@ -264,6 +264,25 @@ describe("Plots", () => {
...
+        var dataset = new Plottable.Dataset(data);
+        piePlot.sectorValue(function(d) { return d; });
+        piePlot.addDataset(dataset).labelsEnabled(true).outerRadius(500);
+        var texts = svg.selectAll("text")[0];
+        texts.forEach((text, index) => {
+          var visibility = d3.select(text).style("visibility");
+          if (index === 1) {
+            assert.strictEqual(visibility, "hidden", "label hidden when cut off");
+          } else {
+            assert.include(["visible", "inherit"], visibility, "label shown when in the renderArea");
```
aicioara: I like this assert, can we make it the same in the test above?

test/plots/piePlotTests.ts
```diff
@@ -264,6 +264,25 @@ describe("Plots", () => {
         });
         svg.remove();
       });
+
+      it("labels outside of the render area are hidden", () => {
+        piePlot.detach();
+        piePlot.removeDataset(simpleDataset);
+        var data = [5000, 5000, 5000];
```
aicioara: Can you explain this test? For the given data it does not make any sense that only one label is hidden.

test/plots/barPlotTests.ts
rcchen: xScale.domain([1, 3]);

Labels are placed in the middle of the bar, forcing the domain slices the labels in the middle.

test/plots/piePlotTests.ts
rcchen: The label for the bottom slice is the first one to get forced out of the SVG (it's directly on an axis, as opposed to at a diagonal).

test/plots/piePlotTests.ts
rcchen: For line 279? The include assert is only for set inclusion.

test/plots/barPlotTests.ts
rcchen: On the contrary, I think this only tests in the width space. I should write another test for the height space (force domain on the yScale)

src/plots/piePlot.ts
```diff
@@ -332,8 +332,15 @@ export module Plots {
           { x: x + measurement.width, y: y + measurement.height }
         ];
 
-        var sliceIndices = corners.map((corner) => this._sliceIndexForPoint(corner));
-        var showLabel = sliceIndices.every((index) => index === datumIndex);
+        var absoluteCenter = { x: this.width() / 2, y: this.height() / 2 };
+        var showLabel = corners.every((corner) => {
+          return Math.abs(corner.x) > absoluteCenter.x || Math.abs(corner.y) > absoluteCenter.y ? false : true;
```
aicioara: return Math.abs(corner.x) <= absoluteCenter.x && Math.abs(corner.y) <= absoluteCenter.y

test/plots/barPlotTests.ts
```diff
@@ -506,6 +506,38 @@ describe("Plots", () => {
         svg.remove();
       });
 
+      it("hides labels properly on the right", () => {
```
aicioara: Let's maybe extract these in their own describe. It is a bit hard to follow the test.

test/plots/barPlotTests.ts
```diff
@@ -694,6 +726,59 @@ describe("Plots", () => {
...
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.renderTo(svg);
+        var texts = svg.selectAll("text")[0];
+        texts.forEach((text, i) => {
+          var selection = d3.select(text);
+          var visibility = selection.style("visibility");
+          if (i === 1) {
+            assert.isTrue(visibility === "visible" || visibility === "inherit", "bar label at index 1 is visible");
+          } else {
+            assert.isTrue(visibility === "hidden", "bar label " + i + " is hidden");
```
aicioara: Again, why are the edge labels invisible?

test/plots/piePlotTests.ts
aicioara: Oh, I see, though it is not so straightforward.

test/plots/piePlotTests.ts
aicioara: I mean the other files
```typescrpit
assert.isTrue(visibility === "visible" || visibility === "inherit", "bar label " + i + " is visible");
```

aicioara:
Moving to owning this

src/plots/piePlot.ts
```diff
@@ -355,8 +355,15 @@ export module Plots {
           { x: x + measurement.width, y: y + measurement.height }
         ];
 
-        var sliceIndices = corners.map((corner) => this._sliceIndexForPoint(corner));
-        var showLabel = sliceIndices.every((index) => index === datumIndex);
+        var absoluteCenter = { x: this.width() / 2, y: this.height() / 2 };
```
jtlan: This is more like "half width" and "half height" right? I think the code would be easier to read that way:
```Typescript
var showLabel = corners.every((corner) => Math.abs(corner.x) <= this.width() / 2 && Math.abs(corner.y) <= this.height() /2 );
```

src/plots/barPlot.ts
```diff
@@ -441,9 +440,31 @@ export module Plots {
...
+          } else {
+            if (!positive) {
+              labelPosition.x = x + w - measurement.width;
+            } else {
+              labelPosition.x = x;
+            }
+          }
+
+          if (labelPosition.x < 0 || labelPosition.x + measurement.width > +this._boundingBox.attr("width") ||
+              labelPosition.y < 0 || labelPosition.y + measurement.height > +this._boundingBox.attr("height")) {
```
jtlan: Instead of `this._boundingBox.attr("width")` and `this._boundingBox.attr("height")` you should be able to use `this.width()` and `this.height()`.

src/plots/piePlot.ts
bluong: Remember to put the github comment below the relevant code.  Putting it above will hide the relevant reviewed block.

aicioara:
Repros: 

Bar Plot
Before: http://jsfiddle.net/5mh3rdko/
After: http://jsfiddle.net/5mh3rdko/1/

Pie Plot:
TODO

PR #2459, opened 2015-07-10
[Plots.Line/Area] Enable interpolation

### LINKED ISSUE #757: Add ability to set interpolation mode on Line and Area Plots
E.g. `d3.svg.area().interpolate("cardinal")`

Perhaps have `Plottable.Plot.Area()` take an options object?

Same for line as well.

rcchen:
image

Adds the following API point to `Plots.Line` and `Plots.Area`:

`.interpolate()`/`.interpolate(interpolate)` - gets or sets the current interpolation mode to the specified string or function. Accepted named interpolation modes are listed [here](https://github.com/mbostock/d3/wiki/SVG-Shapes#line_interpolate). If interpolate is a function, it will be invoked to convert an array of points to an SVG data path string.

Fiddle http://jsfiddle.net/xz9Ljrx6/2/

Fixes #757 

linear - piecewise linear segments, as in a polyline.
linear-closed - close the linear segments to form a polygon.
step - alternate between horizontal and vertical segments, as in a step function.
step-before - alternate between vertical and horizontal segments, as in a step function.
step-after - alternate between horizontal and vertical segments, as in a step function.
basis - a B-spline, with control point duplication on the ends.
basis-open - an open B-spline; may not intersect the start or end.
basis-closed - a closed B-spline, as in a loop.
bundle - equivalent to basis, except the tension parameter is used to straighten the spline.
cardinal - a Cardinal spline, with control point duplication on the ends.
cardinal-open - an open Cardinal spline; may not intersect the start or end, but will intersect other control points.
cardinal-closed - a closed Cardinal spline, as in a loop.
monotone - cubic interpolation that preserves monotonicity in y.

New API points
```typescript
/**
 * Gets the interpolation function associated with the plot.
 *
 * @return {string | (points: Array<[number, number]>) => string)}
 */
interpolator(): string | ((points: Array<[number, number]>) => string);
/**
 * Sets the interpolation function associated with the plot.
 *
 * @param {string | points: Array<[number, number]>) => string} interpolator Interpolation function
 * @return Plots.Line
 */
interpolator(interpolator: string | ((points: Array<[number, number]>) => string)): Plots.Line<X>;
interpolator(interpolator: "linear"): Line<X>;
interpolator(interpolator: "linear-closed"): Line<X>;
interpolator(interpolator: "step"): Line<X>;
interpolator(interpolator: "step-before"): Line<X>;
interpolator(interpolator: "step-after"): Line<X>;
interpolator(interpolator: "basis"): Line<X>;
interpolator(interpolator: "basis-open"): Line<X>;
interpolator(interpolator: "basis-closed"): Line<X>;
interpolator(interpolator: "bundle"): Line<X>;
interpolator(interpolator: "cardinal"): Line<X>;
interpolator(interpolator: "cardinal-open"): Line<X>;
```


src/plots/linePlot.ts
```diff
@@ -21,6 +22,27 @@ export module Plots {
       this.attr("stroke-width", "2px");
     }
 
+    /**
+     * Gets the interpolation function associated with the plot.
+     * 
+     * @return string
+     */
+    public interpolate(): string;
```
aicioara:
```typescript
 string | ((points: number[][]) => string))
```

src/plots/linePlot.ts
```diff
@@ -21,6 +22,27 @@ export module Plots {
       this.attr("stroke-width", "2px");
     }
 
+    /**
+     * Gets the interpolation function associated with the plot.
+     * 
+     * @return string
```
aicioara: amend

src/plots/linePlot.ts
```diff
@@ -106,6 +128,7 @@ export module Plots {
         return d3.svg.line()
                      .x((innerDatum, innerIndex) => xProjector(innerDatum, innerIndex, dataset))
                      .y((innerDatum, innerIndex) => yProjector(innerDatum, innerIndex, dataset))
+                     .interpolate(this.interpolate())
```
aicioara: Just double checking, the linear interpolation is our current non-settable default? (prior to this patch)

src/plots/linePlot.ts
```diff
@@ -3,6 +3,7 @@
 module Plottable {
 export module Plots {
   export class Line<X> extends XYPlot<X, number> {
+    private _interpolationFunction: string | ((points: number[][]) => string) = "linear";
```
aicioara: Do you think it makes sense to type it like this instead
https://github.com/palantir/plottable/blob/develop/typings/d3/d3.d.ts#L2176

It's not really consistent with the rest of our codebase

aicioara:
Can you unit test and / or overlay test this feature and provide a fiddle to this issue description for future reference.

aicioara:
Moving to owning this

aicioara:
Blocking on https://github.com/borisyankov/DefinitelyTyped/pull/5095 for now to avoid hackhacks. If PR is not merged by Thursday, hackhack gets in.

src/plots/areaPlot.ts
```diff
@@ -191,6 +191,8 @@ export module Plots {
                                   .x((innerDatum, innerIndex) => xProjector(innerDatum, innerIndex, dataset))
                                   .y1((innerDatum, innerIndex) => yProjector(innerDatum, innerIndex, dataset))
                                   .y0((innerDatum, innerIndex) => y0Projector(innerDatum, innerIndex, dataset))
+                                  // HACKHACK https://github.com/borisyankov/DefinitelyTyped/pull/5095
```
aicioara: Ignore the hackhack for now. Holding on the merge until PR gets in.

bluong: We have the reference url so I'm now a lot less worried.

That said, I highly prefer to link to issues since pull requests can easily be upped/dropped on the fly more so than issues.  Issues also usually point at the underlying issue more reliably than pull requests.

Also if you could provide a quick tldr description of the hackhack here that'd be great.  At first glance, the only thing I see is an any-cast that is happening because of this hackhack and now I have to click to figure out why.

test/plots/linePlotTests.ts
```diff
@@ -33,6 +33,54 @@ describe("Plots", () => {
...
+        linePlot.renderTo(svg);
+
+        var svgPath: string;
+        svgPath = linePlot.content().select("path").attr("d");
+        assert.lengthOf(svgPath.match(/L/g), data.length - 1, "one line for each pair of consecutive points");
+        assert.isNull(svgPath.match(/V/g), "no vertical lines");
+        assert.isNull(svgPath.match(/H/g), "no horizontal lines");
+
+        linePlot.interpolate("step");
+        linePlot.render();
```
aicioara: Should I make this as part of the `.interpolate()` call?

jtlan: Probably, yeah.

aicioara:
Updated the `d3.d.ts` definition file. Someone seems to have forgotten to run `tsd reinstall -so` because the command also updated the chai-assert. However, Checked it out, as it was outside the scope of this commit. Nothing seems to be broken

src/plots/linePlot.ts
```diff
@@ -21,6 +22,28 @@ export module Plots {
...
+     * @return string | (points: Array<[number, number]>) => string)
+     */
+    public interpolate(): string | ((points: Array<[number, number]>) => string);
+    /**
+     * Sets the interpolation function associated with the plot.
+     *
+     * @param interpolate Interpolation function
+     * @return Plots.Line
+     */
+    public interpolate(interpolate: string | ((points: Array<[number, number]>) => string)): Plots.Line<X>;
```
jtlan: Since "interpolate" is a verb, in a vacuum I would prefer to name this method "`interpolator()`". However, d3 uses "`interpolate`". I think it's okay to depart from D3 here.

Also, is it worth waiting for the bug mentioned [here](https://github.com/borisyankov/DefinitelyTyped/pull/5095#issuecomment-125683305) to get fixed in D3, so we don't have to return a union type?

aicioara:
To preserve, the comment

>Also, is it worth waiting for the bug mentioned [here](https://github.com/borisyankov/DefinitelyTyped/pull/5095#issuecomment-125683305) to get fixed in D3, so we don't have to return a union type?

I am unsure which bug are you talking about. If it is the TS one, I doubt it will be solved soon (or ever). It seems to be a complex compiler issue. The discussion on the mentioned thread has more to do with further fixes which I did not have time to hunt down and implement.

src/plots/linePlot.ts
aicioara: Will do

aicioara:
Why `interpolator` and not `interpolation`?

jtlan:
>I am unsure which bug are you talking about.

I meant the bug in D3 itself, where it doesn't upcast constants.

src/plots/linePlot.ts
```diff
@@ -21,6 +22,28 @@ export module Plots {
...
+    /**
+     * Gets the interpolation function associated with the plot.
+     *
+     * @return string | (points: Array<[number, number]>) => string)
+     */
+    public interpolator(): string | ((points: Array<[number, number]>) => string);
+    /**
+     * Sets the interpolation function associated with the plot.
+     *
+     * @param interpolator Interpolation function
```
jtlan: Wrong param type here (missing `string`), and the TSDoc is incorrect. It should be something like this:
```
@param {string | points: Array<[number, number]>) => string} interpolator
```
Is there a way to list the valid `string` inputs?

plottable.d.ts
```diff
@@ -3056,6 +3056,32 @@ declare module Plottable {
              * @constructor
              */
             constructor();
+            /**
+             * Gets the interpolation function associated with the plot.
+             *
+             * @return {string | (points: Array<[number, number]>) => string)}
+             */
+            interpolator(): string | ((points: Array<[number, number]>) => string);
```
bluong: I remember @jtlan having issues regarding us returning a union type.

plottable.d.ts
```diff
@@ -3056,6 +3056,32 @@ declare module Plottable {
...
+            interpolator(interpolator: "cardinal-open"): Line<X>;
+            interpolator(interpolator: "cardinal-closed"): Line<X>;
+            interpolator(interpolator: "monotone"): Line<X>;
+            /**
+             * Sets the interpolation function associated with the plot.
+             *
+             * @param {string | points: Array<[number, number]>) => string} interpolator Interpolation function
+             * @return Plots.Line
+             */
+            interpolator(interpolator: string | ((points: Array<[number, number]>) => string)): Plots.Line<X>;
```
bluong: This setter seems oddly placed.  We theoretically want a setter for each string as well as a setter for the function, but the union-typed setter shouldn't appear in the .d.ts file I think.

src/plots/linePlot.ts
```diff
@@ -21,6 +22,41 @@ export module Plots {
...
+    public interpolator(interpolator: "cardinal-open"): Line<X>;
+    public interpolator(interpolator: "cardinal-closed"): Line<X>;
+    public interpolator(interpolator: "monotone"): Line<X>;
+    /**
+     * Sets the interpolation function associated with the plot.
+     *
+     * @param {string | points: Array<[number, number]>) => string} interpolator Interpolation function
+     * @return Plots.Line
+     */
+    public interpolator(interpolator: string | ((points: Array<[number, number]>) => string)): Plots.Line<X>;
```
bluong: Can you remove the `string` portion here?

test/plots/linePlotTests.ts
```diff
@@ -31,6 +31,53 @@ describe("Plots", () => {
...
+        linePlot.x((d) => d.x, xScale);
+        linePlot.y((d) => d.y, yScale);
+
+        linePlot.renderTo(svg);
+
+        var svgPath: string;
+        svgPath = linePlot.content().select("path").attr("d");
+        assert.lengthOf(svgPath.match(/L/g), data.length - 1, "one line for each pair of consecutive points");
+        assert.isNull(svgPath.match(/V/g), "no vertical lines");
+        assert.isNull(svgPath.match(/H/g), "no horizontal lines");
```
bluong: Your choice if you want to put this as a comment, but I mainly want to confirm.  Is L/V/H the svg path language for making arbitrary/vertical/horizontal lines?

aicioara: It is, yes

plottable.d.ts
aicioara: There are instances in our codebase where we return unions (say `Plots.Bar.baselineValue()`)

plottable.d.ts
aicioara: This will not work because of this.  https://github.com/Microsoft/TypeScript/issues/1805

Also, current implementation mimics d3.d.ts type signature as can be seen here https://github.com/palantir/plottable/blob/757fix/typings/d3/d3.d.ts#L2223

More no this in a conversation here 
https://github.com/borisyankov/DefinitelyTyped/pull/5095

src/plots/linePlot.ts
aicioara: I believe this is related to the above?

plottable.d.ts
bluong: I would ask him directly.  I personally do not see the problem yet of returning a union type.

@jtlan comment here?

plottable.d.ts
bluong: If we remove the `string` type, though, since we have the overloaded implementation signature at the bottom, shouldnt that allow us to create the appropriate setters without needing to worry about union types?

aicioara: No, because of https://github.com/Microsoft/TypeScript/issues/1805

bluong:
Just need to ask @jtlan about his concern with the union typings.

jtlan:
I think we can leave the union type be here, since it reflects the underlying D3 behavior. The other alternative is to block on a PR to D3 fixing this, but I'm not sure how long that would take.

aicioara:
Discussed with @jtlan in person. Summary is that we cannot wait for all the PRs to D3 and DefinitelyTyped, because that would take at least 6 months. By then we should have Plottable 2.0 and address any issue if there is. 

Apart from that, @jtlan @bluong anything else stopping this from merging / going to QE?

plottable.d.ts
jtlan: Unless a fix can make it into D3 really fast, we're forced to use a union type here.

plottable.d.ts
```diff
@@ -3056,6 +3056,32 @@ declare module Plottable {
              * @constructor
              */
             constructor();
+            /**
+             * Gets the interpolation function associated with the plot.
+             *
+             * @return {string | (points: Array<[number, number]>) => string)}
+             */
+            interpolator(): string | ((points: Array<[number, number]>) => string);
+            interpolator(interpolator: "linear"): Line<X>;
```
jtlan: These signatures should be grouped with the other setter signature, under the TSDoc block for the setter, right?

aicioara: I am unsure how our tools work. I believe we have to copy + paste the code. How do I test this?

jtlan: I just wanted to make sure that the doc and signatures occur in this order:
```
Getter doc
Getter signature
Setter doc
Setter signatures (no particular order)
```

crmorford:
We should maybe include a perf warning? some of these seem to slow things down (anecdotally, cardinal seems to)

aicioara:
I think the slowdown is expected, is it huge?

crmorford:
it was noticeable, even though i wasn't looking for it. maybe i can look into it more objectively.

aicioara:
I updated the fiddle http://jsfiddle.net/xz9Ljrx6/2/

Sorry about that. I should stop using hashes

Issue #746, opened 2014-07-29
ClassScale should exist


teamdandelion:
Think like a color scale, but instead of attaching a color, it attaches css classes.
This way you can use CSS to determine the colors in a way that syncs with legends and with plots.

aicioara:
Related to https://github.com/palantir/plottable/pull/2431 . Maybe we can stop reading CSS colors off the DOM.

PR #2461, opened 2015-07-14
API endpoint to specify opacity of legend

### LINKED ISSUE #2434: Need an API point to specify legend symbol opacity
It'd be nice to have first-class support from Legend (or perhaps a color scale?) for specifying symbol opacity.

yqlu:
Fixes #2434 -- motivation is that opacity of many Plottable components can be specified directly via .attr(), but this is not true for legend.

User can now create Scale.Opacity which can be passed into a legend. If unspecified, opacity is by default 1. User can set whether to apply the opacity scale to just the symbols or the legend text as well.

Demo: http://plnkr.co/edit/5lQh9oanj9bw1gRjPlyp?p=preview

This doesn't cover InterpolatedColorLegend, but we can do something similar easily.

### DIFF
--- a/src/scales/opacityScale.ts
+++ b/src/scales/opacityScale.ts
@@ -27,11 +27,11 @@ export module Scales {
       return this._d3Scale(value);
     }
 
-    public extentOfValues(values: number[]): number[] {
+    public extentOfValues(values: string[]): string[] {
       return Utils.Array.uniq(this._getAllIncludedValues());
     }
 
-    protected _getExtent(): number[] {
+    protected _getExtent(): string[] {
       return Utils.Array.uniq(this._getAllIncludedValues());
     }
 
@@ -39,7 +39,7 @@ export module Scales {
       return this._d3Scale.domain();
     }
 
-    protected _setBackingScaleDomain(values: number[]) {
+    protected _setBackingScaleDomain(values: string[]) {
       this._d3Scale.domain(values);
     }
 
@@ -47,7 +47,7 @@ export module Scales {
       return this._d3Scale.range();
     }
 
-    protected _setRange(values: R[]) {
+    protected _setRange(values: number[]) {
       return this._d3Scale.range(values);
     }
 
--- a/src/scales/opacityScale.ts
+++ b/src/scales/opacityScale.ts
@@ -1,0 +1,56 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Scales {
+  export class Opacity extends Scale<string, number> {
+
+    private _d3Scale: d3.scale.Ordinal<string, number>;
+
+    /**
+     * An Opacity Scale maps string values to opacity values between 0 and 1.
+     *
+     * @constructor
+     */
+    constructor() {
+      super();
+      this._d3Scale = d3.scale.ordinal<string, number>().range([1]);
+    }
+
+    /**
+     * Returns the opacity corresponding to a given string.
+     * If there are not enough opacities in the range(), the scale will recycle values from the start of the range.
+     *
+     * @param {string} value
+     * @returns {number}
+     */
+    public scale(value: string): number {
+      return this._d3Scale(value);
+    }
+
+    public extentOfValues(values: number[]): number[] {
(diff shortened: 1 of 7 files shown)


src/scales/opacityScale.ts
```diff
@@ -0,0 +1,56 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Scales {
+  export class Opacity extends Scale<string, number> {
```
jtlan: I think there's probably a more generalized way to build this. Essentially, this `Scale` maps one-to-one from `String`s to `number`s, but that functionality has applications beyond setting an opacity -- the `number` could have other meanings.

In a more general sense, it may be helpful to have the `Scale` map one-to-one with other types of values, such as `String`s (which would take care of #746), so it might be helpful to leave the *range* generic.

jtlan:
What's the advantage of this implementation over something that's more symmetric with the `symbol()` method:

```Typescript
public opacity(opacity: (datum: any, index: number) => number): Legend;
```

yqlu:
I was thinking of opacity as something analogous to color, but I guess the difference is that colorScales have many uses while this opacityScale is single-purpose.

Something symmetric with `symbol()` would indeed be nicer. Another, even more generic option would be to have `.attr("opacity", function() { ... })` in parity with the rest of the library? 

The drawback to that I can think of is that it might feel arbitrary / inconsistent that `.attr()` in this case will project only onto the symbols but not the text in the legend.

jtlan:
`Legend` is in a weird spot, since it's the only non-`Plot` `Component` that could be considered "data-backed". I'm not sure it makes sense to have the `Legend` labels be part of the controllable attributes though.

jtlan:
@yqlu , can you explain what the use case for this is?

yqlu:
The original issue #2434 is from @istarnx - I understand the motivation is to allow users to specify opacity (and on a broader scale, various styling attributes) for `Legend`s in a first class way like other components. So yeah you're right that #2434 is like a subcase of #746.

It looks like there are several ways to expand the API:

- `.opacity(opacityScale)`
- `.opacity(function(d) { ... })`
- `.attr('opacity', function(d) { ... })`
- `classScale`s
- `.addClass()` or similar, and then use CSS styling

Looking back, the first option (the approach this PR takes) is a bit too special-purpose. @jtlan, what do you think is the most consistent?

jtlan:
> Looking back, the first option (the approach this PR takes) is a bit too special-purpose. @jtlan, what do you think is the most consistent?

I'm not sure it makes sense to be able to specify variable opacity on a `Legend`; the only case I can think of is if there are multiple `Plot` types whose `Dataset`s each appear on the `Legend`, each with a different opacity.

bluong:
Hello @yqlu ,
is it possible for you to word out / visually mock up an example for what you are trying to achieve here?

I'm just having trouble visualizing what you are trying to achieve here

yqlu:
Pinging @istarnx who should have a clearer idea of what a use-case is and can mock one up.

My original thought was if there was a scatter plot of various datasets with varying colors and opacities, you would want the opacity in the legend to match. But I'll defer to you on what the best way to do this would be, whether through a .opacity() or .attr() or classScale, etc?

istarnx:
@yqlu has already nailed the use case--it's precisely what he said. I'd like a API-endorsed way to specify the apparent color, i.e., color + opacity, of a legend entry symbol.

CSS-based solutions aren't enough, and neither is a single opacity setting for all legend entries.

bluong:
@istarnx Just the symbol or the legend entry as a whole?

istarnx:
The symbol. The thing that the color scale--sans opacity--is currently driving.

bluong:
Based on what @istarnx is saying, would a `symbolOpacity` API make sense?

jtlan:
Something like this?
```Typescript
public symbolOpacity(constantOpacity: number);
public symbolOpacity(opacityFunction: (d: any, i: index) => number);
```

yqlu:
That sounds easy to do and relatively lightweight! If you like, I'll close this PR and submit a new one soon based on this.

jtlan:
@yqlu That sounds good. Thanks!

PR #2464, opened 2015-07-15
Fix autoranging on Plots.Segment and Plots.Rectangle

### LINKED ISSUE #2463: Autoranging does not work correctly for Plots.Rectangle and Plots.Segment

rcchen:
Open questions:

* There's currently code duplication between `Plots.Segment` and `Plots.Rectangle` in the way extents are calculated to encompass both `x` and `x2` (or `y` and `y2`). Should this go in a Util function, and what should it be called?

http://jsfiddle.net/hy3vak9k/

Fixes #2463


rcchen:
Closing in favor of #2472

PR #2465, opened 2015-07-15
Travis-CI improvements

rcchen:
Adding `sudo: false` switches us over to Travis' container-based infrastructure, which should give us modest performance boosts in testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
 - '0.12'
-before_install:
-- phantomjs -v
 notifications:
   email: false
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,14 @@
 language: node_js
 node_js:
-- '0.10'
+- '0.12'
+before_install:
+- phantomjs -v
 notifications:
   email: false
+sudo: false
+cache:
+  directories:
+  - node_modules
 env:
   global:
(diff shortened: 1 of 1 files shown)


aicioara:
Can you provide a URL with the documentation? It's hard to review a config file otherwise.

aicioara:
there is still an `npm test` which takes about 2 mins out of our 3 min build time. Can we maybe get rid of that as well? It would be so cool

rcchen:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

rcchen:
If you run `npm test` on your local machine, you'll notice that it encompasses our test suite. I don't think it's possible to throw that out, unless there's a redundancy somewhere that I'm not noticing.

.travis.yml
```diff
@@ -1,8 +1,14 @@
 language: node_js
 node_js:
-- '0.10'
+- '0.12'
+before_install:
+- phantomjs -v
 notifications:
   email: false
+sudo: false
```
aicioara: This is awesome news. It says that we are guaranteed <1 min queue time? Is that correct? If it is, that's amazing.

rcchen:
The benefit that we're mostly going to be realizing from this is that boot times for machines is reduced (time to start of test)

.travis.yml
rcchen: We are guaranteed < 10 sec build starts, and more consistent builds throughout the day :smiley:

aicioara:
Does boot time mean queue + boot time?

aicioara:
Let's merge this and watch it more in the wild. The idea sounds awesome to me.

rcchen:
Queue time is a function of the number of users at any given time, and how fast their builds are going/how fast things are set up/torn down, so queue is going to be variable but overall this should help out.

PR #2466, opened 2015-07-16
Upgrade Typescript to 1.5.0

rcchen:
The main difference (for us) between TS 1.4.1 and TS 1.5 is that touch-config typings are no longer needed, as they are shipped in lib.d.ts.


jtlan:
I don't think we should do this until the 1.5 stable version is out.

andreweduffy:
stable is out! :) http://blogs.msdn.com/b/typescript/archive/2015/07/20/announcing-typescript-1-5.aspx

bluong:
Taking over this PR

test/dispatchers/dispatcherTests.ts
```diff
@@ -32,7 +32,7 @@ describe("Dispatchers", () => {
       var callbackWasCalled = false;
       (<any> dispatcher)._eventToCallback["click"] = () => callbackWasCalled = true;
 
-      var callback = () => null;
+      var callback = () => { return; };
```
jtlan: Returning `null` isn't allowed any more?

bluong: I'm running into an issue with the compiler where leaving it as is has the compiler complaining that there is an implicit `any`.

I could explicitly say that this function returns an `any`, but after thinking about it, since the point of this callback is to do nothing, I decided to change it to just do nothing.

jtlan: Makes sense.

jtlan:
@crmorford Probably a really basic regression check? I don't think anything really changed other than formatting and a few safety checks got safer.

bluong:
@jtlan @crmorford Regression check makes sense.  We are essentially updating the main tech that builds our main library file, but nothing else

crmorford:
regression check went well!

PR #2467, opened 2015-07-16
Upgrade grunt-mocha-phantomjs to 0.6.2

rcchen:
We need this to get our Travis working again because of https://github.com/jdcataldo/grunt-mocha-phantomjs/issues/36


rcchen:
Reverting this one because mocha-phantomjs published v3.5.6 which unlocks the dependency range for Phantom, and we no longer need to be on the "new" version of grunt-mocha-phantomjs

PR #2468, opened 2015-07-16
Revert "Upgrade grunt-mocha-phantomjs to 0.6.2"

rcchen:
Reverts palantir/plottable#2467


rcchen:
Reverting this one because mocha-phantomjs published v3.5.6 which unlocks the dependency range for Phantom, and we no longer need to be on the "new" version of grunt-mocha-phantomjs

PR #2478, opened 2015-07-20
2106 Legend.symbol() doesn't pass in index correctly

ztsai:
index was wrong due to nested selection

### DIFF
--- a/test/components/legendTests.ts
+++ b/test/components/legendTests.ts
@@ -355,4 +355,25 @@ describe("Legend", () => {
 
     svg.remove();
   });
+
+  it("symbol() passes index correctly", () => {
+    var getFactory  = (i: number) => i == 0 ? squareFactory : circleFactory;
+
+    var domain = ["AA", "BB"];
+    color.domain(domain);
+    var squareFactory = Plottable.SymbolFactories.square();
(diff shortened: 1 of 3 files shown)


test/components/legendTests.ts
```diff
@@ -355,4 +355,25 @@ describe("Legend", () => {
...
+    var circleFactory = Plottable.SymbolFactories.circle();
+
+    legend.symbol((d: any, i: number) => getFactory(i));
+    legend.renderTo(svg);
+
+    var symbols = d3.selectAll(symbolSelector);
+    symbols.each(function(d: any, i: number) {
+      var element = d3.select(this);
+      var symbolSize = element.node().getBoundingClientRect().height;
+      assert.strictEqual(element.attr("d"), getFactory(i)(symbolSize));
```
jtlan: I think it should be possible to test this bug with less overhead:

```Typescript
var expectedIndex = 0;
var symbolChecker = (d: any, index: number) => {
  assert.strictEqual(index, expectedIndex, "message");
  expectedIndex++;
  return "";
};
legend.symbol(symbolChecker);
```

ztsai:
branch name changed; new pull request: https://github.com/palantir/plottable/pull/2478

PR #2482, opened 2015-07-21
[JSHint] - removing from repo

bluong:
Now that we have a decently nice set of rules defined through eslint, I think it is safe to remove JSHint from the codebase so that we don't have multiple tools doing the same thing.  I'll continue to look into adding more rules to eslint, but that's in the future


aicioara:
How about `jscs`? Shall we remove that as well?

aicioara:
Please remove `.jshintrc`

bluong:
`jscs` is decently special in that it only looks at code style, which is an area that is different than `jshint` and `eslint` to my understanding.

aicioara:
Can't `eslint` do that? If not, then why? Linting is code style.

PR #2484, opened 2015-07-21
2106 Legend.symbol() doesn't pass in index correctly

### LINKED ISSUE #2106: Legend.symbol() doesn't pass in index correctly
```
legend.symbolFactoryAccessor(function (d, i) {
    console.log(i);
  });
```
logs:
0
0
0

ztsai:
index was wrong due to nested selection

closes #2106

### DIFF
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -277,9 +277,9 @@ export module Components {
         });
       });
 
-      entries.select("path").attr("d", (d: any, i: number) => this.symbol()(d, i)(layout.textHeight * 0.6))
+      entries.select("path").attr("d", (d: any, i: number, j: number) => this.symbol()(d, j)(layout.textHeight * 0.6))
                             .attr("transform", "translate(" + (layout.textHeight / 2) + "," + layout.textHeight / 2 + ")")
-                            .attr("fill", (value: string) => this._colorScale.scale(value) )
+                            .attr("fill", (value: string) => this._colorScale.scale(value))
                             .classed(Legend.LEGEND_SYMBOL_CLASS, true);
 
       var padding = this._padding;
--- a/test/components/legendTests.ts
+++ b/test/components/legendTests.ts
@@ -362,7 +362,7 @@ describe("Legend", () => {
 
     var expectedIndex = 0;
     var symbolChecker = (d: any, index: number) => {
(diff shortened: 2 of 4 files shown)


bluong:
Please make a pull request to DefinitelyTyped and cross reference so we have the consistent diff between the typings file in both places.

bluong:
@jtlan made a comment on the previous iteration of the pull request that can essentially let you dodge the need to check the exact d attribute, which is possibly better.

All we want to check is the index of the entry, so let's verify only the index

src/components/legend.ts
```diff
@@ -277,9 +277,9 @@ export module Components {
         });
       });
 
-      entries.select("path").attr("d", (d: any, i: number) => this.symbol()(d, i)(layout.textHeight * 0.6))
+      entries.select("path").attr("d", (d: any, i: number, j: number) => this.symbol()(d, j)(layout.textHeight * 0.6))
```
bluong: See if you can remove the explicit typings from here.  Not a big deal though.

test/components/legendTests.ts
```diff
@@ -355,4 +355,20 @@ describe("Legend", () => {
...
     svg.remove();
   });
+
+  it("symbol() passes index correctly", () => {
+    var domain = ["AA", "BB", "CC"];
+    color.domain(domain);
+
+    var expectedIndex = 0;
+    var symbolChecker = (d: any, index: number) => {
+      assert.strictEqual(index, expectedIndex, "message");
```
bluong: The message here should be something that's more meangingful.

bluong:
@jtlan Routing the assignment to you since I'm biased on this PR.

test/components/legendTests.ts
jtlan: When I originally typed `"message"`, I originally meant that as an example ... I should have made myself clearer.

aicioara:
@jtlan Just curious, why did we alter `d3.d.ts` ? It seems odd

crmorford:
Congrats on your first plottable PR @ztsai :)

ztsai:
@aicioara the diff looks different because our d3.d.ts was not up to date, but the raw files would be the same

aicioara:
I see, thank you.

Also, congrats for the first PR @ztsai !

PR #2485, opened 2015-07-21
Add a flag on Numeric axes to approximate the width of the labels.

PeterFaiman:
Part of #2433

Approximating the width reduces measure() calls, which are expensive. Using a simple approximation speeds things up on my machine by 30% or more, depending on how many labels are measured to render the graph.

JSFiddle: http://jsfiddle.net/uk3b4ykj/1/

API Changes:

* `Plottable.Axes.Numeric.usesTextWidthApproximation(enable: boolean)` has been added as a getter/setter to enable approximation

This approximation is particularly suited to the Numeric axis, as numbers are all basically the same size in most fonts.

A test running 20 graphs takes between 400 and 500 ms to run.

head-20

With approximate text measuring, it takes between 250 and 350 ms.

branch-20

This is definitely due to the measurement calls, as can be seen in the below timeline of a single graph render.

head-1

And the corresponding improvement.

branch-1

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1905,13 +1905,13 @@ declare module Plottable {
              *
              * @returns {boolean} The current approximate text width setting.
              */
-            approximateTextWidth(): boolean;
+            usesTextWidthApproximation(): boolean;
             /**
              * Sets the approximate text width setting.
              *
              * @param {boolean} enable faster but less accurate computation of text widths.
              */
-            approximateTextWidth(enable: boolean): any;
+            usesTextWidthApproximation(enable: boolean): any;
         }
     }
 }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1913,7 +1913,7 @@ declare module Plottable {
              * Additionally, very abnormal fonts may not approximate reasonably.
              *
              * @param {boolean} The new text width approximation setting.
-             * @returns {Axes.Numeric} The calling Plottable.Axes.Numeric.
+             * @returns {Plottable.Axes.Numeric} The calling Plottable.Axes.Numeric.
              */
             usesTextWidthApproximation(enable: boolean): Axes.Numeric;
         }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1903,13 +1903,16 @@ declare module Plottable {
             /**
              * Gets the approximate text width setting.
              *
-             * @returns {boolean} The current approximate text width setting.
+             * @returns {boolean} The current text width approximation setting.
              */
             usesTextWidthApproximation(): boolean;
             /**
-             * Sets the approximate text width setting.
+             * Sets the approximate text width setting. Approximating text width
+             * measurements can drastically speed up plot rendering, but the plot may
+             * have extra white space that would be eliminated by exact measurements.
+             * Additionally, very abnormal fonts may not approximate reasonably.
              *
-             * @param {boolean} enable faster but less accurate computation of text widths.
+             * @param {boolean} The new text width approximation setting.
              */
             usesTextWidthApproximation(enable: boolean): void;
         }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1913,8 +1913,9 @@ declare module Plottable {
              * Additionally, very abnormal fonts may not approximate reasonably.
              *
              * @param {boolean} The new text width approximation setting.
+             * @returns {Axes.Numeric} The calling Plottable.Axes.Numeric.
              */
-            usesTextWidthApproximation(enable: boolean): void;
+            usesTextWidthApproximation(enable: boolean): Axes.Numeric;
         }
     }
 }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1900,6 +1900,18 @@ declare module Plottable {
              * @returns {Numeric} The calling Numeric Axis.
              */
             tickLabelPosition(position: string): Numeric;
+            /**
+             * Gets the approximate text width setting.
+             *
+             * @returns {boolean} The current approximate text width setting.
+             */
+            approximateTextWidth(): boolean;
+            /**
+             * Sets the approximate text width setting.
+             *
+             * @param {boolean} enable faster but less accurate computation of text widths.
+             */
+            approximateTextWidth(enable: boolean): any;
         }
     }
 }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1911,7 +1911,7 @@ declare module Plottable {
              *
              * @param {boolean} enable faster but less accurate computation of text widths.
              */
-            usesTextWidthApproximation(enable: boolean): any;
+            usesTextWidthApproximation(enable: boolean): void;
         }
     }
 }
--- a/src/axes/numericAxis.ts
+++ b/src/axes/numericAxis.ts
@@ -281,13 +281,16 @@ export module Axes {
     /**
      * Gets the approximate text width setting.
      *
-     * @returns {boolean} The current approximate text width setting.
+     * @returns {boolean} The current text width approximation setting.
      */
     public usesTextWidthApproximation(): boolean;
     /**
-     * Sets the approximate text width setting.
+     * Sets the approximate text width setting. Approximating text width
+     * measurements can drastically speed up plot rendering, but the plot may
+     * have extra white space that would be eliminated by exact measurements.
+     * Additionally, very abnormal fonts may not approximate reasonably.
      *
-     * @param {boolean} enable faster but less accurate computation of text widths.
+     * @param {boolean} The new text width approximation setting.
      */
     public usesTextWidthApproximation(enable: boolean): void;
     public usesTextWidthApproximation(enable?: boolean): boolean {
--- a/src/axes/numericAxis.ts
+++ b/src/axes/numericAxis.ts
@@ -5,7 +5,7 @@ export module Axes {
   export class Numeric extends Axis<number> {
 
     private _tickLabelPositioning = "center";
-    private _approximateTextWidth = false;
+    private _usesTextWidthApproximation = false;
     private _measurer: SVGTypewriter.Measurers.Measurer;
     private _wrapper: SVGTypewriter.Wrappers.Wrapper;
 
@@ -30,7 +30,7 @@ export module Axes {
     }
 
     protected _computeWidth() {
-      var maxTextWidth = this._approximateTextWidth ? this._computeApproximateTextWidth() : this._computeExactTextWidth();
+      var maxTextWidth = this._usesTextWidthApproximation ? this._computeApproximateTextWidth() : this._computeExactTextWidth();
 
       if (this._tickLabelPositioning === "center") {
         this._computedWidth = this._maxLabelTickLength() + this.tickLabelPadding() + maxTextWidth;
@@ -283,18 +283,18 @@ export module Axes {
      *
      * @returns {boolean} The current approximate text width setting.
      */
-    public approximateTextWidth(): boolean;
+    public usesTextWidthApproximation(): boolean;
     /**
      * Sets the approximate text width setting.
      *
      * @param {boolean} enable faster but less accurate computation of text widths.
      */
-    public approximateTextWidth(enable: boolean);
-    public approximateTextWidth(enable?: boolean): boolean {
+    public usesTextWidthApproximation(enable: boolean);
+    public usesTextWidthApproximation(enable?: boolean): boolean {
       if (enable == null) {
-        return this._approximateTextWidth;
+        return this._usesTextWidthApproximation;
       } else {
-        this._approximateTextWidth = enable;
+        this._usesTextWidthApproximation = enable;
       }
     }
 
--- a/src/axes/numericAxis.ts
+++ b/src/axes/numericAxis.ts
@@ -289,7 +289,7 @@ export module Axes {
      *
      * @param {boolean} enable faster but less accurate computation of text widths.
      */
-    public usesTextWidthApproximation(enable: boolean);
+    public usesTextWidthApproximation(enable: boolean): void;
     public usesTextWidthApproximation(enable?: boolean): boolean {
       if (enable == null) {
         return this._usesTextWidthApproximation;
--- a/src/axes/numericAxis.ts
+++ b/src/axes/numericAxis.ts
@@ -291,7 +291,7 @@ export module Axes {
      * Additionally, very abnormal fonts may not approximate reasonably.
      *
      * @param {boolean} The new text width approximation setting.
-     * @returns {Plottable.Axes.Numeric} The calling Plottable.Axes.Numeric.
+     * @returns {Axes.Numeric} The calling Axes.Numeric.
      */
     public usesTextWidthApproximation(enable: boolean): Axes.Numeric;
     public usesTextWidthApproximation(enable?: boolean): any {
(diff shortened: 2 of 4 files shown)


aicioara:
I believe this change can then be leveraged further to improve the performance even more, but it is out of the scope of this PR. Reassigning to get more objective feedback, since I am biased.

aicioara:
Part of #2433

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,26 @@ export module Axes {
       }
     }
 
+    /**
+     * Gets the approximate text width setting.
+     *
+     * @returns {boolean} The current approximate text width setting.
+     */
+    public approximateTextWidth(): boolean;
```
bluong: I'm not a fan of the endpoint name.  If I were to call

`numericAxis.approximateTextWidth()`, what would I expect to be returned?

There's probably a better name, but maybe `useTextWidthApproximation()`?

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,26 @@ export module Axes {
...
+    /**
+     * Sets the approximate text width setting.
+     *
+     * @param {boolean} enable faster but less accurate computation of text widths.
+     */
+    public approximateTextWidth(enable: boolean);
+    public approximateTextWidth(enable?: boolean): boolean {
+      if (enable == null) {
+        return this._approximateTextWidth;
+      } else {
```
bluong: This is not that great of a deal, but I personally am not a fan of the `else` statement here mainly because it forces a nesting for very little reason.

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,26 @@ export module Axes {
...
+        this._approximateTextWidth = enable;
+      }
```
bluong: Under the setter case, if we enable the usage of approximation, do we want to rerender the axis again using the approximation?

src/axes/numericAxis.ts
```diff
@@ -5,6 +5,7 @@ export module Axes {
   export class Numeric extends Axis<number> {
 
     private _tickLabelPositioning = "center";
+    private _approximateTextWidth = false;
```
bluong: Comments regarding naming should be reflected here.

src/axes/numericAxis.ts
```diff
@@ -29,21 +30,36 @@ export module Axes {
...
 
-      if (this._tickLabelPositioning === "center") {
-        this._computedWidth = this._maxLabelTickLength() + this.tickLabelPadding() + maxTextLength;
-      } else {
-        this._computedWidth = Math.max(this._maxLabelTickLength(), this.tickLabelPadding() + maxTextLength);
-      }
+    private _computeApproximateTextWidth(): number {
+      var tickValues = this._getTickValues();
+      var mWidth = this._measurer.measure("M").width;
+      var textLengths = tickValues.map((v: any) => {
```
bluong: I don't think you need to explicitly type v to any.  Not sure

bluong:
This mostly looks great!  Can you fix up the issues @PeterFaiman and add a unit test to verify that this still renders nicely?

bluong:
Also, is it possible to pull this idea to be higher?  This operates only on `NumericAxis` for now.  Can you have this operate at the `Axis` level?

src/axes/numericAxis.ts
PeterFaiman: I copied the above code since I don't know TS very well. I'll investigate the proper typing.

src/axes/numericAxis.ts
PeterFaiman: Agreed. I'll think on it.

src/axes/numericAxis.ts
PeterFaiman: No. Approximation is supposed to be an optimization, causing a re-render defeats that goal. It should be set before the axis is rendered at all, optimally.

PeterFaiman:
Tests coming. I thought about just adding it to Axis, but there are going to be a lot more weird renderings if we do that. It is off by default though. Also, Numeric overrides `_computeWidth` itself, so I'd have to implement it here anyway first. I could potentially implement an approximation for all axes.

bluong:
Tests and we are good to go i think

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
   });
 
+  it("reasonably approximates tick label sizes with approximate measuring", () => {
+    var SVG_WIDTH = 500;
+    var SVG_HEIGHT = 100;
+
+    [[-1, 1],
+     [0, 10],
+     [0, 999999999]
+    ].forEach((domainBounds) => {
```
bluong: I'm personally not a fan of operating on a literal as the above.  Can you wrap the array into a variable and then operate on the variable instead?

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+      var scale = new Plottable.Scales.Linear();
+      scale.domain(domainBounds);
+
+      var svgApprox = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var numericAxisApprox = new Plottable.Axes.Numeric(scale, "left");
+      numericAxisApprox.usesTextWidthApproximation(true);
+      numericAxisApprox.renderTo(svgApprox);
+      var widthApprox = (<any> numericAxisApprox)._computeWidth();
```
bluong: Can we use the the `.width()` API here instead?

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+
+      var svgExact = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var numericAxisExact = new Plottable.Axes.Numeric(scale, "left");
+      numericAxisExact.usesTextWidthApproximation(false);
+      numericAxisExact.renderTo(svgExact);
+      var widthExact = (<any> numericAxisExact)._computeWidth();
```
bluong: Can we use the the `.width()` API here instead?

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+      var scale = new Plottable.Scales.Linear();
+      scale.domain(domainBounds);
+
+      var svgApprox = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var numericAxisApprox = new Plottable.Axes.Numeric(scale, "left");
+      numericAxisApprox.usesTextWidthApproximation(true);
+      numericAxisApprox.renderTo(svgApprox);
+      var widthApprox = (<any> numericAxisApprox)._computeWidth();
+
+      var svgExact = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
```
bluong: Through the rest of our tests, we usually (I think actually always) use one svg to do everything in.  Mainly to keep with consistency in usability and codewise, can you just use one svg?

Maybe like
First, render using no approximation and get the width.
Second, set the flag to use approximation and get the width.
Compare the widths.

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+      var numericAxisExact = new Plottable.Axes.Numeric(scale, "left");
+      numericAxisExact.usesTextWidthApproximation(false);
+      numericAxisExact.renderTo(svgExact);
+      var widthExact = (<any> numericAxisExact)._computeWidth();
+
+      svgApprox.remove();
+      svgExact.remove();
+
+      assert.isTrue(widthApprox < (widthExact * 1.55),
```
bluong: Where is this 1.55 coming from?

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+        "an approximate scale of [" + domainBounds[0] + "," + domainBounds[1] + "] is less than 55% larger than an exact scale");
+      assert.isTrue(widthApprox >= widthExact,
+        "an approximate scale of [" + domainBounds[0] + "," + domainBounds[1] + "] is smaller than an exact scale");
```
bluong: There is an `assert.operator` for usage with `<` and `>=`.  Please use that instead of `assert.isTrue`

test/axes/numericAxisTests.ts
```diff
@@ -431,4 +431,36 @@ describe("NumericAxis", () => {
...
+      var widthApprox = (<any> numericAxisApprox)._computeWidth();
+
+      var svgExact = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var numericAxisExact = new Plottable.Axes.Numeric(scale, "left");
+      numericAxisExact.usesTextWidthApproximation(false);
+      numericAxisExact.renderTo(svgExact);
+      var widthExact = (<any> numericAxisExact)._computeWidth();
+
+      svgApprox.remove();
+      svgExact.remove();
```
bluong: Please put the svg removal at the end of the unit test.  This allows us to easily inspect the svg if the test fails

test/axes/numericAxisTests.ts
PeterFaiman: Probably yeah, .width() just adds in extra space like the margin.

test/axes/numericAxisTests.ts
PeterFaiman: Can I just render over and over to 1 SVG and have stuff work? I wasn't sure.

test/axes/numericAxisTests.ts
bluong: I prefer using the public API where applicable unless the test really needs to query something from the private functions.  It resolves a bunch of headaches later down the road if we decide to change the name of a private function.

test/axes/numericAxisTests.ts
bluong: It should totally work.  To keep things more concise, I don't think you even need to use two numeric axes.  You just need to render one with approximation and then switch if it uses TextWidthApproximation.

test/axes/numericAxisTests.ts
PeterFaiman: I made it up to fit the current runs of the test. It essentially exists to flag if things get way bigger than they already are.

bluong: Let's factor it out to a variable so that this is more transparent

bluong:
This seems good to me!

bluong:
Can you provide a description of the API in the pull request description for QE to reference?

crmorford:
Can anyone explain to me why it pretty consistently uses one less label for the approximation? I can't puzzle through it myself.
screen shot 2015-07-27 at 5 31 59 pm
screen shot 2015-07-27 at 5 31 15 pm

aicioara:
The approximation assumes that each letter is the size of the letter M (which is the wides letter in the alphabet). Hence it believes (by chance) that whatever last label is, it does not fit, so it deletes it.

aicioara:
Unsure if that is an acceptable behavior, but the above is the explanation

crmorford:
thanks @aicioara!

crmorford:
I found something a little weird 
http://jsfiddle.net/uk3b4ykj/3/
The yAxisApprox function is doing something strange to the top xAxis, there are only ticks on the left half and it extends far past the data
screen shot 2015-07-27 at 5 39 55 pm

crmorford:
oh. not a bug
(not showing 1000, which would be interpreted as 4Ms)
http://jsfiddle.net/uk3b4ykj/4/

jtlan:
@crmorford, one neat thing to do to stress-test this is to make a chart with tons of Axes and see if it speeds up at all.

crmorford:
here's another weird one
without useTextWidthApproximation, I can add 10 axes. http://jsfiddle.net/uk3b4ykj/5/
with it uncommented, all the axes collapse http://jsfiddle.net/uk3b4ykj/6/ 
(line 33)

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,29 @@ export module Axes {
...
+     * Additionally, very abnormal fonts may not approximate reasonably.
+     *
+     * @param {boolean} The new text width approximation setting.
+     */
+    public usesTextWidthApproximation(enable: boolean): void;
+    public usesTextWidthApproximation(enable?: boolean): boolean {
+      if (enable == null) {
+        return this._usesTextWidthApproximation;
+      } else {
+        this._usesTextWidthApproximation = enable;
```
bluong: Oh I'm incredibly foolish.

In the Plottable codebase. we try to stick with the fluent style, where getters return the gotten object and where setters return the calling object.  This allows for chaining.

In this scenario, after you set the variable here, please `return this;`.

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,29 @@ export module Axes {
...
+    public usesTextWidthApproximation(): boolean;
+    /**
+     * Sets the approximate text width setting. Approximating text width
+     * measurements can drastically speed up plot rendering, but the plot may
+     * have extra white space that would be eliminated by exact measurements.
+     * Additionally, very abnormal fonts may not approximate reasonably.
+     *
+     * @param {boolean} The new text width approximation setting.
+     */
+    public usesTextWidthApproximation(enable: boolean): void;
```
bluong: In accordance, this `void` should return `Axes.Numeric` instead.

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,29 @@ export module Axes {
...
+     * @returns {boolean} The current text width approximation setting.
+     */
+    public usesTextWidthApproximation(): boolean;
+    /**
+     * Sets the approximate text width setting. Approximating text width
+     * measurements can drastically speed up plot rendering, but the plot may
+     * have extra white space that would be eliminated by exact measurements.
+     * Additionally, very abnormal fonts may not approximate reasonably.
+     *
+     * @param {boolean} The new text width approximation setting.
```
bluong: Please add the `returns` doc that it returns the calling `Axes.Numeric`

plottable.d.ts
```diff
@@ -1900,6 +1900,22 @@ declare module Plottable {
...
+             */
+            usesTextWidthApproximation(): boolean;
+            /**
+             * Sets the approximate text width setting. Approximating text width
+             * measurements can drastically speed up plot rendering, but the plot may
+             * have extra white space that would be eliminated by exact measurements.
+             * Additionally, very abnormal fonts may not approximate reasonably.
+             *
+             * @param {boolean} The new text width approximation setting.
+             * @returns {Axes.Numeric} The calling Plottable.Axes.Numeric.
```
bluong: Modify to `Plottable.Axes.Numeric`

bluong:
Just the comment above!

src/axes/numericAxis.ts
```diff
@@ -262,6 +278,31 @@ export module Axes {
...
+     * @returns {Plottable.Axes.Numeric} The calling Plottable.Axes.Numeric.
```
bluong: Sorry I think I was unclear...

Can you replace with
`{Axes.Numeric} The calling Axes.Numeric.`

PeterFaiman: Haha oops!

bluong:
Thanks a lot for the contribution @PeterFaiman !

aicioara:
Nice!

PR #2469, opened 2015-07-16
Gruntfile refactor

aicioara:
Part of #2445

BLUF: Nothing has been changed (functionally). Things have moved from one place to another. Prereq for deeper purge

Done:
- Renamed variables
- Extracted configuration objects
- Inlined variables
- Fixed indentation

Not done:
- Deleting tasks that are not used anymore
- Improve the minification 
- Improve the run time

Issue #2445, opened 2015-07-07
Clean GruntFile


aicioara:
There's lots of stuff inside which is no longer used. Make it pretty.

rcchen:
Might just end up doing this along with moving to Gulp (#2362)

aicioara:
There is an open discussion on the issue above which is now part of this

aicioara:
Self note. 

`npm test` currently takes `1:45.08`

Improvements: 
- Tslint is super slow, submit pull request 
- There is a small chance we are compiling twice
- SauceLabs can run in parallel after the compilation is done

aicioara:
Self note.

`grunt dev-compile` currently takes `9.3 s`. This might not be able to cut, but the watch looks promising

aicioara:
Check progress here https://github.com/palantir/plottable/compare/gruntPurge?expand=1

aicioara:
`1:38` build POC
https://travis-ci.org/palantir/plottable/builds/71370993

aicioara:
http://stackoverflow.com/questions/26411084/how-to-log-colored-messages-with-grunt-log

aicioara:
Test building 
https://travis-ci.org/palantir/plottable/builds/72038100
vs
https://travis-ci.org/palantir/plottable/builds/72545428

aicioara:
Assure backwards compatibility with TS1.4 for d.ts

aicioara:
I think I can close this. The Gruntfile looks pretty good now.

PR #2486, opened 2015-07-21
Improved incremental build time of watch 

aicioara:
Part of #2445 

Refactored the Gruntfile even more as a followup of #2469 and part of my hack week project.

As part of this change:
- Improved incremental build time  | Before `9.127 s` | After `3.321 s`
- Typescript is creating a `.baseDir.ts` file. Added to gitignore. Known issue. https://github.com/grunt-ts/grunt-ts/issues/77
- Tests no longer get recompiled upon src changes. This will be caught in the travis CI, which makes sense to me.

This change has been briefly tested, but I want to see how it behaves in the wild. Ping @aicioara for any discovered issues. In particular, my addition `spawn: false` in the watch config was documented as unstable.

Not included in this change:
- The incremental build time can be improved even more by somehow caching the `tsc` command. Notice that `2 s` out of the required `3.321s` are just booting the tsc (which as a dependency parses the `lib.d.ts` file). If we can use the built in watch of the TSC, that would be huge. However `grunt-ts` is lacking this feature, or rather reimplements it inefficiently. 

Pre-req for TS1.4 compatibility checking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ test/tests.js
 
 # Visual Studio Code
 .settings/
+
+# Typescript
+.baseDir.ts
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,7 +19,6 @@ module.exports = function(grunt) {
     test: {
       src: ["test/*.ts", "typings/**/*.d.ts", "build/plottable.d.ts"],
       outDir: "build/test/",
-      // watch: "test",
       options: {
         target: "es5",
         sourceMap: false,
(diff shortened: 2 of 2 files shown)


jtlan:
> In particular, my addition `spawn: false` in the watch config was documented as unstable.

Can you explain what this means?

aicioara:
https://github.com/gruntjs/grunt-contrib-watch#user-content-optionsspawn

jtlan:
I'm a bit concerned about the spawn part, but we should try it out. we can always revert just that part if it doesn't pan out.

jtlan:
[NOQE] because no source changes.

PR #2490, opened 2015-07-22
Removed plottable_testing.css

aicioara:
Old and not used anywhere as far as I can tell.

@crmorford can you confirm?


aicioara:
Even if used, it should probably belong to the quicktests folder

PR #2496, opened 2015-07-22
[SelectionBoxLayer] - Allow installation of scales

### LINKED ISSUE #2475: Drag box selection in data space rather than pixel space
When selecting, have the option to select in data space rather than pixel space. Be careful with the crimp line feature here as well as log scales. 

@benduffield has some code for this, reach out to him.

bluong:
Allowing installation of scales onto the `SelectionBoxLayer` allows for `SelectionBoxLayer`s to be data-backed.  This essentially means that the component can react to changes on the scale, re-rendering to place the box in correct location.

JSFiddle: http://jsfiddle.net/bluong63/zajmnedm/

API Changes:
```typescript
/**
 * Gets the x scale for this SelectionBoxLayer.
 */
xScale(): QuantitativeScale<number | { valueOf(): number; }>;
/**
 * Sets the x scale for this SelectionBoxLayer.
 *
 * @returns {SelectionBoxLayer} The calling SelectionBoxLayer.
 */
xScale(xScale: QuantitativeScale<number | { valueOf(): number; }>): SelectionBoxLayer;
/**
 * Gets the y scale for this SelectionBoxLayer.
 */
yScale(): QuantitativeScale<number | { valueOf(): number; }>;
/**
 * Sets the y scale for this SelectionBoxLayer.
 *
 * @returns {SelectionBoxLayer} The calling SelectionBoxLayer.
*/
yScale(yScale: QuantitativeScale<number | { valueOf(): number; }>): SelectionBoxLayer;
/**
 * Gets the data values backing the left and right edges of the box.
 *
 * Returns an undefined array if the edges are not backed by a scale.
 */
xExtent(): (number | { valueOf(): number; })[];
/**
 * Gets the data values backing the top and bottom edges of the box.
 *
 * Returns an undefined array if the edges are not backed by a scale.
 */
yExtent(): (number | { valueOf(): number; })[];
```

Closes #2475


src/components/selectionBoxLayer.ts
```diff
@@ -10,10 +10,18 @@ export module Components {
       topLeft: { x: 0, y: 0 },
       bottomRight: { x: 0, y: 0 }
     };
+    private _boxDataBounds: Bounds = {
```
jtlan: What's the difference between these `Bounds` and the other one? If these are in data-space, then a more descriptive name should be used (the "data" here is ambiguous). In addition, `Bounds` are defined as containing `Point`s which contain `number`s, which means you couldn't use it to hold other kinds of data (like `Date`s).

jtlan:
With the current implementation, there are going to be typing errors introduced because of the `<any>`s, and because `Bounds` expect `number`s and not all `Scale`s produce them.

bluong:
I'll be in process to correct the implementation.

At the very least, does the approach from an API usability standpoint make sense?

src/components/selectionBoxLayer.ts
```diff
@@ -111,6 +131,59 @@ export module Components {
...
+    /**
+     * Gets the x scale for this SelectionBoxLayer.
+     */
+    public xScale(): QuantitativeScale<any>;
+    /**
+     * Sets the x scale for this SelectionBoxLayer.
+     * 
+     * @returns {SelectionBoxLayer} The calling SelectionBoxLayer.
+     */
+    public xScale(xScale: QuantitativeScale<any>): SelectionBoxLayer;
```
jtlan: The `any` here seems pretty iffy. Is there a way to parameterize on the type of scale?

jtlan:
The `destroy()` call should disconnect from the `Scale`s.

jtlan:
Will there be a way to retrieve the current bounds/extents in data-space? What will get returned for `XDragBoxLayer` and `YDragBoxLayer`?

bluong:
Not currently, but we can supply these endpoints if you/users think this is important.

The definition of bounds should be 100% unchanged from before this pull request

jtlan:
I'm not sure what the point of this feature is, if the scaled values aren't accessible. Is it just to avoid the user having to store the data-space values on their own?

src/components/selectionBoxLayer.ts
```diff
@@ -111,6 +131,101 @@ export module Components {
...
+      this._bindBoxDataValues();
+      return this;
+    }
+
+    /**
+     * Gets the data value backing the left edge of the box.
+     *
+     * Returns null if the edge is not backed by a scale.
+     */
+    public boxLeftDataValue() {
```
jtlan: Instead of having `boxLeftDataValue()` and `boxRightDataValue()`, how about having `xExtent()`?

Also, what gets returned if `xScale()` hasn't been set?

src/components/selectionBoxLayer.ts
```diff
@@ -111,6 +131,101 @@ export module Components {
...
+    public boxRightDataValue() {
+      return this._boxRightDataValue;
+    }
+
+    /**
+     * Gets the data value backing the top edge of the box.
+     *
+     * Returns null if the edge is not backed by a scale.
+     */
+    public boxTopDataValue() {
```
jtlan: Instead of having `boxTopDataValue()` and `boxBottomDataValue()`, how about having `yExtent()`?

Also, what gets returned if `xScale()` hasn't been set?

src/components/selectionBoxLayer.ts
bluong: I was personally considering `xExtent()` until I thought of this approach.  The advantages of this approach is that now the things being returned are immutable and a lot more direct on what the returned values are.

The disadvantages are a slightly longer namespace and more endpoints.

What do you think?

src/components/selectionBoxLayer.ts
```diff
@@ -111,6 +131,84 @@ export module Components {
...
+      return this;
+    }
+
+    /**
+     * Gets the data values for the Bounds of the box.
+     *
+     * An undefined value will be returned if a scale is not backing a given edge, i.e.
+     * no xScale is backing the left / right edges of the box.
+     */
+    public dataBounds() {
```
jtlan: What's the return type here? The values might not be `number`s, so this object doesn't contain `Point`s, and is therefore not a `Bounds` object...

jtlan:
I think it makes more sense to have separate `xExtent()` and `yExtent()` methods -- consider what happens on `XDragBoxLayer` and `YDragBoxLayer`.

src/components/selectionBoxLayer.ts
bluong: Yes it is not a `Bounds` object, the item returned is just `{topLeft: {x: numbery, y: numbery}, bottomRight: {x: numbery, y: numbery}}`

jtlan: But the name suggests that it returns a `Bounds` object, while returning something that's only *similar* a `Bounds`. Possibly that could lead to confusion.

jtlan: Further, what gets returned for an `XDragBoxLayer`? Would I expect `null` for the Y data values?

test/components/selectionBoxLayerTests.ts
```diff
@@ -75,4 +75,111 @@ describe("SelectionBoxLayer", () => {
...
+    var bottomRight: Plottable.Point = {
+      x: 250,
+      y: 300
+    };
+    sbl.bounds({
+      topLeft: topLeft,
+      bottomRight: bottomRight
+    });
+
+    assert.strictEqual(sbl.boxXExtent()[0], 400, "data value maps correctly");
```
jtlan: What about the other end of the extent?

src/components/selectionBoxLayer.ts
```diff
@@ -111,6 +131,83 @@ export module Components {
...
+      this._bindBoxDataValues();
+      return this;
+    }
+
+    /**
+     * Gets the data values backing the left and right edges of the box.
+     *
+     * Returns an undefined array if the edges are not backed by a scale.
+     */
+    public boxXExtent() {
```
jtlan: I don't think we need the "box" here -- for example, `bounds()` doesn't have "box" in front of it.

jtlan:
Can you add some unit tests for the Error-throwing functionality?

jtlan:
@crmorford, I'd like to get your opinions on the Typescript signatures and return values of these methods.

jtlan:
I wonder if instead of
```Typescript
xScale<D extends number | { valueOf(): number; >(xScale: QuantitativeScale<D>): SelectionBoxLayer;
```
we could just do
```Typescript
xScale(xScale: QuantitativeScale<number | { valueOf(): number; }>): SelectionBoxLayer;
```

Seems like less overhead for the user.

plottable.d.ts
```diff
@@ -2331,6 +2331,51 @@ declare module Plottable {
...
+            yScale<D extends number | {
+                valueOf(): number;
+            }>(yScale: QuantitativeScale<D>): SelectionBoxLayer;
+            /**
+             * Gets the data values backing the left and right edges of the box.
+             *
+             * Returns an undefined array if the edges are not backed by a scale.
+             */
+            xExtent(): {
+                valueOf(): number;
```
jtlan: Is this going to compile under TS1.4? Right now, I think this will cause a compiler error in Typescript 1.4:
```Typescript
var xExtent = <number[]> selectionBoxLayer.xExtent();
```
because TS1.4 doesn't consider `{ valueOf(): number }` convertible to `number`.

bluong:
There was some trouble after using Typescript 1.4 due to the generic.  After thinking about it, I'm unsure of the usefulness of the generic and I took it out.

jtlan:
Can you add a comment as to why the explicit return type is required?

bluong:
The error was along the lines of

With the generic type `<D extends number | { valueOf(): number}>`, Typescript 1.4 compiler errors when setting to a variable of type `number | {valueOf(): number}`.

Specfically I think it erred on the line

`this._xScale = xScale;`

  This to me seems like a bug on Typescript that got fixed in a recent version.

crmorford:
Talked offline:
Calling dbl.xExtent() on a YDragBoxLayer (and vice versa) should throw an error instead of returning [null, null]

crmorford:
Just nitpicking now, but:
screen shot 2015-07-29 at 2 04 27 pm
instead of 'y scales' and 'y extent' can we use 'yScales' and 'yExtent' to closer match the API?

crmorford:
after the errors are fixed, ready to merge

PR #2500, opened 2015-07-23
add api for keyup event listener

### LINKED ISSUE #2393: Key interaction should have a "onKeyHold" endpoint
It'd be convenient to be able to know when a key was being held down, for example in issue #2011 holding the ctrl key to change which interaction you're using.

ztsai:
added key interaction api:
```onKeyRelease(keyCode: number, callback: KeyCallback)```
```offKeyRelease(keyCode: number, callback: KeyCallback)```

example: http://jsfiddle.net/5c6djgco/2/

closes #2393

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3607,6 +3607,19 @@ declare module Plottable {
              * @return {Dispatchers.Key} The calling Key Dispatcher.
              */
             offKeyDown(callback: KeyCallback): Key;
+            /** Registers a callback to be called whenever a key is released.
+             *
+             * @param {KeyCallback} callback
+             * @return {Dispatchers.Key} The calling Key Dispatcher.
+             */
+            onKeyUp(callback: KeyCallback): Key;
+            /**
+             * Removes the callback to be called whenever a key is released.
+             *
+             * @param {KeyCallback} callback
+             * @return {Dispatchers.Key} The calling Key Dispatcher.
+             */
+            offKeyUp(callback: KeyCallback): Key;
         }
     }
 }
@@ -3735,6 +3748,24 @@ declare module Plottable {
              * @returns {Interactions.Key} The calling Key Interaction.
              */
             offKeyPress(keyCode: number, callback: KeyCallback): Key;
+            /**
+             * Adds a callback to be called when the key with the given keyCode is
+             * released and the user is moused over the Component.
+             *
+             * @param {number} keyCode
+             * @param {KeyCallback} callback
+             * @returns {Interactions.Key} The calling Key Interaction.
+             */
+            onKeyRelease(keyCode: number, callback: KeyCallback): Key;
+            /**
+             * Removes a callback that would be called when the key with the given keyCode is
+             * released and the user is moused over the Component.
+             *
+             * @param {number} keyCode
+             * @param {KeyCallback} callback
+             * @returns {Interactions.Key} The calling Key Interaction.
+             */
+            offKeyRelease(keyCode: number, callback: KeyCallback): Key;
         }
     }
 }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3750,7 +3750,7 @@ declare module Plottable {
             offKeyPress(keyCode: number, callback: KeyCallback): Key;
             /**
              * Adds a callback to be called when the key with the given keyCode is
-             * released and the user is moused over the Component.
+             * released if the key was pressed with the mouse inside of the Component.
              *
              * @param {number} keyCode
              * @param {KeyCallback} callback
@@ -3759,7 +3759,7 @@ declare module Plottable {
             onKeyRelease(keyCode: number, callback: KeyCallback): Key;
             /**
              * Removes a callback that would be called when the key with the given keyCode is
-             * released and the user is moused over the Component.
+             * released if the key was pressed with the mouse inside of the Component.
              *
              * @param {number} keyCode
              * @param {KeyCallback} callback
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3577,8 +3577,8 @@ declare module Plottable {
 
 
 declare module Plottable {
+    type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
     module Dispatchers {
-        type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
         class Key extends Dispatcher {
             /**
              * Gets a Key Dispatcher. If one already exists it will be returned;
@@ -3725,7 +3725,6 @@ declare module Plottable {
 
 
 declare module Plottable {
-    type KeyCallback = (keyCode: number) => void;
     module Interactions {
         class Key extends Interaction {
             protected _anchor(component: Component): void;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3577,8 +3577,8 @@ declare module Plottable {
 
 
 declare module Plottable {
-    type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
     module Dispatchers {
+        type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
         class Key extends Dispatcher {
             /**
              * Gets a Key Dispatcher. If one already exists it will be returned;
@@ -3725,6 +3725,7 @@ declare module Plottable {
 
 
 declare module Plottable {
+    type KeyCallback = (keyCode?: number) => void;
     module Interactions {
         class Key extends Interaction {
             protected _anchor(component: Component): void;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3578,7 +3578,7 @@ declare module Plottable {
 
 declare module Plottable {
     module Dispatchers {
-        type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
+        type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
         class Key extends Dispatcher {
             /**
              * Gets a Key Dispatcher. If one already exists it will be returned;
@@ -3725,7 +3725,7 @@ declare module Plottable {
 
 
 declare module Plottable {
-    type KeyCallback = (keyCode?: number) => void;
+    type KeyCallback = (keyCode: number) => void;
     module Interactions {
         class Key extends Interaction {
             protected _anchor(component: Component): void;
--- a/src/dispatchers/keyDispatcher.ts
+++ b/src/dispatchers/keyDispatcher.ts
@@ -7,6 +7,7 @@ export module Dispatchers {
   export class Key extends Dispatcher {
     private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
     private _keydownCallbacks: Utils.CallbackSet<KeyCallback>;
+    private _keyupCallbacks: Utils.CallbackSet<KeyCallback>;
 
     /**
      * Gets a Key Dispatcher. If one already exists it will be returned;
@@ -32,9 +33,10 @@ export module Dispatchers {
       super();
 
       this._eventToCallback["keydown"] = (e: KeyboardEvent) => this._processKeydown(e);
-
+      this._eventToCallback["keyup"] = (e: KeyboardEvent) => this._processKeyup(e);
       this._keydownCallbacks = new Utils.CallbackSet<KeyCallback>();
-      this._callbacks = [this._keydownCallbacks];
+      this._keyupCallbacks = new Utils.CallbackSet<KeyCallback>();
+      this._callbacks = [this._keydownCallbacks, this._keyupCallbacks];
     }
 
     /**
@@ -59,8 +61,33 @@ export module Dispatchers {
       return this;
     }
 
+    /** Registers a callback to be called whenever a key is released.
+     *
+     * @param {KeyCallback} callback
+     * @return {Dispatchers.Key} The calling Key Dispatcher.
+     */
+    public onKeyUp(callback: KeyCallback): Key {
+      this._setCallback(this._keyupCallbacks, callback);
+      return this;
+    }
+
+    /**
+     * Removes the callback to be called whenever a key is released.
+     *
+     * @param {KeyCallback} callback
+     * @return {Dispatchers.Key} The calling Key Dispatcher.
+     */
+    public offKeyUp(callback: KeyCallback): Key {
+      this._unsetCallback(this._keyupCallbacks, callback);
+      return this;
+    }
+
     private _processKeydown(event: KeyboardEvent) {
       this._keydownCallbacks.callCallbacks(event.keyCode, event);
+    }
+
+    private _processKeyup(event: KeyboardEvent) {
+      this._keyupCallbacks.callCallbacks(event.keyCode, event);
     }
   }
 }
--- a/src/dispatchers/keyDispatcher.ts
+++ b/src/dispatchers/keyDispatcher.ts
@@ -2,7 +2,7 @@
 
 module Plottable {
 export module Dispatchers {
-  export type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
+  export type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
 
   export class Key extends Dispatcher {
     private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
--- a/src/dispatchers/keyDispatcher.ts
+++ b/src/dispatchers/keyDispatcher.ts
@@ -1,9 +1,9 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export module Dispatchers {
-  export type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
+export type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
 
+export module Dispatchers {
   export class Key extends Dispatcher {
     private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
     private _keydownCallbacks: Utils.CallbackSet<KeyCallback>;
--- a/src/dispatchers/keyDispatcher.ts
+++ b/src/dispatchers/keyDispatcher.ts
@@ -1,9 +1,9 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
-
 export module Dispatchers {
+  export type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
+
   export class Key extends Dispatcher {
     private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Key";
     private _keydownCallbacks: Utils.CallbackSet<KeyCallback>;
--- a/src/interactions/keyInteraction.ts
+++ b/src/interactions/keyInteraction.ts
@@ -15,7 +15,7 @@ export module Interactions {
     private _keyReleaseKeyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
 
     private _mouseMoveCallback = (point: Point) => false; // HACKHACK: registering a listener
-    private _keyDowned: { [keyCode: number]: boolean } = {};
+    private _downedKeys = new Plottable.Utils.Set();
     private _keyDownCallback = (keyCode: number) => this._handleKeyDownEvent(keyCode);
     private _keyUpCallback = (keyCode: number) => this._handleKeyUpEvent(keyCode);
 
@@ -47,15 +47,15 @@ export module Interactions {
         if (this._keyPressKeyCodeCallbacks[keyCode]) {
           this._keyPressKeyCodeCallbacks[keyCode].callCallbacks(keyCode);
         }
-        this._keyDowned[keyCode] = true;
+        this._downedKeys.add(keyCode);
       }
     }
 
     private _handleKeyUpEvent(keyCode: number) {
-      if (this._keyDowned[keyCode] && this._keyReleaseKeyCodeCallbacks[keyCode]) {
+      if (this._downedKeys.has(keyCode) && this._keyReleaseKeyCodeCallbacks[keyCode]) {
         this._keyReleaseKeyCodeCallbacks[keyCode].callCallbacks(keyCode);
       }
-      this._keyDowned[keyCode] = false;
+      this._downedKeys.delete(keyCode);
     }
 
     /**
--- a/src/interactions/keyInteraction.ts
+++ b/src/interactions/keyInteraction.ts
@@ -1,7 +1,7 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export type KeyCallback = (keyCode?: number) => void;
+export type KeyCallback = (keyCode: number) => void;
 
 export module Interactions {
   export class Key extends Interaction {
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: plottable.d.ts [hunk 1]
```diff
@@ -3724,6 +3737,24 @@ declare module Plottable {
              * @returns {Interactions.Key} The calling Key Interaction.
              */
             offKeyPress(keyCode: number, callback: KeyCallback): Key;
+            /**
+             * Adds a callback to be called when the key with the given keyCode is
+             * released and the user is moused over the Component.
```
bluong: Very much would want to consider if the user needs to be necessarily moused over the Component.

Let's say that I mouse over a Component, press a key, then mouse out, then release.

I think most would expect the callback for release to be called.

src/interactions/keyInteraction.ts
```diff
@@ -11,10 +11,12 @@ export module Interactions {
      */
     private _positionDispatcher: Plottable.Dispatchers.Mouse;
     private _keyDispatcher: Plottable.Dispatchers.Key;
-    private _keyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
+    private _keydownKeyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
+    private _keyupKeyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
```
bluong: Let's rename these variables to be more relatable to their corresponding endpoints.

Maybe
`_keyPressCallbacks` and
`_keyReleaseCallbacks`

bluong:
Should probably iterate with design / QE to see what the behavior should be specifically for keyRelease.  The fact that release can trigger even though a press may not occur sounds a bit strange to me.

bluong:
Also tests need to be written

crmorford:
Q: if a key is pressed when mouse is in a component, and released outside it, should release callback be fired?
A: In case I would expect the release callback to fire.

Q: if a key is pressed outside of a component and released inside, should release callback be fired?
A: In this case I wouldn't expect a release callback.

For me, where the key press starts feels far more important than where the key press ends.

Another thing to think about is if the key press should be fired if the whole page loses focus. I think I'd expect the callback to fire when I lose focus on the page, if that's possible. Not sure if that's detectable.

plottable.d.ts
bluong: Please rewrite the documentation to express the intended behavior.

src/interactions/keyInteraction.ts
```diff
@@ -11,10 +11,13 @@ export module Interactions {
      */
     private _positionDispatcher: Plottable.Dispatchers.Mouse;
     private _keyDispatcher: Plottable.Dispatchers.Key;
-    private _keyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
+    private _keyPressKeyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
+    private _keyReleaseKeyCodeCallbacks: { [keyCode: string]: Utils.CallbackSet<KeyCallback> } = {};
```
bluong: I don't think the `KeyCode` strings here are necessary

src/interactions/keyInteraction.ts
```diff
@@ -11,10 +11,13 @@ export module Interactions {
...
 
     private _mouseMoveCallback = (point: Point) => false; // HACKHACK: registering a listener
-    private _keyDownCallback = (keyCode: number) => this._handleKeyEvent(keyCode);
+    private _keyDowned: { [keyCode: number]: boolean } = {};
```
bluong: Please use a `Utils.Set`

test/interactions/keyInteractionTests.ts
```diff
@@ -139,5 +139,136 @@ describe("Interactions", () => {
...
+        "callback for \"a\" was not called when \"a\" key was released after being pressed outside");
+
+      $target.simulate("keydown", { keyCode: aCode });
+      TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+      $target.simulate("keyup", { keyCode: aCode });
+      assert.isFalse(aCallbackCalled,
+        "callback for \"a\" was not called when \"a\" key was released inside of component after being pressed outside");
+
+      svg.remove();
+    });
```
bluong: I see a huge test suite here that is decently not too readable.  Is it possible to split this test up so that the test suite is concise and more organized?  Using a mixture of `describe` blocks or so can accomplish this.

bluong:
Mostly looks good.  Please use Plottable's `Set` and have the test suite be more readable.

src/interactions/keyInteraction.ts
ztsai: sorry, I don't understand why it is unnecessary. Different keys should be able to have different callbacks, right?

bluong: Yes, but I personally think `_keyPressCallbacks` and `_keyReleaseCallbacks` is enough of a description.

test/interactions/keyInteractionTests.ts
aicioara: Please don't nest too many describes where possible.

src/interactions/keyInteraction.ts
bluong: I still think the above is shorter and more concise

bluong: While we're modifying this code, I think we can invest in a little effort of code cleanliness.  The way I'm reading the above, a `d3.map()` or a `Utils.Map<string, Utils.CallbackSet<KeyCallback>>` would also work as the type for the above here and it would be in my view more readable and more consitent with the rest of the codebase.

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,69 @@ var assert = chai.assert;
...
+        assert.isFalse(bCallbackCalled, "callback for \"b\" was not called when \"a\" key was pressed");
+      });
 
-        keyInteraction.attachTo(component);
+      it("only fires callback for \"b\" when \"b\" key is pressed", () => {
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        $target.simulate("keydown", { keyCode: bCode });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"b\" key was pressed");
+        assert.isTrue(bCallbackCalled, "callback for \"b\" was called when \"b\" key was pressed");
+      });
```
bluong: This test is redundant I feel.  Remove please.

test/dispatchers/keyDispatcherTests.ts
```diff
@@ -22,6 +22,34 @@ describe("Dispatchers", () => {
...
 
       ked.offKeyDown(callback); // clean up
+
+      keyDowned = false;
+      $("body").simulate("keydown", { keyCode: keyCodeToSend });
+      assert.isFalse(keyDowned, "nothing happens when a key was pressed");
     });
+
+    it("triggers callback on keyup", () => {
+          var ked = Plottable.Dispatchers.Key.getDispatcher();
```
aicioara: Indentation. We use 2 spaces.

test/dispatchers/keyDispatcherTests.ts
```diff
@@ -22,6 +22,34 @@ describe("Dispatchers", () => {
...
+            assert.strictEqual(code, keyCodeToSend, "correct keycode was passed");
+            assert.isNotNull(e, "key event was passed to the callback");
+          };
+
+          ked.onKeyUp(callback);
+
+          $("body").simulate("keyup", { keyCode: keyCodeToSend });
+          assert.isTrue(keyUped, "callback when a key was release");
+
+          ked.offKeyUp(callback); // clean up
```
aicioara: I know it's been used above as well and we don't have a tslint rule for this, but let's avoid comments at the end of the line (move above)

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,62 @@ var assert = chai.assert;
...
-      keyInteraction.attachTo(component);
-
-      var $target = $(component.background().node());
-
-      TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
-      $target.simulate("keydown", { keyCode: bCode });
-      assert.isTrue(bCallbackCalled, "callback for \"b\" was called when \"b\" key was pressed");
-
+    afterEach(() => {
       svg.remove();
```
bluong: `svg.remove()` should not be placed in an `afterEach()` block.  This allows us to see the svg if the test fails.

test/interactions/keyInteractionTests.ts
```diff
@@ -101,43 +72,127 @@ describe("Interactions", () => {
...
+          var aCallback2Called = false;
+          var aCallback2 = () => aCallback2Called = true;
+
+          keyInteraction.onKeyPress(aCode, aCallback1);
+          keyInteraction.onKeyPress(aCode, aCallback2);
+
+          TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+          $target.simulate("keydown", { keyCode: aCode });
+          assert.isTrue(aCallback1Called, "callback 1 for \"a\" was called when \"a\" key was pressed");
+          assert.isTrue(aCallback1Called, "callback 2 for \"b\" was called when \"a\" key was pressed");
```
bluong: The above two asserts are the same...  Please modify.

test/interactions/keyInteractionTests.ts
```diff
@@ -101,43 +72,127 @@ describe("Interactions", () => {
...
+
+          keyInteraction.offKeyPress(aCode, aCallback1);
+          aCallback1Called = false;
+          aCallback2Called = false;
+          $target.simulate("keydown", { keyCode: aCode });
+          assert.isFalse(aCallback1Called, "callback 1 for \"a\" was disconnected from the interaction");
+          assert.isTrue(aCallback2Called, "callback 2 for \"a\" is still connected to the interaction");
```
bluong: I don't see the purpose of this block here.  The name of this unit test is "multiple keyPress callbacks are possible" but this block here refers to the removal of keyPress callbacks.  Either the unit test needs to be split up or it needs to be renamed

test/interactions/keyInteractionTests.ts
```diff
@@ -101,43 +72,127 @@ describe("Interactions", () => {
...
+      });
 
-        keyInteraction.onKeyPress(aCode, aCallback1);
-        keyInteraction.onKeyPress(aCode, aCallback2);
-        keyInteraction.attachTo(component);
+      it("fires callback if key was released after being pressed", () => {
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        $target.simulate("keydown", { keyCode: aCode });
+        $target.simulate("keyup", { keyCode: aCode });
+        assert.isTrue(aCallbackCalled, "callback for \"a\" was called when \"a\" key was released");
```
bluong: Can you assert that `aCallbackCalled` is false at the start of this test?

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,62 @@ var assert = chai.assert;
...
-      assert.isTrue(aCallbackCalled, "callback for \"a\" was called when \"a\" key was pressed");
-      assert.isFalse(bCallbackCalled, "callback for \"b\" was not called when \"a\" key was pressed");
-
-      aCallbackCalled = false;
-      $target.simulate("keydown", { keyCode: bCode });
-      assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"b\" key was pressed");
-      assert.isTrue(bCallbackCalled, "callback for \"b\" was called when \"b\" key was pressed");
-
-      TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
+      $target = $(component.background().node());
```
aicioara: This sounds over specific for a beforeEach(). I would just duplicate and inline this bit in all test.

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,62 @@ var assert = chai.assert;
...
+      });
+
+      afterEach(() => {
+        keyInteraction.offKeyPress(aCode, aCallback);
+        keyInteraction.onKeyPress(bCode, bCallback);
+      });
+
+      it("only fires callback for \"a\" when \"a\" key is pressed", () => {
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        $target.simulate("keydown", { keyCode: aCode });
```
aicioara: In `testMethods.ts` There are a couple of methods used for triggering mouse events. We need a method for keyboard events as well and use that for simulate. Also, if possible, don't use jQuery, but d3.

test/interactions/keyInteractionTests.ts
```diff
@@ -101,43 +72,127 @@ describe("Interactions", () => {
...
-        assert.isTrue(aCallback1Called, "callback 2 for \"b\" was called when \"a\" key was pressed");
+        TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
+        $target.simulate("keyup", { keyCode: aCode });
+        assert.isTrue(aCallbackCalled, "callback for \"a\" was called when \"a\" key was released");
+      });
 
-        keyInteraction.offKeyPress(aCode, aCallback1);
-        aCallback1Called = false;
-        aCallback2Called = false;
+      it("doesn't fire callback if  key was released after being pressed outside", () => {
```
bluong: double white space

test/interactions/keyInteractionTests.ts
```diff
@@ -101,43 +72,127 @@ describe("Interactions", () => {
...
+        TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
         $target.simulate("keydown", { keyCode: aCode });
-        assert.isFalse(aCallback1Called, "callback 1 for \"a\" was disconnected from the interaction");
-        assert.isTrue(aCallback2Called, "callback 2 for \"a\" is still connected to the interaction");
+        $target.simulate("keyup", { keyCode: aCode });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"a\" key was released");
+      });
 
-        svg.remove();
+      it("doesn't fire callback  key was released inside after being pressed outside", () => {
```
bluong: double white space please remove

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,62 @@ var assert = chai.assert;
...
-
-        var keyInteraction = new Plottable.Interactions.Key();
-
-        var aCode = 65; // "a" key
-
-        var aCallbackCalled = false;
-        var aCallback = () => aCallbackCalled = true;
-
+    describe("onKeyPress", () => {
+      beforeEach(() => {
```
aicioara: I don't like nested `beforeEach()`, as expressed here https://github.com/palantir/plottable/issues/2504 . Now it's fine I agree. But let another 6 months pass and see the mess. (as a reference look at `barPlotTests.ts`)

My suggestion is to move these and the onKeyRelease callbacks inside the outer beforeEach, rename aCallback to aPressedCallback (etc). 

This is a personal opinion and I'd like @bluong feedback on it as well.

aicioara:
Sorry I jumped on this CR like that, but seen the nesting in the testing which concerned me.

test/interactions/keyInteractionTests.ts
bluong: I personally used to like to have a describe block for every method as then there is a nice grouping in interacting with that functionality, and so I tried to use this idea where I could.

After thinking about it again though, I think that might have been a mistake as the function could technically be used outside the block.

bluong:
@aicioara it's better if jump onto the CR.  It's actually an issue from my PoV if we restrict the review of these pull requests to one person, which is why I'm personally not a fan of the Assignee model.

bluong:
@ztsai Lots of conversation up there, but the gist of it is sorta we want more organization in the unit tests.

test/interactions/keyInteractionTests.ts
```diff
@@ -4,91 +4,62 @@ var assert = chai.assert;
...
+      });
 
-        var $target = $(component.background().node());
+      it("does not fire callback when the key is pressed outside of the component", () => {
+        TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
+        $target.simulate("keydown", { keyCode: aCode });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when not moused over the Component");
+      });
 
+      it("canceling keyPress callbacks is possible", () => {
```
jtlan: Probably "removing keyPress callbacks"; "canceling" refers to something else in the context of events.

bluong:
Still need the indentation.  After we get that, we should be fine here.

test/dispatchers/keyDispatcherTests.ts
```diff
@@ -21,7 +21,35 @@ describe("Dispatchers", () => {
       $("body").simulate("keydown", { keyCode: keyCodeToSend });
       assert.isTrue(keyDowned, "callback when a key was pressed");
 
-      ked.offKeyDown(callback); // clean up
+      ked.offKeyDown(callback);
+
+      keyDowned = false;
+      $("body").simulate("keydown", { keyCode: keyCodeToSend });
```
aicioara: Why don't we use the bellow written trigger fake keyboard event?

test/dispatchers/keyDispatcherTests.ts
```diff
@@ -21,7 +21,35 @@ describe("Dispatchers", () => {
...
+      var keyUped = false;
+      var callback = (code: number, e: KeyboardEvent) => {
+        keyUped = true;
+        assert.strictEqual(code, keyCodeToSend, "correct keycode was passed");
+        assert.isNotNull(e, "key event was passed to the callback");
+      };
+
+      ked.onKeyUp(callback);
+
+      $("body").simulate("keyup", { keyCode: keyCodeToSend });
```
aicioara: Same

test/dispatchers/keyDispatcherTests.ts
```diff
@@ -21,7 +21,35 @@ describe("Dispatchers", () => {
...
+      assert.isTrue(keyUped, "callback when a key was release");
+
+      ked.offKeyUp(callback);
+
+      keyUped = false;
+      $("body").simulate("keyup", { keyCode: keyCodeToSend });
```
aicioara: Same

test/interactions/keyInteractionTests.ts
```diff
@@ -4,140 +4,271 @@ var assert = chai.assert;
...
+    var svg: d3.Selection<void>;
+    var component: Plottable.Component;
+    var keyInteraction: Plottable.Interactions.Key;
+    var aCode: number;
+    var bCode: number;
+
+    var aCallbackCalled: boolean;
+    var bCallbackCalled: boolean;
+    var aCallback: (asd: number) => void;
+    var bCallback: (asd: number) => void;
```
aicioara: Both aCallback and bCallback take no arguments.

test/interactions/keyInteractionTests.ts
```diff
@@ -4,140 +4,271 @@ var assert = chai.assert;
...
-        $target.simulate("keydown", { keyCode: aCode });
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode);
+        TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aCode);
         assert.isFalse(aCallback1Called, "callback 1 for \"a\" was disconnected from the interaction");
         assert.isTrue(aCallback2Called, "callback 2 for \"a\" is still connected to the interaction");
 
+        keyInteraction.offKeyRelease(aCode, aCallback2);
         svg.remove();
+      });
```
aicioara: Can we add an extra test in which we test the argument passed to the callback.

```typescript
type KeyCallback = (keyCode: number) => void;
```

test/testMethods.ts
```diff
@@ -217,6 +217,13 @@ module TestMethods {
     target.node().dispatchEvent( e );
   }
 
+  export function triggerFakeKeyboardEvent(type: string, target: d3.Selection<void>, keyCode: number) {
+    var event = <KeyboardEvent> document.createEvent("Events");
+    event.initEvent(type, true, true);
+    event.keyCode = keyCode;
+    target.node().dispatchEvent( event );
```
aicioara: It's weird we have no tslint rule around this. Can you remove the whitespace around `(` and `)`

aicioara:
The testing file architecture is amazing. Super clear to read. Nice work. 

Above are just minor things that caught my eye while reading through. Please forward back to @bluong once done.

src/dispatchers/keyDispatcher.ts
```diff
@@ -1,12 +1,13 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export module Dispatchers {
-  export type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
+export type KeyCallback = (keyCode?: number, event?: KeyboardEvent) => void;
```
bluong: We should move this `KeyCallback` back to being under `Plottable.Dispatchers.KeyCallback`.

Otherwise this constitutes an API break.

src/interactions/keyInteraction.ts
```diff
@@ -1,8 +1,6 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export type KeyCallback = (keyCode: number) => void;
```
bluong: Let's put this back...

bluong:
We should fix the types.  Otherwise we are good to go.

test/testMethods.ts
```diff
@@ -214,7 +214,15 @@ module TestMethods {
     e.metaKey = false;
     e.ctrlKey = false;
     e.shiftKey = false;
-    target.node().dispatchEvent( e );
+    target.node().dispatchEvent(e);
+  }
+
+  export function triggerFakeKeyboardEvent(type: string, target: d3.Selection<void>, keyCode: number) {
```
jtlan: Isn't there a JQuery simulation method that does this?

plottable.d.ts
```diff
@@ -3578,7 +3578,7 @@ declare module Plottable {
 
 declare module Plottable {
     module Dispatchers {
-        type KeyCallback = (keyCode: number, event: KeyboardEvent) => void;
```
bluong: Why is this optional?

plottable.d.ts
```diff
@@ -3712,7 +3725,7 @@ declare module Plottable {
 
 
 declare module Plottable {
-    type KeyCallback = (keyCode: number) => void;
```
bluong: Why is this optional?

test/testMethods.ts
ztsai: yeah, but @aicioara requested a change to do it without JQuery

plottable.d.ts
ztsai: it's optional cuz we don't use event in keyInteraction (sorry forgot to revert that in this commit, it might make more sense after the revert)

test/testMethods.ts
aicioara: @jtlan why use jQuery?

plottable.d.ts
bluong: Yes, but the `KeyInteraction` does not utilize this `KeyCallback`.  It still is not clear to me why we need optional parameters here.

plottable.d.ts
bluong: I don't see why this is optional.

PR #2501, opened 2015-07-23
[grunt-ts][grunt-tslint]Upgrade ts tools

### LINKED ISSUE #2493: Upgrade grunt-ts
Upgrade grunt-ts to the newest version. We are at version 1.12.1, while latest is 4.0.2. 

Blocking until we figure out what do we do about TS 1.5 as the latest version of grunt-ts uses ts 1.5

bluong:
Now that we have Typescript 1.5, we should upgrade our tools so that they utilize Typescript 1.5.

After this pull request merges, please run `npm install`.

Closes #2493


package.json
```diff
@@ -15,9 +15,8 @@
     "grunt-cli": "0.1.6",
     "grunt-contrib-watch": "~0.5.3",
     "load-grunt-tasks": "~0.2.0",
-    "grunt-ts": "~1.12.1",
-    "grunt-tslint": "2.2.0-beta",
-    "tslint": "2.2.0-beta",
```
aicioara: I think this will break. `grunt-tslint` does not specify `tslint` as a dependency, but rather as a dev dependency, so I doubt it will download the dependency. Can you please double check
```bash
rm -rf node_modules && npm install ^^ grunt tslint
````

aicioara:
Now that I think of it I think it will work, but just double check

bluong:
Tested! It works!

I was curious because I remember I never needed `tslint` before as a dependency so I think this was something that was added very recently by mistake.

aicioara:
Yeah, before it was only a peer dependency, now it is a dev dependency.

Issue #2385, opened 2015-06-24
Axis labels flicker between different patterns when panning


jtlan:
Repro: http://jsfiddle.net/zvuxboa5/ . Try panning around.

**Actual**: `Axis` labels flicker between even and odd values when panning around
**Expected**: `Axis` does not flicker

aicioara:
That is the expected behavior. Let's take the following scenario:

We have an axis showing 700, 800, 900. Now we pan more, but the 1000 will not fit in the given space, so we would want to show 800, 1000 (later 1200).

However, I agree that the flickering is distracting.

PR #2502, opened 2015-07-24
add more ticks to the linear part of modified log scale

### LINKED ISSUE #2387: Modified log axis ticks are insufficient for the default domain
More ticks, please. :)

ztsai:
closes #2387

http://jsfiddle.net/ztsai/1zfz0tg2/1/

### DIFF
--- a/test/scales/modifiedLogScaleTests.ts
+++ b/test/scales/modifiedLogScaleTests.ts
@@ -180,14 +180,11 @@ describe("Scales", () => {
       assert.operator(betweenPivots.length, ">", 0, "should be ticks between -base and base");
     });
 
-    it("ticks() is always non-empty", () => {
-      var desiredExtents: number[] = [];
-      scale.addIncludedValuesProvider((scale: Plottable.Scale<number, number>) => desiredExtents);
-      [[2, 9], [0, 1], [1, 2], [0.001, 0.01], [-0.1, 0.1], [-3, -2]].forEach((extent) => {
-        desiredExtents = extent;
-        scale.autoDomain();
+    it("ticks() always has more than 2 ticks", () => {
+      [null, [2, 9], [0, 1], [1, 2], [0.001, 0.01], [-0.1, 0.1], [-3, -2]].forEach((domain) => {
+        scale.domain(domain);
         var ticks = scale.ticks();
-        assert.operator(ticks.length, ">", 0);
+        assert.operator(ticks.length, ">", 2);
       });
     });
   });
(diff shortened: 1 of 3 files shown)


test/scales/modifiedLogScaleTests.ts
```diff
@@ -190,5 +190,11 @@ describe("Scales", () => {
         assert.operator(ticks.length, ">", 0);
       });
     });
+
+    it("default domain has more than 2 ticks", () => {
+      scale.domain([200, -100]);
```
jtlan: This isn't the default domain, so we should remove this line.

crmorford:
Probably unrelated, but do we know why do many of these tick labels aren't being drawn when they have plenty of room? http://jsfiddle.net/rtdxr3xv/5/

crmorford:
The [10, -10] case has 3 ticks, but there's only room for one of them http://jsfiddle.net/rtdxr3xv/7/. 
Maybe it's out of scope for this, but it'd be awesome if it looked more like the [9, -9] case http://jsfiddle.net/rtdxr3xv/8/

crmorford:
it looks really good on teeny axes though! http://jsfiddle.net/rtdxr3xv/9/

bluong:
@crmorford 
Likely the above is unrelated, but it should definitely be addressed eventually.  The scope of this pull request is to see what ticks pop out using the default domain.  I think this means that you should be using only the default domain to test out the axes, in which case the [-10, 10] case is invalid as well as [9, -9].

That said, I do 100% agree that we should probably rethink our tick generation logic.  If you think the above makes sense, I think we should make an issue for the problems you're seeing and then adjust your process accordingly.

jtlan:
@crmorford wrote:
>Probably unrelated, but do we know why do many of these tick labels aren't being drawn when they have plenty of room? http://jsfiddle.net/rtdxr3xv/5/

>Edit: something's weird, because on the reversed domain, a lot of ticks are drawn http://jsfiddle.net/rtdxr3xv/6/

Probably due to the same cause as #2385.

bluong:
I take my comment back.

My comment is relevant if the scope of this pull request is to specifically solve the issue at #2387 .

However, based on the title of the pull request, the scope of the pull request encompasses more than just the relevant issue, also including cases that @crmorford has presented.

@ztsai if you are planning to just solve the relevant issue than adjust your pull request description title accordingly.  However, if you are expanding the scope to what is stated in the pull request title, proceed as normal.

PR #2508, opened 2015-07-25
[Grunt] No more noisy output from the concat pipeline

### LINKED ISSUE #2445: Clean GruntFile

### LINKED ISSUE #2507: Grunt is too verbose
screen shot 2015-07-24 at 4 55 10 pm

aicioara:
Before
screen shot 2015-07-24 at 4 55 10 pm

After
screen shot 2015-07-24 at 4 59 15 pm

Inspiration
https://github.com/gruntjs/grunt/issues/895

This is simply removing the `Running "foo:bar" task`, for watch only. The header is still used for all our CI stuff. 

Closes #2507

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -362,11 +362,17 @@ module.exports = function(grunt) {
   grunt.registerTask("dist-compile", ["test", "uglify", "compress"]);
 
   grunt.registerTask("commitjs", ["dist-compile", "gitcommit:built"]);
-  grunt.registerTask("default", ["connect", "dev-compile", "watch"]);
+  grunt.registerTask("default", ["connect", "dev-compile", "watch-silent"]);
 
   grunt.registerTask("test", ["dev-compile", "test-local"]);
   grunt.registerTask("test-local", ["blanket_mocha", "ts:verifyDefinitionFiles", "lint"]);
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
+
+  grunt.registerTask("watch-silent", function() {
+    // HACKHACK #2508
+    grunt.log.header = function() {};
+    grunt.task.run(["watch"]);
+  });
 
   grunt.registerTask("lint", ["parallelize:tslint", "jscs", "eslint"]);
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -369,7 +369,7 @@ module.exports = function(grunt) {
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
 
   grunt.registerTask("watch-silent", function() {
-    // HACKHACK #2508
+    // Surpresses the "Running 'foo' task" messages
     grunt.log.header = function() {};
     grunt.task.run(["watch"]);
   });
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -369,7 +369,7 @@ module.exports = function(grunt) {
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
 
   grunt.registerTask("watch-silent", function() {
-    // Surpresses the "Running 'foo' task" messages
+    // Surpresses the "Running 'foo' task" messages. More info at #2507.
     grunt.log.header = function() {};
     grunt.task.run(["watch"]);
   });
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -369,7 +369,7 @@ module.exports = function(grunt) {
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
 
   grunt.registerTask("watch-silent", function() {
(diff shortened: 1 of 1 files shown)


bluong:
I see a File "plottable.js" created.  Is there a way to supress those as well?

Gruntfile.js
```diff
@@ -362,12 +362,18 @@ module.exports = function(grunt) {
...
   grunt.registerTask("commitjs", ["dist-compile", "gitcommit:built"]);
-  grunt.registerTask("default", ["connect", "dev-compile", "watch"]);
+  grunt.registerTask("default", ["connect", "dev-compile", "watch-silent"]);
 
   grunt.registerTask("test", ["dev-compile", "test-local"]);
   grunt.registerTask("test-local", ["blanket_mocha", "ts:verifyDefinitionFiles", "lint"]);
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
 
+  grunt.registerTask("watch-silent", function() {
+    // HACKHACK #2508
```
bluong: The HACKHACK should have a description as well as  an issue to point to.  Pointing at the source pull request is not that helpful.

Gruntfile.js
```diff
@@ -362,12 +362,18 @@ module.exports = function(grunt) {
...
+    grunt.log.header = function() {};
+    grunt.task.run(["watch"]);
```
bluong: Can you explain what's going on here?

aicioara:
Yeah, it's in the description, with link to what I want to do and everything
> This is simply removing the Running "foo:bar" task, for watch only. The header is still used for all our CI stuff.

Basically, wrapper around the watch which also disables the headers

bluong:
Please answer through the comments.

Also can you link the task website from the pull request?

Gruntfile.js
aicioara: Yeah, it's in the description, with link to what I want to do and everything

This is simply removing the Running "foo:bar" task, for watch only. The header is still used for all our CI stuff.
Basically, wrapper around the watch which also disables the headers

aicioara:
> I see a File "plottable.js" created. Is there a way to supress those as well?

bluong:
Do you know where the success message is coming from?

My personal problem is at the beginning I see
"File "plottable.js" created." twice with no context on why this file is being created twice.

I think we should change these strings so that the build is more transparent on what is happening.  Note that before, it was very transparent, since it seems that plottable.js is first created with the .ts files and then it is concatentated with svgTypewriter.

aicioara:
@bluong All the build messages are still shown at all times, except the watch.

Gruntfile.js
```diff
@@ -362,12 +362,18 @@ module.exports = function(grunt) {
...
-  grunt.registerTask("default", ["connect", "dev-compile", "watch"]);
+  grunt.registerTask("default", ["connect", "dev-compile", "watch-silent"]);
 
   grunt.registerTask("test", ["dev-compile", "test-local"]);
   grunt.registerTask("test-local", ["blanket_mocha", "ts:verifyDefinitionFiles", "lint"]);
   grunt.registerTask("test-sauce", ["connect", "saucelabs-mocha"]);
 
+  grunt.registerTask("watch-silent", function() {
+    // Surpresses the "Running 'foo' task" messages. More info at #2507.
+    grunt.log.header = function() {};
```
bluong: Oh so this is not a HACKHACK?  If it's not why is an issue being addressed here?

Also what does `grunt.log.header` do?

aicioara:
Also, for completeness, this is what generates the "file xx created"
https://github.com/gruntjs/grunt-contrib-concat/blob/master/tasks/concat.js#L117

Gruntfile.js
aicioara: Generates strings like
```
Running "watch-silent" task
```

aicioara: it is not really a hackhack. The issue is for providing more information of what does the `grunt.log.header` do

bluong: Hmm, pointing at the issue I don't think reveals any more information.  I think if you want to put a comment there, I would just put the first part of the comment.  I think the "More info..." is unnecessary, especially if this will close #2507

bluong:
Ultimately I think this'll be a multistep process to clean up our build system.  Let's clean up  the comment that I pointed out and I think we should be ok.  Can you also reach out to @jtlan just to make sure that this pull request is desirable?  It mostly looks good to me.

aicioara:
Sounds good

jtlan:
Seems like an improvement -- even in the case of errors, I don't usually care too much which task failed so long as I know what needs to be fixed.

jtlan:
[NOQE] because no code changes.

aicioara:
Also, as an aside, this will probably not exist when we migrate to Gulp

PR #2511, opened 2015-07-28
fix pie chart selection bug

### LINKED ISSUE #2491: Mouse cursor position is not correct for pie chart selection.
The mouse cursor position does not correspond to the slice of the pie chart that is selected on click. The mouse cursor position is several pixels off and on click, selects either a different slice nearby or no slices at all.

Reproduce:
1. Create a pie chart with 5 slices - where one slice has a value of 500 and the other slices have values around 5.
2. Try to select one of the small slices.

See Slate ticket 2428

ztsai:
closes #2491

before: http://jsfiddle.net/pghLh8da/4/
after: http://jsfiddle.net/pghLh8da/3/

### DIFF
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -213,7 +213,7 @@ export module Plots {
       var center = { x: this.width() / 2, y: this.height() / 2 };
       var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
       var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
-      var angle = Math.acos(-adjustedQueryPoint.y / (1 + radius));
+      var angle = Math.acos(-adjustedQueryPoint.y / radius);
       if (adjustedQueryPoint.x < 0) {
         angle = Math.PI * 2 - angle;
       }
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -212,29 +212,8 @@ export module Plots {
     public entitiesAt(queryPoint: Point) {
       var center = { x: this.width() / 2, y: this.height() / 2 };
       var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
-      var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
-      var angle = Math.acos(-adjustedQueryPoint.y / radius);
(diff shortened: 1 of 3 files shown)


src/plots/piePlot.ts
```diff
@@ -213,7 +213,7 @@ export module Plots {
       var center = { x: this.width() / 2, y: this.height() / 2 };
       var adjustedQueryPoint = { x: queryPoint.x - center.x, y: queryPoint.y - center.y };
       var radius = Math.sqrt(Math.pow(adjustedQueryPoint.x, 2) + Math.pow(adjustedQueryPoint.y, 2));
-      var angle = Math.acos(-adjustedQueryPoint.y / (1 + radius));
```
jtlan: ... I thought this was suspicious the first time I saw it; @rcchen, what was the reasoning behind this?

src/plots/piePlot.ts
```diff
@@ -213,7 +213,7 @@ export module Plots {
...
+      var angle = Math.acos(-adjustedQueryPoint.y / radius);
       if (adjustedQueryPoint.x < 0) {
         angle = Math.PI * 2 - angle;
       }
```
jtlan: Instead of going through the trouble of using `acos()` followed by an if-check, could we use [`atan2()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2)?

src/plots/piePlot.ts
rcchen: https://github.com/palantir/plottable/pull/2340#discussion-diff-32654964

...I added instead of multiplied the two magnitudes...

rcchen: Also https://github.com/palantir/plottable/blob/pieSelect/src/plots/piePlot.ts#L302

ztsai:
talked about using ```atan2()``` offline; the range for the function is (-pi, pi], and it would still require a check to map the result to [0, 2*pi)

bluong:
? the extent of both ranges is 2*pi.  Why can't we do a simple mapping from (-pi,pi] -> [0, 2*pi)

jtlan:
>? the extent of both ranges is 2 * pi. Why can't we do a simple mapping from (-pi,pi] -> [0, 2*pi)

The origin and direction of the angles is different, I think.

bluong:
In which case, we can still do a mapping no?

crmorford:
Do we know why this was only happening on small slices, originally, and not larger ones?

ztsai:
My guess is that the radius is only off by one pixel, so the wrong angel wouldn't off by a lot

PR #2513, opened 2015-07-28
Add a license field to package.json.

PeterFaiman:
I just noticed this, and decided to fix it because why not? Unless there is a reason this is not set already.

Currently npm install warns with the following message:

`npm WARN package.json plottable.js@1.4.0 No license field.`

Plottable is MIT licensed, according to the LICENSE file, and I don't like this warning message.


aicioara:
Agree

PR #2520, opened 2015-07-29
Autorange smooth

### LINKED ISSUE #2439: Autorange only uses points in view to compute bounds
For example, for:

    var line = new Plottable.Plots.Line()
                    .x(xData, xScale)
                    .y(yData, yScale)
                    .autorangeMode("y")

when the view is zoomed in, only the y-values that are in the x-range are considered. This causes lines to be clipped, or, when zoomed fully, to default back to the (-1, 1) range.

A possible solution might be to also use the point where graph lines cross the left/right boundary for the computation of the range.

aicioara:
Closes #2439

Feature request from Contour. Worked closely with the customer to get the correct behavior. 

Affects: 
- Line Plot
- Area Plot
- Stacked Area Plot

Pan and Zoom to get a feeling of the new behavior.
Without feature enabled: http://jsfiddle.net/2cghgtfk/3/
With feature enabled: http://jsfiddle.net/2cghgtfk/4/

New API points

`QuantitativeScale`
```typescript
/**
 * Gets whether or not the scale snaps its domain to nice values.
 */
snappingDomainEnabled(): boolean;
/**
 * Sets whether or not the scale snaps its domain to nice values.
 */
snappingDomainEnabled(snappingDomainEnabled: boolean): QuantitativeScale<D>;
```

`Plots.Line`
```typescript
/**
 * Gets whether or not the autoranging is done smoothly.
 */
autorangeSmooth(): boolean;
/**
 * Sets whether or not the autorange is done smoothly.
 *
 * Smooth autoranging is done by making sure lines always exit on the left / right side of the plot
 * and deactivating the nice domain feature on the scales
 */
autorangeSmooth(autorangeSmooth: boolean): Plots.Line<X>;
```

QE Notes: 
- autoRangeSmooth is not dependent on autoRangeMode being called on any plots in the group. 
- autoRangeSmooth does not make sense without autorangeMode "x" or "y", because there is no autoranging happening 
- autoRangeSmooth works in both directions now
- autoRangeSmooth does not care if the line plot does not look like a line plot anymore (stars also get smoothly autoranged)
- We don't need new endpoints to indicate the direction because that is inferred from the autorangeMode


jtlan:
Can we get it to work when the `padProportion()` of the scale is not 0?

aicioara:
Hmm

aicioara:
It is possible if we add a new API point for the quantitativeScale, where we deactivate the nice domaining, and call it from the autorangeSmooth setter.

Do you agree with the proposed solution?

jtlan:
>It is possible if we add a new API point for the quantitativeScale, where we deactivate the nice domaining, and call it from the autorangeSmooth setter.

Does the user have to manually re-enable this if they disable `autorangeSmooth`?

aicioara:
Nope. autorangeSmooth will make all the required calls in the QScale API

aicioara:
@jtlan Done!

src/scales/quantitativeScale.ts
```diff
@@ -152,7 +153,28 @@ export class QuantitativeScale<D> extends Scale<D, number> {
...
+    if (this._niceDomainEnabled) {
+      return this._niceDomain([newMin, newMax]);
+    }
+    return ([newMin, newMax]);
+  }
+
+  /**
+   * Gets whether or not the scale will have a nice domain.
+   */
+  public niceDomain(): boolean;
```
jtlan: This name might be confusing, since it could suggest that it returns something of type `D[]`, similar to `domain()`. Is there a different name we could use?

src/scales/quantitativeScale.ts
```diff
@@ -152,7 +153,28 @@ export class QuantitativeScale<D> extends Scale<D, number> {
...
+  /**
+   * Sets whether or not the scale will have a nice domain.
+   */
+  public niceDomain(niceDomainEnabled: boolean): QuantitativeScale<D>;
```
jtlan: Generally our convention has been to have param name in a getter-setter carry the same name as the method itself, so `niceDomain` in this case.

aicioara: I think this will be solved by renaming the method. The convention is also to assign the parameter name to the instance variable with the same name, but in our case _niceDomain() is taken

src/scales/quantitativeScale.ts
aicioara: How about 
- nicifyDomain() (is this even a word?)
- prettifyDomain() (will it be confusing to use a synonym?)
- enableNiceDomain() (convention?) 

This was all I could think of. Do you like any of the suggestions?

aicioara: How about 
- nicifyDomain() (is this even a word?)
- prettifyDomain() (will it be confusing to use a synonym?)
- enableNiceDomain() (convention?) 

This was all I could think of. Do you like any of the suggestions?

src/plots/linePlot.ts
```diff
@@ -21,10 +23,121 @@ export module Plots {
       this.attr("stroke-width", "2px");
     }
 
+    public y(): Plots.AccessorScaleBinding<number, number>;
+    public y(y: number | Accessor<number>): Plots.Line<X>;
+    public y(y: number | Accessor<number>, yScale: Scale<number, number>): Plots.Line<X>;
+    public y(y?: number | Accessor<number> | number | Accessor<number>, yScale?: Scale<number, number>): any {
+      if (yScale instanceof QuantitativeScale) {
+        (<QuantitativeScale<number>>yScale).niceDomain(!this._autorangeSmooth);
```
jtlan: One risk with this auto-set-nice feature is that if there are multiple `Plots.Line` depending on the same y-scale, then `autorangeSmooth()` has to be set on all of them individually, otherwise the order of assembly matters.

src/plots/linePlot.ts
```diff
@@ -21,10 +23,121 @@ export module Plots {
...
+      }
+      return super.y(y, yScale);
+    }
+
+    public autorangeSmooth(): boolean;
+    public autorangeSmooth(autorangeSmooth: boolean): Plots.Line<X>;
```
jtlan: TSDoc?

src/scales/quantitativeScale.ts
jtlan: The issue with "prettify" is that it's a verb -- "`prettifiesDomain()`" would make more sense, as other other getter-setters have adjectives for names.

"smoothsDomain"?

aicioara: smoothsDomain is the opposite of niceDomain

src/plots/linePlot.ts
aicioara: Can you elaborate?

aicioara:
Saving one of the comments from above

@jtlan 
>One risk with this auto-set-nice feature is that if there are multiple Plots.Line depending on the same y-scale, then autorangeSmooth() has to be set on all of them individually, otherwise the order of assembly matters.

Can you elaborate?

jtlan:
```Typescript
linePlot1.y(yAccessor, yScale);
linePlot1.autorangeSmooth(true);
linePlot2.y(yAccessor, yScale);
```
Since `linePlot2` was bound to `yScale` after `autorangeSmooth(true)` was set, it undoes the `niceDomain()` setting.

aicioara:
Let me play with this

aicioara:
Fiddle here. http://jsfiddle.net/30zd21g3/ It does indeed

aicioara:
Well, it makes sense to me that all plots should agree on the smooth autoranging. I don't think there is a clean way to send a warning when this is not happening

aicioara:
Looking for previous instances of this scenario, `.autorangeMode()` is having an all or nothing approach. If one plot does not agree on the autorange mode, it falls back to the default.

aicioara:
Can't find an elegant solution for this. 

>the issue is that both the plot and the scale have to agree in order for magic to happen
but user can manually change the scale
so it will break anyway 
because the plot and scale are not connected in any way

crmorford:
Why is this only for line?

aicioara:
That was the feature request. It also doesn't make as much sense for other plots I believe. We can also release this as a candidate, get some feedback and then extend to other plots in the future if we want.

However, not sure if this feature makes sense in the context of a bar plot, segment plot or rectangle plot.

crmorford:
Ah, I just thought from the description that it was line (but not area). I tried it on area and it worked though.

aicioara:
Ah yea, my bad. Line, Area, StackedArea. They should all work.

crmorford:
there's been no improvement for pan-zoom in the y direction. http://jsfiddle.net/2cghgtfk/7/ is choppy, while the closest example in the x-direction is really smooth http://jsfiddle.net/2cghgtfk/6/

aicioara:
Whoaa... I did not even know that was possible. 

I cannot support that because I don't know the orientation of the plot. As an analogy, if I just read the code without seeing the data, I could not have told that what was going to output was a horizontal line plot. 

This is just smoothing left / right. Maybe I should document the assumptions better?

aicioara:
Let me play with this for 5 mins though

aicioara:
Unless there is a way to infer what orientation the plot is having, the feature you request is not possible. The only way to infer what sort of orientation is wished is by looking at the `autorangeMode` and check if it is "x" or "y". However, that opens way to so many edge cases. My suggestion is roll it out like this, open a new issue for supporting horizontal line plots and go from there. This PR will get way too complicated otherwise.

crmorford:
Discussion w/ Christina offline
* autoRangeSmooth shouldn't be dependent on autoRangeMode being called on all(/any) plots in the group.

aicioara:
I guess something like this... 
http://jsfiddle.net/2cghgtfk/12/

aicioara:
Doable, but I want to separate PRs

aicioara:
@bluong suggested `snapsDomain()` -> `snappedDomain()` / `snappingDomain()`, such that we don't use a verb. I'll make this change as well.

aicioara:
Doing it properly now

src/plots/linePlot.ts
```diff
@@ -61,6 +140,136 @@ export module Plots {
...
+      let edgeIntersectionPoints = this._getEdgeIntersectionPoints();
+      let includedValues: number[];
+      if (this.autorangeMode() === "y") {
+        includedValues = edgeIntersectionPoints.left.concat(edgeIntersectionPoints.right).map((point) => point.y);
+      } else { // === "x"
+        includedValues = edgeIntersectionPoints.top.concat(edgeIntersectionPoints.bottom).map((point) => point.x);
+      }
+
+      let maxIncludedValue = Math.max.apply(this, includedValues);
+      let minIncludedValue = Math.min.apply(this, includedValues);
```
jtlan: Shouldn't we use `Utils.Math.min` and `Utils.Math.max` here?

src/plots/linePlot.ts
```diff
@@ -61,6 +140,136 @@ export module Plots {
...
+        bottom: [],
+        left: [],
+        right: []
+      };
+      let leftX = xScale.scale(xScale.domain()[0]);
+      let rightX = xScale.scale(xScale.domain()[1]);
+      let bottomY = yScale.scale(yScale.domain()[0]);
+      let topY = yScale.scale(yScale.domain()[1]);
+
+      this.datasets().forEach((dataset) => {
```
jtlan: This looks like an `O(total number of data points)` algorithm, because we don't have an assurance that the points are sorted in either X or Y. I'm beginning to think that `Line` should have been built like `Bar`, with an explicit orientation.

What's the most common use-case here? Can we add the smoothing in other directions later? Right now it looks like  the most common use case (left-to-right sorted) has its runtime reduced from `log(n)` to `(n)` because of the need to make it universal. We have precedent for considering this -- `entityNearest()` uses X-then-Y logic, for example.

aicioara: I think you meant the other way round. Reduce from `O(n)` to `O(log(n))` by assuming the points are ordered and doing 2 binary searches.

1. From previous feedback from QE and PM, it seemed like such assumptions might not be reasonable. It is not clear from the method name that the points have to be ordered, and it is very hard to explode gracefully if the points are unordered, so the user might not know what is going on.

2. In-memory operations are not as costly as DOM manipulation or rendering. (I mean by 1-2 orders of magnitude). I'd prefer not to optimize early, but I will profile this and see if you're right.

aicioara: Following up item number 2:

There does not seem to be a difference
- Before: http://jsfiddle.net/qb7bycjh/1/
- After: http://jsfiddle.net/qb7bycjh/2

aicioara: Getting more data-driven:
- Before: 240 ms
- After: 220 ms

(Weird, it's actually faster now)

aicioara: More data:

The total time for this method to run is `8.8ms` per flush in a plot of 5500 points. That is 3.6% of the time.

aicioara: Let me update my example to remove the scatter plot. For some reason that takes longer to render

aicioara: - Before: http://jsfiddle.net/qb7bycjh/5/
- After: http://jsfiddle.net/qb7bycjh/4/

aicioara:
Bumping
> @bluong suggested snapsDomain() -> snappedDomain() / snappingDomain(), such that we don't use a verb. 

@jtlan which one you prefer.

jtlan:
`snapsDomain()` seems fine to me.

src/plots/linePlot.ts
```diff
@@ -22,6 +31,76 @@ export module Plots {
...
+      if (autorangeSmooth == null) {
+        return this._autorangeSmooth;
+      }
+      this._autorangeSmooth = autorangeSmooth;
+
+      if (this.autorangeMode() === "x" && this.x() && this.x().scale && this.x().scale instanceof QuantitativeScale) {
+        (<QuantitativeScale<X>>this.x().scale).snapsDomain(!autorangeSmooth);
+      }
+
+      if (this.autorangeMode() === "y" && this.y() && this.y().scale && this.y().scale instanceof QuantitativeScale) {
```
jtlan: This line and other instances of similar if-checks seem like they could be factored out into a helper function?

src/plots/linePlot.ts
```diff
@@ -61,6 +140,136 @@ export module Plots {
...
+        let prevX: number, prevY: number, currX: number, currY: number;
+        for (let i = 1; i < data.length; i++) {
+          prevX = currX || xScale.scale(this.x().accessor(data[i - 1], i - 1, dataset));
+          prevY = currY || yScale.scale(this.y().accessor(data[i - 1], i - 1, dataset));
+
+          currX = xScale.scale(this.x().accessor(data[i], i, dataset));
+          currY = yScale.scale(this.y().accessor(data[i], i, dataset));
+
+          // If values crossed left edge
+          if ((prevX < leftX) === (leftX <= currX)) {
```
jtlan: If `prevX === leftX === currX`, then this will return `false`.

test/scales/linearScaleTests.ts
```diff
@@ -118,6 +118,29 @@ describe("Scales", () => {
...
+
+      it("domain snapping works", () => {
+        let scale = new Plottable.Scales.Linear();
+        scale.addIncludedValuesProvider(function() {
+          return [1.123123123, 3.123123123];
+        });
+
+        assert.deepEqual(scale.domain(), [1, 3.2], "domain snapping works");
+        scale.snapsDomain(false);
+        assert.deepEqual(scale.domain(), [1.073123123, 3.173123123], "domain snapping can be deactivated");
```
jtlan: Rather than using magic numbers here, can you set `paddingProportion(0)`, then assert that the domain matches the included values?

aicioara: `padProportion(0)` was actually misleading in this case as it did not catch one bug. The scale was snapping when it shouldn't, and setting padProportion to 0 does not catch that. Should I try to compute the value instead? so I don't use magic numbers

jtlan: Maybe, yeah.

src/plots/linePlot.ts
aicioara: if `prevX === leftX` then the point is already included, hence inclusion is not needed any more. This means that `<=` is redundant and should be `<`.
Furthermore, if `prevX === leftX === currX`, then the intersection is a line. How do we treat this case? The method should return a single point.

src/plots/linePlot.ts
aicioara: The problem is that they are slightly different. Not sure if it is worth extracting it

src/plots/linePlot.ts
```diff
@@ -22,6 +31,78 @@ export module Plots {
       this.attr("stroke-width", "2px");
     }
 
+    public x(): Plots.AccessorScaleBinding<X, number>;
+    public x(x: number | Accessor<number>): Line<X>;
+    public x(x: X | Accessor<X>, xScale: Scale<X, number>): Line<X>;
+    public x(x?: number | Accessor<number> | X | Accessor<X>, xScale?: Scale<X, number>): any {
```
jtlan: I find that we often need to do additional work in property setters, leading to overriding in the subclass with a duplication of superclass logic. Instead of duplicating the logic each time we have subclass-specific code in the setter, does it make sense to make a `protected` `_setX()` method that could be overridden independently?

This refactor shouldn't be done in this PR, of course.

src/plots/linePlot.ts
```diff
@@ -22,6 +31,78 @@ export module Plots {
...
+      if (x == null) {
+        return super.x();
+      } else {
+        if (xScale == null) {
```
jtlan: Why this `if`-check? Wouldn't just calling `super.x(x, xScale)` work?

src/plots/linePlot.ts
aicioara: That makes sense.

src/plots/linePlot.ts
aicioara: Casting. Just doing what you suggested yields 
```
Using the custom compiler : ./node_modules/typescript/bin/tsc
/Users/acioara/dev/plottable/src/plots/linePlot.ts(41,17): error TS2345: Argument of type 'number | Accessor<number> | X | Accessor<X>' is not assignable to parameter of type 'X | Accessor<X>'.
  Type 'number' is not assignable to type 'X | Accessor<X>'.
    Type 'number' is not assignable to type 'Accessor<X>'.
```

test/plots/linePlotTests.ts
```diff
@@ -375,4 +375,312 @@ describe("Plots", () => {
...
+        let expectedTop = base + y2 * x1 / x2;
+
+        x1 = (xScale.domain()[0] - data[0].x) + (xScale.domain()[1] - xScale.domain()[0]) * (1 + yScale.padProportion() / 2);
+        let expectedBottom = base + y2 * x1 / x2;
+
+        assert.closeTo(yScale.domain()[0], expectedBottom, 0.001, "smooth autoranging forces the domain to include the line (left)");
+        assert.closeTo(yScale.domain()[1], expectedTop, 0.001, "smooth autoranging forces the domain to include the line (right)");
+      });
+
+      it("smooth autoranging works (called before after rendering)", () => {
```
jtlan: Is it before or after?

test/plots/linePlotTests.ts
```diff
@@ -375,4 +375,312 @@ describe("Plots", () => {
...
+        let expectedTop = base + y2 * x1 / x2;
+
+        x1 = (xScale.domain()[0] - data[0].x) + (xScale.domain()[1] - xScale.domain()[0]) * (1 + yScale.padProportion() / 2);
+        let expectedBottom = base + y2 * x1 / x2;
+
+        assert.closeTo(yScale.domain()[0], expectedBottom, 0.001, "smooth autoranging forces the domain to include the line (left)");
+        assert.closeTo(yScale.domain()[1], expectedTop, 0.001, "smooth autoranging forces the domain to include the line (right)");
+      });
+
+      it("smooth autoranging works (called before before autorangeMode)", () => {
```
jtlan: Two "before"s

src/plots/linePlot.ts
jtlan: The point is already included by some other `includedValuesProvider`, I take it?

Also, since the line returns `false`, we avoid the division by zero, so this edge case might be okay. Maybe we should just leave it like this?

src/plots/linePlot.ts
jtlan: How come it works in `y()` a few lines later?

aicioara: Different signature because line plot is more specific (there is no `Y` type). I can make it the same if you want for consistency

test/plots/linePlotTests.ts
aicioara: lol, that's weird.

src/plots/linePlot.ts
jtlan: No, that's fine, makes sense now.

crmorford:
Can I get a summary of what happened on this PR?
I asked for this: 
* autoRangeSmooth shouldn't be dependent on autoRangeMode being called on all(/any) plots in the group.
* autoRangeSmooth should work in both directions. We should make endpoints that indicate direction before sending out something that only works in x-direction

Did both of these things happen?

aicioara:
- autoRangeSmooth is not dependent on autoRangeMode being called on any plots in the group. 
- autoRangeSmooth does not make sense without autorangeMode "x" or "y", because there is no autoranging happening 
- autoRangeSmooth works in both directions now
- autoRangeSmooth does not care if the line plot does not look like a line plot anymore (stars also get smoothly autoranged)
- We don't need new endpoints to indicate the direction because that is inferred from the autorangeMode

Hope it makes sense. I add these to the description. Ask me if unclear

crmorford:
Hey, you added comments to the top, but it doesn't seem like you updated the api description up there? (I think "snapsDomain" is inaccurate, I remember seeing "snappingDomainEnabled" in the code.)

It's nice to keep those updated so that they mirror the status of the pull request. It also helps me get the PR through much much quicker when I can trust the summary instead of reading through all the code.

aicioara:
I don't know why I keep doing that. I added it to my checklist for the future. Apologies. It should be corrected now

crmorford:
thanks :) starting QE now

crmorford:
**Not a blocker**
want to note that you get some jerky/flickery behavior when a scale is both
* included in a pan zoom interaction
* has autorange turned on

crmorford:
@aicioara i'm having trouble figuring out what snappingDomainEnabled is doing. What's a scenario where turning it on makes a difference?

aicioara:
Sure. By default, a scale is trying to make it's domain nice, by extending the domain to nice values (say 13.4 instead of 13.4123124134712). Let me see if I can provide you a fiddle

aicioara:
Scenario for the snappingDomainEnabled http://jsfiddle.net/a8bfjvfk/1/

Please note that it is enabled by default.

crmorford:
**autorange**
y-direction linear: http://jsfiddle.net/2cghgtfk/14/
y-direction modLog: http://jsfiddle.net/2cghgtfk/15/
x-direction linear: http://jsfiddle.net/2cghgtfk/4/
x-direction modLog: http://jsfiddle.net/2cghgtfk/16/
x-direction time: http://jsfiddle.net/2cghgtfk/17/
y-direction and x-direction:  considering that this is abuse of the API, it looks good http://jsfiddle.net/2cghgtfk/18/
autorange(x, y), panzoom x http://jsfiddle.net/2cghgtfk/19/
autorange(x, y), panzoom y http://jsfiddle.net/2cghgtfk/20/
autorange(x, y), panzoom x,y  http://jsfiddle.net/2cghgtfk/18/
non-sorted data
x http://jsfiddle.net/2cghgtfk/21/ pretty amazing
y http://jsfiddle.net/2cghgtfk/22/
x, y http://jsfiddle.net/2cghgtfk/23/

summary: 
* setting panzoom(null, yScale) --> set .autorangeMode("x")
*  setting panzoom(xScale, null) --> set .autorangeMode("y")
* setting autorangeMode on "x" and "y" with panzoom is bad
* setting panzoom(xscale, yscale) and an autorangeMode isn't super good
but these don't affect this PR, I'm mostly just documenting in case I come back to these tests and wonder why they feel funny

**snappingDomainEnabled**
time scales http://jsfiddle.net/a8bfjvfk/3/
very small ranges http://jsfiddle.net/a8bfjvfk/2/ I'm concerned about it choosing 0.000009999999999999999 as a nice value
during panzoom -- anecdotally, looks fine. no concerns.

crmorford:
this is the only case i'm concerned about: 
http://jsfiddle.net/a8bfjvfk/2/ chooses 0.000009999999999999999 as a nice value
can you explain why it returns that instead of 0.00001?

aicioara:
That is the previous behavior and it is what d3 is returning, can't return that. Let me see if I can get you an example.

aicioara:
Here http://jsfiddle.net/7bucu5nb/

PR #2526, opened 2015-07-31
[GlobalInitialization] - assert variable defined in only one place

bluong:
In playing around with the variables, I found that we actually redefine the `assert` variable for every test file we make... which seems incredibly silly.  We should instead put this in globalInitialization.ts since that is a global initialization file.

From a practical standpoint, removing this redundancy reduces the overhead in creating a new test file.


aicioara:
looks good

PR #2529, opened 2015-07-31
[Reference.ts] - Remove reference lines

bluong:
Removing these lines seems to have no effect on the compiler.  My personal guess is that `reference.ts` has already defined the files that need to be compiled and that restating it is redundant?

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1,4 +1,3 @@
-
 declare module Plottable {
     module Utils {
         module Math {
@@ -58,7 +57,6 @@ declare module Plottable {
     }
 }
 
-
 declare module Plottable {
     module Utils {
         /**
@@ -75,7 +73,6 @@ declare module Plottable {
         }
     }
 }
-
 
 declare module Plottable {
     module Utils {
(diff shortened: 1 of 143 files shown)


aicioara:
Agree. Can you double check the following: 

- Without the PR, while `grunt watch`ing, try to change a single file (any src/**/*.ts), and check how many files get recompiled in the `build/` (check the timestamp of the `touch`)
- With the PR do the same.

Ideally, only the changed file should be recompiled in the build for both versions.

jtlan:
Reassigning to @aicioara, who has done more work with the tooling than I have.

aicioara:
Yeah, just test the above and it should be good to go.

bluong:
In both scenarios, it seems we build all files.

bluong:
Note that I will be attempting to move so that the Gruntfile.js file is a lot more clear than what it is now.  I have my own personal suspicions on how we set up our compiler, for instance, and while we will move to using Gulp likely, this cleanup should facilitate that migration.

aicioara:
Sounds good

PR #2535, opened 2015-08-03
[Gridlines] - Allow to be rendered on its own

### LINKED ISSUE #2532: Gridlines don't expand to fill available space
Repro: http://jsfiddle.net/f2yqb8ek/

**Actual**: `Scales` backing `Gridlines` don't have their `range()` set correctly, so the lines are bunched up on one side.
**Expected**: `Scales` have their `range()` set appropriately.

bluong:
Currently, `Gridline`s are only able to render to our expectations when it is grouped with something like a `Plots.Scatter` or similar.  However, it would be nice if we could render these `Gridline`s by themselves without heavy reliance on a grouped component.

JSFiddle: http://jsfiddle.net/bluong63/z1pnmgL2/

Fixes #2532

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2202,6 +2202,7 @@ declare module Plottable {
             destroy(): Gridlines;
             protected _setup(): void;
             renderImmediately(): Gridlines;
+            computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number): Gridlines;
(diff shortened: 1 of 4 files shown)


crmorford:
art http://jsfiddle.net/z1pnmgL2/4/ 
+1

PR #2537, opened 2015-08-03
draw stroke with an additional path for pie plot

### LINKED ISSUE #2492: Pie chart slice selection border is not consistent between slices
The slice selection border is not consistent between pie slices and at the inner radius.

The last slice works fine because the slice selection border overlaps all other slices. But each of the other slices have this issue.

See Slate ticket 2429

ztsai:
closes #2492 
demo: http://jsfiddle.net/ztsai/g5spysum/7/

### DIFF
--- a/src/drawers/arcDrawer.ts
+++ b/src/drawers/arcDrawer.ts
@@ -3,18 +3,30 @@
 module Plottable {
 export module Drawers {
   export class Arc extends Drawer {
-    private _isOutline: boolean;
 
-    constructor(dataset: Dataset, isOutline = false) {
+    constructor(dataset: Dataset) {
       super(dataset);
-      this._className = "arc " + (isOutline ? "outline" : "fill");
+      this._className = "arc fill";
       this._svgElementName = "path";
-      this._isOutline = isOutline;
     }
 
     protected _applyDefaultAttributes(selection: d3.Selection<any>) {
       super._applyDefaultAttributes(selection);
-      selection.style(this._isOutline ? "fill" : "stroke", "none");
+      selection.style("stroke", "none");
+    }
+  }
+
+  export class ArcOutline extends Drawer {
+
+    constructor(dataset: Dataset) {
+      super(dataset);
+      this._className = "arc outline";
+      this._svgElementName = "path";
+    }
+
+    protected _applyDefaultAttributes(selection: d3.Selection<any>) {
+      super._applyDefaultAttributes(selection);
+      selection.style("fill", "none");
     }
   }
 }
--- a/src/drawers/arcDrawer.ts
+++ b/src/drawers/arcDrawer.ts
@@ -3,13 +3,19 @@
 module Plottable {
 export module Drawers {
   export class Arc extends Drawer {
+    private _isOutline: boolean;
 
-    constructor(dataset: Dataset) {
+    constructor(dataset: Dataset, isOutline = false) {
       super(dataset);
-      this._className = "arc";
+      this._className = "arc " + (isOutline ? "outline" : "fill");
       this._svgElementName = "path";
+      this._isOutline = isOutline;
     }
 
+    protected _applyDefaultAttributes(selection: d3.Selection<any>) {
+      super._applyDefaultAttributes(selection);
+      selection.style(this._isOutline ? "fill" : "stroke", "none");
+    }
   }
 }
 }
--- a/src/drawers/arcOutlineDrawer.ts
+++ b/src/drawers/arcOutlineDrawer.ts
@@ -2,17 +2,18 @@
 
 module Plottable {
 export module Drawers {
-  export class Arc extends Drawer {
+
+  export class ArcOutline extends Drawer {
 
     constructor(dataset: Dataset) {
       super(dataset);
-      this._className = "arc fill";
+      this._className = "arc outline";
       this._svgElementName = "path";
     }
 
     protected _applyDefaultAttributes(selection: d3.Selection<any>) {
       super._applyDefaultAttributes(selection);
-      selection.style("stroke", "none");
+      selection.style("fill", "none");
     }
   }
 }
--- a/src/drawers/arcOutlineDrawer.ts
+++ b/src/drawers/arcOutlineDrawer.ts
@@ -14,6 +14,7 @@ export module Drawers {
     protected _applyDefaultAttributes(selection: d3.Selection<any>) {
       super._applyDefaultAttributes(selection);
       selection.style("fill", "none");
+      selection.style("stroke-linejoin", "round");
     }
   }
 }
--- a/src/drawers/arcOutlineDrawer.ts
+++ b/src/drawers/arcOutlineDrawer.ts
@@ -14,7 +14,6 @@ export module Drawers {
     protected _applyDefaultAttributes(selection: d3.Selection<any>) {
       super._applyDefaultAttributes(selection);
       selection.style("fill", "none");
-      selection.style("stroke-linejoin", "round");
     }
   }
 }
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -11,6 +11,7 @@ export module Plots {
     private _endAngles: number[];
     private _labelFormatter: Formatter = Formatters.identity();
     private _labelsEnabled = false;
+    private _strokeDrawers: Utils.Map<Dataset, Drawers.Arc>;
 
     /**
      * @constructor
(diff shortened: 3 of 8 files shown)


src/drawers/arcDrawer.ts
```diff
@@ -3,13 +3,19 @@
 module Plottable {
 export module Drawers {
   export class Arc extends Drawer {
+    private _isOutline: boolean;
 
-    constructor(dataset: Dataset) {
+    constructor(dataset: Dataset, isOutline = false) {
```
jtlan: It seems like it would be better to make a subclass for this instead of using a `boolean` parameter.

src/plots/piePlot.ts
```diff
@@ -42,10 +50,20 @@ export module Plots {
         return this;
       }
       this._updatePieAngles();
+      var strokeDrawer = new Drawers.Arc(dataset, true);
+      if (this._isSetup) {
+       strokeDrawer.renderArea(this._renderArea.append("g"));
+      }
+      this._strokeDrawers.set(dataset, strokeDrawer);
```
jtlan: You'll want to `delete()` the entry from the `Set` in `removeDataset()`.

ztsai: it's deleted in ```_removeDatasetNodes```

bluong:
Can you please explain why you're drawing an ArcOutline on top of the `Plots.Pie`?  This seems incredibly odd that we need to run to `Drawer` for this answer

src/drawers/arcDrawer.ts
```diff
@@ -6,10 +6,28 @@ export module Drawers {
 
     constructor(dataset: Dataset) {
       super(dataset);
-      this._className = "arc";
+      this._className = "arc fill";
```
bluong: Why are we adding this class name?

src/drawers/arcDrawer.ts
```diff
@@ -6,10 +6,28 @@ export module Drawers {
...
       this._svgElementName = "path";
     }
 
+    protected _applyDefaultAttributes(selection: d3.Selection<any>) {
+      super._applyDefaultAttributes(selection);
+      selection.style("stroke", "none");
+    }
+  }
+
+  export class ArcOutline extends Drawer {
```
bluong: If we want this class, it should be in a different file.

bluong: We have to think about the name of this class.  Not sure about `ArcOutline`.  Maybe we should name this thing to be `Arc` and rename the odd `Arc` to be `Sector` maybe?

src/drawers/arcDrawer.ts
jtlan: Distinguishing between the fill and the stroke.

src/drawers/arcDrawer.ts
jtlan: We can't rename the existing classes without an API break.

jtlan:
Split the `ArcOutline` `Drawer` into its own file, and I think that should be it.

jtlan:
Repro focusing on new CSS classes: http://jsfiddle.net/249fe23e/

crmorford:
http://jsfiddle.net/v0b2ur4L/

crmorford:
Is there a way to get all the centers looking the same?



vs

crmorford:
I also predict slate being unhappy with the way multi select looks :( 
http://jsfiddle.net/v0b2ur4L/3/
it doesn't seem predictable which colors are going to overlap which other colors.

crmorford:
Slate says the overlap would be better if the z-ordering was based on click order. it would feel more intuitive that way.

crmorford:
Talked to Christina -- she said ideally, the stroke would be on the inside of each pie slice. I didn't think SVG allowed that so we looked it up... sounds like it might be in the works? http://stackoverflow.com/questions/7241393/can-you-control-how-an-svgs-stroke-width-is-drawn

We might have to do this not entirely in css, in order to do the click-ordering matches z-ordering. Then we can watch for SVG2 and use the inner-stroke when it comes out. https://svgwg.org/specs/strokes/#SpecifyingStrokeAlignment

ztsai:
talked to Natasha offline, and she said the current implementation works fine with their current design
drawing stroke inside of each pie slice might be an issue if user updates stroke's property via css as we do not re-render upon css changes
I will look into the linejoin issue in the center tho

ztsai:
I set the linejoin to "round" because I think it looks best with single select and multi-select
Let me know if anyone feels it makes more sense to set it to something else

- round
screen shot 2015-08-17 at 3 08 58 pm screen shot 2015-08-17 at 3 09 47 pm

- miter
screen shot 2015-08-17 at 3 09 27 pm screen shot 2015-08-17 at 3 09 32 pm

- bevel
screen shot 2015-08-17 at 3 09 54 pm screen shot 2015-08-17 at 3 09 57 pm

src/drawers/arcOutlineDrawer.ts
```diff
@@ -0,0 +1,21 @@
...
+    constructor(dataset: Dataset) {
+      super(dataset);
+      this._className = "arc outline";
+      this._svgElementName = "path";
+    }
+
+    protected _applyDefaultAttributes(selection: d3.Selection<any>) {
+      super._applyDefaultAttributes(selection);
+      selection.style("fill", "none");
+      selection.style("stroke-linejoin", "round");
```
jtlan: Do we want the user to have control over this? If so, we should define it in **plottable.css** instead of here.

test/plots/piePlotTests.ts
```diff
@@ -13,6 +13,29 @@ describe("Plots", () => {
...
+      let dataset = new Plottable.Dataset(data);
+      piePlot.addDataset(dataset);
+      piePlot.renderTo(svg);
+      let renderArea = (<any> piePlot)._renderArea;
+      let arcPaths = renderArea.selectAll(".arc");
+      let arcFillPaths = renderArea.selectAll(".arc.fill");
+      let arcOutlinePaths = renderArea.selectAll(".arc.outline");
+      assert.strictEqual(arcPaths.size(), 4, "2 paths per datum");
+      assert.strictEqual(arcFillPaths.size(), 2, "1 fill path per datum");
+      assert.strictEqual(arcOutlinePaths.size(), 2, "1 outline path1 per datum");
```
jtlan: "path1"?

test/plots/piePlotTests.ts
```diff
@@ -351,14 +374,14 @@ describe("Plots", () => {
     describe("selections", () => {
       it("retrieves all dataset selections with no args", () => {
         let allSectors = piePlot.selections();
-        assert.strictEqual(allSectors.size(), 2, "all sectors retrieved");
+        assert.strictEqual(allSectors.size(), 2 * 2, "all sectors retrieved");
```
jtlan: Can we assert that the returned `selection` includes the correct number of fill and stroke `<path>`s?

test/plots/piePlotTests.ts
```diff
@@ -351,14 +374,14 @@ describe("Plots", () => {
...
 
         svg.remove();
       });
 
       it("retrieves correct selections", () => {
         let allSectors = piePlot.selections([simpleDataset]);
-        assert.strictEqual(allSectors.size(), 2, "all sectors retrieved");
+        assert.strictEqual(allSectors.size(), 2 * 2, "all sectors retrieved");
```
jtlan: Same here.

jtlan:
Can you also add a test that verifies that `entities()` returns `Entities` whose `selection` includes both the area and the path?

test/plots/piePlotTests.ts
```diff
@@ -13,6 +13,51 @@ describe("Plots", () => {
...
+      ];
+      let dataset = new Plottable.Dataset(data);
+      piePlot.addDataset(dataset);
+      piePlot.renderTo(svg);
+
+      let entities = piePlot.entities();
+      entities.forEach((entity) => {
+        assert.strictEqual(entity.selection.size(), 2, "each entity selection has 2 paths");
+        assert.lengthOf(entity.selection.filter("fill"), 1, "each entity selection has 1 fill path");
+        assert.lengthOf(entity.selection.filter("outline"), 1, "each entity selection has 1 stroke path");
```
jtlan: Shouldn't the filters here be `".fill"` and `".outline"`?

D3 selections always have size 1. The correct way check their size is this:

```Typescript
assert.strictEquals(selection.size(), expectedSize, message);
```

crmorford:
http://jsfiddle.net/g5spysum/10/
Gonna ask Sara to look at this and see what she thinks. Center looks good to me though!

crmorford:
following slate talk, giving +1

PR #2539, opened 2015-08-03
[Plots.Scatter] - Adding `entitiesIn()` endpoint

### LINKED ISSUE #2519: entitiesIn() for Plots.Scatter
`Plots.Scatter` is probably the `Plot` that needs `entitiesIn()` the most.

bluong:
Similar to how `Plots.Bar` has an `entitiesIn()` endpoint, `Plots.Scatter` can now use this endpoint to find all of the entities within the given range or with the given bounds.

API Changes:
```typescript
/**
 * Gets the Entities that intersect the Bounds.
 *
 * @param {Bounds} bounds
 * @returns {PlotEntity[]}
 */
entitiesIn(bounds: Bounds): PlotEntity[];
/**
 * Gets the Entities that intersect the area defined by the ranges.
 *
 * @param {Range} xRange
 * @param {Range} yRange
 * @returns {PlotEntity[]}
 */
entitiesIn(xRange: Range, yRange: Range): PlotEntity[];
```

JSFiddle: http://jsfiddle.net/bluong63/o3s4wst5/

Closes #2519

### DIFF
--- a/src/plots/scatterPlot.ts
+++ b/src/plots/scatterPlot.ts
@@ -169,14 +169,12 @@ export module Plots {
         dataXRange = (<Range> xRangeOrBounds);
         dataYRange = yRange;
       }
-      var attrToProjector = this._generateAttrToProjector();
       return this.entities().filter((entity) => {
         var datum = entity.datum;
         var index = entity.index;
         var dataset = entity.dataset;
-        var translate = d3.transform(attrToProjector["transform"](datum, index, dataset)).translate;
-        var x = translate[0];
-        var y = translate[1];
+        var x = Plot._scaledAccessor(this.x())(datum, index, dataset);
+        var y = Plot._scaledAccessor(this.y())(datum, index, dataset);
         return dataXRange.min <= x && x <= dataXRange.max && dataYRange.min <= y && y <= dataYRange.max;
       });
     }
--- a/src/plots/scatterPlot.ts
+++ b/src/plots/scatterPlot.ts
@@ -142,6 +142,44 @@ export module Plots {
         symbolProjector(datum, index, dataset)(sizeProjector(datum, index, dataset));
       return propertyToProjectors;
     }
+
+    /**
+     * Gets the Entities that intersect the Bounds.
+     *
+     * @param {Bounds} bounds
+     * @returns {PlotEntity[]}
+     */
+    public entitiesIn(bounds: Bounds): PlotEntity[];
+    /**
+     * Gets the Entities that intersect the area defined by the ranges.
+     *
+     * @param {Range} xRange
+     * @param {Range} yRange
+     * @returns {PlotEntity[]}
+     */
+    public entitiesIn(xRange: Range, yRange: Range): PlotEntity[];
+    public entitiesIn(xRangeOrBounds: Range | Bounds, yRange?: Range): PlotEntity[] {
+      var dataXRange: Range;
+      var dataYRange: Range;
+      if (yRange == null) {
+        var bounds = (<Bounds> xRangeOrBounds);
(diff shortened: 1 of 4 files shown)


test/plots/scatterPlotTests.ts
```diff
@@ -140,6 +140,58 @@ describe("Plots", () => {
...
+          .y((d: any) => d.y, yScale)
+          .addDataset(dataset)
+          .addDataset(dataset2);
+      plot.renderTo(svg);
+
+      var entities = plot.entitiesIn({ topLeft: {
+                                         x: xScale.scale(1),
+                                         y: yScale.scale(1)
+                                       },
+                                       bottomRight: {
```
jtlan: Why is `bottomRight` the same as `topLeft?

jtlan:
Should we have a test case where the selection zone passes through a point without including its center?

test/plots/scatterPlotTests.ts
bluong: I wanted the bounds to hit the point exactly without touching any other points.  I could add a tiny bit of error to both sides of x/y but I think it's unnecessary.

bluong:
I can write one with not much trouble.

test/plots/scatterPlotTests.ts
```diff
@@ -140,6 +140,80 @@ describe("Plots", () => {
...
+      var entities = plot.entitiesIn({ min: xScale.scale(1), max: xScale.scale(1) },
+                                     { min: yScale.scale(1), max: yScale.scale(1) });
+
+      assert.lengthOf(entities, 1, "only one entity has been retrieved");
+      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
+
+      svg.remove();
+    });
+
+    it("cannot retrieve entities if range does not contain center", () => {
```
jtlan: "Entities are only returned if their center is contained in the ranges" would be a better description.

src/plots/scatterPlot.ts
```diff
@@ -142,6 +142,44 @@ export module Plots {
...
+      } else {
+        dataXRange = (<Range> xRangeOrBounds);
+        dataYRange = yRange;
+      }
+      var attrToProjector = this._generateAttrToProjector();
+      return this.entities().filter((entity) => {
+        var datum = entity.datum;
+        var index = entity.index;
+        var dataset = entity.dataset;
+        var translate = d3.transform(attrToProjector["transform"](datum, index, dataset)).translate;
```
jtlan: Instead of this, would it make more sense to use the `x()` and `y()` applied accessors? Then we don't have to go to the DOM.

bluong: This doesn't go to the DOM either since we use the `attrToProjector` to get the right naswer

jtlan: Ah. But it'd be fewer layers of misdirection if we just used `x()` and `y()` directly, right?

bluong: Possibly.  I'll make the switch and see how it works.

src/plots/scatterPlot.ts
```diff
@@ -142,6 +142,42 @@ export module Plots {
...
+      } else {
+        dataXRange = (<Range> xRangeOrBounds);
+        dataYRange = yRange;
+      }
+      return this.entities().filter((entity) => {
+        var datum = entity.datum;
+        var index = entity.index;
+        var dataset = entity.dataset;
+        var x = Plot._scaledAccessor(this.x())(datum, index, dataset);
+        var y = Plot._scaledAccessor(this.y())(datum, index, dataset);
```
jtlan: Declaring the scaled accessors outside would be slightly more efficient.

crmorford:
http://jsfiddle.net/o3s4wst5/2/

PR #2546, opened 2015-08-04
Deferred Rendering fix

### LINKED ISSUE #2538: Deferred rendering + PanZoom on only one scale behaves weirdly
Good: http://jsfiddle.net/gp522apr/
Bad: http://jsfiddle.net/gp522apr/1/

aicioara:
Closes #2538

Before: http://jsfiddle.net/ys358yxy/2/
After: http://jsfiddle.net/ys358yxy/1/

The issue was that an unanchored plot is not setting up the scale correctly (because it does not know how much space it has in the SVG) and we were not taking that into account. With this fix will only cache domains for anchored plots

Also as part of this PR: 
- Added tests for deferredRendering
- Fixed more issues caught by the tests and not observed through manual inspection
- `onDetachCallbacks` on the component are not called unless the component was already anchored
- added tests for the item above
- Rearranged the test file for XYPlot. I did not like that the beforeEach was doing way too much for the basic use cases of the XYPlot. Also, I did not like creating a new XYPlot category, so I added 1 more level of nesting to allow my DeferredRendering to live under the XYPlot tag

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -450,10 +450,10 @@ export class Component {
 
     if (this._isAnchored) {
       this._element.remove();
+      this._isAnchored = false;
+      this._onDetachCallbacks.callCallbacks(this);
     }
-    this._isAnchored = false;
 
-    this._onDetachCallbacks.callCallbacks(this);
     return this;
   }
 
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -31,8 +31,8 @@ export class XYPlot<X, Y> extends Plot {
     var _deltaY = 0;
     var _scalingX = 1;
     var _scalingY = 1;
-    var _lastSeenDomainX: X[] = null;
-    var _lastSeenDomainY: Y[] = null;
+    var _lastSeenDomainX: X[] = [null, null];
+    var _lastSeenDomainY: Y[] = [null, null];
     var _timeoutReference = 0;
     var _deferredRenderingTimeout = 500;
 
@@ -69,6 +69,9 @@ export class XYPlot<X, Y> extends Plot {
     };
 
     var _lazyDomainChangeCallbackY = (scale: Scale<Y, any>) => {
+      console.log("cached domain Y", this._cachedDomainY);
+
+
       if (!this._isAnchored) {
         return;
       }
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -49,6 +49,8 @@ export class XYPlot<X, Y> extends Plot {
         this._cachedDomainY = _lastSeenDomainY;
         _deltaX = 0;
         _deltaY = 0;
+        _scalingX = 1;
+        _scalingY = 1;
         this.render();
         this._renderArea.attr("transform", "translate(0, 0) scale(1, 1)");
       }, _deferredRenderingTimeout);
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -111,7 +111,7 @@ export class XYPlot<X, Y> extends Plot {
       return this._deferredRendering;
     }
 
-    if (deferredRendering) {
+    if (deferredRendering && this._isAnchored) {
       if (this.x() && this.x().scale) {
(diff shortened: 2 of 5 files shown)


src/components/component.ts
```diff
@@ -450,10 +450,10 @@ export class Component {
 
     if (this._isAnchored) {
       this._element.remove();
+      this._isAnchored = false;
+      this._onDetachCallbacks.callCallbacks(this);
```
bluong: Can you explain why these changes should be in this pull request?  It is not that clear from the description why this is part of the fix.  Also, I feel that we should ask if `detach()` on un-anchored components should call the callbacks.

src/plots/xyPlot.ts
```diff
@@ -49,6 +49,8 @@ export class XYPlot<X, Y> extends Plot {
         this._cachedDomainY = _lastSeenDomainY;
         _deltaX = 0;
         _deltaY = 0;
+        _scalingX = 1;
+        _scalingY = 1;
```
bluong: Is this the majority of the fix?

bluong:
Mainly a bunch of questions

src/plots/xyPlot.ts
```diff
@@ -109,7 +111,7 @@ export class XYPlot<X, Y> extends Plot {
       return this._deferredRendering;
     }
 
-    if (deferredRendering) {
+    if (deferredRendering && this._isAnchored) {
```
aicioara: This is the main fix. There is no point in cacheing the domain when the plot is not anchored, because the domain of the scale will be something default (like [0, 1])

src/plots/xyPlot.ts
aicioara: This is just good practice. It fixed the issue initially, but then the line bellow addressed some edge cases. 

I was not resetting _scalingXY before because I was wrongly assuming that both my callbacks (bellow) are going to be called every time. I was not considering the case where pan and zoom is in just one direction. 

I hope it makes sense

src/components/component.ts
aicioara: Main reason why this is here is because I was playing with things to fix this. This change makes sense to me, why would we call the callbacks if not detaching is performed. 

I agree that the change does not make sense in the context of this PR, so I am happy to drop it if you have a strong preference towards me doing so. 

However, I think the change is needed and it's an overhead to open a new PR

aicioara:
Addressed

src/components/component.ts
bluong: Let's loop to QE for this specific part.  While it is overhead to open a new PR, it keeps a separation of features that is very handy for testing and reviewing (note that I commented on this part in the first place, because I was confused how it fit into the whole mix), but luckily this is a decently small pull request that adding this context is not difficult.

bluong:
Is there a test we can attach to this?  We essentially have to add a test for the change in functionality for `detach()` part as well.

bluong:
I would like a test regarding the actual bugfix with deferredRendering

aicioara:
Added tests. There was no easy way to wait for the deferred rendering to finish and check the cached domain. If you can think of any let me know.

aicioara:
Ignore that, I hacked it.

src/plots/xyPlot.ts
```diff
@@ -31,8 +31,8 @@ export class XYPlot<X, Y> extends Plot {
     var _deltaY = 0;
     var _scalingX = 1;
     var _scalingY = 1;
-    var _lastSeenDomainX: X[] = null;
-    var _lastSeenDomainY: Y[] = null;
+    var _lastSeenDomainX: X[] = [null, null];
+    var _lastSeenDomainY: Y[] = [null, null];
```
bluong: Minor, but would this work better?
`var _lastSeenDomainX: [X, X] = [null, null];`

aicioara: It will not work because `scale.domain()` returns `X[]` and 
```
Type 'X[]' is not assignable to type '[X, X]'.
  Property '0' is missing in type 'X[]'.
```

bluong: Ah ok.

PR #2548, opened 2015-08-04
XDragBoxLayer and YDragBoxLayer don't error when destroy()-ed.

bluong:


jtlan:
Would it be better to just not have the `xScale()` and `yScale()` getters throw errors?

bluong:
I didn't necessarily want to have a functional change to patch this up.  That said, I'm unsure what approach to take.

crmorford:
This hotfix is about the error "XDragBoxLayer has no yScale" and vice versa, although you can't (and shouldn't) set the yScale. Correct?

crmorford:
Anyway, here's something weird: http://jsfiddle.net/fee6p9w0/4/
The last line of the fiddle is "dragBoxLayer.destroy()", so it should be gone. But you can still drag and reach the call back .onDrag(), as evidenced by the console.logs.

crmorford:
Summary (a la Justin)
User perspective: You should be able to call .destroy() without getting any errors.

crmorford:
QE fiddles
DragBoxLayer http://jsfiddle.net/fee6p9w0/7/
SelectionBoxLayer http://jsfiddle.net/fee6p9w0/8/
XDragBoxLayer http://jsfiddle.net/fee6p9w0/9/
YDragBoxLayer http://jsfiddle.net/fee6p9w0/10/

crmorford:
No errors, everything seems to work correctly whether xScale and yScale are set or not.

PR #2550, opened 2015-08-04
API endpoint to specify opacity of legend (take 2)

yqlu:
Follow-up from PR #2461 and addresses issue #2434 

Currently there is no way to update the opacity of symbols in the legend in an API-endorsed way. This is desirable if you, for example, built a scatter plot of various datasets with varying colors and opacities, and would like the opacity in the legend to match.

This PR provides the following API endpoints on `Legend`:
```typescript
/**
 * Gets the opacity of the symbols of the Legend.
 *
 * @returns {(datum: any, index: number) => number}
 */
symbolOpacity(): (datum: any, index: number) => number;

/**
 * Sets the opacity of the symbols of the Legend.
 *
 * @param {number | ((datum: any, index: number) => number)} symbolOpacity
 * @returns {Legend} The calling Legend
 */
symbolOpacity(symbolOpacity: number | ((datum: any, index: number) => number)): Legend;
```

See http://plnkr.co/edit/nyEuE8L7rekfoftjeOSt?p=preview for demo.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2133,15 +2133,15 @@ declare module Plottable {
              */
             symbol(symbol: (datum: any, index: number) => SymbolFactory): Legend;
             /**
-             * Gets the function determining the opacity of the symbols of the Legend.
+             * Gets the opacity of the symbols of the Legend.
              *
              * @returns {(datum: any, index: number) => number}
              */
             symbolOpacity(): (datum: any, index: number) => number;
             /**
-             * Sets the function determining the opacity of the symbols of the Legend.
+             * Sets the opacity of the symbols of the Legend.
              *
-             * @param {(datum: any, index: number) => number} opacityFunction
+             * @param {number | ((datum: any, index: number) => number)} symbolOpacity
              * @returns {Legend} The calling Legend
              */
             symbolOpacity(symbolOpacity: number | ((datum: any, index: number) => number)): Legend;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2132,6 +2132,26 @@ declare module Plottable {
              * @returns {Legend} The calling Legend
              */
             symbol(symbol: (datum: any, index: number) => SymbolFactory): Legend;
+            /**
+             * Gets the function determining the opacity of the symbols of the Legend.
+             *
+             * @returns {(datum: any, index: number) => number}
+             */
+            symbolOpacity(): (datum: any, index: number) => number;
+            /**
+             * Sets the function determining the symbols of the Legend.
+             *
+             * @param {(datum: any, index: number) => number} opacityFunction
+             * @returns {Legend} The calling Legend
+             */
+            symbolOpacity(opacityFunction: (datum: any, index: number) => number): Legend;
+            /**
+             * Sets the function determining the symbols of the Legend.
+             *
+             * @param {number} constantOpacity
+             * @returns {Legend} The calling Legend
+             */
+            symbolOpacity(constantOpacity: number): Legend;
             fixedWidth(): boolean;
             fixedHeight(): boolean;
         }
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2139,19 +2139,12 @@ declare module Plottable {
              */
             symbolOpacity(): (datum: any, index: number) => number;
             /**
-             * Sets the function determining the symbols of the Legend.
+             * Sets the function determining the opacity of the symbols of the Legend.
              *
              * @param {(datum: any, index: number) => number} opacityFunction
              * @returns {Legend} The calling Legend
              */
-            symbolOpacity(opacityFunction: (datum: any, index: number) => number): Legend;
-            /**
-             * Sets the function determining the symbols of the Legend.
-             *
-             * @param {number} constantOpacity
-             * @returns {Legend} The calling Legend
-             */
-            symbolOpacity(constantOpacity: number): Legend;
+            symbolOpacity(symbolOpacity: number | ((datum: any, index: number) => number)): Legend;
             fixedWidth(): boolean;
             fixedHeight(): boolean;
         }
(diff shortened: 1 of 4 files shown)


src/components/legend.ts
```diff
@@ -280,6 +282,7 @@ export module Components {
       entries.select("path").attr("d", (d: any, i: number, j: number) => this.symbol()(d, j)(layout.textHeight * 0.6))
                             .attr("transform", "translate(" + (layout.textHeight / 2) + "," + layout.textHeight / 2 + ")")
                             .attr("fill", (value: string) => this._colorScale.scale(value))
+                            .attr("opacity", (d: any, i: number) => this.symbolOpacity()(d, i))
```
bluong: Can you check if the opacity based on the index of the datum actually works correctly.  We've had a bug in the past where the indices were not being piped through correctly.

src/components/legend.ts
```diff
@@ -326,6 +329,38 @@ export module Components {
       }
     }
 
+    /**
+     * Gets the function determining the opacity of the symbols of the Legend.
+     *
+     * @returns {(datum: any, index: number) => number}
+     */
+    public symbolOpacity(): (datum: any, index: number) => number;
```
bluong: We would need to figure out if we want to always return a function here or maybe union type it between a number and a function

src/components/legend.ts
```diff
@@ -326,6 +329,38 @@ export module Components {
...
+     */
+    public symbolOpacity(opacityFunction: (datum: any, index: number) => number): Legend;
+    /**
+     * Sets the function determining the symbols of the Legend.
+     *
+     * @param {number} constantOpacity
+     * @returns {Legend} The calling Legend
+     */
+    public symbolOpacity(constantOpacity: number): Legend;
+    public symbolOpacity(opacity?: number|((datum: any, index: number) => number)): any {
```
bluong: Can you put whitespace in between the union types?

src/components/legend.ts
yqlu: If you inspect the elements in the third legend in the plunker I created, an opacity based on the index works:

```
legend.symbolOpacity(function(d,i) {return 1 - i/5;})
```

I can codify this up into a unit test (the current unit test has a function based on datum).

bluong: Yeah, the main reason I'm concerned is because in the past (As you can see from 3 lines up) we were handing in the wrong index for whatever reason.

src/components/legend.ts
yqlu: Done

src/components/legend.ts
yqlu: Haha, good catch -- I wrote a unit test on the indices and it had the same issue. I've updated it to do the same thing as symbol, passing in d and j instead of d and i.

src/components/legend.ts
yqlu: Mmm good point. I chose to do it this way because I noticed that accessors in plots are always converted to functions:

```
> var rect = new Plottable.Plots.Rectangle().x(5);
> rect.x().accessor
function (){return n}
```

This would be consistent with d3 (`Plot._bind()` calls `d3.functor()`, after all) but might be confusing to the first time user. What do you think?

jtlan: Upcasting makes the most sense -- the typing information supplied by Typescript should protect users in that language.

src/components/legend.ts
```diff
@@ -280,6 +282,7 @@ export module Components {
       entries.select("path").attr("d", (d: any, i: number, j: number) => this.symbol()(d, j)(layout.textHeight * 0.6))
                             .attr("transform", "translate(" + (layout.textHeight / 2) + "," + layout.textHeight / 2 + ")")
                             .attr("fill", (value: string) => this._colorScale.scale(value))
+                            .attr("opacity", (d: any, i: number, j: number) => this.symbolOpacity()(d, j))
```
bluong: Did there wind up being a bug where we needed to supply `j` as well?

src/components/legend.ts
```diff
@@ -326,6 +329,38 @@ export module Components {
...
+     * @returns {Legend} The calling Legend
+     */
+    public symbolOpacity(opacityFunction: (datum: any, index: number) => number): Legend;
+    /**
+     * Sets the function determining the symbols of the Legend.
+     *
+     * @param {number} constantOpacity
+     * @returns {Legend} The calling Legend
+     */
+    public symbolOpacity(constantOpacity: number): Legend;
```
bluong: Feel free to combine the above two setters into a single union type

Let's also call the parameter `symbolOpacity`

src/components/legend.ts
yqlu: Yes. I'm still not sure how the indices are being passed around, but when I wrote up the index unit tests for symbolOpacity() in the same style as symbol(), I realized the index was off and doing the same thing here with the j solved it.

src/components/legend.ts
yqlu: Done!

plottable.d.ts
```diff
@@ -2132,6 +2132,19 @@ declare module Plottable {
...
              */
             symbol(symbol: (datum: any, index: number) => SymbolFactory): Legend;
+            /**
+             * Gets the function determining the opacity of the symbols of the Legend.
+             *
+             * @returns {(datum: any, index: number) => number}
+             */
+            symbolOpacity(): (datum: any, index: number) => number;
+            /**
+             * Sets the function determining the opacity of the symbols of the Legend.
```
bluong: Can we change the documentation to 
"Sets the opacity of the symbols of the Legend."

crmorford:
cool :) http://jsfiddle.net/xzp3yp4m/4/

crmorford:
ps i qe'd this on every major browser + ipad

bluong:
Really neat example

PR #2551, opened 2015-08-04
[TSLint] - enforce no-var-keyword on scaleTests

bluong:
Fun thing to see is that I'm trying to move towards biasing towards a `const` usage over a `let` usage and as a result some test code is a bit more precise on what its testing.

### DIFF
--- a/test/scales/linearScaleTests.ts
+++ b/test/scales/linearScaleTests.ts
@@ -1,43 +1,42 @@
 ///<reference path="../testReference.ts" />
-/* tslint:disable: no-var-keyword */
 
 describe("Scales", () => {
   describe("Linear Scales", () => {
     it("extentOfValues() filters out invalid numbers", () => {
-      var scale = new Plottable.Scales.Linear();
-      var expectedExtent = [0, 1];
-      var arrayWithBadValues: any[] = [null, NaN, undefined, Infinity, -Infinity, "a string", 0, 1];
-      var extent = scale.extentOfValues(arrayWithBadValues);
+      const scale = new Plottable.Scales.Linear();
+      const expectedExtent = [0, 1];
+      const arrayWithBadValues: any[] = [null, NaN, undefined, Infinity, -Infinity, "a string", 0, 1];
+      const extent = scale.extentOfValues(arrayWithBadValues);
       assert.deepEqual(extent, expectedExtent, "invalid values were filtered out");
     });
 
     it("autoDomain() defaults to [0, 1]", () => {
-      var scale = new Plottable.Scales.Linear();
+      const scale = new Plottable.Scales.Linear();
       scale.autoDomain();
-      var d = scale.domain();
+      const d = scale.domain();
       assert.strictEqual(d[0], 0);
       assert.strictEqual(d[1], 1);
     });
 
     it("autoDomain() expands single value to [value - 1, value + 1]", () => {
-      var scale = new Plottable.Scales.Linear();
-      var singleValue = 15;
+      const scale = new Plottable.Scales.Linear();
+      const singleValue = 15;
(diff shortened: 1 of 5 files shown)


aicioara:
TBH, I don't really like this TSLint rule. I like my `var`'s. Also, this is a huge conflict generator.

test/scales/linearScaleTests.ts
```diff
@@ -188,9 +187,9 @@ describe("Scales", () => {
     });
 
     describe("autoranging behavior", () => {
-      var data: any[];
-      var dataset: Plottable.Dataset;
-      var scale: Plottable.Scales.Linear;
+      let data: any[];
+      let dataset: Plottable.Dataset;
+      let scale: Plottable.Scales.Linear;
```
aicioara: Why is this scale a `let` and not a `const`? I just feel like this `let` / `const` will lead to inconsistencies

aicioara:
I am against this change as a whole. I feel like we should not get overexcited and jump on new TS1.5 features like that.

My arguments against it are: 
- The change is big and conflict prone
- The change is time consuming and it is not justified 
- It introduces inconsistencies between when we use const and when we use let (as you can see, you tend to overuse the const, what if that is actually a variable? 1. I have to always think what sort of purpose a private (!) variable will serve. 2. If I get it wrong, there will always be 1 extra line change per change (const -> let)
- I can't see the benefits of forcing devs into this yet. Let's give it some time, see how we are actually using `const` and `let` in the wild, and then conclude if it makes sense. So far I have rarely seen anyone using it, even though we upgraded 2 weeks ago.

Need more opinion on this @jtlan @ztsai 

If at one more dev agrees with this change, then sure. I reviewed the code change and it is okay.

aicioara:
Reassigning this. If either @jtlan or @ztsai are okay with the change, it's a +1 from me.

test/scales/linearScaleTests.ts
bluong: If we let a variable be uninitialized to a value, that's essentially a definition that `const` should not be used to that variable.

aicioara: Which is weird, right? Since we use this variable as a const

bluong:
Being able to place bounds onto a variable is in my view extremely important.  The fact that the `var` keyword is scoped at the function level instead of the block level is something that can bite many a new javascript developer since, from what I gather, many of the common programming langauges essentially have the block level be the default.

In terms of how we have got bitten.
```typescript
for (var test = 0; test < 3; test++) {
  Plottable.Utils.Window.warn(test.toString());
}
var test: number;
Plottable.Utils.Window.warn(test.toString());
```

The above compiles, meaning that a developer later could get bitten by the fact that `test` is already initialized to a value that came from the above for loop.

bluong:
In terms of how we should use `const` and `let`, to my understanding, there are many variables that unknowingly we use that are essentially `const`s.  This tells a reader of the code that this value is not going to be changed, which is great if your code is meant to be read almost essentially as some form of procedural flow.

bluong:
A good last point I want to make is that "thinking about the purpose of a variable" is actually something that I think we should do more.  If the purpose of a variable is more embedded into the actual writing of the variable, in my view, it becomes immensely easier to read.

aicioara:
Removing the -1 to avoid confusion. 

I agree with your points above. I do not agree with forcing the devs to use this already, until we see how it bites back (basically do not agree with the tslint rule)

bluong:
We've already seen it bite us when a dev accidentally used a variable that he thought was scoped to inside a for loop.  Thankfully, we caught the bug before it was seen in the wild.

ztsai:
I am in favor of using ```const```/```let``` when possible to avoid scope confusion
I am not sure how big of conflict it would cause, but I think for this kind of change it's better to do it sooner than later; the longer we wait the more we have to change in the future

PR #2558, opened 2015-08-04
No more flickering for Bar Plot pan zoom

### LINKED ISSUE #1280: Scales in bar plots don't account for bar width

### LINKED ISSUE #2545: Glitch when panning and zooming clustered bars
<a href="http://www.youtube.com/watch?feature=player_embedded&v=JS0WA17XLnc
" target="_blank">Pan-Zoom Glitch</a>

aicioara:
Closes #2545
Closes #1280

This was actually a feature, but it has been signaled as a bug so many times, that it no longer makes sense.

Before: http://jsfiddle.net/m6amh0vv/3/
After: http://jsfiddle.net/m6amh0vv/2/

The test changed was this (this shows why the feature made sense more or less):
Before: http://jsfiddle.net/x52jL23n/2/
After: http://jsfiddle.net/x52jL23n/3/

QE Notes:
- The main behavior to be tested besides the flickering is autoDomaining when the bars are very close to the edge of the plot. The old way of keeping the bars in the view was to shrink them. The new way of doing it is expanding the domain. I played around with horizontal and vertical plots, with or without padding proportion


jtlan:
Slate originally requested the auto-width feature. With this new change, the bars can be cut off if the spacing between them is larger than the spacing between the two closest bars (http://jsfiddle.net/godesfyk/), so deleting this feature will cause a regression.

Instead, can we fix it by addressing this issue: https://github.com/palantir/plottable/issues/1280 ?

aicioara:
@jtlan You mean solve that in addition to this?

For a pan zoom (or manually setting the domain) it makes sense to cut off bars, as that is what the user wishes. Instead you want to make the auto-domain smarter?

jtlan:
>Instead you want to make the auto-domain smarter?

This. The width was set in pixels originally because it was difficult to include the effects of bar width when autodomain()-ing, since the width could be set in pixel-space only.

aicioara:
Sorry, was that a yes or a no to my question?

jtlan:
That'd be a "yes". The original way of fixing the problem involved shrinking the bars to make them fit the domain, rather than expanding the domain to fit the bars. However, the way we do autodomaining has gotten a bit cleaner as a result of the 1.0.0 refactor, so it's probably time to revisit this issue.

aicioara:
Sounds good

src/plots/barPlot.ts
```diff
@@ -397,6 +397,24 @@ export module Plots {
       }
     }
 
+    /**
+     * Makes sure the extent takes into account the widths of the bars
+     */
+    protected _computeExtent(dataset: Dataset, accScaleBinding: Plots.AccessorScaleBinding<any, any>, filter: Accessor<boolean>): any[] {
```
jtlan: Could this logic have been added to `_extentsForProperty()` instead? That method is already `protected`.

aicioara: Yup, I'll have a look.

src/plots/barPlot.ts
```diff
@@ -397,6 +397,35 @@ export module Plots {
...
+      let accScaleBinding: Plots.AccessorScaleBinding<any, any>;
+      if (property === "x" && this._isVertical) {
+        accScaleBinding = this.x();
+      } else if (property === "y" && !this._isVertical) {
+        accScaleBinding = this.y();
+      } else {
+        return extents;
+      }
+
+      if (!(accScaleBinding.scale instanceof QuantitativeScale)) {
```
jtlan: Potential for error if `accScaleBinding` is `undefined`, which is possible if `x()` or `y()` hasn't been set up by the user.

crmorford:
please stop using hashes instead of branch names. this example is from a week old commit 
http://jsfiddle.net/m6amh0vv/2/

aicioara:
The fiddles are also for future reference. When the branch is deleted, it is pretty hard to find out what happened during a PR. Will try not to use them in the future. Note that the second pair of fiddles which are actually for QE do not use hashes.

crmorford:
bar http://jsfiddle.net/a3mmsbdt/7/
stacked bar http://jsfiddle.net/m6amh0vv/4/
waterfall (derived from bar) http://jsfiddle.net/a3mmsbdt/5/
constrained domain from start http://jsfiddle.net/a3mmsbdt/6/
overlaying looks good too

jasperSpeicher:
Thank you!

PR #2559, opened 2015-08-04
check touch positions for zooming in iPad

### LINKED ISSUE #2510: iPad: Pan and Zoom interactions within one chart affect other charts
Panning/zooming within a single chart causes other charts within the document to pan/zoom. This happens on an iPad when multiple charts with Pan and Zoom enabled are within a single document. 

Version: plottable v1.4.0

ztsai:
closes #2510 
http://jsfiddle.net/ztsai/w82shn5r/

### DIFF
--- a/test/interactions/panZoomInteractionTests.ts
+++ b/test/interactions/panZoomInteractionTests.ts
@@ -6,6 +6,7 @@ describe("Interactions", () => {
     var SVG_WIDTH = 400;
     var SVG_HEIGHT = 500;
 
+    var component: Plottable.Component;
     var eventTarget: d3.Selection<void>;
 
     var xScale: Plottable.QuantitativeScale<number>;
@@ -15,7 +16,7 @@ describe("Interactions", () => {
     beforeEach(() => {
       svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
 
-      var component = new Plottable.Component();
+      component = new Plottable.Component();
       component.renderTo(svg);
 
       xScale = new Plottable.Scales.Linear();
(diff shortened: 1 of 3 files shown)


bluong:
We'll need a test with this

bluong:
Can you test what happens in this scenario?  I start with fingers outside the svg and then I move those fingers into the svg without lifting them off the ipad?

ztsai:
the scenario seems almost impossible since if you try to zoom outside then the whole page will zoom instead; I don't think you can have both end points being inside of one component

test/interactions/panZoomInteractionTests.ts
```diff
@@ -181,6 +181,29 @@ describe("Interactions", () => {
       svg.remove();
     });
 
+    it("pinching inside one component does not affect another component", () => {
+      var svg2 = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      var component2 = new Plottable.Component();
+      component2.renderTo(svg2);
```
bluong: We don't need the component to be in a separate svg.  The original PanZoomInteraction would fail even if the components were in the same svg.  Let's just stick to 1 svg for simplicity.

test/interactions/panZoomInteractionTests.ts
```diff
@@ -181,6 +181,29 @@ describe("Interactions", () => {
...
+
+      var startPoint = { x: SVG_WIDTH / 4, y: SVG_HEIGHT / 2 };
+      var startPoint2 = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 2 };
+      TestMethods.triggerFakeTouchEvent( "touchstart", eventTarget, [startPoint, startPoint2], [0, 1] );
+
+      var endPoint = { x: SVG_WIDTH * 3 / 4, y: SVG_HEIGHT / 2 };
+      TestMethods.triggerFakeTouchEvent("touchmove", eventTarget, [endPoint], [1] );
+      TestMethods.triggerFakeTouchEvent("touchend", eventTarget, [endPoint], [1] );
+      assert.deepEqual(xScale.domain(), [SVG_WIDTH / 16, SVG_WIDTH * 5 / 16], "xScale inside target component transforms via pinch");
+      assert.deepEqual(xScale2.domain(), [0, SVG_WIDTH / 2], "xScale outside of target component does not transform via pinch");
```
bluong: Let's factor out this array into a variable that's used twice.

bluong:
Well the scenario I'm interested in is what happens if you start with your touches being outside the component space and then move them into the component space.

bluong:
There's a scenario to check as well as minor test fixes.

ztsai:
for that scenario, it won't affect the component

crmorford:
http://jsfiddle.net/w82shn5r/5/
vs
http://jsfiddle.net/w82shn5r/4/
looks good

crmorford:
Today something is happening with http://jsfiddle.net/w82shn5r/4/ that is confusing/concerning.
it has one yaxis, and 2 xaxes, and at first you can move them no problem. But for some reason, by the time you've panned/zoomed on all of them, they freeze up.

for example
1. pan on the yAxis for a while
2. pan on top xAxis for a while
3. pan on bottom xAxis for a while
4. try to pan anywhere, it won't work.

this works ~50% of the time :/ i don't have a foolproof repo

ztsai:
does that happen on develop?

crmorford:
Hm yeah. It takes a little longer, but it does freeze up eventually. I'll +1 this and we should look into that seperately.

PR #2561, opened 2015-08-04
TSLint file

aicioara:
All our lint files are dot prefixed. Getting tslint in line with those.


bluong:
What is very likely is that the tslint project https://github.com/palantir/tslint itself uses a tslint.json file and thus we have our filename mirror that.

Do we want to roll with our own configuration or just go with what they do?

aicioara:
I don't see a reason to mirror the name. It's normal for them to call it like that in an example because
1. It is a JSON file
2. It is a TSLINT file

The fact that they allow us to specify the file ourself gives us freedom to rename it to whatever.

bluong:
I think to keep in line with the rest of our checking configuration files, can we name the file
`.tslintrc`?

PR #2562, opened 2015-08-05
[TSLint] - enforce no-var-keyword on drawerTests

bluong:


test/drawers/drawerTests.ts
```diff
@@ -22,17 +21,17 @@ class MockAnimator implements Plottable.Animator {
...
 describe("Drawers", () => {
   describe("Abstract Drawer", () => {
-    var oldTimeout: any;
-    var timings: number[] = [];
-    var svg: d3.Selection<void>;
-    var drawer: Plottable.Drawer;
+    let oldTimeout: any;
+    let timings: number[] = [];
+    let svg: d3.Selection<void>;
+    let drawer: Plottable.Drawer;
```
aicioara: This is so weird. You declared it as a `const` above. That's why I don't like this change.

test/drawers/drawerTests.ts
```diff
@@ -22,17 +21,17 @@ class MockAnimator implements Plottable.Animator {
 }
 
 function createMockDrawer(dataset: Plottable.Dataset) {
-  var drawer = new Plottable.Drawer(dataset);
+  const drawer = new Plottable.Drawer(dataset);
```
aicioara: `let`!

test/drawers/drawerTests.ts
```diff
@@ -57,77 +56,77 @@ describe("Drawers", () => {
...
+      const selection = drawer.selectionForIndex(1);
       assert.strictEqual(selection.node(), circles[0][1], "correct selection gotten");
       svg.remove();
     });
 
     it("totalDrawTime()", () => {
-      var svg = TestMethods.generateSVG(300, 300);
-      var drawer = createMockDrawer(null);
+      const svg = TestMethods.generateSVG(300, 300);
+      const drawer = createMockDrawer(null);
```
aicioara: `let`

test/drawers/lineDrawerTests.ts
```diff
@@ -1,17 +1,16 @@
...
-      var linePlot = new Plottable.Plots.Line();
+      const svg = TestMethods.generateSVG(300, 300);
+      const data = [{a: 12, b: 10}, {a: 13, b: 24}, {a: 14, b: 21}, {a: 15, b: 14}];
+      const dataset = new Plottable.Dataset(data);
+      const xScale = new Plottable.Scales.Linear();
+      const yScale = new Plottable.Scales.Linear();
+      const linePlot = new Plottable.Plots.Line();
 
-      var drawer = new Plottable.Drawers.Line(dataset);
+      const drawer = new Plottable.Drawers.Line(dataset);
```
aicioara: `let`

aicioara:
Since we decided to use this, let's setup some rules of usage. 

I suggest variables should be `let` unless obviously `const`. It's not obvious to me that an `svg` or a `drawer` should be `const`

test/drawers/drawerTests.ts
bluong: We should be thinking of the scope of this variable.

The scope of this variable is only to the inside of the function, and thus the intention of this variable is to hold specifically a reference of a mock drawer that is created and returned.

test/drawers/drawerTests.ts
bluong: It's declared as `let` here because we do not know the value of this variable (it is not initialized to a value here)

The same variable name is declared as a `const` above because these two variables are fundamentally different variables.  The `drawer` variable in `createMockDrawer()` is used to hold a creation of a new mock drawer.  However, the `drawer` variable in this describe suite is used to switch between creations of new drawers to use across unit tests.

aicioara: But they refer to the same thing. You actually return the const and then you use it as a let

bluong: Note that the `let` and the `const` does not return to the object side of the expressions but specifically the reference side of the expressions.  Returning a `const` and then using it as a `let` is I think an incorrect interpretation of what's going on here.

If I were to name `createMockDrawer()`'s `drawer` variable to be `mockDrawer`, would this still run into the same issues?

test/drawers/drawerTests.ts
ztsai: but ```drawer``` is modified on line 25

aicioara: What is a `const` anyway? A constant reference to a variable or a reference to a constant variable?

bluong: `drawer` is modified on line 25 but the underlying object reference is still the same.

ztsai: yeah, now that I looked at some documents, it seems like it's a constant reference to a variable not the other way around

bluong: Let's say that I have the situation
```typescript
const foo = bar;
```
The `const` in the above only affects the usage of `foo`.  It does not care about what `bar` is (even if `bar` is mutable or `immutable`)

How does it affect the usage of `foo`?  Mainly, that `foo` cannot be switched to some different thing.  This makes it so that
```typescript
foo = baz;
```
is disallowed.  The switching portion is key here.

bluong: In regards to how the switching idea is key.  Let's say I have the below code
```typescript
let tester = foo;
tester([1, 3, 4]);
tester = bar;
tester([1, 3, 4]);
```

The semantics of the above would essentially read:
I have a variable tester that can be switched out to other testers to operate on some array.  Another way to word it is that essentially the value of `tester` changes drastically between lines.

However, what we can do is reduce the switching with `const`
```typescript
const fooTester = foo;
fooTester([1, 3, 4]);
const barTester = bar;
barTester([1, 3, 4]);
```

The semantics become a lot more clear, in my opinion.

aicioara: Let's take the following dummy suggestion

```typescript
const x = [];
x.push(2);
```

`x` changed fundamentally, yet it is not clear to the reader who assumes const gives safety. Hence, `const` does not give safety.

Recapping: using const:
- Gives no safety, even worse it misleads the reader into thinking that the state of the variable will not change
- Introduces mental overhead. Every time you have to think what will be the usage of the variable

Suggestion:
- Use `const` only for primitive types. Or in the cases where you would normally TYPE_THE_OBJECT_ALL_CAPS_CUZ_ITS_CONST

Aside:
- We are diverging. We were supposed to do this change only because `let` > `var` because of block scope. Now we are doing a completely different change.

aicioara:
Copying this here to make it easier to find

Let's take the following dummy suggestion

```typescript
const x = [];
x.push(2);
```

`x` changed fundamentally, yet it is not clear to the reader who assumes const gives safety. Hence, `const` does not give safety.

Recapping: using const:
- Gives no safety, even worse it misleads the reader into thinking that the state of the variable will not change
- Introduces mental overhead. Every time you have to think what will be the usage of the variable
- The benefit of `const` is *only* that it disallows variable reassignments. Well, the blocks are generally 20-30 lines long at most, you can see that yourself.

Suggestion:
- Use `const` only for primitive types. Or in the cases where you would normally TYPE_THE_OBJECT_ALL_CAPS_CUZ_ITS_CONST

Aside:
- We are diverging. We were supposed to do this change only because `let` > `var` because of block scope. Now we are doing a completely different change.

bluong:
The area of `const` resides in the same space as `let` though, and since the goal of these sets of pull requests is essentially to make `var` into something better, we should definitely consider using `const`.

In terms of no safety, the way I am interpreting this is that when you switch out a variable's value, a) you can be changing the entire definition of the variable or b) the variable's name and usage is broad/flexible enough to support the switching.  However, when I think about modifying an object, you still still have the same underlying object.  The reference is still the same.

In terms of mental overhead, this is something that is probably more better experienced than discussed.  The decision on having to make a variable prefaced by `let` or `const` is, from my limited experience, not much overhead at all.  Most (with probably around an 75% rate or higher) variables probably can be written as `const`'s and it seems to be that variables that need to be prefaced with `let` are very very often obvious.

bluong:
There is also a valid argument to be made that we should be actively making variables as tightly scoped.  I don't think it's too much trouble to upcast the variable to `let` status and even when we do, the trouble may be warranted as we should be rethinking the contract behind the variable.

ztsai:
I am not sure if it's just me but when I see ```const``` I would think it's a constant variable not a constant reference. Personally I would prefer not to mutate a ```const``` array/object as I think that's hard to read and I don't see much advantage of doing it

ztsai:
@jtlan what do you think?

PR #2564, opened 2015-08-05
[TSLint] - enforce no-var-keyword on dispatcherTests

bluong:


aicioara:
Is there a reason why we don't do a global search replace var -> let ?

bluong:
I'll do that in the next pull request maybe.  I might as well finish off this pull request

PR #2566, opened 2015-08-05
Root <svg> has pointer-events: visible.

### LINKED ISSUE #2454: Clicks/drags on background in Safari desktop don't register
This looks like it was introduced in the v1.2.0 release.

Go to this link in safari:
http://plottablejs.org/components/dragboxes/regular/

click and drag on a bar: works.
Click and drag starting on the background of the chart: does not work.

Can't repro in Firefox or Chrome.

Works fine in the v1.1.0 release.

jtlan:
Default property appears to be `"auto"` -- on Safari this doesn't behave correctly.

The root `<svg>` will now have `pointer-events` set to `"visible"`, which should be what we want:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pointer-events

Close #2454.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -284,7 +284,7 @@ module.exports = function(grunt) {
         testname: "Plottable Sauce Unit Tests",
         browsers: [{
           browserName: "firefox",
-          platform: "linux"
+          platform: "WIN7"
         }, {
           browserName: "chrome",
           platform: "linux"
--- a/plottable.css
+++ b/plottable.css
@@ -41,6 +41,7 @@
 
 svg.plottable {
   display : block; /* SVGs must be block elements for width/height calculations to work in Firefox. */
+  pointer-events: visible;
 }
 
 .plottable .background-fill {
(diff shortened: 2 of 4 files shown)


bluong:
Can you explain what you mean by the default property being "auto" and why this on Safari does not behave correctly?

jtlan:
On all browsers, `pointer-events` defaults to `"auto"`: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events
Of importance is this line describing `"auto"`:
>The element behaves as it would if the pointer-events property was not specified. In SVG content, this value and the value visiblePainted have the same effect.

And for `"visiblePainted"`:
>SVG only. The element can only be the target of a mouse event when the visibility property is set to visible and when the mouse cursor is over the interior (i.e., 'fill') of the element and the fill property is set to a value other than none, or when the mouse cursor is over the perimeter (i.e., 'stroke') of the element and the stroke property is set to a value other than none.

Possibly all the other browsers had it wrong; in either case, `pointer-events: "visible"` seems like the correct setting.

jtlan:
Closing for alternate attempt with cleaner history.

Issue #2076, opened 2015-05-15
Consistent way to check for absence of arguments.


jtlan:
Currently in our codebase we have a number of getter-setters:
```Typescript
public property(): Property; // getter
public property(property: Property): Class; // setter
```
... but we don't seem to have a consistent way to check for the absence of the argument to determine which version of the function to execute. The most common version is this:
```Typescript
if (property == null) {
  return this._property; // getter functionality
}
// setter logic
```
However, sometimes we do this instead:
```Typescript
if (property === undefined) {
  return this._property; // getter functionality
}
// setter logic
```
We should pick a convention and stick with it.

jtlan:
For reference, the Typescript compiler uses this mechanism when a default value for an argument is specified:
```Typescript
if (property === void 0) {
  ...
}
```
This has the advantage of not being susceptible to trolling [when someone redefines `undefined`](http://stackoverflow.com/questions/8783510/javascript-how-dangerous-is-it-really-to-assume-undefined-is-not-overwritten), but it seems like that only happens on non-ES5-compliant browsers.

aicioara:
I would still use double equal 
```typescript
if (property == void 0) {
  ...
}
```

That is because `null !== void 0`. And because sometimes we do things like 
```typescript
new Plottable.Interactions.PanZoom(null, yScale)
```

And also because we want to be backward compatible to users who decided to pass null instead of leaving the argument empty.

PR #2574, opened 2015-08-06
label.text() only accepts strings

### LINKED ISSUE #2140: Setting Label text to a non-string should throw a meaningful error
This works
```typescript
var label = new Plottable.Components.Label("1");
```

This does not
```typescript
var label = new Plottable.Components.Label(1);
```

TODO: Please document with current error message. 

Typescript disallows that indeed, but we might make it more friendly for our JS users.

**Edit** After debating, we decided that this will not be solved by stringifying the number, but instead throw a meaningful message. More information on #2442

crmorford:
fixes #2140 

Labels used to accept all types, now they will throw an error on any type that isn't a string.

### DIFF
--- a/test/components/labelTests.ts
+++ b/test/components/labelTests.ts
@@ -2,6 +2,12 @@
 
 describe("Labels", () => {
 
+  it("text() only accepts strings", () => {
+    assert.throws(() => new Plottable.Components.Label(<any> 23), Error);
+    assert.throws(() => new Plottable.Components.Label(<any> new Date()), Error);
+    assert.doesNotThrow(() => new Plottable.Components.Label("string"), Error, "accepts strings");
+  });
+
   it("Standard text title label generates properly", () => {
     var svg = TestMethods.generateSVG(400, 80);
     var label = new Plottable.Components.TitleLabel("A CHART TITLE");
--- a/test/components/labelTests.ts
+++ b/test/components/labelTests.ts
@@ -3,9 +3,10 @@
 describe("Labels", () => {
 
   it("text() only accepts strings", () => {
-    assert.throws(() => new Plottable.Components.Label(<any> 23), Error);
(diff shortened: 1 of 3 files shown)


test/components/labelTests.ts
```diff
@@ -2,6 +2,12 @@
 
 describe("Labels", () => {
 
+  it("text() only accepts strings", () => {
+    assert.throws(() => new Plottable.Components.Label(<any> 23), Error);
+    assert.throws(() => new Plottable.Components.Label(<any> new Date()), Error);
+    assert.doesNotThrow(() => new Plottable.Components.Label("string"), Error, "accepts strings");
```
ztsai: consider using a more descriptive message

ztsai:
plottable.js is not included
did you run grunt or grunt dev-compile?

crmorford:
haha nope.
do i run that right before i add and commit?

jtlan:
`label.text()` --> okay (acts as getter)
`label.text(undefined)` --> okay (acts as getter)
`label.text(null)` --> `Error` (acts as setter)

Does this make sense? Usually `null` and `undefined` are equivalent inputs to our methods.

bluong:
Relevant issue: https://github.com/palantir/plottable/issues/2076

I think we want to eventually move so that our argument checking works at the `void 0` level, which I believe is equivalent to `undefined`.  However, as it stands now, having the change in behavior is inconsistent with the behavior of the rest of the code.

If we want to set a precedent here, then we should continue with this idea and then roll the other endpoints appropriately.  Otherwise, we should have `null`/`undefined` work the same way.

jtlan:
I mean, the previous effect would have been this:
```Javascript
label.text(null); // weird error later on downstream
```

While not part of the original spec for the bug, I think defending against `null` input is in the spirit of the change (allowing only valid `string`s into `Label`).

jtlan:
GIL. Is it freaky Thursday?

PR #2576, opened 2015-08-06
trigger keypress callback only once when the key was pressed

### LINKED ISSUE #2499: onKeyPress is listening to keydown event
as mentioned in comment in #2393, onKeyPress is fired repeatedly while the key is pressed which is a bug
```onKeyPress()``` should be listening to ```keypress``` and we should add another api for ```keydown```

### PR DESCRIPTION
**Pull request by ztsai**
closes #2499 
demo: http://jsfiddle.net/sqa7dw6c/4/

EDIT: talked offline; since there's a use case requiring the functionality, will fix the auto-repeating behavior.  Will add ```onKeydown```/revise ```OnKeyPress``` if there's new feature requests

### DIFF
--- a/src/interactions/keyInteraction.ts
+++ b/src/interactions/keyInteraction.ts
@@ -16,7 +16,7 @@ export module Interactions {
 
     private _mouseMoveCallback = (point: Point) => false; // HACKHACK: registering a listener
     private _downedKeys = new Plottable.Utils.Set();
-    private _keyDownCallback = (keyCode: number) => this._handleKeyDownEvent(keyCode);
+    private _keyDownCallback = (keyCode: number, event: KeyboardEvent) => this._handleKeyDownEvent(keyCode, event);
     private _keyUpCallback = (keyCode: number) => this._handleKeyUpEvent(keyCode);
 
     protected _anchor(component: Component) {
@@ -41,9 +41,9 @@ export module Interactions {
       this._keyDispatcher = null;
     }
 
-    private _handleKeyDownEvent(keyCode: number) {
+    private _handleKeyDownEvent(keyCode: number, event: KeyboardEvent) {
       let p = this._translateToComponentSpace(this._positionDispatcher.lastMousePosition());
-      if (this._isInsideComponent(p)) {
+      if (this._isInsideComponent(p) && !event.repeat) {
         if (this._keyPressCallbacks[keyCode]) {
           this._keyPressCallbacks[keyCode].callCallbacks(keyCode);
         }
--- a/test/interactions/keyInteractionTests.ts
+++ b/test/interactions/keyInteractionTests.ts
@@ -196,6 +196,12 @@ describe("Interactions", () => {
         TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aCode);
         assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"a\" key was released");
 
+        TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode);
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aCode);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode, { repeat : true });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"a\" key was released");
         keyInteraction.offKeyRelease(aCode, aCallback);
         svg.remove();
       });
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by jtlan** [comment]
>I saw a note stating that `keypress` event isn't standard and it might have different behaviors across browsers/versions/platforms

Can you link the source? Is it [this](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)? That page suggests that the auto-repeat behavior depends on the platforming, meaning that `keyPress` might not be that helpful for everyone.

**Comment by jtlan** [comment]
Looking at the logic, it seems like our version of `keyPress` is case-insensitive since it relies on `keydown` events. Is that correct?

**Comment by ztsai** [comment]
I didn't save the source, but it's probably from [here](https://api.jquery.com/keypress/)

And yes, the current implementation is case-insensitive and I think that makes more sense as ```onKeyRelease``` and ```onKeyPress``` are often used in pairs; but we can iterate on that

I'll test the behaviors on different platforms, and send out another RFC
I will aslo check if ```keypress``` has ```repeat``` attribute in case we want to listen to ```keypress``` event for case-sensitivity but force it to not auto repeat in all browsers/platforms

File: src/interactions/keyInteraction.ts [hunk 1]
```diff
@@ -1,7 +1,7 @@
 ///<reference path="../reference.ts" />
 
 module Plottable {
-export type KeyCallback = (keyCode: number) => void;
+export type KeyCallback = (keyCode: number, event?: KeyboardEvent) => void;
```
jtlan: This is a bit odd -- most of our `Interaction` callbacks don't expose the low-level event.

ztsai:
is there a reason why we were not passing event back to callback?  or could we pass ```repeat```, ```shiftKey```, ```ctrlKey```, ```metakey```.. etc in an object?

jtlan:
> is there a reason why we were not passing event back to callback?

A lot of the Plottable `Interaction`s serve as abstractions to the actual events -- for example, there isn't actually a `Pointer` event, `Drag` and `Click` are actually particular sequences of mouse/touch events, etc. The goal is essentially to present the user with a flexible API for specifying responses to events, without dropping into the low-level details.

test/testMethods.ts
```diff
@@ -225,12 +225,18 @@ module TestMethods {
...
   }
 
-  export function triggerFakeKeyboardEvent(type: string, target: d3.Selection<void>, keyCode: number) {
+  export function triggerFakeKeyboardEvent(type: string, target: d3.Selection<void>, keyCode: number, options?: {[key: string]: any}) {
     let event = <KeyboardEvent> document.createEvent("Events");
     event.initEvent(type, true, true);
     event.keyCode = keyCode;
+    if (options != null ) {
+      for (var key in options) {
+        if (options.hasOwnProperty(key)) {
```
jtlan: Using `Object.keys()` will allow you to avoid having to do this check yourself:
```Typescript
Object.keys(options).forEach((key) => (<any> event)[key] = options[key] );
```

test/interactions/keyInteractionTests.ts
```diff
@@ -196,6 +196,12 @@ describe("Interactions", () => {
         TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aCode);
         assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"a\" key was released");
 
+        TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode);
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aCode);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode, { repeat : true });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when \"a\" key was released");
```
jtlan: It's not clear how this differs from the above sequence of fake events and `assert()`. I would pull this out into a separate test for the exact functionality that was implemented, which is that `KeyPress` callbacks are only triggered once, even if there are multiple keydown events. Pseudocode:
```
Mouse over
Key Down
Assert callback called
Key Down again
Assert callback not called a second time
```

test/interactions/keyInteractionTests.ts
```diff
@@ -116,6 +116,36 @@ describe("Interactions", () => {
...
+
+        aCallbackCalled = false;
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode, { repeat : true });
+        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when keydown was fired the second time");
+
+        keyInteraction.offKeyPress(aCode, aCallback);
+        svg.remove();
+      });
+
+      it("does not fire callback pressing the key when mouse was outside and moved inside", () => {
```
jtlan: "does not fire the callback if the key is pressed and held down outside, then the mouse moved inside"

PR #2579, opened 2015-08-07
Formatters provided with non-integer precision now throw an error

### LINKED ISSUE #2527: Audit all formatters
We are unsure what happens when we pass wrong values to `Plottable.Formatters`. 
- Decimal values 2.3
- Values bigger than 20
- 0
- Negative values.

Make sure we have consistency

aicioara:
Closes #2527

Previously setting a non-integer precision to `Formatters.general` or `Formatters.siSuffix` was silently failing. Now it throws an error.

The reason this PR exists is because the behavior was inconsistent across formatters, in particular, the `Formatters.fixed` was truncating the precision behind the scenes.


src/core/formatters.ts
```diff
@@ -60,7 +60,7 @@ export module Formatters {
    * @returns {Formatter} A formatter that displays exactly [precision] decimal places.
    */
   export function fixed(precision = 3) {
-    verifyPrecision(precision);
+    precision = verifyAndAdjustPrecision(precision);
```
bluong: My personal approach if I need to verify and then possibly modify the input precision would be to have the function be called `ensure`.  In this case, I would modify the function name to be `ensurePrecision`.

bluong:
Related to the topic of input sanitization and verification, I remember that the stance we took ultimately is to error if the user were to put in a bad input.  Do we want to take that stance here (would require a change on all the formatters)?  Or do we want to just seek consistency here?

aicioara:
Would that be a breaking change? If the users were relying on buggy behavior? 

If it is not, I totally agree with throwing an error.

src/core/formatters.ts
aicioara: sounds a bit weird. `ensurePrecision` sounds like if we put a precision of (say) `2.3`, we want to ensure that the precision is that, which is not true. We are actually amending the precision.

However, this is dependent on my comment bellow. If we decide to just throw a warning, this can be renamed back to `verifyPrecision()`

bluong:
Well it would not be an API breaking change, so I believe we are still within the bounds of a minor version bump.  We should do a small discussion regarding if we want to switch to a more strict checking model (implementation should hopefully be straightforward)

src/core/formatters.ts
bluong: Agree with second comment, and I think I agree with the first comment.

If we do go with essentially a sanitizing idea, maybe `sanitizePrecision()`?  It does not signal well if the input were to error though.

aicioara: I like `sanitizePrecision()`

aicioara:
So what is your decision on this? Should we throw an error? Or go with silent conversion 

I like the `sanitizePrecision` idea

bluong:
We should hold a small discussion regarding consistency vs continuing with a known precedent.  I personally like trying to go forward with the precedent, but it may be disruptive to some folks.

aicioara:
Made the rename suggested. Regarding throwing an error, we should hold a small discussion. 

TL;DR: When a user specifies a non-integer precision (like 2.8) for a formatter, should we truncate the value (to 2) or throw an error?

aicioara:
Following an offline discussion, we are no longer respecting the JavaScript ecosystem and instead throwing an error for non-integer precision

src/core/formatters.ts
```diff
@@ -269,6 +268,10 @@ export module Formatters {
     if (precision < 0 || precision > 20) {
       throw new RangeError("Formatter precision must be between 0 and 20");
     }
+
+    if (precision !== Math.floor(precision)) {
+      throw new RangeError("Formatter precision must be an integer number");
```
bluong: "integer number" is redundant. "integer" is enough.

crmorford:
double checking: i should make sure that for *all* formatters, non-integer precisions cause errors?

aicioara:
that is correct. 

Not only non-integer, but also precision greater than 20 or less than 0 should throw error for all formatters. Precision of 0 is not part of this PR, but if you spot something weird with that, please flag. Also, all integers between 0 and 20 should behave correctly and consistently.

crmorford:
[ ] identity (no precision)
[ ] general
[ ] fixed
[ ] currency
[ ] percentage
[ ]  SI
[ ] shortscale

crmorford:
http://jsfiddle.net/1h335t30/12/

PR #2583, opened 2015-08-07
[Typings] - update chai typings

bluong:
Ran into an issue where the `chai-assert.d.ts` file was declaring a variable that we were actually overriding.  Updating the typings appears to have removed that variable.


bluong:
Not this is not necessarily required.  I found a workaround.  We should probably still do this at one point, but its not a prioirty.

PR #2584, opened 2015-08-07
Var to let

bluong:
Except for one place, this is essentially find-replace on var to let.

The one place is just that two for loops were using the same variable.  So, I just redeclared it using let.


jtlan:
Can you comment in the spot where the code was changed? And in what way was this broken before?

src/axes/timeAxis.ts
```diff
@@ -483,7 +482,7 @@ export module Axes {
       this._renderLabellessTickMarks(labelLessTicks);
 
       this._hideOverflowingTiers();
-      for (i = 0; i < tierConfigs.length; ++i) {
```
bluong: A modification was needed here.

The reason was because `i` was defined in the above for loop, which means that this `i` has lost scope of the previous scope after moving to `let`

jtlan: I see. Previously it was redeclared due to a scoping conflict, so `let` resolves that.

jtlan:
Just a regression pass, I guess?

bluong:
Still side branch stuff.

PR #2590, opened 2015-08-10
[HOTFIX] detach() callbacks should called even if not anchored()

jtlan:
`detach()` callbacks should be called whenever `detach()` is called, to make sure that the `Component` and its parent correctly disassociate.

#### Testing Notes
The bug manifested when `detach()`-ing `Component`s that were in a `Group` or `Table` before the `Group` or `Table` was rendered.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -3292,9 +3292,9 @@ var Plottable;
             this.parent(null);
             if (this._isAnchored) {
                 this._element.remove();
-                this._isAnchored = false;
-                this._onDetachCallbacks.callCallbacks(this);
             }
+            this._isAnchored = false;
+            this._onDetachCallbacks.callCallbacks(this);
             return this;
         };
         /**
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -450,9 +450,9 @@ export class Component {
 
     if (this._isAnchored) {
       this._element.remove();
-      this._isAnchored = false;
-      this._onDetachCallbacks.callCallbacks(this);
     }
+    this._isAnchored = false;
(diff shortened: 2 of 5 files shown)


aicioara:
Can we make a followup ticket where we address the problem properly? It feels weird to me to call `detach` callbacks when no actual detaching is performed.

aicioara:
Can we add a test which covers this case

```typescript
c.onDetach(callback);

c.detach();
assert.isTrue(callbackWasCalled);

callbackWasCalled = false

c.detach();
assert.isTrue(callbackWasCalled);

```

jtlan:
What is the suggested test case supposed to check for?

aicioara:
I am personally unsure what is going to happen in that scenario. I don't see anything covering the scenario

crmorford:
tables: destroy and detach http://jsfiddle.net/rzspuLvx/
groups: destroy and detach http://jsfiddle.net/rzspuLvx/1/
detach then add to group http://jsfiddle.net/rzspuLvx/2/
detach then add to table http://jsfiddle.net/rzspuLvx/3/
destroy then add to group http://jsfiddle.net/rzspuLvx/4/ (FAILS -- expected)
destroy then add to table http://jsfiddle.net/rzspuLvx/5/ (FAILS -- expected)

crmorford:
detach after render http://jsfiddle.net/rzspuLvx/6/
detach and reattach, after render http://jsfiddle.net/rzspuLvx/7/
destroy after render http://jsfiddle.net/rzspuLvx/8/
destroy and reattach, after render http://jsfiddle.net/rzspuLvx/9/ (FAILS -- expected)

Issue #1640, opened 2015-02-17
Slider (FR)


crmorford:
Create a slider that can be dragged along an axis. Slate wants this to normalize time axes.

jtlan:
Probably two parts to this:
* A "flag" that sits on the axis
* A line that is drawn on the chart itself

crmorford:
* Flag shape & size approved by designer
* Flag looks good on small charts and large charts
* 'Slider' line still 1) takes up all of plot when plot height increases AND 2) doesn't go out of bounds if plot height decreases
* 'Slider' line moves in sync with flag
* Flag can be grabbed and dragged
* Slider can be grabbed and dragged (? discuss)
* Two slider+flag can be placed on the same axis
* 2 Sliders + flags can be placed on x and y axis 
* Slider callback that gives pixel position is accurate on Time/Category/Numeric

PR #2593, opened 2015-08-10
Add GuideLineLayer

jtlan:
Added new class `GuideLineLayer`. This `Component` fully-occupies a cell in a `Table` (like `SelectionBoxLayer`), and draws either a vertical or horizontal line at a specified position in data- or pixel-space:

`DragLineLayer` will extend this class. This PR serves as step one towards #1640.

screen shot 2015-08-10 at 2 58 24 pm

The class is generic on the type of value it would take in data-space:
```Typescript
class GuideLineLayer<D> extends Component {
  ...
}
```
`D` would be `number` if the `value()` of the line is a `number`, for example; it should also match the type of input to the `QuantitativeScale` used with the `GuideLineLayer`. `void` should be used if the `GuideLineLayer` is only used in pixel-space.

Features the following API points:
```Typescript
constructor(orientation: string); // "vertical" or "horizontal"

/**
 * Gets the QuantitativeScale on the GuideLineLayer.
 */
scale(): QuantitativeScale<D>;
/**
 * Sets the QuantitativeScale on the GuideLineLayer.
 * If value() was the last property set, pixelPosition() will be updated according to the new scale.
 * If pixelPosition() was the last property set, value() will be updated according to the new scale.
 */
scale(scale: QuantitativeScale<D>): GuideLineLayer<D>;

/**
 * Gets the value of the guide line in data-space.
 */
value(): D;
/**
 * Sets the value of the guide line in data-space.
 * If the GuideLineLayer has a scale, pixelPosition() will be updated now and whenever the scale updates.
 */
value(value: D): GuideLineLayer<D>;

/**
 * Gets the position of the guide line in pixel-space.
 */
pixelPosition(): number;
/**
 * Sets the position of the guide line in pixel-space.
 * If the GuideLineLayer has a scale, the value() will be updated now and whenever the scale updates.
 */
pixelPosition(pixelPosition: number): GuideLineLayer<D>;
```

CSS styles should be applied to the following element to style the line:
```CSS
.plottable .guide-line-layer line.guide-line {
  ...
}
```

Demo: http://jsfiddle.net/fza2vdcc/

### DIFF
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -10,7 +10,10 @@ export module Components {
     private _value: D;
     private _scale: QuantitativeScale<D>;
     private _pixelPosition: number;
-    private _scaleUpdateCallback = () => this._updatePixelPosition();
+    private _scaleUpdateCallback = () => {
+      this._updatePixelPosition();
+      this.render();
+    };
     private _line: d3.Selection<void>;
 
     constructor(orientation: string) {
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -10,10 +10,7 @@ export module Components {
     private _value: D;
     private _scale: QuantitativeScale<D>;
     private _pixelPosition: number;
-    private _scaleUpdateCallback = () => {
-      this._updatePixelPosition();
-      this.render();
-    };
+    private _scaleUpdateCallback: ScaleCallback<QuantitativeScale<D>>;
     private _line: d3.Selection<void>;
 
     constructor(orientation: string) {
@@ -24,6 +21,10 @@ export module Components {
       this._orientation = orientation;
       this._clipPathEnabled = true;
       this.addClass("guide-line-layer");
+      this._scaleUpdateCallback = () => {
+        this._updatePixelPosition();
+        this.render();
+      };
     }
 
     protected _setup() {
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -22,7 +22,7 @@ export module Components {
       this._clipPathEnabled = true;
       this.addClass("guide-line-layer");
       this._scaleUpdateCallback = () => {
-        this._updatePixelPosition();
+        this._syncValueAndPixelPosition();
         this.render();
       };
     }
@@ -53,7 +53,7 @@ export module Components {
 
     public renderImmediately() {
       super.renderImmediately();
-      this._updatePixelPosition();
+      this._syncValueAndPixelPosition();
       this._guideLine.attr({
         x1: this._isVertical() ? this.pixelPosition() : 0,
         y1: this._isVertical() ? 0 : this.pixelPosition(),
@@ -63,9 +63,13 @@ export module Components {
       return this;
     }
 
-    private _updatePixelPosition() {
-      if (this.scale() != null && this.value() != null) {
-        this._pixelPosition = this.scale().scale(this.value());
+    private _syncValueAndPixelPosition() {
+      if (this.scale() != null) {
+        if (this.value() != null) {
+          this._pixelPosition = this.scale().scale(this.value());
+        } else if (this.pixelPosition() != null) {
+          this._value = this.scale().invert(this.pixelPosition());
+        }
       }
     }
 
@@ -93,7 +97,7 @@ export module Components {
       }
       this._scale = scale;
       this._scale.onUpdate(this._scaleUpdateCallback);
-      this._updatePixelPosition();
+      this._syncValueAndPixelPosition();
       this.render();
       return this;
     }
@@ -117,7 +121,7 @@ export module Components {
         return this._value;
       }
       this._value = value;
-      this._updatePixelPosition();
+      this._syncValueAndPixelPosition();
       this.render();
       return this;
     }
@@ -141,9 +145,7 @@ export module Components {
         return this._pixelPosition;
       }
       this._pixelPosition = pixelPosition;
-      if (this.scale() != null) {
-        this._value = this.scale().invert(pixelPosition);
-      }
+      this._syncValueAndPixelPosition();
       this.render();
       return this;
     }
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -1,0 +1,30 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Components {
+  export class GuideLineLayer<D> extends Component {
+
+    constructor(orientation: string) {
+      super();
+    }
+
+    public scale(): QuantitativeScale<D>;
+    public scale(scale: QuantitativeScale<D>): GuideLineLayer<D>;
+    public scale(scale?: QuantitativeScale<D>): any {
+      return;
+    }
+
+    public value(): D;
+    public value(value: D): GuideLineLayer<D>;
+    public value(value?: D): any {
+      return;
+    }
+
+    public pixelPosition(): number;
+    public pixelPosition(pixelPosition: number): GuideLineLayer<D>;
+    public pixelPosition(pixelPosition?: number): any {
+      return;
+    }
+  }
+}
+}
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -22,7 +22,7 @@ export module Components {
       this._clipPathEnabled = true;
       this.addClass("guide-line-layer");
       this._scaleUpdateCallback = () => {
-        this._syncValueAndPixelPosition();
+        this._setPixelPositionFromValue();
         this.render();
       };
     }
@@ -65,7 +65,7 @@ export module Components {
 
     public renderImmediately() {
       super.renderImmediately();
-      this._syncValueAndPixelPosition();
+      this._setPixelPositionFromValue();
       this._guideLine.attr({
         x1: this._isVertical() ? this.pixelPosition() : 0,
         y1: this._isVertical() ? 0 : this.pixelPosition(),
@@ -75,13 +75,15 @@ export module Components {
       return this;
     }
 
-    private _syncValueAndPixelPosition() {
-      if (this.scale() != null) {
-        if (this.value() != null) {
-          this._pixelPosition = this.scale().scale(this.value());
-        } else if (this.pixelPosition() != null) {
-          this._value = this.scale().invert(this.pixelPosition());
-        }
+    private _setPixelPositionFromValue() {
+      if (this.scale() != null && this.value() != null) {
+        this._pixelPosition = this.scale().scale(this.value());
+      }
+    }
+
+    private _setValueFromPixelPosition() {
+      if (this.scale() != null && this.pixelPosition() != null) {
+        this._value = this.scale().invert(this.pixelPosition());
       }
     }
 
@@ -110,7 +112,8 @@ export module Components {
       }
       this._scale = scale;
       this._scale.onUpdate(this._scaleUpdateCallback);
-      this._syncValueAndPixelPosition();
+      this._setPixelPositionFromValue();
+      this._setValueFromPixelPosition();
       this.redraw();
       return this;
     }
@@ -134,7 +137,7 @@ export module Components {
         return this._value;
       }
       this._value = value;
-      this._syncValueAndPixelPosition();
+      this._setPixelPositionFromValue();
       this.render();
       return this;
     }
@@ -158,7 +161,7 @@ export module Components {
         return this._pixelPosition;
       }
       this._pixelPosition = pixelPosition;
-      this._syncValueAndPixelPosition();
+      this._setValueFromPixelPosition();
       this.render();
       return this;
     }
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -51,6 +51,18 @@ export module Components {
       return true;
     }
 
+    public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
+      super.computeLayout(origin, availableWidth, availableHeight);
+      if (this.scale() != null) {
+        if (this._isVertical()) {
+          this.scale().range([0, this.width()]);
+        } else {
+          this.scale().range([this.height(), 0]);
+        }
+      }
+      return this;
+    }
+
     public renderImmediately() {
       super.renderImmediately();
       this._syncValueAndPixelPosition();
@@ -99,7 +111,7 @@ export module Components {
       this._scale = scale;
       this._scale.onUpdate(this._scaleUpdateCallback);
       this._syncValueAndPixelPosition();
-      this.render();
+      this.redraw();
       return this;
     }
 
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -162,6 +162,9 @@ export module Components {
       if (pixelPosition == null) {
         return this._pixelPosition;
       }
+      if (!Utils.Math.isValidNumber(pixelPosition)) {
+        throw new Error("pixelPosition must be a finite number");
+      }
       this._pixelPosition = pixelPosition;
       this._mode = PropertyMode.PIXEL;
       this._syncPixelPositionAndValue();
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -82,6 +82,7 @@ export module Components {
     /**
      * Sets the QuantitativeScale on the GuideLineLayer.
      * If value() has been set, pixelPosition() will be updated according to the new scale.
+     * Otherwise, if pixelPosition() has been set but value() has not, value() will be set.
      *
      * @param {QuantitativeScale<D>} scale
      * @return {GuideLineLayer<D>} The calling GuideLineLayer.
@@ -103,13 +104,13 @@ export module Components {
     }
 
     /**
-     * Gets the position of the guide line in data-space.
+     * Gets the value of the guide line in data-space.
      *
      * @return {D}
      */
     public value(): D;
     /**
-     * Sets the position of the guide line in data-space.
+     * Sets the value of the guide line in data-space.
      * If the GuideLineLayer has a scale, pixelPosition() will be updated.
      *
      * @param {D} value
--- a/src/components/guideLineLayer.ts
+++ b/src/components/guideLineLayer.ts
@@ -11,7 +11,7 @@ export module Components {
     private _scale: QuantitativeScale<D>;
     private _pixelPosition: number;
     private _scaleUpdateCallback: ScaleCallback<QuantitativeScale<D>>;
-    private _line: d3.Selection<void>;
+    private _guideLine: d3.Selection<void>;
 
     constructor(orientation: string) {
       super();
@@ -29,7 +29,7 @@ export module Components {
 
     protected _setup() {
       super._setup();
-      this._line = this.content().append("line").classed("guide-line", true);
+      this._guideLine = this.content().append("line").classed("guide-line", true);
     }
 
     protected _sizeFromOffer(availableWidth: number, availableHeight: number) {
@@ -54,7 +54,7 @@ export module Components {
     public renderImmediately() {
       super.renderImmediately();
       this._updatePixelPosition();
-      this._line.attr({
+      this._guideLine.attr({
         x1: this._isVertical() ? this.pixelPosition() : 0,
         y1: this._isVertical() ? 0 : this.pixelPosition(),
         x2: this._isVertical() ? this.pixelPosition() : this.width(),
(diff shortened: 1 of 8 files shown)


jtlan:
Found a bug. Doesn't update on scale domain change...

jtlan:
By the way, neither `SelectionBoxLayer`/`DragBoxLayer` nor `GuideLineLayer` renders correctly on their own if given a data-space position, for the same reason as #2532.

bluong:
I am personally a fan of the consistent model of Components being able to render independently.  It'd allow for some easier testing and more flexibility

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    private _value: D;
+    private _scale: QuantitativeScale<D>;
+    private _pixelPosition: number;
+    private _scaleUpdateCallback = () => {
+      this._updatePixelPosition();
+      this.render();
+    };
+    private _line: d3.Selection<void>;
+
+    constructor(orientation: string) {
```
bluong: Supposedly you can type the constructors like the below:

```typescript
constructor(orientation: "vertical");
constructor(orientation: "horizontal");
constructor(orientation: string);
```

Note the fact that the `string` version is still forced to appear.  Is this preferable?  Precedent set by `linePlot.interpolator()`

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    public static ORIENTATION_HORIZONTAL = "horizontal";
+
+    private _orientation: string;
+    private _value: D;
+    private _scale: QuantitativeScale<D>;
+    private _pixelPosition: number;
+    private _scaleUpdateCallback = () => {
+      this._updatePixelPosition();
+      this.render();
+    };
```
bluong: I'm personally surprised they allow code blocks like this in instance variable initialization.

So far I don't have an opinion on this

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    private _line: d3.Selection<void>;
```
bluong: name this `_guideLine`?

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+      return this._orientation === GuideLineLayer.ORIENTATION_VERTICAL;
+    }
+
+    public fixedWidth() {
+      return true;
+    }
+
+    public fixedHeight() {
+      return true;
+    }
```
bluong: Why does this have a `fixedWidth()` and a `fixedHeight()`?  I see it on `SelectionBoxLayer` and I sorta remember we needed this for a reason, but I forget why

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    public renderImmediately() {
+      super.renderImmediately();
+      this._updatePixelPosition();
```
bluong: Just curious why this updating needs to be done before rendering

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Gets the QuantitativeScale on the GuideLineLayer.
+     *
+     * @return {QuantitativeScale<D>}
+     */
+    public scale(): QuantitativeScale<D>;
+    /**
+     * Sets the QuantitativeScale on the GuideLineLayer.
+     * If the value of the guide line in data-space has been set,
+     * the position of the guide line in pixel-space will be updated.
```
bluong: Let's say that the position is already set, should the value be updated?

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Gets the position of the guide line in data-space.
+     *
+     * @return {D}
+     */
+    public value(): D;
+    /**
+     * Sets the position of the guide line in data-space.
+     * If the GuideLineLayer has a scale, the position of the line
+     * in pixel-space will be updated accordingly.
```
bluong: Do we want to go into technical detail here?

Would
```typescript
If the GuideLineLayer has a scale, then position() will be updated accordingly.
```

work better?

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Gets the position of the guide line in pixel-space.
+     *
+     * @return {number}
+     */
+    public pixelPosition(): number;
+    /**
+     * Sets the position of the guide line in pixel-space.
+     * If the GuideLineLayer has a scale, the position of the line
+     * in data-space will be updated accordingly.
```
bluong: Same comments that maybe just saying
```typescript
value() will be updated accordingly.
```

might be easier to understand.

test/components/guideLineLayerTests.ts
```diff
@@ -0,0 +1,378 @@
...
+    assert.strictEqual(gll.value(), expectedValue, "getter returns the set value");
+  });
+
+  it("pixelPosition()", () => {
+    let gll = new Plottable.Components.GuideLineLayer<number>("vertical");
+    let expectedPosition = 5;
+    assert.isUndefined(gll.pixelPosition(), "returns undefined before any pixel position is set");
+    assert.strictEqual(gll.pixelPosition(expectedPosition), gll, "setter returns the calling GuideLineLayer");
+    assert.strictEqual(gll.pixelPosition(), expectedPosition, "getter returns the set pixel position");
+  });
```
bluong: Do we allow `pixelPosition` and `value` to be set independently even with no backing scale?

src/components/guideLineLayer.ts
jtlan: No additional type safety is provided by the additional signatures, so I'm not sure how that helps. I should add documentation to the constructor though.

bluong:
Curious on the coordination behavior `scale` `value` and `position`

src/components/guideLineLayer.ts
jtlan: I can move it to the `constructor`.

src/components/guideLineLayer.ts
jtlan: Sure.

src/components/guideLineLayer.ts
jtlan: It's so that it doesn't ask for extra space in computations -- otherwise it would expand the height taken up by any `Axis` it was grouped with, for example.

src/components/guideLineLayer.ts
jtlan: So it accounts for any changes to the `range()` of the `QuantitativeScale`.

src/components/guideLineLayer.ts
jtlan: Yes, to make sure that the pixel-space and data-space locations are in agreement with regards to the scale.

src/components/guideLineLayer.ts
jtlan: `pixelPosition()`? I'm trying to describe the behavior rather than the API points, but maybe that would be clearer.

test/components/guideLineLayerTests.ts
jtlan: There's no translation between them with no scale, so yes. `value()` takes precedence though, so adding a scale will cause `pixelPosition()` to be updated to match the scaled `value()`.

src/components/guideLineLayer.ts
bluong: `Gridlines` probably has to follow this example also.  I'll mark that.

src/components/guideLineLayer.ts
bluong: Implementation currently only has the pixel-space position being set.  Let's modify the data-space value accordingly then?

src/components/guideLineLayer.ts
bluong: Describing the behavior instead of API points is perfectly fine; I just think the language of data-space and pixel-space could be possibly cryptic.

test/components/guideLineLayerTests.ts
bluong: Not sure how much of a big deal this is, but should be get a 3rd opinion on if `value` and `pixelPosition` can be set without a scale

src/components/guideLineLayer.ts
jtlan: No, data-space `value()` takes precedence. So `pixelPosition()` gets adjusted to match.

src/components/guideLineLayer.ts
jtlan: It's probably less important for `Gridlines`, since I don't think it'll ever be combined with `Axis`.

jtlan:
@jtlan wrote:
>By the way, neither `SelectionBoxLayer`/`DragBoxLayer` nor `GuideLineLayer` renders correctly on their own if given a data-space position, for the same reason as #2532. Should that be considered a bug?

I'm going to backtrack on this. Setting `range()` in a lot of places increases the odds of an emergent bug because now we have to make sure the logic is the same in all places.

bluong:
> @jtlan wrote:
> 
>     By the way, neither SelectionBoxLayer/DragBoxLayer nor GuideLineLayer renders correctly on their own if given a data-space position, for the same reason as #2532. Should that be considered a bug?
> 
> I'm going to backtrack on this. Setting range() in a lot of places increases the odds of an emergent bug because now we have to make sure the logic is the same in all places.

Currently, it seems that our `XYPlot`s and components that utilize scales to render in a data-based style do this scale configuration, mainly because so that our users don't have to do some needless configuration.  I don't see why we need to have users configure this component when we know how the range is going to be used pretty consistently.

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Gets the QuantitativeScale on the GuideLineLayer.
+     *
+     * @return {QuantitativeScale<D>}
+     */
+    public scale(): QuantitativeScale<D>;
+    /**
+     * Sets the QuantitativeScale on the GuideLineLayer.
+     * If value() has been set, pixelPosition() will be updated according to the new scale.
```
bluong: Let's add a message regarding that `value()` will be updated

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+      this._scale = scale;
+      this._scale.onUpdate(this._scaleUpdateCallback);
+      this._syncValueAndPixelPosition();
+      this.render();
+      return this;
+    }
+
+    /**
+     * Gets the position of the guide line in data-space.
```
bluong: `Gets the value...` ?

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * @return {D}
+     */
+    public value(): D;
+    /**
+     * Sets the position of the guide line in data-space.
```
bluong: `Sets the value...`

test/components/guideLineLayerTests.ts
```diff
@@ -0,0 +1,401 @@
...
+
+      let gll = new Plottable.Components.GuideLineLayer<number>("vertical");
+      let value = 0.5;
+      gll.value(value);
+      let setPosition = -999;
+      gll.pixelPosition(setPosition);
+      let expectedPosition = linearScale.scale(value);
+      gll.scale(linearScale);
+      assert.strictEqual(gll.pixelPosition(), expectedPosition, "setting the scale updates the pixel position");
+      assert.notStrictEqual(gll.pixelPosition(), setPosition, "originally-set pixel position was overridden");
```
bluong: The above two tests are a bit redundant and would also fail if `expectedPosition === setPosition`.  Can we rewrite this small portion?

test/components/guideLineLayerTests.ts
```diff
@@ -0,0 +1,401 @@
...
+      assert.isTrue(gll.fixedWidth(), "fixed width");
+      assert.isTrue(gll.fixedHeight(), "fixed height");
+
+      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      gll.anchor(svg);
+      gll.computeLayout({x: 0, y: 0}, SVG_WIDTH, SVG_HEIGHT);
+      assert.strictEqual(gll.width(), SVG_WIDTH, "accepted all offered width");
+      assert.strictEqual(gll.height(), SVG_HEIGHT, "accepted all offered height");
+      svg.remove();
+    });
```
bluong: Is this essentially 2 different tests?

test/components/guideLineLayerTests.ts
```diff
@@ -0,0 +1,401 @@
...
+  describe("rendering (vertical)", () => {
+    let SVG_WIDTH = 400;
+    let SVG_HEIGHT = 300;
+
+    it("has an effective size of 0, but will occupy all offered space", () => {
+      let gll = new Plottable.Components.GuideLineLayer<void>("vertical");
+      let request = gll.requestedSpace(SVG_WIDTH, SVG_HEIGHT);
+      TestMethods.verifySpaceRequest(request, 0, 0, "does not request any space");
+      assert.isTrue(gll.fixedWidth(), "fixed width");
+      assert.isTrue(gll.fixedHeight(), "fixed height");
```
bluong: This seems to test more that GLL requests 0 space, rather than having an effective size of 0.

test/components/guideLineLayerTests.ts
```diff
@@ -0,0 +1,401 @@
...
+    });
+
+    it("renders correctly given a pixel position", () => {
+      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      let expectedPosition1 = SVG_WIDTH / 2;
+      let gll = new Plottable.Components.GuideLineLayer<number>("vertical");
+      gll.pixelPosition(expectedPosition1);
+      gll.renderTo(svg);
+
+      assert.strictEqual(gll.content().selectAll("line").size(), 1, "exactly one line is drawn");
```
bluong: For this and the rest of the tests, can we use the more specific ".guide-line" identifier?

bluong:
Test semantics and standalone

test/components/guideLineLayerTests.ts
jtlan: If `expectedPosition === setPosition`, then the test would pass whether or not this functionality worked. It makes sense to error in that case to guard against changes to the code, right?

test/components/guideLineLayerTests.ts
jtlan: No, since the combination of "space-accepted" and "space used" is critical to the `GuideLineLayer`'s behavior during layout; therefore it makes sense to test them together.

test/components/guideLineLayerTests.ts
jtlan: I'll rename the test.

test/components/guideLineLayerTests.ts
bluong: I'm not sure about erroring, but yes with your current implementation, the only way for this test to be valid is to have a place that states the test is invalid if `expectedPosition === setPosition`.  I was thinking of using an `assert` statement.  However, I'm not 100% sure.  If you have a good reason to use an error, then let's use that.

jtlan: Logically, the test test derives from the following logic:
```
Set a position.
Set a value.
Set a scale.
The originally-set position gets overridden by the scaled value.
```
The last statement essentially translates into two asserts:
1. The old value was overridden.
2. The new value was set.

jtlan:
Found a bug :/

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private _syncValueAndPixelPosition() {
+      if (this.scale() != null) {
+        if (this.value() != null) {
+          this._pixelPosition = this.scale().scale(this.value());
+        } else if (this.pixelPosition() != null) {
+          this._value = this.scale().invert(this.pixelPosition());
+        }
+      }
```
bluong: You can avoid some of the nesting here by having
```typescript
if (this.scale() == null) {
  return;
}
```

It just makes the following code easier to read

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,174 @@
...
+      });
+      return this;
+    }
+
+    private _syncValueAndPixelPosition() {
+      if (this.scale() != null) {
+        if (this.value() != null) {
+          this._pixelPosition = this.scale().scale(this.value());
+        } else if (this.pixelPosition() != null) {
+          this._value = this.scale().invert(this.pixelPosition());
```
bluong: You may want to separate out the coordination above into 2 separate functions

jtlan: Just about to do that.

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    private _setPixelPositionFromValue() {
+      if (this.scale() != null && this.value() != null) {
+        this._pixelPosition = this.scale().scale(this.value());
+      }
+    }
+
+    private _setValueFromPixelPosition() {
+      if(this.scale() != null && this.pixelPosition() != null) {
```
bluong: surprised tslint didn't catch this (maybe I need to relook at the rules), please put a space after `if`

crmorford:
set a scale
set a pixel position (100)
call GuideLineLayer.value()
render plot
call GuideLineLayer.value() 

expected: line has pixelposition 100, like i set
actual: line isn't on graph. it's at the correct pixel position if value == 100

http://jsfiddle.net/1734wvs5/2/

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,179 @@
...
+    public static ORIENTATION_VERTICAL = "vertical";
+    public static ORIENTATION_HORIZONTAL = "horizontal";
+
+    private _orientation: string;
+    private _value: D;
+    private _scale: QuantitativeScale<D>;
+    private _pixelPosition: number;
+    private _scaleUpdateCallback: ScaleCallback<QuantitativeScale<D>>;
+    private _guideLine: d3.Selection<void>;
+    private _primaryProperty = "value";
```
bluong: Let's use an enum instead of a string for this

bluong: We should probably name the instance variable.  Note that your commit messages make some instance of a "mode", maybe "propertyMode"?

src/components/guideLineLayer.ts
```diff
@@ -0,0 +1,180 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Components {
+  enum PropertyMode { value, pixel };
```
bluong: To my understanding, enum's are not variable, and thus so to not confuse them with other variables, they should not be undercased

crmorford:
Can we get a clearer error for bad input to pixelPosition?
http://jsfiddle.net/1734wvs5/13/

screen shot 2015-08-14 at 11 35 23 am

screen shot 2015-08-14 at 11 35 31 am

crmorford:
No error when value doesn't match scale 
example: .scale(time) & .value(number) http://jsfiddle.net/1734wvs5/18/
or vice versa http://jsfiddle.net/1734wvs5/19/

crmorford:
**before/after render tests**
set scale, set value, get pixelpos b/a renderhttp://jsfiddle.net/1734wvs5/1/
set scale, set pixelpos, get value b/a render http://jsfiddle.net/1734wvs5/2/
update scale b/a render http://jsfiddle.net/1734wvs5/4/
update value b/a render http://jsfiddle.net/1734wvs5/5/
update pixelpos b/a render http://jsfiddle.net/1734wvs5/6/
destroy() b/a render http://jsfiddle.net/1734wvs5/7/ and http://jsfiddle.net/1734wvs5/8/

**try to break**
put in group with legend http://jsfiddle.net/1734wvs5/24/
set pixelpos out of svg http://jsfiddle.net/1734wvs5/9/
set value/scale, set conflicting pixelpos http://jsfiddle.net/1734wvs5/10/
set pixelpos, set conflicting value/scale http://jsfiddle.net/1734wvs5/11/
render alone http://jsfiddle.net/1734wvs5/12/
--give conflicting scale type and value type http://jsfiddle.net/1734wvs5/18/ and http://jsfiddle.net/1734wvs5/19/
--give pixelpos NaN http://jsfiddle.net/1734wvs5/13/
nonsense word in constructor http://jsfiddle.net/1734wvs5/14/
no value for constructor http://jsfiddle.net/1734wvs5/15/
don't set scale/val/pixelpos before render http://jsfiddle.net/1734wvs5/16/

**real life*
multiple gll (crosshair) http://jsfiddle.net/1734wvs5/20/
have gll follow pointer http://jsfiddle.net/1734wvs5/20/
click on axis, plot to move gll (test on ipad) http://jsfiddle.net/1734wvs5/20/
panzoom http://jsfiddle.net/1734wvs5/21/, http://jsfiddle.net/1734wvs5/22/ locks on screen if called pixelPosition last, moves with panzoom if value called last. cool!
negative numbers http://jsfiddle.net/1734wvs5/22/
multiple on time axis http://jsfiddle.net/1734wvs5/17/

*css*
http://jsfiddle.net/1734wvs5/25/

crmorford:
ok all functionality looks good, i just want clearer errors from the two comments i made.

jtlan:
> Can we get a clearer error for bad input to pixelPosition?

Sure, I can check that the input is a valid number.

> No error when value doesn't match scale 

This is kind of difficult, since the correct type-checking only really occurs in Typescript -- JS can't easily catch this sort of problem.

PR #2595, opened 2015-08-11
add labels for Rectangle plot

### LINKED ISSUE #2126: Labels on Plots.Rectangle
I wanna put the values in these boxes
screen shot 2015-05-20 at 4 13 32 pm
A lot of plots online do this. It seems like a very common use case

ztsai:
```Rectangle.label(label?: Accessor<string>)```: get/set label accessor
```Rectangle.labelsEnabled(enabled?: boolean)```: get/set label enabled status
demo: http://jsfiddle.net/ztsai/qcjrxLzh/3/
closes #2126

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -181,7 +181,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Bar<X, Y>;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled === undefined) {
+      if (enabled == null) {
         return this._labelsEnabled;
       } else {
         this._labelsEnabled = enabled;
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -181,7 +181,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Bar<X, Y>;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled == null) {
+      if (enabled === undefined) {
         return this._labelsEnabled;
       } else {
         this._labelsEnabled = enabled;
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -173,7 +173,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Pie;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled === undefined) {
+      if (enabled == null) {
         return this._labelsEnabled;
       } else {
         this._labelsEnabled = enabled;
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -173,7 +173,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Pie;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled == null) {
+      if (enabled === undefined) {
         return this._labelsEnabled;
       } else {
         this._labelsEnabled = enabled;
--- a/src/plots/rectanglePlot.ts
+++ b/src/plots/rectanglePlot.ts
@@ -306,11 +306,13 @@ export module Plots {
     }
 
     /**
-     * Gets the Accessor for label.
+     * Gets the accessor for labels.
+     *
+     * @returns {Accessor<string>}
      */
     public label(): Accessor<string>;
     /**
-     * Sets label to the result of an Accessor.
+     * Sets the text of labels to the result of an Accessor.
      *
      * @param {Accessor<string>} label
      * @returns {Plots.Rectangle} The calling Rectangle Plot.
@@ -327,9 +329,9 @@ export module Plots {
     }
 
     /**
-     * Get whether rectangle labels are enabled.
+     * Gets whether labels are enabled.
      *
-     * @returns {boolean} Whether rectangles should display labels or not.
+     * @returns {boolean}
      */
     public labelsEnabled(): boolean;
     /**
--- a/src/plots/rectanglePlot.ts
+++ b/src/plots/rectanglePlot.ts
@@ -5,8 +5,8 @@ export module Plots {
   export class Rectangle<X, Y> extends XYPlot<X, Y> {
     private static _X2_KEY = "x2";
     private static _Y2_KEY = "y2";
-    private static _LABEL_KEY = "label";
     private _labelsEnabled = false;
+    private _label: Accessor<string> = null;
 
     /**
      * A Rectangle Plot displays rectangles based on the data.
@@ -308,7 +308,7 @@ export module Plots {
     /**
      * Gets the Accessor for label.
      */
-    public label(): AccessorScaleBinding<any, string>;
+    public label(): Accessor<string>;
     /**
      * Sets label to the result of an Accessor.
      *
@@ -318,10 +318,10 @@ export module Plots {
     public label(label: Accessor<string>): Plots.Rectangle<X, Y>;
     public label(label?: Accessor<string>): any {
       if (label == null) {
-        return this._propertyBindings.get(Rectangle._LABEL_KEY);
+        return this._label;
       }
 
-      this._bindProperty(Rectangle._LABEL_KEY, label, null);
+      this._label = label;
       this.render();
       return this;
     }
@@ -419,7 +419,7 @@ export module Plots {
       let measurer = new SVGTypewriter.Measurers.Measurer(labelArea);
       let writer = new SVGTypewriter.Writers.Writer(measurer);
       data.forEach((datum, datumIndex) => {
-        let label = "" + this.label().accessor(datum, datumIndex, dataset);
+        let label = "" + this.label()(datum, datumIndex, dataset);
         let measurement = measurer.measure(label);
 
         let x = attrToProjector["x"](datum, datumIndex, dataset);
--- a/src/plots/rectanglePlot.ts
+++ b/src/plots/rectanglePlot.ts
@@ -334,6 +334,7 @@ export module Plots {
     public labelsEnabled(): boolean;
     /**
      * Sets whether labels are enabled.
+     * Labels too big to be contained in the rectangle, cut off by edges, or blocked by other rectangles will not be shown.
      *
      * @param {boolean} labelsEnabled
      * @returns {Rectangle} The calling Rectangle Plot.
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: src/plots/rectanglePlot.ts [hunk 1]
```diff
@@ -251,6 +253,50 @@ export module Plots {
...
+    public label(): AccessorScaleBinding<any, string>;
+    /**
+     * Sets label to the result of an Accessor.
+     *
+     * @param {Accessor<string>} label
+     * @returns {Plots.Rectangle} The calling Rectangle Plot.
+     */
+    public label(label: Accessor<string>): Plots.Rectangle<X, Y>;
+    public label(label?: Accessor<string>): any {
+      if (label === undefined) {
```
jtlan: `== null` is typical.

src/plots/rectanglePlot.ts
```diff
@@ -251,6 +253,50 @@ export module Plots {
...
+    public labelsEnabled(): boolean;
+    /**
+     * Sets whether labels are enabled.
+     *
+     * @param {boolean} labelsEnabled
+     * @returns {Rectangle} The calling Rectangle Plot.
+     */
+    public labelsEnabled(enabled: boolean): Plots.Rectangle<X, Y>;
+    public labelsEnabled(enabled?: boolean): any {
+      if (enabled === undefined) {
```
jtlan: `== null`.

test/plots/rectanglePlotTests.ts
```diff
@@ -403,6 +403,80 @@ describe("Plots", () => {
 
         svg.remove();
       });
+
+    });
+
+    describe("Rectangle Plot With Labels", () => {
+      let svg: d3.Selection<any>;
```
jtlan: `<void>` is probably more accurate, since no data is being bound to it.

test/plots/rectanglePlotTests.ts
```diff
@@ -403,6 +403,80 @@ describe("Plots", () => {
...
+      let rectanglePlot: Plottable.Plots.Rectangle<any, any>;
```
jtlan: `<number, number>`, judging by the `Scale`s you use later?

src/plots/rectanglePlot.ts
```diff
@@ -251,6 +253,50 @@ export module Plots {
       });
     }
 
+    /**
+     * Gets the Accessor for label.
+     */
+    public label(): AccessorScaleBinding<any, string>;
```
jtlan: I realize that this is the case on `symbol()` from `Plots.Scatter` as well, but should we be returning an `AccessorScaleBinding` when there's never a `Scale`? It helps with consistency, but it's kind of odd.

jtlan:
I expect we'll get some use cases involving labels on the top/bottom or left/right edges of each rectangle. Is that something that could be addressed by modifying this API, or is that a completely different feature?

ztsai:
do you mean having more than one label for each entities?

jtlan:
>do you mean having more than one label for each entities?

That's more of an implementation-level description; I'm just suggesting that a user might want to make something that looks like the image, and we should figure out what we want to do if that request comes, even if it's "we come up with a different API point for it".

ztsai:
talked offline; it's probably better to add new API points upon new features requests

jtlan:
Potential future API ideas that don't require modifying the existing API (just noting for future references):

#### Add more property setters for other positions
```Typescript
labelLeft();
labelRight();
labelTop();
labelBottom();
```
This adds a lot of methods, but since they're all fairly similar I don't think it adds much *API complexity*.

#### Add an optional field to specify which position's labels to get/set
```Typescript
label(position = "center");
label(label: Accessor<string>, position="center");
```
Doesn't add more methods, but effectively adds a stringy enum at the back of the signature, when it makes more sense at the front. Of course this is also possible:
```Typescript
label(position = "center");
label(position: string, label: Accessor<string>);
label(label: Accessor<string>); // behaves like label("center", accessor)
```

src/plots/rectanglePlot.ts
jtlan: Example of returning `AccessorScaleBinding` even when there's no `Scale`: `symbol()` (`Plots.Scatter`)

Example of not returning a binding: `interpolator()` (`Plots.Line`)

I think if there's no foreseeable case in which a `Scale` makes sense, we shouldn't return an `AccessorScaleBinding`.

src/plots/rectanglePlot.ts
```diff
@@ -251,6 +253,50 @@ export module Plots {
...
+     * @param {Accessor<string>} label
+     * @returns {Plots.Rectangle} The calling Rectangle Plot.
+     */
+    public label(label: Accessor<string>): Plots.Rectangle<X, Y>;
+    public label(label?: Accessor<string>): any {
+      if (label == null) {
+        return this._propertyBindings.get(Rectangle._LABEL_KEY);
+      }
+
+      this._bindProperty(Rectangle._LABEL_KEY, label, null);
```
jtlan: Since the label-drawing code is independent from the main part of the drawing, does it even make sense to bind this as a property? In particular, there's no `Scale`, so there's not really a point in storing the binding so we can compute the extent or something.

src/plots/piePlot.ts
```diff
@@ -173,7 +173,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Pie;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled === undefined) {
+      if (enabled == null) {
```
bluong: Technically outside scope of pull request

src/plots/barPlot.ts
```diff
@@ -181,7 +181,7 @@ export module Plots {
      */
     public labelsEnabled(enabled: boolean): Bar<X, Y>;
     public labelsEnabled(enabled?: boolean): any {
-      if (enabled === undefined) {
+      if (enabled == null) {
```
bluong: Technically outside scope of pull request

src/plots/rectanglePlot.ts
bluong: We probably need to revisit the property API for a later v2.0.

Let's consider what a label is supposed to do and what properties are supposed to do and see if the two have aligned goals.

My interpretation is that properties (using `x()` and `y()` as examples) derive aspects from the data and then can be driven into a scale so that the data is displayed properly.

Labels, however, derive aspects from the data and should be perfectly fine without the need of a scale.

Thus, label should not be a property.

Hopefully that makes sense.

bluong: Just restating when we need to make a property; if we have a need to port data values through a scale to make them visually presentable, then we should make that aspect a property.

jtlan: I mean, it does affect a property, just not one that requires a scale.

test/plots/rectanglePlotTests.ts
```diff
@@ -441,6 +441,80 @@ describe("Plots", () => {
...
+          { x: 0, y: 0, x2: 1, y2: 1, val: 1 },
+          { x: 0, y: 1, x2: 1, y2: 2, val: 2 }
+        ];
+        dataset = new Plottable.Dataset(DATA);
+        rectanglePlot.addDataset(dataset);
+        rectanglePlot.x((d: any) => d.x, xScale)
+                     .y((d: any) => d.y, yScale)
+                     .x2((d: any) => d.x2)
+                     .y2((d: any) => d.y2)
+                     .label((d: any) => d.val)
```
jtlan: Theoretically, this shouldn't compile, because `val` is a `number`, not a `string`. However, Typescript doesn't flag this because of the way <any>s work:
```Typescript
(d: any) => d.val; // Identifies as (d: any) => any
```

Not sure what to do about this.

src/plots/rectanglePlot.ts
```diff
@@ -303,6 +305,50 @@ export module Plots {
...
+        return this._label;
+      }
+
+      this._label = label;
+      this.render();
+      return this;
+    }
+
+    /**
+     * Get whether rectangle labels are enabled.
```
jtlan: "Gets whether labels are enabled."

 Also, on getters for simple values the description after the `@returns` can be safely omitted.

src/plots/rectanglePlot.ts
```diff
@@ -303,6 +305,50 @@ export module Plots {
       return intersected;
     }
 
+    /**
+     * Gets the Accessor for label.
+     */
+    public label(): Accessor<string>;
+    /**
+     * Sets label to the result of an Accessor.
```
jtlan: "Sets the text of labels to the result of an Accessor."

src/plots/rectanglePlot.ts
```diff
@@ -303,6 +305,50 @@ export module Plots {
       return intersected;
     }
 
+    /**
+     * Gets the Accessor for label.
```
jtlan: "Gets the accessor for labels"?

crmorford:
1. using accessor from one plot for another plot http://jsfiddle.net/fza2vdcc/
2. switch labels on/off with click http://jsfiddle.net/hhdqno45/1/
3. switch labels with click http://jsfiddle.net/hhdqno45/2/
4. tight fit in x-direction http://jsfiddle.net/hhdqno45/4/
5. tight fit in y-direction http://jsfiddle.net/hhdqno45/5/
6. label cut off by edge http://jsfiddle.net/hhdqno45/6/ (still shows up: this is inconsistent behavior from other plots, we've fixed this on bar and pie recently)
7. label cut off by other rectangles http://jsfiddle.net/hhdqno45/7/ (shouldn't show up on blue rect)
8. change css http://jsfiddle.net/hhdqno45/8/ (labels can be written outside of rectangle)
9. date label http://jsfiddle.net/hhdqno45/9/

I'd like 6 and 7 to be fixed. I don't remember how we usually handle 8, we might have to talk with the team to see what we do in other plots when this happens.

src/plots/rectanglePlot.ts
```diff
@@ -354,6 +400,85 @@ export module Plots {
...
+          let verticalOffset = (height - measurement.height) / 2;
+          x += horizontalOffset;
+          y += verticalOffset;
+
+          let xLabelRange = { min: x, max: x + measurement.width };
+          let yLabelRange = { min : y, max: y + measurement.height };
+          if (xLabelRange.min < xMin || xLabelRange.max > xMax || yLabelRange.min < yMin || yLabelRange.max > yMax) {
+            return;
+          }
+          if (this._overlayLabel(xLabelRange, yLabelRange, datumIndex, datasetIndex, dataToDraw)) {
```
jtlan: Since `_overlayLabel` is an `O(number of points)` call, drawing labels takes `O(n^2)`. This might cause difficulties with larger datasets.

src/plots/rectanglePlot.ts
```diff
@@ -354,6 +400,85 @@ export module Plots {
...
+            textRotation: 0
+          });
+        }
+      });
+    }
+
+    private _overlayLabel(labelXRange: Range, labelYRange: Range, datumIndex: number, datasetIndex: number,
+                          dataToDraw: Utils.Map<Dataset, any[]>) {
+      let attrToProjector = this._generateAttrToProjector();
+      for (let i = datasetIndex; i < this.datasets().length; i ++ ) {
```
jtlan: Might be more efficient to store `this.datasets()` off in a separate variable.

jtlan:
@christinaforney asked if we could document the cases where labels get hidden, so probably add a line to that effect in the documentation for `labelsEnabled()`.

crmorford:
The hiding by other rectangles looks really good :)
http://jsfiddle.net/hhdqno45/10/
wasd to move blue rectangle around

PR #2597, opened 2015-08-11
Contextual label placement

### LINKED ISSUE #2578: More display options for labels on Plots.Bar
Currently, labels on `Plots.Bar` are drawn within the bar itself, and are hidden if they are too small. However, it's often exactly in the circumstances of small bars that labels are most useful. It'd be good to have an option to display the labels *above* the bars instead of within them.

okal:
Possibly closes #2578.

Still working out what's broken in the tests. `grunt test` output, even with the `--force` flag doesn't tell me which tests are broken. Also need to add tests for a few things. Was sort of exploring as I went, so I didn't do Test First.

*Missing*
- [x] Similar functionality for horizontal bar charts

*Tests Needed*

Please add to the list if you can think of anything I've missed.

- [x] Confirm that labels are shown off the bar when the bar is too short.

image


jtlan:
@okal, thanks for the PR! This looks pretty neat.

>Still working out what's broken in the tests

If you go to `localhost:9999/test/tests.html`, the tests will run in your browser, making it easier to figure out what's going wrong.

> Confirm that labels aren't drawn when the value for the primary accessor is `0`.

Might be better to allow the user to supply an `Accessor` for the labels? This way "`0`" can be returned if for some reason the user wants that displayed.

> Confirm that labels are shown above the bar when the bar is too short.

Should the label be drawn below if the bar value is negative? That behavior would be symmetric with respect to positive labels, but also scatters the labels across the `Plot`, making them harder to read.

src/plots/barPlot.ts
```diff
@@ -440,10 +439,16 @@ export module Plots {
...
           }
 
-          let showLabel = true;
           let labelPosition = {
             x: x,
-            y: positive ? y : y + h - measurement.height
+            y: y
+          };
+
+          if (showLabelAboveBar) {
```
pld: looks like this is duplicating 7935-7942, should it be abstracted to an fn?

pld:
About drawing negative labels, another potential problem with this is that you'd need to ensure they don't overlap with ticks and axis labels in a hard to read/messy way.

jtlan:
> About drawing negative labels, another potential problem with this is that you'd need to ensure they don't overlap with ticks and axis labels in a hard to read/messy way.

That'd be a problem for positive labels too, if the `Axis` were on top. However, `Plots` have a `clipPath`, which cuts off anything that would be drawn outside the space the `Plot` has been allocated. Any logic that makes positive labels fit could also be used to make negative labels fit.

pld:
Good point

src/plots/barPlot.ts
okal: :+1:

okal:
@jtlan The in-browser test runner is really cool! Thanks for the tip.

About showing or not showing the zero, I'm starting to think a formatter may be ideal, since this is an entirely presentational concern. Maybe with a default that hides zeroes. Any thoughts on that?

I see what you mean, regarding showing negative labels before bars. I think your concern applies to the in-bar labels as well, though. Let me set them below and then we can base further discussion off the result.

@pld This is what they look like right now.

http://jsfiddle.net/okal/j5L2o6e4/7/embedded/result/

okal:
Here's how it now looks for negative labels, as of 56c1c80.

image

jtlan:
> I'm starting to think a formatter may be ideal

That might be the best way to go about it, since the `Plots.Bar` already has a `labelFormatter()` API point -- currently it defaults to the identity formatter. I'd vouch for leaving it that way to minimize confusion with this change.

plottable.css
```diff
@@ -65,17 +65,17 @@ svg.plottable {
 }
 
 .plottable .label-area text {
-  fill: #32313F;
```
jtlan: The colors used here were selected by our designers; please don't change them. The recommended way to change the styles for your use case is to load a separate CSS file that overrides them.

plottable.css
```diff
@@ -65,17 +65,17 @@ svg.plottable {
...
+  fill: white;
   font-family: "Helvetica Neue", sans-serif;
   font-size: 14px;
 }
 
-.plottable .light-label text {
```
jtlan: Renaming the CSS classes is technically API break, so we have to leave them as-is. You can still add the other class, but I'd leave the original one.

jtlan:
@okal, can you add an API point that lets users choose between inside and outside labels? That would give some additional flexibility, I think. Something like this:
```Typescript
public labelPosition(): string;
public labelPosition(labelPosition: string): Plots.Bar<X, Y>;
```
where the allowable inputs are `"inside"` and `"outside"`?

plottable.css
okal: Apologies. I'll switch them back.

plottable.css
okal: Cool, will do.

okal:
@jtlan That would be an API break unless "inside" is the default. It'd also be unclear, IMO, because the previous behaviour is a subset of the current behaviour. I think "outside" implies different, rather than expanded behaviour. Also, not quite a positioning issue, more about visibility.

What do you think about something like this?

```javascript
public showOuterLabels(): false;
public showOuterLabels(showOuterLabels: boolean): Plots.Bar<X, Y>;
```

The default behaviour (`false`) would be what the current API supports. Obviously speculative on my part, but I suspect more people would want outer labels to be visible than not, but that would be an API break, so this might be a good compromise. So existing code wouldn't need to change anything.

okal:
@jtlan I think these new changes address your concerns. The API is identical as long if you don't call `plot.displayLabelsOffBar(true)`. Let me know what you think. Hoping we can merge it today.

crmorford:
Just a heads up: StackedBar is a descendant of Bar, so this could get tricky down the road. Imagine if the lowest bar doesn't have enough space, so it wants its label above the bar. The bar directly above it does have enough space, so it wants to put its label inside that bar. However, those are the same space.
labelconflict

jtlan:
> the previous behaviour is a subset of the current behaviour

Somehow, I missed that. I think this is a net improvement overall, so I don't think we need an API point for this -- having it on by default should be fine.

plottable.css
```diff
@@ -78,6 +78,14 @@ svg.plottable {
   fill: #32313F;
 }
 
+.plottable .off-bar-label text {
+  fill: #5279c7;
+}
+
+.plottable .on-bar-label text {
+  fill: white;
```
jtlan: Probably we should leave this class empty -- with this class, all on-bar labels are white, instead of color-dependent, which looks odd if the bars are light-colored:

screen shot 2015-08-17 at 1 25 28 pm

plottable.css
```diff
@@ -78,6 +78,14 @@ svg.plottable {
   fill: #32313F;
 }
 
+.plottable .off-bar-label text {
+  fill: #5279c7;
```
jtlan: This should probably be `#32313F`, which is the color used for the `.dark-label` class, as well as text on `Label`, `Legend`, and `Axis`.

src/plots/barPlot.ts
```diff
@@ -420,51 +445,65 @@ export module Plots {
         let text = this._labelFormatter(primaryAccessor(d, i, dataset)).toString();
         let w = attrToProjector["width"](d, i, dataset);
         let h = attrToProjector["height"](d, i, dataset);
-        let x = attrToProjector["x"](d, i, dataset);
-        let y = attrToProjector["y"](d, i, dataset);
+        let baselineX = attrToProjector["x"](d, i, dataset);
+        let baselineY = attrToProjector["y"](d, i, dataset);
```
jtlan: Rather than "baseline", let's use "base" to avoid confusion with the actual `baseline` of the `Plots.Bar`.

src/plots/barPlot.ts
```diff
@@ -420,51 +445,65 @@ export module Plots {
...
-            x: x,
-            y: positive ? y : y + h - measurement.height
+          let getY = () => {
+            let addend = 0;
+            if (this._isVertical)  {
+                addend += offset;
+                if (showLabelOffBar && positive) {
+                    addend += (offset - h);
+                }
+                if (showLabelOffBar && !positive) {
```
jtlan: Could these `if`s be done by nesting?

test/plots/barPlotTests.ts
```diff
@@ -634,32 +634,18 @@ describe("Plots", () => {
         barPlot.renderTo(svg);
       });
 
-      it("hides labels properly on the right", () => {
```
jtlan: What happened to this test? Previously, there was functionality that hid labels if they would be cut off by the edges of the `Plot` area to avoid confusion (consider a case where the last few digits of "100000" are cut off), which seems to have gone missing.

jtlan:
A few things left to do:
* Restore functionality that hides labels if they would extend off the `Plot`
* Fix CSS classes
* Remove `displayLabelsOffBar()` and make the new behavior the default

Other than that, this feature looks pretty good!

plottable.css
okal: Cool. I think I misunderstood earlier. I'll fix that.

plottable.css
okal: Will change it :+1:

test/plots/barPlotTests.ts
okal: Aah, I see what you mean. It was a misunderstanding on my part.

okal:
Thansk, @jtlan. Looking into the fixes in a bit.

okal:
Pending tasks:

* [ ] Restore removed tests.
* [ ] Rename baselineX/baseLineY to baseX/baseY.
* [ ] Set off-bar label colour to `#32313F`, to be overriden in individual implementations where necessary.
* [ ] Remove flag to enable off-bar labels. They will be on by default, as long as labels are enabled.
* [ ] Remove `.on-bar-label` colour.

jtlan:
Labels that are cut off by the screen no longer hide, whereas they used to:
http://jsfiddle.net/rwtttq4y/1/

I think I can find the line where this is happening.

src/plots/barPlot.ts
```diff
@@ -420,68 +420,78 @@ export module Plots {
...
-              xAlign: xAlign,
-              yAlign: yAlign,
-              textRotation: 0
-          };
-          writer.write(text, w, h, writeOptions);
+        let color = attrToProjector["fill"](d, i, dataset);
+        let dark = Utils.Color.contrast("white", color) * 1.6 < Utils.Color.contrast("black", color);
+        g.classed(dark ? "dark-label" : "light-label", true);
+
+        let hideLabel = (x + measurement.width > this.width() || (positive ? y + measurement.height : y + h) > this.height());
```
jtlan: This isn't quite the right calculation, since half the bar might have to be added, depending on the orientation of the `Plot`. The original code took that into account:
```Typescript
let showLabel = true;
let labelPosition = {
  x: x,
  y: positive ? y : y + h - measurement.height
};

if (this._isVertical) {
  labelPosition.x = x + w / 2 - measurement.width / 2;
} else {
  if (!positive) {
    labelPosition.x = x + w - measurement.width;
  } else {
    labelPosition.x = x;
  }
}

if (labelPosition.x < 0 || labelPosition.x + measurement.width > this.width() ||
    labelPosition.y < 0 || labelPosition.y + measurement.height > this.height()) {
  showLabel = false;
}
```

jtlan:
@okal Restore the missing tests, fix the label-hiding logic, and we should be good to go! It might be worth merging in the latest **`develop`** as well, since the bar-width logic has changed slightly in cases where bars are cut off.

okal:
@jtlan Cool. Sorry this is taking so long. Working on other stuff at the moment, so I haven't been able to prioritize it.

okal:
@jtlan Should be good now.

test/plots/barPlotTests.ts
```diff
@@ -788,23 +820,6 @@ describe("Plots", () => {
         plot.renderTo(svg);
       });
 
-      it("hides labels outside of the visible render area (horizontal)", () => {
```
jtlan: Can you restore this test too?

jtlan:
I thought I noticed weirdness with labels in the horizontal orientation, but turns out that's an existing issue (#2658), which should probably be solved separately from this PR.

test/plots/barPlotTests.ts
okal: Done.

okal: Curious though. Why is this necessary? Why make it `"hidden"` if it's already outside the viewport?

crmorford:
Horizontal StackedBar + Time isn't looking good :( 
http://jsfiddle.net/ts720tdr/6/

jtlan:
screen shot 2015-08-24 at 12 44 18 pm


@okal , it looks like the issue is caused by labels being drawn outside of bars that are farther down in the stack. Maybe on stacked bars, labels that don't fit should just be hidden?

crmorford:
here: http://jsfiddle.net/ts720tdr/11/
it seems like there should be room to place the label to the left of the top bar
(there are 4 bars, you can't see the other 3 because I set a tight domain)

crmorford:
http://jsfiddle.net/ts720tdr/13/
looks like the .light-label class overrides the .off-bar-label fill. @jtlan is this expected/okay? it was a surprise to me.
screen shot 2015-08-24 at 1 27 55 pm
screen shot 2015-08-24 at 1 28 06 pm

crmorford:
Not enough space in y-direction http://jsfiddle.net/ts720tdr/
panzoom y  (really cool seeing the numbers hop on and off) http://jsfiddle.net/ts720tdr/1/
panzoom x http://jsfiddle.net/ts720tdr/2/
negative zero http://jsfiddle.net/ts720tdr/3/
negative bars http://jsfiddle.net/ts720tdr/4/
time (formatted) http://jsfiddle.net/ts720tdr/5/
stackedBar + time http://jsfiddle.net/ts720tdr/6/ *
stackedBar horizontal http://jsfiddle.net/ts720tdr/7/ * 
stackedBar vertical http://jsfiddle.net/ts720tdr/9/
numbers too wide for bars http://jsfiddle.net/ts720tdr/10/
room to left of bar http://jsfiddle.net/ts720tdr/11/
css labels http://jsfiddle.net/ts720tdr/13/ *
css labels 2 http://jsfiddle.net/ts720tdr/14/
doesn't break waterfall http://jsfiddle.net/ts720tdr/16/
clustered bar http://jsfiddle.net/ts720tdr/17/ & http://jsfiddle.net/ts720tdr/18/
ModLog scale http://jsfiddle.net/ts720tdr/19/

okal:
@crmorford Had the week off work. Will address pending issues this week.

okal:
@jtlan I agree. Hiding on StackedBars sounds sensible to me. Having the labels there isn't helpful.
@crmorford That's not the behaviour _I_ intended. I'll switch the order `.off-bar-label text` and `.light-label text` order around.
Hard to tell how I should use the fiddles you posted, though I realize they're for your personal reference. Could you please mark the ones in need of a fix?

okal:
Tasks
- [ ] Hide labels in StackedBars.
- [ ] Prevent `.light-label text` from overriding `.off-bar-label text`.

crmorford:
Sorry @okal, didn't see your comment til now. I'll tag you next to fiddles that need your attention, feel free to ignore any others :)

jtlan:
Hey @okal,

We're pretty eager to add this functionality. Are you planning to get back to this soon? We're also happy to pick it up from here and get it over the finish line. If we don't hear back from you in a day or so we may begin trying to integrate it.

Thanks!

okal:
Hey @jtlan. Haven't had a chance to work on it. We're working on a launch so I had to reprioritize. Yeah, I'd appreciate it if you took it from here. Sorry I couldn't spend more time on this.

jtlan:
Thanks for letting us know, @okal. I'll take it from here.

jtlan:
Recreated at #2779.

PR #2598, opened 2015-08-11
Apply the correct pointer-events styles on DragBoxLayer edges/corners.

jtlan:
Because functional CSS properties (in this case `pointer-events`) were not being set directly on the elements, the resize cursors weren't showing.

On this branch: http://jsfiddle.net/5tw5efp9/
On **`develop`**: http://jsfiddle.net/36euzLjv/

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -10890,7 +10890,8 @@ var Plottable;
                 _super.prototype._setup.call(this);
                 var createLine = function () { return _this._box.append("line").style({
                     "opacity": 0,
-                    "stroke": "pink"
+                    "stroke": "pink",
+                    "pointer-events": "visibleStroke"
                 }); };
                 this._detectionEdgeT = createLine().classed("drag-edge-tb", true);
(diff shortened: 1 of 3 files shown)


aicioara:
As a side note, one thing I don't like is the cursor flickering while dragging. Can be fixed with something like this, but it is out of the scope of this commit and probably inefficient (perf wise, as it triggers reflows of the SVG)

```javascript
neswCorner.mousedown(function() { $('svg').css('cursor', 'nesw-resize') })
neswCorner.mouseup(function() { $('svg').css('cursor', 'default') })
```

crmorford:
This can only happen on desktop browsers right? Since it's a hover-ish visualization.

crmorford:
I agree that the flickering is distracting, but doesn't need to be fixed urgently. We've tended to add on/off switches for the perf/visual appearance tradeoff, could we add andrei's idea as another of those, in a later PR?

Issue #2572, opened 2015-08-06
Add .entitiesIn() to Plots.Segment


christinaforney:
Broken out from parent issue: https://github.com/palantir/plottable/issues/1959

PR #2600, opened 2015-08-12
add entitiesIn() for segment plots

ztsai:
add entitiesIn(bound) / entitiesIn(xRange, yRange) for segment plot
demo: http://jsfiddle.net/ztsai/ydm4znfs/5/
closes: #2572

### DIFF
--- a/src/plots/segmentPlot.ts
+++ b/src/plots/segmentPlot.ts
@@ -88,7 +88,7 @@ export module Plots {
     /**
      * Sets X2 to a constant number or the result of an Accessor.
      * If a Scale has been set for X, it will also be used to scale X2.
-     * 
+     *
      * @param {number|Accessor<number>|Y|Accessor<Y>} y2
      * @returns {Plots.Segment} The calling Segment Plot
      */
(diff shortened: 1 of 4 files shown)


src/plots/segmentPlot.ts
```diff
@@ -172,6 +172,83 @@ export module Plots {
...
+      let corners = [
+        { x: xRange.min, y: yRange.min },
+        { x: xRange.min, y: yRange.max },
+        { x: xRange.max, y: yRange.max },
+        { x: xRange.max, y: yRange.min }];
+      let intersections = corners.filter((point: Point, index: number) => {
+          if (index !== 0) {
+            // return true if border formed by conecting current corner and previous corner intersects with the segment
+            return this._lineIntersectsSegment(startPoint, endPoint, point, corners[index - 1]) &&
+                   this._lineIntersectsSegment(point, corners[index - 1], startPoint, endPoint) ;
```
jtlan: Doesn't this logic miss one of the four edges? Although the end-point check above should cover that case.

test/plots/segmentPlotTests.ts
```diff
@@ -113,5 +113,69 @@ describe("Plots", () => {
...
+        let entities = plot.entitiesIn({
+          topLeft: { x: xScale.scale(0), y: yScale.scale(4.5) },
+          bottomRight: { x: xScale.scale(2.5), y: yScale.scale(3) } });
+        assert.lengthOf(entities, 2, "retrieved 2 entities intersect with the box");
+        assert.strictEqual(entities[0].index, 0, "the entity of index 0 is retrieved");
+        assert.strictEqual(entities[1].index, 1, "the entity of index 1 is retrieved");
+        svg.remove();
+      });
+
+      it("retrieves the entities that intersect with the bounding box", () => {
```
jtlan: How is this test different from the above? Can we rename them to make the distinction clear?

test/plots/segmentPlotTests.ts
```diff
@@ -113,5 +113,69 @@ describe("Plots", () => {
...
+        let entities = plot.entitiesIn(
+          { min: xScale.scale(2.5), max: xScale.scale(4) },
+          { min: yScale.scale(3.5), max: yScale.scale(2.5) });
+        assert.lengthOf(entities, 2, "retrieved 2 entities intersect with the box");
+        assert.strictEqual(entities[0].index, 1, "the entity of index 1 is retrieved");
+        assert.strictEqual(entities[1].index, 2, "the entity of index 2 is retrieved");
+        svg.remove();
+      });
+
+      it("retrieves the entities that intersect with given ranges", () => {
```
jtlan: Same here -- Can we rename this and the above test to make the distinction clear?

src/plots/segmentPlot.ts
ztsai: yeah, it's not possible for a segment to intersect only the last edge with no end points lying inside the box; I can add the check for the last edge for completeness if you think that's necessary

jtlan:
For the testing, I think it would be best if we enumerated the different possible cases, and called them out separately:
1. One endpoint.
screen shot 2015-08-11 at 8 18 50 pm
2. The other endpoint.
screen shot 2015-08-11 at 8 19 06 pm
3. Both endpoints.
screen shot 2015-08-11 at 8 19 16 pm
4. Neither endpoint (intersecting)
screen shot 2015-08-11 at 8 19 00 pm
5. No intersection.
screen shot 2015-08-11 at 8 20 09 pm

Right now, if the test were to fail, it's hard to tell exactly what went wrong and what case we should be investigating. It's also difficult to verify that the correct cases are being covered in testing, because they aren't separated out.

test/plots/segmentPlotTests.ts
ztsai: they are checking selecting different combinations of entities, that's all.  
Actually, is it necessary to check different combinations in unit tests? if so, would it make more sense if I combine the 2 tests in one?

ztsai:
what about different orientations of the line segments?

jtlan:
>what about different orientations of the line segments?

It's probably best to check this for completeness, although I'm not sure it's strictly necessary. There's two potential ways to organize the test battery in that case:
1. One `it()` for each orientation of line (vertical, horizontal, diagonal, ... do we need any more?), each containing an `assert()` for each of the cases listed above, or
2. One `it()` for each case listed above, each containing an `assert()` for each orientation of line. This is closer to the existing test structure, but with the individual cases called out more clearly.

test/plots/segmentPlotTests.ts
```diff
@@ -113,5 +113,130 @@ describe("Plots", () => {
...
+
+        let entities = plot.entitiesIn(
+          { min: xScale.scale(1.5), max: xScale.scale(2.5) },
+          { min: yScale.scale(2.5), max: yScale.scale(1.5) });
+        assert.lengthOf(entities, 0, "no enitites interects with the ranges");
+
+        svg.remove();
+      });
+
+      function checkEneitesInRange(plot: Plottable.Plots.Segment<any, any>, index: number, x1: number, x2: number, y1: number, y2: number) {
```
jtlan: "check**Entities**InRange"?

test/plots/segmentPlotTests.ts
```diff
@@ -113,5 +113,130 @@ describe("Plots", () => {
...
+        checkEneitesInRange(plot, 2, 3.5, 4.5, 2.5, 1.5);
+
+        // horizontal segment
+        checkEneitesInRange(plot, 3, 1.5, 2.5, 1.5, 0.5);
+        checkEneitesInRange(plot, 3, 3.5, 4.5, 1.5, 0.5);
+
+        svg.remove();
+      });
+
+      it("retrieves the entity if both of its enpoints are in the ranges", () => {
```
jtlan: "endpoints"

test/plots/segmentPlotTests.ts
```diff
@@ -113,5 +113,130 @@ describe("Plots", () => {
...
+        let plot = new Plottable.Plots.Segment()
+          .x((d) => d.x, xScale).x2((d) => d.x2)
+          .y((d) => d.y, yScale).y2((d) => d.y2);
+        plot.addDataset(new Plottable.Dataset(data)).renderTo(svg);
+
+        // vertial segment
+        checkEneitesInRange(plot, 0, 0.5, 1.5, 1.5, 0.5);
+        checkEneitesInRange(plot, 0, 0.5, 1.5, 4.5, 3.5);
+
+        // diagonal segment with negative slop
```
jtlan: "slope"?

PR #2605, opened 2015-08-13
DragBoxLayer doesn't show resize/move cursors if not enabled.

### LINKED ISSUE #2542: DragBoxLayer: Do now show alternate cursors when not enabled()
When disabled the selection (XDragBoxLayer) is not resizable/movable (although the mouse pointer changes as if it were resizable). When this is disabled, the cursor should not indicate that it is.

jtlan:
The CSS classes for resizing (`x-resizable`, `y-resizable`, `movable`) are no longer applied if the `DragBoxLayer` is disabled.

With the fix: http://jsfiddle.net/uhpevd7n/1/
On **`develop`**: http://jsfiddle.net/uhpevd7n/

Closes #2542.

### DIFF
--- a/src/components/dragBoxLayer.ts
+++ b/src/components/dragBoxLayer.ts
@@ -289,7 +289,7 @@ export module Components {
 
     // Sets resizable classes. Overridden by subclasses that only resize in one dimension.
     protected _setResizableClasses(canResize: boolean) {
-      if (canResize) {
+      if (canResize && this.enabled()) {
         this.addClass("x-resizable");
         this.addClass("y-resizable");
       } else {
@@ -314,12 +314,16 @@ export module Components {
         return this._movable;
       }
       this._movable = movable;
-      if (movable) {
+      this._setMovableClass();
+      return this;
+    }
+
+    private _setMovableClass() {
+      if (this.movable() && this.enabled()) {
         this.addClass("movable");
       } else {
         this.removeClass("movable");
       }
-      return this;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


jtlan:
There's a known issue here, where if the user disables the `DragBoxLayer` indirectly, by directly changing the state of the `Interaction.Drag`
```Typescript
dragBoxLayer.dragInteraction().enabled(false);
```
then the CSS classes will not be updated correctly. However, there isn't a way to detect this happening from inside `DragBoxLayer`. Maybe we should deprecate the `dragInteraction()` method?

bluong:
This is arguable, but if the user were to dig down into the dragBoxLayer's dragInteraction and then mess with the configuration of an inner object, I think that's at the point where total reliance and liability is at the users end.  (Think about affecting the data array on `Dataset`s.  If the data array object is affected, no updates are sent unless the setter is called.)

bluong:
Using the above logic, since `dragInteraction()` also gives users a lot of capability, deprecating the functionality may be difficult.  One way would be to port the functions on the `Interactions.Drag` to `DragBoxlayer` and duplicate endpoints / code, however this comes with the possible danger and cost of duplication.

src/components/dragBoxLayer.ts
```diff
@@ -289,7 +289,7 @@ export module Components {
 
     // Sets resizable classes. Overridden by subclasses that only resize in one dimension.
     protected _setResizableClasses(canResize: boolean) {
```
bluong: Possibly outside scope of this pull request, but this function really doesn't need a parameter if `canResize === this._resizable`

jtlan: Probably best done as a separate refactor. Plus, it's `protected`, so technically we can't change the API...

bluong: Darn, that's a good point.

Issue #2569, opened 2015-08-06
Add .entitiesAt() to Plots.Scatter


christinaforney:
Broken out from https://github.com/palantir/plottable/issues/1959

jtlan:
One question is if the points are considered "square" or "round" for detection:

#### "square"
The hit area is considered a square with edge of length `size()`, but that means the corners count as part of the circle:
![circleoutside](https://cloud.githubusercontent.com/assets/1440449/9262236/32cac0ba-41c8-11e5-9ecb-d2aedae55be5.png)

#### "round"
The hit area is considered a circle with radius `size()/2`, but that means the corners won't count as part of the square:
![squareoutside](https://cloud.githubusercontent.com/assets/1440449/9262263/6b47c294-41c8-11e5-935d-798a85bd4efd.png)

Things get weirder when shapes like crosses/triangles are considered. Doing a direct hit-test against the path is really difficult.

crmorford:
Obviously a direct hittest would be nice, but I think it's better to give more room for user error (escpecially since scatter plot points are so tiny). I'm in favor of "square", so that every shape will be completely contained in the hitbox.

PR #2608, opened 2015-08-13
add entitiesAt() for scatter plot

ztsai:
closes: #2569
demo: http://jsfiddle.net/fa3mjzew/4/

### DIFF
--- a/src/plots/scatterPlot.ts
+++ b/src/plots/scatterPlot.ts
@@ -180,6 +180,27 @@ export module Plots {
         return dataXRange.min <= x && x <= dataXRange.max && dataYRange.min <= y && y <= dataYRange.max;
       });
     }
+
+    /**
+     * Gets the Entities at a particular Point.
+     *
+     * @param {Point} p
+     * @returns {PlotEntity[]}
+     */
+    public entitiesAt(p: Point) {
+      let xProjector = Plot._scaledAccessor(this.x());
+      let yProjector = Plot._scaledAccessor(this.y());
+      let sizeProjector = Plot._scaledAccessor(this.size());
+      return this.entities().filter((entity) => {
+        let datum = entity.datum;
+        let index = entity.index;
+        let dataset = entity.dataset;
+        let x = xProjector(datum, index, dataset);
+        let y = yProjector(datum, index, dataset);
+        let size = sizeProjector(datum, index, dataset);
+        return x - size / 2  <= p.x && p.x <= x + size / 2 && y - size / 2 <= p.y && p.y <= y + size / 2;
+      });
+    }
   }
 }
 }
--- a/test/plots/scatterPlotTests.ts
+++ b/test/plots/scatterPlotTests.ts
@@ -214,6 +214,71 @@ describe("Plots", () => {
       svg.remove();
     });
 
+    it("can retrieve entities centered at a given point", () => {
+      let svg = TestMethods.generateSVG(400, 400);
+      let xScale = new Plottable.Scales.Linear();
+      let yScale = new Plottable.Scales.Linear();
+
+      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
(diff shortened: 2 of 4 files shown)


test/plots/scatterPlotTests.ts
```diff
@@ -214,6 +214,71 @@ describe("Plots", () => {
...
+
+      svg.remove();
+    });
+
+    it("determines whether an entity contains a given point by its size", () => {
+      let svg = TestMethods.generateSVG(400, 400);
+      let xScale = new Plottable.Scales.Linear();
+      let yScale = new Plottable.Scales.Linear();
+
+      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
```
jtlan: Is this `Dataset` needed for this test? It looks like the later `entitiesAt()` calls are only testing against a point in the other `Dataset`.

test/plots/scatterPlotTests.ts
```diff
@@ -214,6 +214,71 @@ describe("Plots", () => {
...
+      assert.lengthOf(entities, 1, "only one entity has been retrieved");
+      assert.deepEqual(entities[0].datum, {x: 1, y: 2}, "correct datum has been retrieved");
+
+      plot.size(6);
+      entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(2) - 5});
+      assert.lengthOf(entities, 0, "none of the entities is retrieved");
+      svg.remove();
+    });
+
+    it("returns an empty array when no entities contain a given point", () => {
```
jtlan: Isn't this basically the same thing as the last `assert()` in the above test?

test/plots/scatterPlotTests.ts
```diff
@@ -214,6 +214,71 @@ describe("Plots", () => {
       svg.remove();
     });
 
+    it("can retrieve entities centered at a given point", () => {
+      let svg = TestMethods.generateSVG(400, 400);
+      let xScale = new Plottable.Scales.Linear();
+      let yScale = new Plottable.Scales.Linear();
+
+      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
+      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
```
jtlan: This `Dataset` doesn't appear to be used in the test. Can we remove it?

jtlan:
In some of the tests, multiple `Dataset`s are being used, but the scope of the test only requires one to be present, meaning the test setup can probably be streamlined a bit.

However, we should also have a test that confirms that `entitiesAt()` correctly returns overlapping points from multiple `Dataset`s.

src/plots/scatterPlot.ts
```diff
@@ -180,6 +180,27 @@ export module Plots {
...
+      let yProjector = Plot._scaledAccessor(this.y());
+      let sizeProjector = Plot._scaledAccessor(this.size());
+      return this.entities().filter((entity) => {
+        let datum = entity.datum;
+        let index = entity.index;
+        let dataset = entity.dataset;
+        let x = xProjector(datum, index, dataset);
+        let y = yProjector(datum, index, dataset);
+        let size = sizeProjector(datum, index, dataset);
+        return x - size / 2  <= p.x && p.x <= x + size / 2 && y - size / 2 <= p.y && p.y <= y + size / 2;
```
jtlan: I made [a comment](https://github.com/palantir/plottable/issues/2569#issuecomment-130851513) on the root issue about whether we should be treating the points as "square" (the logic here) or "round". That would require a change to this line and probably also the test code.

crmorford:
That's an amazing demo fiddle!
I was going to test different shapes, sizes, and overlapping symbols, and you hit them all with one test. 
Super helpful, thank you :)

crmorford:
tested on all browsers, with multiple datasets, etc. looks good :)

PR #2619, opened 2015-08-17
[Axes.Time] - includes the margin in its height

### LINKED ISSUE #2042: [timeAxis] orientation() does not update margin
The gutter place is different, depending on the constructor of the axis
http://jsfiddle.net/77p5bera/13/ -> The gutter is between axis and gridlines
http://jsfiddle.net/77p5bera/14/ -> The gutter is at the bottom.
Difference between the two is the constructor. However, the orient() should behave like the constructor value of the orientation

bluong:
`Axes.Time` now appropriately includes the margin in its space usage.

This means that if `Axes.Time` is given enough space to include the tiers as well as the margin, the margin is included in the height for the axis.

Fixes #2042

### DIFF
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -309,9 +309,10 @@ export module Axes {
       // we have no leftover tiers
 
       let size = super._sizeFromOffer(availableWidth, availableHeight);
-      size.height = this._tierHeights.reduce((prevValue, currValue, index, arr) => {
+      let tierHeights = this._tierHeights.reduce((prevValue, currValue, index, arr) => {
         return (prevValue + currValue > size.height) ? prevValue : (prevValue + currValue);
       });
+      size.height = Math.min(size.height, tierHeights + this.margin());
       return size;
     }
 
@@ -491,7 +492,7 @@ export module Axes {
     }
 
     private _hideOverflowingTiers() {
-      let availableHeight = this.height();
+      let availableHeight = this.height() - this.margin();
       let usedHeight = 0;
 
       this.content()
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -492,7 +492,7 @@ export module Axes {
     }
 
     private _hideOverflowingTiers() {
-      let availableHeight = this.height() - this.margin();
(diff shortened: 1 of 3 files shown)


jtlan:
Probably closes #2042, right?

src/axes/timeAxis.ts
```diff
@@ -491,7 +492,7 @@ export module Axes {
     }
 
     private _hideOverflowingTiers() {
-      let availableHeight = this.height();
+      let availableHeight = this.height() - this.margin();
```
jtlan: The margin is intended to be "soft", meaning that the main visualization (tiers in this case) can steal space from it if they need it.

bluong: Is the margin intended to be a "soft" margin?  Note that the current implementation has the timeAxis not consider the margin at all

test/axes/timeAxisTests.ts
```diff
@@ -281,4 +282,15 @@ describe("TimeAxis", () => {
 
   });
 
+  it("height includes margin, label padding, and tick length", () => {
+    let svg = TestMethods.generateSVG(400, 400);
+    let xScale = new Plottable.Scales.Time();
+    let xAxis = new Plottable.Axes.Time(xScale, "bottom");
+    xAxis.margin(100);
+    xAxis.renderTo(svg);
```
jtlan: It seems like rather than `renderTo()`, it would be better to check this by `anchor()`ing the `Axis`, then calling `requestedSpace` to see that it asks for enough, right?

bluong:
I didn't consider #2042 but perhaps it does!

bluong:
Now that I've learned that `margin()`'s are only soft margins (can be decreased in size if space doesn't allow for it), this may no longer be a blocker

crmorford:
regarding #2402: 
before http://jsfiddle.net/77p5bera/27/
after: http://jsfiddle.net/77p5bera/29/

there's a visible margin on the "bottom" oriented axis now

crmorford:
Definitely not ignoring margin now http://jsfiddle.net/77p5bera/30/
(click axis to change margin, click plot to change axis orientation)

PR #2620, opened 2015-08-17
Axis annotation

### LINKED ISSUE #2517: Tick Identifier for Axes
Taken from #2476 as well as from #1640 

Some identifier that can be placed at the axis area that be attached to some tick value to signify that the tick value is important.

Note that it should not be difficult to place multiple of these identifiers.  This is so, if a recurring tick value is important, it is not difficult to create an identifier for each tick.

bluong:
Axis annotations are a visualization on top of an axis that places certain tick values into separate containers.  This allows for these ticks to present themselves as a bit more special to the user than other ticks.

Below is a simple example:

screen shot 2015-08-18 at 5 13 37 pm

Note that the annotations will reside in its own annotation area of the axis taking up as much space as the set number of annotation tiers.  Under the situation where annotations were to collide if they are to be put in the same tier, they would be placed in the next tier.  If the annotation were to render outside the annotation area, the annotation is hidden.

Note that the annotations are placed below the axis area in the above example with a "bottom" orientation axis.  They will be placed above the axis area with a "top" orientation axis, right of the axis area with a "right" orientation axis , and left of the axis area with a "left" orientation axis.

To best calculate collisions, "left" and "right" orientation axes have the annotation labels rotated by 90 degrees.

This feature should work with all types of Axes.

The annotations can be styled through css.  To style the annotation circles, you can use the selector `.plottable .axis .annotation-circle`.  To style the annotation lines, you can use the selector `.plottable .axis .annotation-line`.  To style the annotation rectangles, you can use the selector `.plottable .axis .annotation-rect`.

JSFiddle: https://jsfiddle.net/86trca4u/14/

API Additions:
```typescript
/**
 * Gets the annotated ticks
 */
annotatedTicks(): D[];
/**
 * Sets the annotated ticks
 *
 * @returns {Axis} The calling Axis.
 */
annotatedTicks(annotatedTicks: D[]): Axis<D>;
/**
 * Gets the formatter for the annotations
 */
annotationFormatter(): Formatter;
/**
 * Sets the formatter for the annotations
 *
 * @returns {Axis} The calling Axis.
 */
annotationFormatter(annotationFormatter: Formatter): Axis<D>;
/**
 * Gets if annotations are enabled
 */
annotationsEnabled(): boolean;
/**
 * Sets if annotations are enabled
 *
 * @returns {Axis} The calling Axis.
 */
annotationsEnabled(annotationsEnabled: boolean): Axis<D>;
/**
 * Gets the count of annotation tiers to render.
 */
annotationTierCount(): number;
/**
 * Sets the count of annotation tiers to render.
 *
 * @returns {Axis} The calling Axis.
 */
annotationTierCount(annotationTierCount: number): Axis<D>;
```

Protected API Additions:
```typescript
_drawAnnotations(): void;
_removeAnnotations(): void;
_annotationTierHeight(): number;
_axisSizeWithoutMarginAndAnnotations(): number;
```

Closes #2517


src/axes/axis.ts
```diff
@@ -129,6 +139,14 @@ export class Axis<D> extends Component {
     this._tickLabelContainer = this.content().append("g")
                                            .classed(Axis.TICK_LABEL_CLASS + "-container", true);
     this._baseline = this.content().append("line").classed("baseline", true);
+    this._annotationContainer = this.content().append("g")
+                                              .classed("annotation-container", true);
+    this._annotationContainer.append("g").classed("annotation-line-container", true);
+    this._annotationContainer.append("g").classed("annotation-circle-container", true);
+    this._annotationContainer.append("g").classed("annotation-rect-container", true);
```
jtlan: There's a container for each piece of each annotation? Wouldn't it make more sense to have one `<g>` that contains all parts of a single annotation?

bluong: I tried this, and could not get it to work as intended.  The problem is I want the texts and the rectangles to appear on top of all of the lines.  This means that all of the lines must appear first, then the rectangles, then the texts.

I could theoretically have everything reside a single container and then insert each piece in a specified order; I just thought the container idea would be more organized.

src/axes/axis.ts
```diff
@@ -150,9 +168,216 @@ export class Axis<D> extends Component {
...
+      return this._annotationsEnabled;
+    }
+    this._annotationsEnabled = annotationsEnabled;
+    this.render();
+    return this;
+  }
+
+  protected _drawAnnotations() {
+    let labelPadding = 4;
+    let measurements = new Utils.Map<D, SVGTypewriter.Measurers.Dimensions> ();
```
jtlan: Space before the parens? Can we make a TSLint rule for this?

src/axes/axis.ts
```diff
@@ -150,9 +168,216 @@ export class Axis<D> extends Component {
...
+    this.annotatedTicks().forEach((annotatedTick) => {
```
jtlan: Would it make sense to filter on ticks that are in the domain?

src/axes/axis.ts
```diff
@@ -150,9 +168,216 @@ export class Axis<D> extends Component {
...
+        secondaryPosition = 0;
+        break;
+    }
+
+    bindElements(this._annotationContainer.select(".annotation-line-container"), "line", "annotation-line")
+      .attr(this._isHorizontal() ? "x1" : "y1", positionF)
+      .attr(this._isHorizontal() ? "x2" : "y2", positionF)
+      .attr(this._isHorizontal() ? "y1" : "x1", secondaryPosition)
+      .attr(this._isHorizontal() ? "y2" : "x2", offsetF)
+      .attr("visibility", visibilityF);
```
jtlan: Can we use the object signature of `attr()`?

src/axes/axis.ts
```diff
@@ -150,9 +168,216 @@ export class Axis<D> extends Component {
...
+          textRotation: isHorizontal ? 0 : 90
+        };
+        annotationWriter.write(annotationFormatter(annotationLabel),
+                                 isHorizontal ? measurements.get(annotationLabel).width : measurements.get(annotationLabel).height,
+                                 isHorizontal ? measurements.get(annotationLabel).height : measurements.get(annotationLabel).width,
+                                 writeOptions);
+      });
+  }
+
+  protected _axisHeightWithoutMargin() {
```
jtlan: Maybe "`_axisSizeWithoutMargin()`" or some other name that doesn't suggest an orientation?

src/axes/axis.ts
```diff
@@ -150,9 +168,216 @@ export class Axis<D> extends Component {
...
+    let annotationToTier = new Utils.Map<D, number>();
+    let dimension = this._isHorizontal() ? this.width() : this.height();
+    this.annotatedTicks().forEach((annotatedTick) => {
+      let position = this._scale.scale(annotatedTick);
+      let length = measurements.get(annotatedTick).width;
+      if (position < 0 || position + length > dimension) {
+        annotationToTier.set(annotatedTick, -1);
+        return;
+      }
+      let tierHasCollision = (testTier: number) => annotationTiers[testTier].some((testTick) => {
```
jtlan: This results in an quadratic-time algorithm in the number of annotated ticks. Is it possible to store the annotated ticks in sorted order, which should speed this up to `n log n`? Or is `n` small enough here that it's not that relevant?

jtlan:
Something to watch out for: What if the user passes in bad values to `annotatedTicks()`? For `QuantitativeScale`s, these could be `NaN` or +/-`Infinity`; for `Scales.Category` they could be `string`s not in the domain.

src/axes/axis.ts
bluong: `n` is likely small enough that it will not be relevant (text takes up a lot of space inherently, which means the idea of even > 10 annotations cannot give a really good result)

src/axes/axis.ts
```diff
@@ -150,9 +168,227 @@ export class Axis<D> extends Component {
...
+        };
+        annotationWriter.write(annotationFormatter(annotationLabel),
+                                 isHorizontal ? measurements.get(annotationLabel).width : measurements.get(annotationLabel).height,
+                                 isHorizontal ? measurements.get(annotationLabel).height : measurements.get(annotationLabel).width,
+                                 writeOptions);
+      });
+  }
+
+  protected _annotatedTicksInDomain() {
+    return this.annotatedTicks();
```
jtlan: Wouldn't a generalizable implementation be:
```Typescript
let range = this._scale.range();
return this.annotatedTicks().filter((tick) => Utils.Math.inRange(tick, range[0], range[1]));
```

bluong: What about for `Axes.Category`?

jtlan: Still works, since that function operates in pixel-space. Values not in the domain of `Scales.Category` get scaled to `NaN`, which won't be in the `range()`.

bluong: Oh I see.  Ok making the change.

src/axes/timeAxis.ts
```diff
@@ -488,6 +489,10 @@ export module Axes {
         this._hideOverlappingAndCutOffLabels(i);
       }
 
+      if (this.annotationsEnabled()) {
+        this._drawAnnotations();
```
jtlan: Why is this repeated here?

bluong: `Axes.Time` completely overrides `renderImmmediately()` without using the `super` call, which means I need to port the annotation drawing logic there.

jtlan: Weird, but I guess Axis.Time is also visually different.

jtlan:
Awaiting tests.

bluong:
This took me a while, but I hopefully added an extremely comprehensive test suite.  Note that it did help catch around 2 bugs while digging around, and hopefully they are understandable.

src/axes/axis.ts
jtlan: Since you need the sub-containers later, wouldn't it make sense to store references to them?

bluong: I didn't want to add more instance variables into the mix since I can easily access the values from accessing the `_annotationContainer`.  In this case, I do not necessarily need the `_annotationContainer` instance variable and probably access it via d3 code or just shelve these calls into function calls.

All in all, I don't think instance variables are needed here and if we don't want to see this code, we should probably shelve it into a function.

What do you think?

src/axes/axis.ts
```diff
@@ -150,9 +168,238 @@ export class Axis<D> extends Component {
...
+    this.render();
+    return this;
+  }
+
+  /**
+   * Gets if annotations are enabled
+   */
+  public annotationsEnabled(): boolean;
+  /**
+   * Sets if annotations are enabled
```
jtlan: You might want to add some explanation as to exactly how the annotations work. Do you need to call `margin()` to give them enough room? Do they request additional space? etc.

src/axes/axis.ts
```diff
@@ -150,9 +168,238 @@ export class Axis<D> extends Component {
...
+
+    let hiddenAnnotations = new Utils.Set<any>();
+    let numTiers = Math.floor(this.margin() / tierHeight);
+    annotationToTier.forEach((tier, annotation) => {
+      if (tier === -1 || tier >= numTiers) {
+        hiddenAnnotations.add(annotation);
+      }
+    });
+
+    let bindElements = (selection: d3.Selection<any>, elementName: string, className: string) => {
```
jtlan: I feel like the code might be more readable if the D3 update sequence was written out each time, instead of using `bindElements()`...?

bluong: I personally just didn't like looking at the same 3-4 lines of code multiple times that were doing almost the same thing every time.  Also shoring up the code into a single function centralizes the failure points into a single function rather than across multiple implementations of this (I think there were a few times where I accidentally forgot to class the selections or forget to put the `true` in the classed call or use the wrong selector in the `selectAll` etc.)

src/axes/axis.ts
bluong: I will likely need to update the `margin()` documentation.  Annotations in and of themselves do not request space

jtlan: I'd also add a note about it here -- users of annotations might not immediately think to go look at `margin()`.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,615 @@ describe("BaseAxis", () => {
...
+        it("does not render annotations by default", () => {
+          let mockScale = new Mocks.Scale<{}, number>();
+          let axis = new Plottable.Axis(mockScale, "bottom");
+          let svg = TestMethods.generateSVG(300, 300);
+          axis.renderTo(svg);
+          assert.strictEqual(axis.content().selectAll(".annotation-label").size(), 0, "no annotations by default");
+          svg.remove();
+        });
+
+        it("does not have annotations to render by default", () => {
```
jtlan: HOw's this different from the previous test?

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,615 @@ describe("BaseAxis", () => {
...
+
+      it("returns the axis when setting the annotation formatter", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        assert.strictEqual(axis.annotationFormatter(() => ""), axis, "setting the annotation formatter returns calling axis");
+      });
+    });
+
+    describe("rendering annotations", () => {
+      describe("label content", () => {
+        it("does not render annotations by default", () => {
```
jtlan: This test should probably assert that `annotationsEnabled()` defaults to `false` too.

test/mocks.ts
```diff
@@ -26,4 +26,35 @@ module Mocks {
       return true;
     }
   }
+
+  export class Scale<D, R> extends Plottable.Scale<D, R> {
```
jtlan: Why use a mock `Scale` instead of one of the existing ones?

test/axes/timeAxisTests.ts
```diff
@@ -294,4 +294,15 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  describe("axis annotations", () => {
+    describe("formatting annotation ticks", () => {
+      it("formats the dates to [{{abbreviated weekday}} {{abbreviated month}} {{day of month}}, {{year}}]", () => {
```
jtlan: Maybe mention that this is the default `Formatter` somewhere in the test name.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,615 @@ describe("BaseAxis", () => {
...
+        let svg = TestMethods.generateSVG(300, 300);
+        axis.renderTo(svg);
+
+        axis.content().selectAll(".annotation-rect").each(function() {
+          let annotationRect = d3.select(this);
+          let bbox = this.getBBox();
+          let allowVisible = bbox.x >= 0 && bbox.x + bbox.width <= axis.width() &&
+                             bbox.y >= axis.height() - axis.margin() && bbox.y + bbox.height <= axis.height();
+          let visibilityAttr = allowVisible ? "visible" : "hidden";
+          assert.strictEqual(annotationRect.attr("visibility"), visibilityAttr, "annotation rect inside margin area");
```
jtlan: This is a bit subtle, but changing the "semantic flow" of the test would better represent what is being tested. Currently, the test checks that the "insideness" and visibility of each annotation is the same. While that logic does check the functionality, the following logic would better represent the test title and what the test is looking for:
- Check if the rectangle is inside the margin.
  - If so, assert that it is visible
  - If not, assert that it is not visible

Since this structure better mirrors the logical flow of the test's name, it's easier for developers to understand what it's doing and be confident that the test is checking the correct thing.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,615 @@ describe("BaseAxis", () => {
...
+
+            let firstAnnotationRect = axis.content().select(".annotation-rect");
+            assert.strictEqual(TestMethods.numAttr(firstAnnotationRect, "x"),
+              axis.width() - axis.margin(), "rectangle positioned correctly");
+            svg.remove();
+          });
+        });
+
+        describe("additional row offset position", () => {
+          it("offsets additional rows by the placement of colliding rectangles for bottom orientation", () => {
```
jtlan: It's hard for me to understand what this test is doing based on the description. Is this correct:

`"moves rectangles that would overlap to different rows (orientation: bottom)"`

jtlan:
A number of questions abut the tests.

test/axes/baseAxisTests.ts
bluong: The previous test just amounts to saying that since annotations are disabled by default, then no annotations should render since they are disabled.

This test amounts to saying that since there are no annotated ticks; even if annotations are enabled, no annotations should render.

I do agree that the wording is a bit peculiar; how do you think this should be changed.

test/axes/baseAxisTests.ts
bluong: This is already proven in another test.

test/mocks.ts
bluong: This gives more confidence that annotations will work on any type of scale; not just `Time` or `Linear` since all that is required from the scale is knowledge of the domain, range, and a scale mapping.

test/axes/timeAxisTests.ts
bluong: Good point. will do

test/axes/baseAxisTests.ts
bluong: Is this simply a change in variable name from `allowVisible` to `insideMargin`?

test/axes/baseAxisTests.ts
bluong: Yup I'll go with that.  I'd like to add downwards to the end of that to be more clear about behavior.

test/axes/baseAxisTests.ts
jtlan: How about just "there are not annotated ticks by default", and then check that `annotatedTicks()` returns an empty array?

test/axes/baseAxisTests.ts
jtlan: Is it worth having this test then?

test/mocks.ts
jtlan: However, it doesn't test the subclasses of interest (sometimes a problem with `Time`, since `Date`s with the same value are not strict equal), and adds a potential failure mode because now we also have to prove that there aren't problems in the mock.

test/axes/baseAxisTests.ts
jtlan: No, the loop structure would change slightly too. Right now there's a single assert for both branches; it would be better if there was an `if` that switched on whether the rectangle was inside, and then an `assert` inside each branch.

test/axes/baseAxisTests.ts
bluong: There are essentially 2 aspects to our library.

One is setting up the state on the javascript object, and the other is rendering to the DOM.

The previous test is checking the state on the javascript object but this one is checking that this changed state is reflected in the DOM.

test/axes/baseAxisTests.ts
bluong: Same notes as above since checking that `annotatedTicks()` is an empty array by default is already proven earlier.

test/mocks.ts
bluong: The mock is supposed to be extremely simple (Note the previous mock defined above), and if the mock gets too complex then it is not a good mock.

This mock essentially obeys the fundamentals of a what a scale does.  It takes a domain value and then returns a scaled range value.

bluong: The only alternative to not using a mock of the parent class is to make subsequent tests for each of the subclasses, which is quite infeasible as that would multiply the number of tests by the number of subclasses (This already starts to be infeasible at 2...)

test/axes/baseAxisTests.ts
bluong: What's wrong with using the ternary?  Now the semantic reads...
"The visibility of the rectangle is either visible if inside the margin or hidden if not"

test/axes/baseAxisTests.ts
jtlan: Then what's the point of this test?

test/axes/baseAxisTests.ts
jtlan: Shouldn't those be lumped together?

bluong: I thought about doing this and ultimately decided against it.  It is incredibly easy to test things on a javascript object object level, as I don't have to worry about DOM complexities, etc.  This means that I can have test the API on the pure javascript level and be confident in that aspect.  Then when it comes to the DOM, I just need to realize now the DOM takes in the information I feed it and then renders it.

Now, all the above said, if the tests confidently prove that it just purely pipes the state into appropriate DOM elements.  Maybe indeed these default rendering tests do not need to exist.

test/mocks.ts
jtlan: In this case though, the mock isn't a "mock" of the parent class -- it's a subclass, which puts it in the same position of the inheritance tree as `Scales.Linear`. In this case, it can't give any more faith that tests based on it will be more representative of the general case than `Scales.Linear` would.

If the behavior of your code should be independent of the type of `Scale` in the `Axis`, it should work regardless of the type of `Scale` used in tests right?

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,552 @@ describe("BaseAxis", () => {
...
+            svg.remove();
+          });
+        });
+
+      });
+
+      it("hides annotations if rectangles are outside the margin area", () => {
+        let mockScale = new Plottable.Scales.Linear;
+        let annotatedTicks = [50, 51, 150];
+        mockScale.domain(annotatedTicks);
```
jtlan: Oddly enough, this works, but it shouldn't. This does set the domain to `[50, 51, 150]`, but that means that 50 maps to 0 and 51 maps to the `<svg>` width (300 in this case). I'd explicitly set the domain to `[0, SVG_WIDTH]`.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,552 @@ describe("BaseAxis", () => {
...
+        let svg = TestMethods.generateSVG(300, 300);
+        axis.renderTo(svg);
+
+        axis.content().selectAll(".annotation-rect").each(function() {
+          let annotationRect = d3.select(this);
+          let bbox = this.getBBox();
+          let insideMargin = bbox.x >= 0 && bbox.x + bbox.width <= axis.width() &&
+                             bbox.y >= axis.height() - axis.margin() && bbox.y + bbox.height <= axis.height();
+          let visibilityAttr = insideMargin ? "visible" : "hidden";
+          assert.strictEqual(annotationRect.attr("visibility"), visibilityAttr, "annotation rect inside margin area");
```
jtlan: Also, going to revive my earlier comment about semantic flow:
>This is a bit subtle, but changing the "semantic flow" of the test would better represent what is being tested. Currently, the test checks that the "insideness" and visibility of each annotation is the same. While that logic does check the functionality, the following logic would better represent the test title and what the test is looking for:
- Check if the rectangle is inside the margin.
  - If so, assert that it is visible
  - If not, assert that it is not visible

>Since this structure better mirrors the logical flow of the test's name, it's easier for developers to understand what it's doing and be confident that the test is checking the correct thing.

jtlan: The other advantage of a split `assert` is that we can provide a better assert-message; with this line, we instead get a message like this:
```
AssertionError: annotation rect inside margin area: expected 'hidden' to equal 'visible'
```
---
Maybe you could reword the assert-message instead? Right now it's hard to understand what `"hidden"`/`"visible"` have to do with `"annotation rect inside margin area"`. Something like `"annotations that don't fit are hidden"` might be better, but then you can still get the following case:
```
AssertionError: annotations that don't fit are hidden: expected 'hidden' to equal 'visible'
```
which is really confusing unless you realize that the particular annotation being tested is supposed to be visible.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,555 @@ describe("BaseAxis", () => {
...
+      });
+
+      it("can set if annotations are enabled", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        let annotationsEnabled = axis.annotationsEnabled();
+        axis.annotationsEnabled(!annotationsEnabled);
+        assert.strictEqual(axis.annotationsEnabled(), !annotationsEnabled, "can set if annotations are enabled");
+      });
+
+      it("returns the axis when setting if annotations are enabled", () => {
```
jtlan: You can just convert this into one `assert` in the previous test.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,555 @@ describe("BaseAxis", () => {
...
+      });
+
+      it("can set the annotated ticks", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        let annotatedTicks = [20, 30, 40];
+        axis.annotatedTicks(annotatedTicks);
+        assert.deepEqual(axis.annotatedTicks(), annotatedTicks, "can set the annotated ticks");
+      });
+
+      it("returns the axis when setting the annotated ticks", () => {
```
jtlan: Same with combining this with the above test.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,555 @@ describe("BaseAxis", () => {
...
+
+      it("can set the annotation formatter", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        let annotationFormatter = (d: any) => "";
+        axis.annotationFormatter(annotationFormatter);
+        let test = "testString";
+        assert.strictEqual(annotationFormatter(test), axis.annotationFormatter()(test), "can set the annotated ticks");
+      });
+
+      it("returns the axis when setting the annotation formatter", () => {
```
jtlan: Ditto here.

src/axes/axis.ts
```diff
@@ -150,9 +168,242 @@ export class Axis<D> extends Component {
...
+    this.render();
+    return this;
+  }
+
+  /**
+   * Gets the formatter for the annotations
+   */
+  public annotationFormatter(): Formatter;
+  /**
+   * Sets the formatter for the annotations
```
jtlan: Capitalize Formatter.

src/axes/axis.ts
```diff
@@ -150,9 +168,242 @@ export class Axis<D> extends Component {
...
+  /**
+   * Gets if annotations are enabled
+   *
+   * Annotations will render in the margin area of the axis if enabled and if there is enough space
+   */
+  public annotationsEnabled(): boolean;
+  /**
+   * Sets if annotations are enabled
+   *
+   * Annotations will render in the margin area of the axis if enabled and if there is enough space
```
jtlan: "Annotations are drawn in the margin of the Axis." should be sufficient.

src/axes/axis.ts
```diff
@@ -150,9 +168,242 @@ export class Axis<D> extends Component {
...
+  /**
+   * Gets if annotations are enabled.
+   *
+   * Annotations are drawn in the margin of the Axis..
+   */
+  public annotationsEnabled(): boolean;
+  /**
+   * Sets if annotations are enabled.
+   *
+   * Annotations are drawn in the margin of the Axis..
```
jtlan: Find-replace fail :/

crmorford:
http://jsfiddle.net/86trca4u/3/
* Strings that can be coerced to numbers, are. See '150' in the array here. I don't think that's what we want.
* also, 'null' shows up, I think maybe we should filter that out.

crmorford:
values that shouldn't show up as annotations: : http://jsfiddle.net/86trca4u/3/
panzoom with annotations moving in/out & stacking http://jsfiddle.net/86trca4u/4/
heavy labeling & negative labels http://jsfiddle.net/86trca4u/5/
margin(0) http://jsfiddle.net/86trca4u/6/
margin(just enough) http://jsfiddle.net/86trca4u/7/
margin not explicitly set http://jsfiddle.net/86trca4u/8/

crmorford:
http://jsfiddle.net/86trca4u/8/
if the margin isn't explicitly set, annotations don't show up (even if they could fit)

bluong:
@crmorford For the margin not being explicitly set, after inspection of the DOM, the margin is currently set at 15 pixels even though it takes a minimum of I think 23 pixels or so to show a box.  This is why nothing shows up.  I can increase the default margin so that 1 row shows up.  Does that work?

crmorford:
hm i'm not sure if it makes sense to affect the default on every axis so that ones with annotations look good right away.

is it possible to call .margin(23) on an axis, when axis.annotationsEnabled() is called, if .margin() hasn't been explicitly set?

crmorford:
the annotations are escaping the axis when you zoom out on this example http://jsfiddle.net/86trca4u/10/

crmorford:
modlog http://jsfiddle.net/86trca4u/10/
linear http://jsfiddle.net/86trca4u/11/
category http://jsfiddle.net/86trca4u/12/
time http://jsfiddle.net/86trca4u/13/

bluong:
> Strings that can be coerced to numbers, are. See '150' in the array here. I don't think that's what we want.

This is mainly a problem on the existing scales I believe as I don't do any coercion in my code.  We can file an issue to deal with scales doing the coercion.

> also, 'null' shows up, I think maybe we should filter that out.

Probably same as above, but I can filter that out.

bluong:
> is it possible to call .margin(23) on an axis, when axis.annotationsEnabled() is called, if .margin() hasn't been explicitly set?

The main problem is that I can't detect accurately how much height the annotation needs in order to render until right before rendering.  Is there a strict value that you want to use (in case people want to change their fonts)?  Also, would you rather that the margin increases when annotations are enabled or would you rather that the margin changes if margin has not been set and annotations become enabled.

crmorford:
Didn't realize that coercing strings to numbers was still a problem on scales :cry: 
We should just fix that in the right place.

talked offline about a solution to the margin() problem. Going to make a new space for annotations to go in, the margin will keep its original purpose (not holding annotations)

src/axes/axis.ts
```diff
@@ -150,9 +178,266 @@ export class Axis<D> extends Component {
...
+    let scaleRange = this._scale.range();
+    return this.annotatedTicks().filter((tick) => {
+      if (tick == null) {
+        return false;
+      }
+      return Utils.Math.inRange(this._scale.scale(tick), scaleRange[0], scaleRange[1]);
+    });
+  }
+
+  protected _axisSizeWithoutMarginAndAnnotations() {
```
jtlan: This name seems kind of flexible and unwieldy, since it's defined by what's *not* in it rather than what *is*. Could this method be replaced with a combination of other methods that return the height of each part of the `Axis`?

bluong: The things that are vital here are: the label height, the label padding, and the maximum of innerTickLength and outerTickLength.  The most troublesome thing here is the label padding, mainly because every axis hides how they measure their labels (I could abstract that up, but that would require another pr).  Also another very troublesome thing is that timeAxis does this calculation on multiple tiers.

I don't want to have a switch statement on what type of axis `this` is so a `protected` function seems to be the way to go, but the name still eludes me.  The best I thought of was possibly `_tickLabelSize`?

jtlan:
Right now, when annotations are covered and there aren't enough tiers, they get hidden entirely:
screen shot 2015-08-31 at 2 47 49 pm
http://jsfiddle.net/95q0ao2q/

I think it would probably be better if the circle was still displayed, so viewers still know that there's an annotated tick there, just not visible. @crmorford , thoughts?

crmorford:
> I think it would probably be better if the circle was still displayed, so viewers still know that there's an annotated tick there, just not visible. @crmorford , thoughts?

I think I like this idea. I'd have to see it though, it's hard to imagine

bluong:
Slightly modified the code and here is an example with a tier count of 1
screen shot 2015-08-31 at 2 59 02 pm

crmorford:
oh yep i'm a fan of that

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,593 @@ describe("BaseAxis", () => {
...
+          it("places the rectangle at the scaled x position for bottom orientation", () => {
+            let axis = new Plottable.Axis(scale, "bottom");
+            axis.annotatedTicks(annotatedTicks);
+            axis.annotationsEnabled(true);
+
+            let svg = TestMethods.generateSVG(300, 300);
+            axis.renderTo(svg);
+
+            let annotationRects = axis.content().selectAll(".annotation-rect");
+            annotationRects.each(function (d) {
```
jtlan: Can you make sure to check the size of the selection, as per [this best practice](https://github.com/palantir/plottable/wiki/Testing-best-practices#always-check-the-size-of-a-d3-selection)?

Here and elsewhere there are `each()` loops.

src/axes/axis.ts
jtlan: >The things that are vital here are: the label height, the label padding, and the maximum of innerTickLength and outerTickLength.

Not even valid on `Axes.Time`, for that matter. Could we use something like "`_coreAxisHeight`", or is that too ambiguous?

bluong: or possibly `_mainAxisHeight`?

Note that I like these names a lot better than the current name mainly because of concision

src/axes/axis.ts
```diff
@@ -150,9 +178,265 @@ export class Axis<D> extends Component {
...
+    let scaleRange = this._scale.range();
+    return this.annotatedTicks().filter((tick) => {
+      if (tick == null) {
+        return false;
+      }
+      return Utils.Math.inRange(this._scale.scale(tick), scaleRange[0], scaleRange[1]);
+    });
+  }
+
+  protected _coreAxisHeight() {
```
jtlan: Can you doc this with its intended functionality?

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+          axis.annotationsEnabled(true);
+
+          let svg = TestMethods.generateSVG(300, 300);
+          axis.renderTo(svg);
+
+          let annotationRects = axis.content().selectAll(".annotation-rect");
+
+          let annotationCircles = axis.content().selectAll(".annotation-circle");
+
+          assert.strictEqual(annotationCircles.size(), annotationRects.size(), "same number of annotation circles as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+            let annotationCircle = d3.select(this);
+            let correspondingRect = d3.select(annotationRects[0][i]);
+            assert.closeTo(TestMethods.numAttr(annotationCircle, "cx"),
+              TestMethods.numAttr(correspondingRect, "x"), window.Pixel_CloseTo_Requirement, `circle ${i} at start of rect`);
+            assert.closeTo(TestMethods.numAttr(annotationCircle, "cy"), 0, window.Pixel_CloseTo_Requirement, `circle ${i} at the top`);
+          });
+
+          let annotationLines = axis.content().selectAll(".annotation-line");
+
+          assert.strictEqual(annotationLines.size(), annotationRects.size(), "same number of annotation lines as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+            assert.closeTo(TestMethods.numAttr(annotationLine, "x2"),
+              TestMethods.numAttr(correspondingRect, "x"), window.Pixel_CloseTo_Requirement, `line ${i} at start of rect`);
+            assert.closeTo(TestMethods.numAttr(annotationLine, "y1"), 0, window.Pixel_CloseTo_Requirement, `line ${i} starts at the top`);
+            assert.closeTo(TestMethods.numAttr(annotationLine, "y2"),
+              TestMethods.numAttr(correspondingRect, "y"), window.Pixel_CloseTo_Requirement, `line ${i} goes to the rect`);
+          });
+
+          let annotationLabelTexts = axis.content().selectAll(".annotation-label text");
+
+          assert.strictEqual(annotationLabelTexts.size(), annotationRects.size(), "same number of annotation labels as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+          axis.annotationsEnabled(true);
+
+          let svg = TestMethods.generateSVG(300, 300);
+          axis.renderTo(svg);
+
+          let annotationRects = axis.content().selectAll(".annotation-rect");
+
+          let annotationCircles = axis.content().selectAll(".annotation-circle");
+
+          assert.strictEqual(annotationCircles.size(), annotationRects.size(), "same number of annotation circles as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+            let correspondingRect = d3.select(annotationRects[0][i]);
+            assert.closeTo(TestMethods.numAttr(annotationCircle, "cx"),
+              TestMethods.numAttr(correspondingRect, "x"), window.Pixel_CloseTo_Requirement, `circle ${i} at start of rect`);
+            assert.closeTo(TestMethods.numAttr(annotationCircle, "cy"), axis.height(),
+              window.Pixel_CloseTo_Requirement, `circle ${i} at the bottom`);
+          });
+
+          let annotationLines = axis.content().selectAll(".annotation-line");
+
+          assert.strictEqual(annotationLines.size(), annotationRects.size(), "same number of annotation lines as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

test/axes/baseAxisTests.ts
```diff
@@ -206,4 +206,661 @@ describe("BaseAxis", () => {
...
+            assert.closeTo(TestMethods.numAttr(annotationLine, "y1"), axis.height(),
+              window.Pixel_CloseTo_Requirement, `line ${i} starts at the bottom`);
+            assert.closeTo(TestMethods.numAttr(annotationLine, "y2"),
+              TestMethods.numAttr(correspondingRect, "y") + TestMethods.numAttr(correspondingRect, "height"),
+              window.Pixel_CloseTo_Requirement, `line ${i} goes to the rect`);
+          });
+
+          let annotationLabelTexts = axis.content().selectAll(".annotation-label text");
+
+          assert.strictEqual(annotationLabelTexts.size(), annotationRects.size(), "same number of annotation labels as rects");
```
jtlan: Probably better to assert that each selection has the same `size()` as the number of ticks.

src/axes/axis.ts
```diff
@@ -150,9 +178,271 @@ export class Axis<D> extends Component {
...
+    });
+  }
+
+  /**
+   * Retrieves the height of the core pieces of the Axis,
+   * where height is vertical for horizontal Axes and horizontal for vertical Axes.
+   *
+   * The core pieces include the axis labels, the end tick marks, the inner tick marks, and the tick label padding.
+   */
+  protected _coreAxisHeight() {
```
jtlan: "`_coreSize()`"? We know it's on `Axis`, and "height" could be confusing given the existence of `height()` (which *is* direction-dependent).

crmorford:
click to update annotations http://jsfiddle.net/86trca4u/15/
clear annotations by passing [] http://jsfiddle.net/86trca4u/16/
toggle on/off http://jsfiddle.net/86trca4u/17/

crmorford:
duplicated annotations:
when an annotation value is duplicated, only one annotation shows up, in the second tier http://jsfiddle.net/86trca4u/18/
when triplicated (word?) the annotation shows up in the third tier
http://jsfiddle.net/86trca4u/18/

I think if these are included multiple times, they should show up multiple times.

crmorford:
+1 pending duplicated annotations fix
here's a present http://jsfiddle.net/86trca4u/26/

jtlan:
One that uses `XDragBoxLayer` instead: http://jsfiddle.net/pmzwesvc/

jtlan:
Doesn't appear to solve the problem where the ticks are drawn one tier too low:
http://jsfiddle.net/pmzwesvc/

screen shot 2015-09-01 at 5 01 10 pm

jtlan:
Can you add a test for the tier position when there are duplicates?

crmorford:
uhoh, I don't see them showing up on time anymore http://jsfiddle.net/86trca4u/13/

bluong:
It seems that the `Date` string format used is incorrect 
`new Date("2-25-2014")` returns an invalid Date

Issue #2641, opened 2015-08-17
[Axes.Category] - Does not override space functions and variables.


bluong:
At the base `Axis` class, it defines a `requestedSpace()` implementation as well as functions and variables like `_computeWidth()`, `_computeHeight`, `_computedWidth()`, `_computedHeight()`.  `Axes.Category` takes a completely different approach in implementing `requestedSpace()`, which means that the values of the above variables and return value of the functions are invalid.

bluong:
Based on some investigation, there is a bunch of awkwardness with the implementation at the base level, which might mean that the above variables / functions might have to be deprecated to solve this bug.

PR #2642, opened 2015-08-17
[Axis.Category] - correct values for computedWidth and computedHeight

bluong:
QE:  This is a fix on a protected variable and should not have any effect on user-expected behavior.  Regression pass should be fine.

The protected variables `_computedWidth` and `_computedHeight` were not properly set correctly.  Thankfully, while `Axis.Category` routes around this, this means that accessing these variables can lead to errors.  Note that developers are the main user of these variables.

This is mainly a temporary band-aid for #2641 as #2620 is blocked on a reliable way to get the width/height of the axis disregarding the margin.

This affects only protected variables, which we should have probably thought about more carefully when `Axis.Category` overwrote the behavior.

### DIFF
--- a/test/axes/categoryAxisTests.ts
+++ b/test/axes/categoryAxisTests.ts
@@ -179,15 +179,16 @@ describe("Category Axes", () => {
   });
 
   it("_computed space variables should be set on requestedSpace", () => {
-    let svg = TestMethods.generateSVG(300, 300);
+    let svgWidth = 300;
+    let svgHeight = 300;
+    let svg = TestMethods.generateSVG(svgWidth, svgHeight);
     let labels = ["label1", "label2", "label100"];
     let scale = new Plottable.Scales.Category().domain(labels);
     let axis = new Plottable.Axes.Category(scale, "bottom");
     axis.anchor(svg);
-    axis.requestedSpace(300, 300);
+    axis.computeLayout({ x: 0, y: 0 }, svgWidth, svgHeight);
 
-    assert.isNotNull((<any> axis)._computedWidth, "computed width variable should be set to a value");
-    assert.isNotNull((<any> axis)._computedHeight, "computed height variable should be a set to a value");
+    assert.strictEqual((<any> axis)._computedHeight, axis.height() - axis.margin(), "computed height should be height without margin");
 
     svg.remove();
   });
--- a/test/axes/categoryAxisTests.ts
+++ b/test/axes/categoryAxisTests.ts
@@ -178,7 +178,7 @@ describe("Category Axes", () => {
     svg.remove();
   });
 
-  it("_computed space variables should be set on requestedSpace", () => {
+  it("_computedHeight should be set to height without margin", () => {
     let svgWidth = 300;
     let svgHeight = 300;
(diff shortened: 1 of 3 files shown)


jtlan:
@bluong , if this is a band-aid, what do you envision as longer-term solution to this problem?

test/axes/categoryAxisTests.ts
```diff
@@ -177,4 +177,18 @@ describe("Category Axes", () => {
...
+  it("_computed space variables should be set on requestedSpace", () => {
+    let svg = TestMethods.generateSVG(300, 300);
+    let labels = ["label1", "label2", "label100"];
+    let scale = new Plottable.Scales.Category().domain(labels);
+    let axis = new Plottable.Axes.Category(scale, "bottom");
+    axis.anchor(svg);
+    axis.requestedSpace(300, 300);
+
+    assert.isNotNull((<any> axis)._computedWidth, "computed width variable should be set to a value");
+    assert.isNotNull((<any> axis)._computedHeight, "computed height variable should be a set to a value");
```
jtlan: Should we assert something about the value itself, other than not being null?

bluong: My personal issue when I came across this issue was that `Axis.Category` just forgot to set the variable to a value.

I can try to assert a value if you think it's appropriate, it'll however take some more code to establish the space variable values.

jtlan: Maybe we should assert that they match `height()`/`width()`?

bluong: ...And as you suggested to check the actual value amount, I see a bug in my implementation.  I'll roll out a fix and a better test.

bluong:
The longer-term solution to this problem is that we need to revisit these four members, `_computedHeight`, `_computedWidth`, `_computeHeight()`, and `_computeWidth()`, and then have one of two adaptations.

We either need to have `Axes.Category` adapt itself to use those four members (Based on a glance, this seems incorrect)

Or, we need to revamp those four members, possibly to the point of deprecating them, and then come up with a suitable replacement architecture that can accomplish a similar enough task (we want to compute the width and height of the axis without the margin to use in space calculations)

jtlan:
Maybe something like "`_widthWithoutMargin()`"?

bluong:
Possibly, note that this endpoint would be out of scope for this pull request, but we should definitely consider revamping at some close point of time so that the Axis implementations are a lot more aligned.

test/axes/categoryAxisTests.ts
```diff
@@ -177,4 +177,19 @@ describe("Category Axes", () => {
 
     svg.remove();
   });
+
+  it("_computed space variables should be set on requestedSpace", () => {
```
jtlan: This test name seems wrong? We're checking that the variables are set correctly on `computeLayout()`, right?

bluong:
There are talks about if is a good idea to have this pull request merge in as a bandaid to the problem.  For now, I think I'm going to make a workaround in the blocked pull request.

bluong:
Closing this pull request.  We can aim for the full solution at a later time.  For now, I'm going to use a workaround

PR #2648, opened 2015-08-18
The batch datasets setter dispatches updates at the end in batch

### LINKED ISSUE #2418: Unnecessary quadratic slowdown when adding datasets to stacked plot

aicioara:
Part of #2418 

Suggested improvements will come as separate PRs

150 Datasets (as requested) - Speedup `100x`
Before: http://jsfiddle.net/cn9x2p5j/2/ - `3300 ms`
After: http://jsfiddle.net/cn9x2p5j/3/ - `36 ms`

Affected (perf wise):
- Plot.datasets(). This method should be preferred for multiple datasets additions

Unaffected (perf wise):
- Plot.addDataset()
- Plot.removeDataset()

QE notes - affected points:
- Plot.datasets()
- Plot.destroy()
- Plot.addDataset()
- Plot.removeDataset()

QE notes - regression: 
- Everything should behave exactly the same

QE notes - performance: 
- Plot.datasets() should be 100x faster.
- Plot.destroy() should be faster
- Everything else should be exactly the same

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -93,7 +93,8 @@ export class Plot extends Component {
   public destroy() {
     super.destroy();
     this._scales().forEach((scale) => scale.offUpdate(this._renderCallback));
-    this.datasets().forEach((dataset) => this.removeDataset(dataset));
+    this.datasets().forEach((dataset) => this._removeDataset(dataset, false));
+    this._onDatasetUpdate();
   }
 
   protected _createNodesForDataset(dataset: Dataset) {
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -385,7 +385,7 @@ export class Plot extends Component {
    */
   public removeDataset(dataset: Dataset): Plot {
     this._removeDataset(dataset);
-    this._onDatasetUpdate()
+    this._onDatasetUpdate();
     return this;
   }
 
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -361,10 +361,8 @@ export class Plot extends Component {
     return this._addDataset(dataset, true);
   }
 
-  public _addDataset(dataset: Dataset, withUpdate: boolean) {
-    // if (this.datasets().indexOf(dataset) > -1) {
-      this._removeDataset(dataset, false);
-    // };
+  private _addDataset(dataset: Dataset, dispatchUpdate: boolean) {
+    this._removeDataset(dataset, false);
     let drawer = this._createDrawer(dataset);
     this._datasetToDrawer.set(dataset, drawer);
 
@@ -373,7 +371,7 @@ export class Plot extends Component {
     }
 
     dataset.onUpdate(this._onDatasetUpdateCallback);
-    if (withUpdate) {
+    if (dispatchUpdate) {
       this._onDatasetUpdate();
     }
     return this;
@@ -389,7 +387,7 @@ export class Plot extends Component {
     return this._removeDataset(dataset, true);
   }
 
-  public _removeDataset(dataset: Dataset, withUpdate: boolean) {
+  private _removeDataset(dataset: Dataset, dispatchUpdate: boolean) {
     if (this.datasets().indexOf(dataset) === -1) {
       return this;
     }
@@ -397,7 +395,7 @@ export class Plot extends Component {
     this._removeDatasetNodes(dataset);
     dataset.offUpdate(this._onDatasetUpdateCallback);
     this._datasetToDrawer.delete(dataset);
-    if (withUpdate) {
+    if (dispatchUpdate) {
       this._onDatasetUpdate();
     }
     return this;
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -93,8 +93,7 @@ export class Plot extends Component {
   public destroy() {
     super.destroy();
     this._scales().forEach((scale) => scale.offUpdate(this._renderCallback));
-    this.datasets().forEach((dataset) => this._removeDataset(dataset));
-    this._onDatasetUpdate();
+    this.datasets([]);
   }
 
   protected _createNodesForDataset(dataset: Dataset) {
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -363,7 +363,7 @@ export class Plot extends Component {
     return this;
   }
 
-  private _addDataset(dataset: Dataset) {
+  protected _addDataset(dataset: Dataset) {
     this._removeDataset(dataset);
     let drawer = this._createDrawer(dataset);
     this._datasetToDrawer.set(dataset, drawer);
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -388,7 +388,7 @@ export class Plot extends Component {
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


aicioara:
No tests added because functionally nothing changed.

aicioara:
Also this is a straight improvement, so no new API endpoints have been added.

aicioara:
Also moved the `addDatasets()` method lines closer to `datasets()` and `removeDataset()` (no idea why they were separated).

src/plots/plot.ts
```diff
@@ -374,16 +353,50 @@ export class Plot extends Component {
...
    */
   public removeDataset(dataset: Dataset): Plot {
-    if (this.datasets().indexOf(dataset) > -1) {
-      this._removeDatasetNodes(dataset);
-      dataset.offUpdate(this._onDatasetUpdateCallback);
-      this._datasetToDrawer.delete(dataset);
+    return this._removeDataset(dataset, true);
+  }
+
+  private _removeDataset(dataset: Dataset, dispatchUpdate: boolean) {
```
bluong: Instead of having a boolean parameter here, can we have this function not do any updating and then have `removeDataset()` do the updating manually?

src/plots/plot.ts
```diff
@@ -374,16 +353,50 @@ export class Plot extends Component {
...
+   * Adds a Dataset to the Plot.
+   *
+   * @param {Dataset} dataset
+   * @returns {Plot} The calling Plot.
+   */
+  public addDataset(dataset: Dataset) {
+    return this._addDataset(dataset, true);
+  }
+
+  private _addDataset(dataset: Dataset, dispatchUpdate: boolean) {
```
bluong: Instead of having a boolean parameter here, can we have this function not do any updating and then have `addDataset()` do the updating manually?

bluong:
I was actually expecting a totally different solution to the quadratic problem, but this is a good solution as well.

This solution is specifically good since it batches all the dataset updates, which is great for general usage of `_onDatasetUpdate()`, however it does not help if the user were to add their datasets one-by-one.

Stacking is a process that luckily can be done incrementally (adding dataset 3 does not affect the stacking results for datasets 1/2) so I expected the answer to reside in our use of the stacking utils.

bluong:
Let's rename the pull request title and description, as they are not descriptive.  This seems to be more aligned with "The batch datasets setter dispatches updates at the end in batch" or something similar.

bluong:
Up to you if you want to do the incremental stacking but it is out of scope of this pull request

aicioara:
Will do incremental stacking as well. Thanks for the suggestion.

aicioara:
Changed my mind. Will do it in a separate PR. Will amend the rest and this is no longer closing the issue, but instead it is part of it.

src/plots/plot.ts
```diff
@@ -93,29 +93,8 @@ export class Plot extends Component {
...
-    let drawer = this._createDrawer(dataset);
-    this._datasetToDrawer.set(dataset, drawer);
-
-    if (this._isSetup) {
-      this._createNodesForDataset(dataset);
-    }
-
-    dataset.onUpdate(this._onDatasetUpdateCallback);
+    this.datasets().forEach((dataset) => this._removeDataset(dataset));
     this._onDatasetUpdate();
```
bluong: Instead why not
`this.datasets([]);`

crmorford:
Great summary! thanks :)

crmorford:
Something weird happening on Area 
http://jsfiddle.net/cn9x2p5j/4/
screen shot 2015-08-19 at 5 32 35 pm
and Stacked Area
http://jsfiddle.net/cn9x2p5j/5/
screen shot 2015-08-19 at 5 33 18 pm

aicioara:
Thank you for catching that. Should be fixed now.

test/plots/piePlotTests.ts
```diff
@@ -187,7 +187,8 @@ describe("Plots", () => {
           { value: 5000 },
           { value: 5000 }];
         let dataset = new Plottable.Dataset(data);
-        piePlot.addDataset(dataset).outerRadius(500);
+        piePlot.addDataset(dataset)
+        piePlot.outerRadius(500);
```
aicioara: This is because addDataset returns `Plottable.Plot` now (notice the removal in the `Plottable.Plots.Pie`). 

Is this an API breaking change? If so the code I should add is.
```typescript
// this is in piePlot.ts
public addDataset(dataset: Dataset) {
  super.addDataset(dataset);
  return this;
```

test/plots/piePlotTests.ts
```diff
@@ -187,7 +187,8 @@ describe("Plots", () => {
           { value: 5000 },
           { value: 5000 }];
         let dataset = new Plottable.Dataset(data);
-        piePlot.addDataset(dataset).outerRadius(500);
+        piePlot.addDataset(dataset);
+        piePlot.outerRadius(500);
```
aicioara: This is because addDataset returns `Plottable.Plot` now (notice the removal in the `Plottable.Plots.Pie`). 

Is this an API breaking change? If so the code I should add is.
```typescript
// this is in piePlot.ts
public addDataset(dataset: Dataset) {
  super.addDataset(dataset);
  return this;
}
```

bluong: For javascript no, for typescript yes even though the intention through typescript is visible.  I'd like to say that it's not an API breaking change, but we can do a quick survey.

crmorford:
destroy b/a render is ok
speeds up removing too http://jsfiddle.net/cn9x2p5j/6/
regression pass looks good

PR #2649, opened 2015-08-18
skip negative values in pie plot

### LINKED ISSUE #2611: Pie should blow up on negative values
Plottable currently throws a warning when the user gives negative values for a pie plot. We should blow up the plot instead, since the result is misleading:
http://jsfiddle.net/k5yaLhyp/6/
screen shot 2015-08-13 at 2 41 30 pm

ztsai:
closes #2611 
http://jsfiddle.net/ztsai/k5yaLhyp/7/

### DIFF
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -278,23 +278,31 @@ export module Plots {
       let sectorValueAccessor = Plot._scaledAccessor(this.sectorValue());
       let ds = this.datasets()[0];
       let data = dataToDraw.get(ds);
-      let filteredData = data.filter((d, i) => {
-        let value = sectorValueAccessor(d, i, ds);
-        return Plottable.Utils.Math.isValidNumber(value) && value >= 0;
-      });
+      let filteredData = data.filter((d, i) => Pie._isValidData(sectorValueAccessor(d, i, ds)));
       dataToDraw.set(ds, filteredData);
       return dataToDraw;
     }
 
+    protected static _isValidData(value: any) {
+      return Plottable.Utils.Math.isValidNumber(value) && value >= 0;
+    }
+
     protected _pixelPoint(datum: any, index: number, dataset: Dataset) {
+      let scaledValueAccessor = Plot._scaledAccessor(this.sectorValue());
+      if (!Pie._isValidData(scaledValueAccessor(datum, index, dataset))) {
+        return { x: NaN, y: NaN};
+      }
+
       let innerRadius = Plot._scaledAccessor(this.innerRadius())(datum, index, dataset);
       let outerRadius = Plot._scaledAccessor(this.outerRadius())(datum, index, dataset);
       let avgRadius = (innerRadius + outerRadius) / 2;
 
-      let scaledValueAccessor = Plot._scaledAccessor(this.sectorValue());
       let pie = d3.layout.pie()
                          .sort(null)
-                         .value((d: any, i: number) => scaledValueAccessor(d, i, dataset))(dataset.data());
+                         .value((d: any, i: number) => {
+                           let value = scaledValueAccessor(d, i, dataset);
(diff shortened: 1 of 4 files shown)


jtlan:
I think it might be better to gracefully degrade by treating negative numbers the same way we treat `NaN`, `null`, or `undefined` -- by not drawing the slice. Other `Plot`s don't throw errors on bad data, so throwing errors on negative values seems odd.

src/plots/piePlot.ts
```diff
@@ -234,11 +234,8 @@ export module Plots {
       if (this.datasets().length === 0) { return; }
       let sectorValueAccessor = Plot._scaledAccessor(this.sectorValue());
       let dataset = this.datasets()[0];
-      let data = dataset.data().filter((d, i) => Plottable.Utils.Math.isValidNumber(sectorValueAccessor(d, i, dataset)));
```
jtlan: ... this didn't use `_getDataToDraw()` before? How strange.

crmorford:
I got this to happen
screen shot 2015-08-19 at 3 30 54 pm
Sadly it was with random data :( http://jsfiddle.net/j5b9f60c/1/

crmorford:
http://jsfiddle.net/j5b9f60c/2/
here's something strange, it's reusing a color. this doesn't happen every time
screen shot 2015-08-19 at 3 32 54 pm

crmorford:
Plot.entities() logs 6 objects in this case (my assumption was that it would only log 3)
http://jsfiddle.net/j5b9f60c/3/
is this expected?

ztsai:
just a QE note: 
- if there's "NaN" in the value, all the entries after that will not be returned by entities()
- if value is a string that can be parsed as a number, it is not drawable, but it will be included in entities()

jtlan:
>if value is a string that can be parsed as a number, it is not drawable, but it will be included in entities()

Is this also true of `NaN`, `null`, and `undefined` values?

ztsai:
`null`, and `NaN` will mess up the calculation, and any data after those entries will not be returned
`undefined` and string of number will have a corresponding entity, and any other strings will not not
after the fix, any of the above should not have a corresponding entity

src/plots/piePlot.ts
```diff
@@ -249,20 +246,31 @@ export module Plots {
...
     }
 
+    protected static _isValidData(value: any) {
+      return Plottable.Utils.Math.isValidNumber(value) && value >= 0;
+    }
+
     protected _pixelPoint(datum: any, index: number, dataset: Dataset) {
+      let scaledValueAccessor = Plot._scaledAccessor(this.sectorValue());
+      if (!Pie._isValidData(scaledValueAccessor(datum, index, dataset))) {
+        return { x: NaN, y: NaN};
```
jtlan: This isn't immediately relevant because `_pixelPoint` isn't invoked for invalid slices, but I wonder what we should be returning in these cases. A few options I see:
- A valid point, with invalid `x` and `y` properties
  - `NaN`s? `null`s? `undefined`s?
- `null` instead of a `Point`

The second option is more blatant, but runs the risk of blowing up downstream code if they aren't expecting invalid values.

crmorford:
NaN http://jsfiddle.net/j5b9f60c/5/
null http://jsfiddle.net/j5b9f60c/6/
undefined http://jsfiddle.net/j5b9f60c/7/
Infinity http://jsfiddle.net/j5b9f60c/8/
-Infinity http://jsfiddle.net/j5b9f60c/9/
"##" http://jsfiddle.net/j5b9f60c/10/

non-blocker, but should be flagged: 
most of those examples had to have lines of code commented out because passing undefined/NaN/null/etc... through broke them. For example, returning d.x for .attr("fill", ...) causes a text.trim error.
labels also don't play nice with bad data

ztsai:
we could add safeguard for bad data, but I am not sure if that's necessary

Issue #2636, opened 2015-08-17
[Testing Audit] Rectangle Plot and Segment Plot


crmorford:

bluong:
Pull requests merged.  Closing

PR #2653, opened 2015-08-19
Refactor rectangle plot tests - Pass 1

aicioara:
Part of #2636

As part of this: 
- No nested `beforeEach()`es
- Correct semantics of the test titles
screen shot 2015-08-20 at 4 21 49 pm

### DIFF
--- a/test/plots/rectanglePlotTests.ts
+++ b/test/plots/rectanglePlotTests.ts
@@ -22,13 +22,17 @@ describe("Plots", () => {
         plot.renderTo(svg);
 
         let cells = plot.content().selectAll("rect");
-        assert.strictEqual(cells.size(), data.length);
+        assert.strictEqual(cells.size(), data.length, "There is one rectangle for each piece of data");
         cells.each(function(d, i) {
           let cell = d3.select(this);
-          assert.closeTo(+cell.attr("height"), 50, 0.5, "Cell height is correct");
-          assert.closeTo(+cell.attr("width"), 50, 0.5, "Cell width is correct");
-          assert.closeTo(+cell.attr("x"), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
-          assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
+          assert.closeTo(+cell.attr("height"), 50, window.Pixel_CloseTo_Requirement,
+            "Cell height is correct");
+          assert.closeTo(+cell.attr("width"), 50, window.Pixel_CloseTo_Requirement,
+            "Cell width is correct");
+          assert.closeTo(+cell.attr("x"), 25 + 50 * i, window.Pixel_CloseTo_Requirement,
+            "Cell x coordinate is correct");
+          assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), window.Pixel_CloseTo_Requirement,
+            "Cell y coordinate is correct");
         });
         svg.remove();
       });
--- a/test/plots/rectanglePlotTests.ts
+++ b/test/plots/rectanglePlotTests.ts
@@ -257,53 +257,53 @@ describe("Plots", () => {
         assert.strictEqual(cellBV.attr("fill"), "#777777", "cell 'BV' color is correct");
       };
 
-      it("renders correctly", () => {
-        let xScale = new Plottable.Scales.Category();
-        let yScale = new Plottable.Scales.Category();
-        let colorScale = new Plottable.Scales.InterpolatedColor();
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Category;
+      let yScale: Plottable.Scales.Category;
+      let colorScale: Plottable.Scales.InterpolatedColor;
+
+      beforeEach(() => {
+        svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+        xScale = new Plottable.Scales.Category();
+        yScale = new Plottable.Scales.Category();
+        colorScale = new Plottable.Scales.InterpolatedColor();
         colorScale.range(["black", "white"]);
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let gridPlot = new Plottable.Plots.Rectangle();
-        gridPlot.addDataset(new Plottable.Dataset(DATA))
-                .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
-                .y((d: any) => d.y, yScale);
-        gridPlot.renderTo(svg);
-        VERIFY_CELLS((<any> gridPlot)._renderArea.selectAll("rect")[0]);
+      });
+
+      afterEach(() => {
         svg.remove();
       });
 
-      it("renders correctly when data is set after construction", () => {
-        let xScale = new Plottable.Scales.Category();
-        let yScale = new Plottable.Scales.Category();
-        let colorScale = new Plottable.Scales.InterpolatedColor();
-        colorScale.range(["black", "white"]);
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let dataset = new Plottable.Dataset();
-        let gridPlot = new Plottable.Plots.Rectangle();
-        gridPlot.addDataset(dataset)
+      it("renders correctly", () => {
+        let plot = new Plottable.Plots.Rectangle();
+        plot.addDataset(new Plottable.Dataset(DATA))
                 .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
+        plot.x((d: any) => d.x, xScale)
+                .y((d: any) => d.y, yScale);
+        plot.renderTo(svg);
+        VERIFY_CELLS((<any> plot)._renderArea.selectAll("rect")[0]);
+      });
+
+      it("renders correctly when data is set after construction", () => {
+        let dataset = new Plottable.Dataset();
+        let plot = new Plottable.Plots.Rectangle();
+        plot.addDataset(dataset)
+                .attr("fill", (d) => d.magnitude, colorScale);
+        plot.x((d: any) => d.x, xScale)
                 .y((d: any) => d.y, yScale)
                 .renderTo(svg);
         dataset.data(DATA);
-        VERIFY_CELLS((<any> gridPlot)._renderArea.selectAll("rect")[0]);
-        svg.remove();
+        VERIFY_CELLS((<any> plot)._renderArea.selectAll("rect")[0]);
       });
 
       it("renders correctly when there isn't data for every spot", () => {
         let CELL_HEIGHT = 50;
         let CELL_WIDTH = 100;
-        let xScale = new Plottable.Scales.Category();
-        let yScale = new Plottable.Scales.Category();
-        let colorScale = new Plottable.Scales.InterpolatedColor();
-        colorScale.range(["black", "white"]);
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
         let dataset = new Plottable.Dataset();
-        let gridPlot = new Plottable.Plots.Rectangle();
-        gridPlot.addDataset(dataset)
+        let plot = new Plottable.Plots.Rectangle();
+        plot.addDataset(dataset)
                 .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
+        plot.x((d: any) => d.x, xScale)
                 .y((d: any) => d.y, yScale)
                 .renderTo(svg);
         let data = [
@@ -313,7 +313,7 @@ describe("Plots", () => {
           {x: "D", y: "Z", magnitude: 24}
         ];
         dataset.data(data);
-        let cells = (<any> gridPlot)._renderArea.selectAll("rect")[0];
+        let cells = (<any> plot)._renderArea.selectAll("rect")[0];
         assert.strictEqual(cells.length, data.length);
         for (let i = 0; i < cells.length; i++) {
           let cell = d3.select(cells[i]);
@@ -322,25 +322,19 @@ describe("Plots", () => {
           assert.strictEqual(cell.attr("width"), String(CELL_WIDTH), "Cell width is correct");
           assert.strictEqual(cell.attr("height"), String(CELL_HEIGHT), "Cell height is correct");
         }
-        svg.remove();
       });
 
       it("can invert y axis correctly", () => {
-        let xScale = new Plottable.Scales.Category();
-        let yScale = new Plottable.Scales.Category();
-        let colorScale = new Plottable.Scales.InterpolatedColor();
-        colorScale.range(["black", "white"]);
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let gridPlot = new Plottable.Plots.Rectangle();
-        gridPlot.addDataset(new Plottable.Dataset(DATA))
+        let plot = new Plottable.Plots.Rectangle();
+        plot.addDataset(new Plottable.Dataset(DATA))
                 .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
+        plot.x((d: any) => d.x, xScale)
                 .y((d: any) => d.y, yScale)
                 .renderTo(svg);
 
         yScale.domain(["U", "V"]);
 
-        let cells = (<any> gridPlot)._renderArea.selectAll("rect")[0];
+        let cells = (<any> plot)._renderArea.selectAll("rect")[0];
         let cellAU = d3.select(cells[0]);
         let cellAV = d3.select(cells[2]);
         cellAU.attr("fill", "#000000");
@@ -352,7 +346,7 @@ describe("Plots", () => {
         cellAV.attr("y", "0");
 
         yScale.domain(["V", "U"]);
-        cells = (<any> gridPlot)._renderArea.selectAll("rect")[0];
+        cells = (<any> plot)._renderArea.selectAll("rect")[0];
         cellAU = d3.select(cells[0]);
         cellAV = d3.select(cells[2]);
         cellAU.attr("fill", "#000000");
@@ -363,9 +357,12 @@ describe("Plots", () => {
         cellAV.attr("x", "0");
         cellAV.attr("y", "100");
 
-        svg.remove();
       });
     });
+
+
+
+
 
     describe("Rectangle Plot - selections()", () => {
       let SVG_WIDTH = 400;
@@ -383,15 +380,15 @@ describe("Plots", () => {
         let colorScale = new Plottable.Scales.InterpolatedColor();
         colorScale.range(["black", "white"]);
         let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let gridPlot = new Plottable.Plots.Rectangle();
+        let plot = new Plottable.Plots.Rectangle();
         let dataset = new Plottable.Dataset(DATA);
-        gridPlot.addDataset(dataset)
+        plot.addDataset(dataset)
                 .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
+        plot.x((d: any) => d.x, xScale)
                 .y((d: any) => d.y, yScale);
-        gridPlot.renderTo(svg);
+        plot.renderTo(svg);
 
-        let allCells = gridPlot.selections();
+        let allCells = plot.selections();
         assert.strictEqual(allCells.size(), 4, "all cells retrieved");
 
         svg.remove();
@@ -403,15 +400,15 @@ describe("Plots", () => {
         let colorScale = new Plottable.Scales.InterpolatedColor();
         colorScale.range(["black", "white"]);
         let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let gridPlot = new Plottable.Plots.Rectangle();
+        let plot = new Plottable.Plots.Rectangle();
         let dataset = new Plottable.Dataset(DATA);
-        gridPlot.addDataset(dataset)
+        plot.addDataset(dataset)
                 .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
+        plot.x((d: any) => d.x, xScale)
                 .y((d: any) => d.y, yScale);
-        gridPlot.renderTo(svg);
+        plot.renderTo(svg);
 
-        let allCells = gridPlot.selections([dataset]);
+        let allCells = plot.selections([dataset]);
         assert.strictEqual(allCells.size(), 4, "all cells retrieved");
         let selectionData = allCells.data();
         assert.includeMembers(selectionData, DATA, "data in selection data");
@@ -425,16 +422,16 @@ describe("Plots", () => {
         let colorScale = new Plottable.Scales.InterpolatedColor();
         colorScale.range(["black", "white"]);
         let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let gridPlot = new Plottable.Plots.Rectangle();
+        let plot = new Plottable.Plots.Rectangle();
         let dataset = new Plottable.Dataset(DATA);
-        gridPlot.addDataset(dataset)
+        plot.addDataset(dataset)
           .attr("fill", (d) => d.magnitude, colorScale);
-         gridPlot.x((d: any) => d.x, xScale)
+         plot.x((d: any) => d.x, xScale)
           .y((d: any) => d.y, yScale);
-        gridPlot.renderTo(svg);
+        plot.renderTo(svg);
 
         let dummyDataset = new Plottable.Dataset([]);
-        let allCells = gridPlot.selections([dataset, dummyDataset]);
+        let allCells = plot.selections([dataset, dummyDataset]);
         assert.strictEqual(allCells.size(), 4, "all cells retrieved");
         let selectionData = allCells.data();
         assert.includeMembers(selectionData, DATA, "data in selection data");
--- a/test/plots/rectanglePlotTests.ts
+++ b/test/plots/rectanglePlotTests.ts
@@ -14,7 +14,7 @@ describe("Plots", () => {
       ];
       let VERIFY_CELLS = (cells: d3.Selection<any>) => {
         assert.strictEqual(cells[0].length, 5);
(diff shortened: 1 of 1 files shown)


aicioara:
[NOQE] As no JS changes

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-    let VERIFY_CELLS = (cells: d3.Selection<any>) => {
-      assert.strictEqual(cells[0].length, 5);
-      cells.each(function(d: any, i: number) {
-        let cell = d3.select(this);
-        assert.closeTo(+cell.attr("height"), 50, 0.5, "Cell height is correct");
-        assert.closeTo(+cell.attr("width"), 50, 0.5, "Cell width is correct");
-        assert.closeTo(+cell.attr("x"), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
-        assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
+    describe("RectanglePlot - basics", () => {
+      let DATA = [
```
bluong: Since we're starting the refactor here...

I don't think `DATA` should be all-capped

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-    ];
-    let VERIFY_CELLS = (cells: d3.Selection<any>) => {
-      assert.strictEqual(cells[0].length, 5);
-      cells.each(function(d: any, i: number) {
-        let cell = d3.select(this);
-        assert.closeTo(+cell.attr("height"), 50, 0.5, "Cell height is correct");
-        assert.closeTo(+cell.attr("width"), 50, 0.5, "Cell width is correct");
-        assert.closeTo(+cell.attr("x"), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
-        assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
+    describe("RectanglePlot - basics", () => {
```
bluong: I feel the "RectanglePlot - " prefix is redundant.

bluong: I am really not a fan of the basics description.  It really tells the reader nothing.

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
+      let DATA = [
+        { x: 0, y: 0, x2: 1, y2: 1 },
+        { x: 1, y: 1, x2: 2, y2: 2 },
+        { x: 2, y: 2, x2: 3, y2: 3 },
+        { x: 3, y: 3, x2: 4, y2: 4 },
+        { x: 4, y: 4, x2: 5, y2: 5 }
+      ];
+      let VERIFY_CELLS = (cells: d3.Selection<any>) => {
```
bluong: Not a fan of the all caps

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-      assert.lengthOf(entities, 1, "found only one entity when querying a point inside the second rectangle");
-      assert.strictEqual(entities[0].index, 1, "entity retrieved is at index 1");
-      entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
-      assert.lengthOf(entities, 1, "found only one entity when querying a point inside the first rectangle");
-      assert.strictEqual(entities[0].index, 0, "entity retrieved is at index 0");
-      svg.remove();
-    });
+      afterEach(() => {
+        svg.remove();
+      });
```
bluong: We made a conscious decision not to put the `svg.remove()` statement in an `afterEach` block since `afterEach` will run even if the test has failed.  This allows us to quickly up the test on the browser to see what has failed visually without modifying the underlying test.

If there is a way to not run `afterEach()` if the test has failed then that works too.

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-
-      let entities = plot.entitiesIn({
-        topLeft: { x: xScale.scale(1.5), y: yScale.scale(2.5) },
-        bottomRight: { x: xScale.scale(2.5), y: yScale.scale(1.5) } });
-      assert.lengthOf(entities, 2, "retrieved 2 entities intersect with the box");
-      assert.strictEqual(entities[0].index, 1, "the entity of index 1 is retrieved");
-      assert.strictEqual(entities[1].index, 2, "the entity of index 2 is retrieved");
-      svg.remove();
-    });
+      it("renders correctly", () => {
```
bluong: This test description is really vague...

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
+        assert.lengthOf(entities, 2, "two entities when querying a point in both");
+        assert.strictEqual(entities[0].index, 0, "entity retrieved is at index 0");
+        assert.strictEqual(entities[1].index, 1, "entity retrieved is at index 1");
+        entities = plot.entitiesAt({ x: xScale.scale(4), y: yScale.scale(4) });
+        assert.lengthOf(entities, 1, "found only one entity when querying a point inside the second rectangle");
+        assert.strictEqual(entities[0].index, 1, "entity retrieved is at index 1");
+        entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
+        assert.lengthOf(entities, 1, "found only one entity when querying a point inside the first rectangle");
+        assert.strictEqual(entities[0].index, 0, "entity retrieved is at index 0");
+      });
```
bluong: Does this test not encapsulate the above test?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-      plot.x(function(d) { return d.x; }, xScale);
-      plot.y(function(d) { return d.y; }, yScale);
-      plot.y2(function(d) { return d.y2; });
-      plot.addDataset(new Plottable.Dataset(staggeredData));
-      plot.autorangeMode("y");
-      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-      plot.renderTo(svg);
+      afterEach(() => {
+        svg.remove();
+      });
```
bluong: Same comment about `svg.remove()`

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
 
-      assert.deepEqual(yScale.domain(), [0, 2], "y domain includes both visible segments");
+      it("autorangeMode(\"x\")", () => {
```
bluong: I'm unsure how I feel about the escaping here, it just makes it hard to read.

Would
"autorangeMode on x" be appropriate?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
+        let staggeredData = [
+          { y: "A", x: 0, x2: 1 },
+          { y: "B", x: 1, x2: 2 }
+        ];
```
bluong: I like renaming the data variable to be more accurate, but why is this staggered?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
 
-      xScale.domain(["A"]);
-      assert.deepEqual(yScale.domain(), [0, 1], "y domain includes only the visible segment (first)");
+        let xScale = new Plottable.Scales.Linear();
+        let yScale = new Plottable.Scales.Category();
+        xScale.padProportion(0);
```
bluong: Just curious on why this needs to be done.

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-        assert.isFalse(Plottable.Utils.Math.isNaN(+sel.attr("height")),
-          "height attribute should be valid for rectangle # " + i + ". Currently " + sel.attr("height"));
-        assert.isFalse(Plottable.Utils.Math.isNaN(+sel.attr("width")),
-          "width attribute should be valid for rectangle # " + i + ". Currently " + sel.attr("width"));
       });
 
-      svg.remove();
-    });
-  });
+      it("autorangeMode(\"y\")", () => {
```
bluong: Same comment as above on unit test names

test/plots/rectanglePlotTests.ts
bluong: Also the test name is not aligned with a scenario.  Can we rename the test so that it matches with a wanted scenario?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
-      let gridPlot = new Plottable.Plots.Rectangle();
-      gridPlot.addDataset(dataset)
-              .attr("fill", (d) => d.magnitude, colorScale);
-      gridPlot.x((d: any) => d.x, xScale)
-              .y((d: any) => d.y, yScale)
-              .renderTo(svg);
-      dataset.data(DATA);
-      VERIFY_CELLS((<any> gridPlot)._renderArea.selectAll("rect")[0]);
-      svg.remove();
+    describe("RectanglePlot - fail safe tests", () => {
```
bluong: Same comment as above that the prefix here is redundant

test/plots/rectanglePlotTests.ts
```diff
@@ -2,444 +2,418 @@
...
+      it("illegal rectangles don't get displayed", () => {
```
bluong: Can you better define what is illegal?

test/plots/rectanglePlotTests.ts
aicioara: Yup, by doing refactor in other files I've seen it's not even consistent. I'll amend

test/plots/rectanglePlotTests.ts
aicioara: Agree with point1. What is your suggestion instead of basics? Maybe "general usage".

bluong: It would be best if we can narrow down what "general usage" and "basics" mean...

Maybe "renders rectangles if x/y/width/height properties are set"?  I think the reason why our testing infrastructure has slowly decayed is like because we chose ambiguous names in the past

bluong:
Things that I am seeing after this refactor that we may want to consider
* I do not see too much point in having a RectanglePlot prefix for every describe/test.  It is boilerplate and you can see in the shown picture that it is quite obvious that the test is about a RectanglePlot.  This would be more relevant if RectanglePlot was testing behavior with other components... but then that test is no longer a unit test.
* Make the names as precise and as accurate as possible.  I believe poor naming is what got use into a state where we need a refactor due to poor decisions that were justified because of poor naming.  The names have gotten slightly better, but it can definitely be improved.

Less important, but it would help structure our tests
* It has always been interesting to me why we have words like `it` and `describe` but we simply just use them as unit test endpoints and not really semantically.  I am currently not sure how to use the describe block semantically, but the it blocks are almost certainly for things like `it ("informs the user of a success when something good happens")`.  It'd be great if we can structure our tests to be more aligned with our testing software.

test/plots/rectanglePlotTests.ts
aicioara: Not really. For the above test, the rectangles did not overlap (which is the common case for a rectangle plot). For this one they do. Since this is what the test is testing, I amended the title of the test only and left both tests in

test/plots/rectanglePlotTests.ts
aicioara: I like it too, but I don't know what "staggered" means. It is from the previous codebase. Will rename to data for now as it is less ambiguous.

test/plots/rectanglePlotTests.ts
aicioara: To avoid computing the padding as well. 

You look at the data and say "oh yea, lowest value is 0, highest one is 2, so I expect the domain to be [0, 2] (instead of [-0.2, 2.2])" Make it as easier for the reader to figure out why is the answer of the assert the one it is.

aicioara:
For your second point, you basically want each test title to begin with a verb. Doable.

aicioara:
Will merge as well as there are some conflicts with `develop`

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+    describe("Basic usage", () => {
+      let data = [
+        { x: 0, y: 0, x2: 1, y2: 1 },
+        { x: 1, y: 1, x2: 2, y2: 2 },
+        { x: 2, y: 2, x2: 3, y2: 3 },
+        { x: 3, y: 3, x2: 4, y2: 4 },
+        { x: 4, y: 4, x2: 5, y2: 5 }
+      ];
+      let verifyCells = (cells: d3.Selection<any>) => {
+        assert.strictEqual(cells[0].length, 5);
```
bluong: Shouldn't 5 be `data.length`?  I know this is carryover from the previous implementation, but if we want to do a revamp, let's see what's the best state we can put this test suite in.

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        cells.each(function(d, i) {
+          let cell = d3.select(this);
+          assert.closeTo(+cell.attr("height"), 50, 0.5, "Cell height is correct");
+          assert.closeTo(+cell.attr("width"), 50, 0.5, "Cell width is correct");
+          assert.closeTo(+cell.attr("x"), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
+          assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
+        });
+      };
```
bluong: Please verify but it seems `verifyCells` is only used to check rectangle widths and heights in a single unit test.  Can we move this there?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
-    });
+      it("retrieves the correct entity under a point", () => {
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesAt({ x: xScale.scale(2.5), y: yScale.scale(2.5) });
+        assert.lengthOf(entities, 1, "found only one entity when querying a point inside the third rectangle");
+        assert.strictEqual(entities[0].index, 2, "entity retrieved is at index 2");
+        svg.remove();
+      });
```
bluong: Is this test encapsulated by the following test?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesIn(
+          {min: xScale.scale(1.5), max: xScale.scale(2.5)},
+          {min: yScale.scale(2.5), max: yScale.scale(1.5)});
+        assert.lengthOf(entities, 2, "retrieved 2 entities intersect with the box");
+        assert.strictEqual(entities[0].index, 1, "the entity of index 1 is retrieved");
+        assert.strictEqual(entities[1].index, 2, "the entity of index 2 is retrieved");
+        svg.remove();
+      });
```
bluong: I would personally wrap the above into a describe block like
`describe("retrieving entities")`

How would that work?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
 
-      let plot = new Plottable.Plots.Rectangle();
-      plot.x(function(d) { return d.x; }, xScale);
-      plot.x2(function(d) { return d.x2; });
-      plot.y(function(d) { return d.y; }, yScale);
-      plot.addDataset(new Plottable.Dataset(staggeredData));
-      plot.autorangeMode("x");
-      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-      plot.renderTo(svg);
+      it("adjusts the xScale domain with respect to the yScale domain when autorangeMode on x", () => {
```
bluong: "when autorangeMode is set to x"?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
-      ];
+        yScale.domain(["B"]);
+        assert.deepEqual(xScale.domain(), [1, 2], "x domain includes only the visible segment (second)");
+        svg.remove();
+      });
 
-      let xScale = new Plottable.Scales.Category();
-      let yScale = new Plottable.Scales.Linear();
-      yScale.padProportion(0);
+      it("adjusts the yScale domain with respect to the xScale domain when autorangeMode on y", () => {
```
bluong: Same as above

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+            "height attribute should be valid for rectangle # " + i + ". Currently " + sel.attr("height"));
+          assert.isFalse(Plottable.Utils.Math.isNaN(+sel.attr("width")),
+            "width attribute should be valid for rectangle # " + i + ". Currently " + sel.attr("width"));
+        });
+
+        svg.remove();
+      });
+    });
+
+    describe("Grids", () => {
```
bluong: What does "Grids" mean?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+      let SVG_WIDTH = 400;
+      let SVG_HEIGHT = 200;
```
bluong: Let's camelcase this while we're here?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        plot.y((d: any) => d.y, yScale);
+        plot.renderTo(svg);
+        dataset.data(data);
+        verifyCells(plot.content().selectAll("rect")[0]);
+        svg.remove();
+      });
+
+      it("renders correctly when there isn't data for every spot", () => {
+        let CELL_HEIGHT = 50;
+        let CELL_WIDTH = 100;
```
bluong: more need to camelcase

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        cellAV = d3.select(cells[2]);
+        cellAU.attr("fill", "#000000");
+        cellAU.attr("x", "0");
+        cellAU.attr("y", "0");
+
+        cellAV.attr("fill", "#ffffff");
+        cellAV.attr("x", "0");
+        cellAV.attr("y", "100");
+        svg.remove();
+      });
```
bluong: There are a ton of attribute setters here with 0 assertions being made.  What is this test doing here?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        plot.attr("fill", (d) => d.magnitude, colorScale);
+        plot.x((d: any) => d.x, xScale);
+        plot.y((d: any) => d.y, yScale);
+        plot.renderTo(svg);
+        dataset.data(data);
+        verifyCells(plot.content().selectAll("rect")[0]);
+        svg.remove();
+      });
+
+      it("renders correctly when there isn't data for every spot", () => {
```
bluong: What does the test name mean?

bluong: I see a lot of "renders correctly".  Maybe this should be in a describe block to talk about rendering?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
-        let dataset = new Plottable.Dataset(DATA);
-        gridPlot.addDataset(dataset)
-                .attr("fill", (d) => d.magnitude, colorScale);
-        gridPlot.x((d: any) => d.x, xScale)
-                .y((d: any) => d.y, yScale);
-        gridPlot.renderTo(svg);
-
-        let allCells = gridPlot.selections();
+        let allCells = plot.selections();
         assert.strictEqual(allCells.size(), 4, "all cells retrieved");
```
bluong: Do we want to check the contents?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        colorScale.range(["black", "white"]);
+
+        plot = new Plottable.Plots.Rectangle<string, string>();
+        plot.attr("fill", (d) => d.magnitude, colorScale);
+        plot.x((d: any) => d.x, xScale);
+        plot.y((d: any) => d.y, yScale);
+
+        dataset = new Plottable.Dataset(data);
+        plot.addDataset(dataset);
+        plot.renderTo(svg);
```
bluong: Why do we have this decently specific `beforeEach`?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
-                     .label((d: any) => d.val)
-                     .renderTo(svg);
+        plot = new Plottable.Plots.Rectangle<number, number>();
+        dataset = new Plottable.Dataset(data);
+        plot.addDataset(dataset);
+        plot.x((d: any) => d.x, xScale)
+          .y((d: any) => d.y, yScale)
+          .x2((d: any) => d.x2)
+          .y2((d: any) => d.y2)
+          .label((d: any) => d.val);
```
bluong: Why this specific `beforeEach`?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
       });
 
-      it("rectangle labels disabled by default", () => {
+      it("displays rectangle labels by default", () => {
```
bluong: does not display?

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,486 @@
...
+        let svg = TestMethods.generateSVG(150, 300);
+        plot.renderTo(svg);
         let texts = svg.selectAll("text")[0].map((n: any) => d3.select(n).text());
         assert.lengthOf(texts, 0, "by default, no labels are drawn");
         svg.remove();
       });
 
-      it("rectangle labels render properly", () => {
-        rectanglePlot.labelsEnabled(true);
+      it("renders correct number of labels with the correct text", () => {
```
bluong: This sounds like two tests?

bluong:
I'm learning how to create a test hierarchy as well as I'm reading this.  I think naming and organization are still the biggest issues here.  While I remember we had a discussion about being careful with our `beforeEach()` usage, there are a few places here where we make assumptions using `beforeEach()`.  Also we should try to use `describe` blocks for better shelving.

test/plots/rectanglePlotTests.ts
aicioara: Not really. This was supposed to be renders correctly, but I wanted to be more explicit. Whenever we check for the correct text, we should check for the correct number of labels, otherwise we run into the problem of `assert for all (nothing) === true`. My suggestion, I will delete the "correct number of labels bit from the test"

test/plots/rectanglePlotTests.ts
aicioara: Correct.

test/plots/rectanglePlotTests.ts
aicioara: Not really. For the above test, the rectangles did not overlap (which is the common case for a rectangle plot). For this one they do. Since this is what the test is testing, I amended the title of the test only and left both tests in.

Commented on this before as well, but the comment got outdated.

test/plots/rectanglePlotTests.ts
aicioara: Why this change? That is a constant.

test/plots/rectanglePlotTests.ts
aicioara: Lol, this test was so broken.

test/plots/rectanglePlotTests.ts
aicioara: Because all this behavior is tested already above. Do you think it is too specific?

test/plots/rectanglePlotTests.ts
aicioara: same argument as above

test/plots/rectanglePlotTests.ts
aicioara: That would result in duplication. Maybe I'll give it a break until second pass, when I see how I deal with this in other scenarios.

test/plots/rectanglePlotTests.ts
aicioara: Done

test/plots/rectanglePlotTests.ts
aicioara: Done

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,519 @@
...
+        let xScale = new Plottable.Scales.Linear();
+        let yScale = new Plottable.Scales.Linear();
+        let plot = new Plottable.Plots.Rectangle<number, number>();
+        plot.x((d) => d.x, xScale).x2((d) => d.x2);
+        plot.y((d) => d.y, yScale).y2((d) => d.y2);
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.renderTo(svg);
+
+        let cells = plot.content().selectAll("rect");
+        assert.strictEqual(cells.size(), data.length);
```
bluong: I think we prefer a message here.

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,519 @@
...
+        cells.each(function(d, i) {
+          let cell = d3.select(this);
+          assert.closeTo(+cell.attr("height"), 50, 0.5, "Cell height is correct");
+          assert.closeTo(+cell.attr("width"), 50, 0.5, "Cell width is correct");
+          assert.closeTo(+cell.attr("x"), 25 + 50 * i, 0.5, "Cell x coordinate is correct");
+          assert.closeTo(+cell.attr("y"), 25 + 50 * (cells[0].length - i - 1), 0.5, "Cell y coordinate is correct");
```
bluong: Unsure but I think we want to use window.Pixel_CloseToRequirement here instead of 0.5

test/plots/rectanglePlotTests.ts
```diff
@@ -2,511 +2,519 @@
...
-      let data1 = [
-        { x: "A", y: 1, y2: 2, v: 1 },
-        { x: "B", y: 2, y2: 3, v: 2 },
-        { x: "C", y: 3, y2: NaN, v: 3 },
-        { x: "D", y: 4, y2: 5, v: 4 },
-        { x: "E", y: 5, y2: 6, v: 5 },
-        { x: "F", y: 6, y2: 7, v: 6 }
+    describe("Plots based on Category Scales", () => {
+      let SVG_WIDTH = 400;
+      let SVG_HEIGHT = 200;
```
bluong: use `const` instead

aicioara: outside the scope of this PR. Will do that once we can enforce it with TSLINT

bluong:
I think this is a huge step up, but probably still quite a bit of work left to go.  Let's do these minor changes and then accumulate more ideas to do later

aicioara:
I will do a second pass anyway.

PR #2654, opened 2015-08-19
Refactor segment plot tests

aicioara:
Part of #2636

As part of this: 
- No nested `beforeEach()`es
- Added `afterEach()` to save some LOCs
screen shot 2015-08-20 at 4 20 48 pm

### DIFF
--- a/test/plots/segmentPlotTests.ts
+++ b/test/plots/segmentPlotTests.ts
@@ -45,8 +45,8 @@ describe("Plots", () => {
         plot.addDataset(new Plottable.Dataset(data));
         plot.renderTo(svg);
 
-        plot.content().selectAll("line")[0].forEach((line) => {
-          let lineSelection = d3.select(line);
+        plot.content().selectAll("line").each(function() {
+          let lineSelection = d3.select(this);
           assert.strictEqual(lineSelection.attr("x1"), lineSelection.attr("x2"), "line is vertical");
         });
         svg.remove();
@@ -60,8 +60,8 @@ describe("Plots", () => {
         plot.addDataset(new Plottable.Dataset(data));
         plot.renderTo(svg);
 
-        plot.content().selectAll("line")[0].forEach((line) => {
-          let lineSelection = d3.select(line);
+        plot.content().selectAll("line").each(function() {
+          let lineSelection = d3.select(this);
           assert.strictEqual(lineSelection.attr("y1"), lineSelection.attr("y2"), "line is horizontal");
         });
         svg.remove();
--- a/test/plots/segmentPlotTests.ts
+++ b/test/plots/segmentPlotTests.ts
@@ -19,10 +19,6 @@ describe("Plots", () => {
         yScale = new Plottable.Scales.Linear();
       });
 
-      afterEach(() => {
-        svg.remove();
-      });
-
       it("renders a line properly", () => {
         let plot = new Plottable.Plots.Segment();
         plot.x((d) => d.x, xScale);
@@ -38,6 +34,7 @@ describe("Plots", () => {
         assert.strictEqual(+line.attr("x2"), 437.5, "x2 is correct");
         assert.strictEqual(+line.attr("y1"), 437.5, "y1 is correct");
         assert.strictEqual(+line.attr("y2"), 62.5, "y2 is correct");
+        svg.remove();
       });
 
       it("renders vertical lines when x2 is not set", () => {
@@ -52,6 +49,7 @@ describe("Plots", () => {
           let lineSelection = d3.select(line);
           assert.strictEqual(lineSelection.attr("x1"), lineSelection.attr("x2"), "line is vertical");
         });
+        svg.remove();
       });
 
       it("renders horizontal lines when y2 is not set", () => {
@@ -66,6 +64,7 @@ describe("Plots", () => {
           let lineSelection = d3.select(line);
           assert.strictEqual(lineSelection.attr("y1"), lineSelection.attr("y2"), "line is horizontal");
         });
+        svg.remove();
       });
 
       it("autorangeMode(\"x\")", () => {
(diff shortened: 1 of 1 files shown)


aicioara:
[NOQE] as no JS changes

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
+      it("renders a line properly", () => {
+        let plot = new Plottable.Plots.Segment();
+        plot.x((d) => d.x, xScale);
+        plot.x2((d) => d.x2);
+        plot.y((d) => d.y, yScale);
+        plot.y2((d) => d.y2);
+        plot.addDataset(new Plottable.Dataset([data[0]]));
+        plot.renderTo(svg);
+        renderArea = (<any> plot)._renderArea;
+        let lineSelection = d3.select(renderArea.selectAll("line")[0][0]);
```
jtlan: Probably better to just use `content()`:
```Typescript
let lineSelection = plot.content().select("line");
```

Probably should also assert that there is only one line.

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
-    });
+        plot.renderTo(svg);
 
-    it("autorangeMode(\"y\")", () => {
-      let staggeredData = [
-        { x: 0, y: 0, y2: 1 },
-        { x: 1, y: 1, y2: 2 }
-      ];
-      yScale.padProportion(0);
+        assert.deepEqual(xScale.domain(), [0, 2], "y domain includes both visible segments");
```
jtlan: "x Domain", right?

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
+
+        yScale.domain([-0.5, 0.5]);
+        assert.deepEqual(xScale.domain(), [0, 1], "y domain includes only the visible segment (first)");
```
jtlan: "x Domain", right?

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
 
-      let plot = new Plottable.Plots.Segment();
-      plot.x(function(d) { return d.x; }, xScale);
-      plot.y(function(d) { return d.y; }, yScale);
-      plot.y2(function(d) { return d.y2; });
-      plot.addDataset(new Plottable.Dataset(staggeredData));
-      plot.autorangeMode("y");
-      plot.renderTo(svg);
+        yScale.domain([0.5, 1.5]);
+        assert.deepEqual(xScale.domain(), [1, 2], "y domain includes only the visible segment (second)");
```
jtlan: "x Domain", right?

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
+        xScale.domain([-0.5, 0.5]);
+        assert.deepEqual(yScale.domain(), [0, 1], "y domain includes only the visible segment (first)");
+
+        xScale.domain([0.5, 1.5]);
+        assert.deepEqual(yScale.domain(), [1, 2], "y domain includes only the visible segment (second)");
+      });
     });
 
-    describe("entitiesIn() returns segments that intersect with the given constraints", () => {
+    describe("SegmentPlot - entitiesIn()", () => {
```
jtlan: It's nested inside the a `describe("SegmentPlot)"`, so it's probably not worth repeating the `Plot` type here.

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
+        xScale = new Plottable.Scales.Linear();
+        yScale = new Plottable.Scales.Linear();
+        plot = new Plottable.Plots.Segment<number, number>()
           .x((d) => d.x, xScale).x2((d) => d.x2)
           .y((d) => d.y, yScale).y2((d) => d.y2);
         plot.addDataset(new Plottable.Dataset(data)).renderTo(svg);
+      });
+
+      afterEach(() => {
+        svg.remove();
```
jtlan: I've been wondering if we should do this, instead of having a separate `svg.remove()` at the end of each block. Opinions?

aicioara: I was actually thinking about the same thing. What are the advantages of doing that in your opinion?

To me, it seems like an overhead to need to remove the svg all the time. The removal of the SVG has nothing to do with the `segmentPlot` itself. Instead, it is part of the cleaning process. Think of it as you setup the sandbox, you do your testing and you remove the sandbox. The sandbox removal should have nothing to do with the testing itself. Maybe we should test the assertion that the svg can be removed somewhere else and then start using it as an axiom (which we already do).

Can't really see the advantages.

test/plots/segmentPlotTests.ts
```diff
@@ -2,159 +2,168 @@
...
-      plot.renderTo(svg);
+      it("renders vertical lines when x2 is not set", () => {
+        let plot = new Plottable.Plots.Segment();
+        plot.x((d) => d.x, xScale);
+        plot.y((d) => d.y, yScale);
+        plot.y2((d) => d.y2);
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.renderTo(svg);
+
+        plot.content().selectAll("line")[0].forEach((line) => {
```
jtlan: The "more d3" way to iterate over would be to do:
```Typescript
plot.content().selectAll("line").each(function() {
  let lineSelection = d3.select(this); // "this" refers to the DOM node inside the loop
  ...
});
```
However, I think the syntax you used is clearer. We should standardize around one or the other.

test/plots/segmentPlotTests.ts
jtlan: The only difficulty is if there's a test that requires multiple `<svg>`s -- some of the `Component` tests do this. In that case though we can just clean up the second `<svg>` at the end of that test.

jtlan: Also, if `afterEach()` executes even if the test fails, it will remove the `<svg>` of the failed test, making it harder to see what went wrong.

aicioara: Makes sense. Will amend.

aicioara:
Applying the feedback from the other 2 PRs

Issue #2623, opened 2015-08-17
[Testing Audit] SelectionBoxLayer, GuideLineLayer, and descendants


crmorford:

aicioara:
Note for self, this means:
- `dragBoxLayerTests.ts`
- `guidelinesLayerTests.ts`
- `selectionBoxLayerTests.ts`
- `xDragBoxLayerTests.ts`
- `yDragBoxLayerTests.ts`

PR #2655, opened 2015-08-19
Refactor drag box layer tests - Part 1

aicioara:
Part of #2623 

As part of this: 
- No nested `beforeEach()`es
- Added `afterEach()` to save some LOCs
screen shot 2015-08-20 at 4 28 31 pm

[NOQE] as no JS changes

### DIFF
--- a/test/components/dragBoxLayerTests.ts
+++ b/test/components/dragBoxLayerTests.ts
@@ -132,6 +132,34 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
+      it("can register two onDragStart callbacks on the same component", () => {
+        dbl.renderTo(svg);
+
+        let callbackDragStart1Called = false;
+        let callbackDragStart2Called = false;
+
+        let callbackDragStart1 = () => callbackDragStart1Called = true;
+        let callbackDragStart2 = () => callbackDragStart2Called = true;
+
+        dbl.onDragStart(callbackDragStart1);
+        dbl.onDragStart(callbackDragStart2);
+
+        let target = dbl.background();
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+
+        assert.isTrue(callbackDragStart1Called, "the callback 1 for drag start was called");
+        assert.isTrue(callbackDragStart2Called, "the callback 2 for drag start was called");
+
+        dbl.offDragStart(callbackDragStart1);
+        callbackDragStart1Called = false;
+        callbackDragStart2Called = false;
+
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackDragStart1Called, "the callback 1 for drag start was disconnected");
+        assert.isTrue(callbackDragStart2Called, "the callback 2 for drag start is still connected");
+        svg.remove();
+      });
+
       it("calls the onDrag callback", () => {
         dbl.renderTo(svg);
 
@@ -155,6 +183,34 @@ describe("Interactive Components", () => {
 
         TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
         assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+        svg.remove();
+      });
+
+      it("can register two onDrag callbacks on the same component", () => {
+        dbl.renderTo(svg);
+
+        let callbackDrag1Called = false;
+        let callbackDrag2Called = false;
+
+        let callbackDrag1 = () => callbackDrag1Called = true;
+        let callbackDrag2 = () => callbackDrag2Called = true;
+
+        dbl.onDrag(callbackDrag1);
+        dbl.onDrag(callbackDrag2);
+
+        let target = dbl.background();
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+
+        assert.isTrue(callbackDrag1Called, "the callback 1 for drag was called");
+        assert.isTrue(callbackDrag2Called, "the callback 2 for drag was called");
+
+        dbl.offDrag(callbackDrag1);
+        callbackDrag1Called = false;
+        callbackDrag2Called = false;
+
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackDrag1Called, "the callback 1 for drag was disconnected");
+        assert.isTrue(callbackDrag2Called, "the callback 2 for drag is still connected");
         svg.remove();
       });
 
@@ -183,31 +239,31 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
-      it("can register two callbacks for the same event", () => {
+      it("can register two onDragEnd callbacks on the same component", () => {
         dbl.renderTo(svg);
 
-        let callbackDragStart1Called = false;
-        let callbackDragStart2Called = false;
+        let callbackDragEnd1Called = false;
+        let callbackDragEnd2Called = false;
 
-        let callbackDragStart1 = () => callbackDragStart1Called = true;
-        let callbackDragStart2 = () => callbackDragStart2Called = true;
+        let callbackDragEnd1 = () => callbackDragEnd1Called = true;
+        let callbackDragEnd2 = () => callbackDragEnd2Called = true;
 
-        dbl.onDragStart(callbackDragStart1);
-        dbl.onDragStart(callbackDragStart2);
+        dbl.onDragEnd(callbackDragEnd1);
+        dbl.onDragEnd(callbackDragEnd2);
 
         let target = dbl.background();
         TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
 
-        assert.isTrue(callbackDragStart1Called, "the callback 1 for drag start was called");
-        assert.isTrue(callbackDragStart2Called, "the callback 2 for drag start was called");
+        assert.isTrue(callbackDragEnd1Called, "the callback 1 for drag end was called");
+        assert.isTrue(callbackDragEnd2Called, "the callback 2 for drag end was called");
 
-        dbl.offDragStart(callbackDragStart1);
-        callbackDragStart1Called = false;
-        callbackDragStart2Called = false;
+        dbl.offDragEnd(callbackDragEnd1);
+        callbackDragEnd1Called = false;
+        callbackDragEnd2Called = false;
 
         TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
-        assert.isFalse(callbackDragStart1Called, "the callback 1 for drag start was disconnected");
-        assert.isTrue(callbackDragStart2Called, "the callback 2 for drag start is still connected");
+        assert.isFalse(callbackDragEnd1Called, "the callback 1 for drag end was disconnected");
+        assert.isTrue(callbackDragEnd2Called, "the callback 2 for drag end is still connected");
         svg.remove();
       });
 
--- a/test/components/dragBoxLayerTests.ts
+++ b/test/components/dragBoxLayerTests.ts
@@ -7,12 +7,24 @@ describe("Interactive Components", () => {
       let SVG_WIDTH = 400;
       let SVG_HEIGHT = 400;
 
-      var svg: d3.Selection<void>;
+      let svg: d3.Selection<void>;
       let dbl: Plottable.Components.DragBoxLayer;
+      let quarterPoint: Plottable.Point;
+      let halfPoint: Plottable.Point;
 
       beforeEach(() => {
         svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
         dbl = new Plottable.Components.DragBoxLayer();
+
+        quarterPoint = {
+          x: SVG_WIDTH / 4,
+          y: SVG_HEIGHT / 4
+        };
+        halfPoint = {
+          x: SVG_WIDTH / 2,
+          y: SVG_HEIGHT / 2
+        };
+
       });
 
       afterEach(() => {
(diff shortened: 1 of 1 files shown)


test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
-        y: SVG_HEIGHT / 2
-      };
 
-      let target = dbl.background();
-      TestMethods.triggerFakeDragSequence(target, startPoint, endPoint);
-      assert.isTrue(dbl.boxVisible(), "box is drawn on drag");
-      let bounds = dbl.bounds();
-      assert.deepEqual(bounds.topLeft, startPoint, "top-left point was set correctly");
-      assert.deepEqual(bounds.bottomRight, endPoint, "bottom-right point was set correctly");
+    describe("DragBoxLayer - basics", () => {
```
ztsai: maybe a better description?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+        halfPoint = {
+          x: SVG_WIDTH / 2,
+          y: SVG_HEIGHT / 2
+        };
+      });
 
-      let target = dbl.background();
-      TestMethods.triggerFakeDragSequence(target, targetPoint, targetPoint);
+      afterEach(() => {
+        svg.remove();
```
ztsai: this will get called even the test fails; I remember being told that this is not desired

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
-      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-      let dbl = new Plottable.Components.DragBoxLayer();
-      dbl.renderTo(svg);
-      TestMethods.verifyClipPath(dbl);
-      let clipRect = (<any> dbl)._boxContainer.select(".clip-rect");
-      assert.strictEqual(TestMethods.numAttr(clipRect, "width"), SVG_WIDTH, "the clipRect has an appropriate width");
-      assert.strictEqual(TestMethods.numAttr(clipRect, "height"), SVG_HEIGHT, "the clipRect has an appropriate height");
-      svg.remove();
-    });
+      it("dismisses on click", () => {
```
ztsai: I don't think the test matches the description
the dragbox was never drawn to begin with

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
-        let corner = d3.select(this);
-        assert.strictEqual(corner.attr("r"), radius, "corner radius was set correctly");
-      });
-
-      // HACKHACK: chai-assert.d.ts has the wrong signature
-      (<any> assert.throws)(() => dbl.detectionRadius(-1), Error, "", "rejects negative values");
 
-      svg.remove();
-    });
+      it("detectionRadius()", () => {
```
ztsai: the description doesn't tell much

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+        assert.doesNotThrow(() => dbl.detectionRadius(3), Error, "can set detection radius before anchoring");
```
ztsai: "cannot" set detection radius?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+        dbl.renderTo("svg");
 
-    it("onDragStart()", () => {
-      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-      let dbl = new Plottable.Components.DragBoxLayer();
-      dbl.renderTo(svg);
+        let radius = 5;
+        dbl.detectionRadius(radius);
+        assert.strictEqual(dbl.detectionRadius(), radius, "can retrieve the detection radius");
```
ztsai: are you testing that it can retrieve detection radius or it's set properly?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
-      let startPoint = {
-        x: SVG_WIDTH / 4,
-        y: SVG_HEIGHT / 4
-      };
-      let endPoint = {
-        x: SVG_WIDTH / 2,
-        y: SVG_HEIGHT / 2
-      };
+        // HACKHACK: chai-assert.d.ts has the wrong signature
+        (<any> assert.throws)(() => dbl.detectionRadius(-1), Error, "", "rejects negative values");
```
ztsai: can you verify that the hack is still needed?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
 
-    it("onDragStart()", () => {
-      let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-      let dbl = new Plottable.Components.DragBoxLayer();
-      dbl.renderTo(svg);
+        let radius = 5;
+        dbl.detectionRadius(radius);
+        assert.strictEqual(dbl.detectionRadius(), radius, "can retrieve the detection radius");
+        let edges = dbl.content().selectAll("line");
+        edges.each(function() {
```
ztsai: have you drawn anything?
I think `edges` is just an empty selection...

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+          let edge = d3.select(this);
+          assert.strictEqual(edge.style("stroke-width"), 2 * radius, "edge width was set correctly");
+        });
+        let corners = dbl.content().selectAll("circle");
+        corners.each(function() {
```
ztsai: same here

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
-        callbackCalled = true;
-      };
-      dbl.onDragStart(callback);
+      it("destroy() does not error if scales are not inputted", () => {
+        assert.doesNotThrow(() => dbl.destroy(), Error, "can destroy");
+      });
 
-      let target = dbl.background();
-      TestMethods.triggerFakeDragSequence(target, startPoint, endPoint);
+      it("onDragStart()", () => {
```
ztsai: description

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
 
-      let startPoint = {
-        x: SVG_WIDTH / 4,
-        y: SVG_HEIGHT / 4
-      };
-      let endPoint = {
-        x: SVG_WIDTH / 2,
-        y: SVG_HEIGHT / 2
-      };
+      it("onDrag()", () => {
```
ztsai: description

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
 
-      TestMethods.triggerFakeDragSequence(target, startPoint, endPoint);
-      assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+      });
 
-      svg.remove();
-    });
+      it("onDragEnd()", () => {
```
ztsai: description

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+      });
 
-      assert.isTrue(callbackCalled, "the callback was called");
-      assert.deepEqual(receivedBounds.topLeft, startPoint, "top-left point was set correctly");
-      assert.deepEqual(receivedBounds.bottomRight, endPoint, "bottom-right point was set correctly");
-      dbl.offDragEnd(callback);
-      callbackCalled = false;
+      it("multiple drag interaction callbacks", () => {
```
ztsai: can you split this into 3 test cases?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
         assert.isFalse(dbl.enabled(), "drag box layer reports when it is disabled");
-        svg.remove();
       });
 
       it("disables box when enabled(false)", () => {
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let dbl = new Plottable.Components.DragBoxLayer();
         dbl.enabled(false);
         dbl.renderTo(svg);
         assert.isFalse(dbl.boxVisible(), "box is hidden initially");
```
ztsai: don't you need to fake a drag sequence before assertion?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,241 @@
...
+        svg.remove();
+      });
 
-    describe("enabling/disabling", () => {
       it("enabled(boolean) properly modifies the state", () => {
-        let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
-        let dbl = new Plottable.Components.DragBoxLayer();
         assert.isTrue(dbl.enabled(), "drag box layer is enabled by default");
         assert.strictEqual(dbl.enabled(false), dbl, "enabled(boolean) returns itself");
         assert.isFalse(dbl.enabled(), "drag box layer reports when it is disabled");
```
ztsai: maybe a better assertion message here

test/components/dragBoxLayerTests.ts
```diff
@@ -317,7 +272,6 @@ describe("Interactive Components", () => {
       });
```
ztsai: this is not in the test description 
can you either update the description or move this into another test?

test/components/dragBoxLayerTests.ts
```diff
@@ -328,7 +282,10 @@ describe("Interactive Components", () => {
...
```
ztsai: same here

ztsai:
just wondering, is it necessary to put `DragBoxLayer` as a prefix in all describe blocks?  after all it's the name of the parent block

test/components/dragBoxLayerTests.ts
aicioara: Nice catch

test/components/dragBoxLayerTests.ts
aicioara: Good catch!

test/components/dragBoxLayerTests.ts
aicioara: It is split above. This test just checks if having all 3 callbacks in the same time work. What are the 3 test cases you were thinking about?

ztsai: I thought by multiple callbacks it means that you can register more than one callback onDrag/onDragStart/onDragEnd
I think you are testing 2 things in one test case here:
one is you can register more than one callback for each event, and another is that all 3 events work together?

aicioara: Indeed. I'll split it into 2

aicioara:
Amazing review. Thank you very much. Managed to apply suggestions from here to my other refactor PRs.

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,272 @@
...
+        TestMethods.verifyClipPath(dbl);
+        let clipRect = (<any> dbl)._boxContainer.select(".clip-rect");
+        assert.strictEqual(TestMethods.numAttr(clipRect, "width"), SVG_WIDTH, "the clipRect has an appropriate width");
+        assert.strictEqual(TestMethods.numAttr(clipRect, "height"), SVG_HEIGHT, "the clipRect has an appropriate height");
+        svg.remove();
+      });
+
+      it("can get and set the detection radius", () => {
+        assert.strictEqual(dbl.detectionRadius(), 3, "there is a default detection radius");
+        assert.doesNotThrow(() => dbl.detectionRadius(4), Error, "can set detection radius before anchoring");
```
ztsai: "cannot set detection"

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,272 @@
...
-      TestMethods.triggerFakeDragSequence(target, startPoint, endPoint);
-      assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackCalled, "the callback was detached from the dragoBoxLayer and not called");
+        svg.remove();
+      });
 
-      svg.remove();
-    });
+      it("can register two callbacks for the same event", () => {
```
ztsai: do you need to test the same thing for the other 2 events?

test/components/dragBoxLayerTests.ts
```diff
@@ -2,284 +2,272 @@
...
+        dbl.offDrag(callbackDrag);
+        dbl.offDragEnd(callbackDragEnd);
+
+        callbackDragStartCalled = false;
+        callbackDragCalled = false;
+
+        TestMethods.triggerFakeDragSequence(target, quarterPoint, halfPoint);
+        assert.isFalse(callbackDragStartCalled, "the callback for drag start was disconnected");
+        assert.isFalse(callbackDragCalled, "the callback for drag was disconnected");
+        assert.isTrue(callbackDragEndCalled, "the callback for drag end is still connected");
```
ztsai: this is true because you didn't set it to false not because it's still connected(you unregistered callback above)

test/components/dragBoxLayerTests.ts
aicioara: `doesNotThrow()`. I got tricked by this 3 times while refactoring. The signature is a bit weird.

test/components/dragBoxLayerTests.ts
aicioara: That would not scale. If I have (say) 10 events, that would be 45 test cases C(2 5)

test/components/dragBoxLayerTests.ts
aicioara: nice catch

test/components/dragBoxLayerTests.ts
ztsai: it would just be one per event, and in each test you call onEventName twice to register 2 different callbacks
I think it's necessary as the code path is different for each event dispatching

aicioara: That calculation was an oversight from my side. However I still think that would be too much. Semantically, all we want to do is check if the callbackSet calls all callbacks or just one. Can you iterate on the "different patch" idea and how is that not caught in the other tests?

ztsai: in theory, each line of the code should be tested. (I don't think we have coverage test yet?)
right now each `onEvent` is an one-liner calling ` this._eventCallbacks.add(callback);`
what if someone makes a mistake by emptying `this._eventCallbacks` before adding for one of the events, it will not be caught if we only test multiple callback for one event

aicioara: I agree with the above. However I don't want to run into the issue of overtesting our code. That would make tests
- Unmaintainable
- Too long to run
- Redundant

Coverage wise, I don't see how adding the 2 extra tests would improve the effective line of code coverage. (but this is not really an argument)

@jtlan @bluong comments?

jtlan: @ztsai We do have coverage tests (see http://localhost:9999/test/coverage.html). However, I think it's best to think of coverage percentage as necessary but not sufficient for good tests.

I understand the concern about over-coverage. However, in this case, the ability to register multiple callbacks is a key feature of the `onEvent()` methods, so I think it is worth testing.

aicioara: For completeness, the second part of the comment was for me and I applied

PR #2656, opened 2015-08-19
allow InterpolatedColorLegend to span entire space

### LINKED ISSUE #1568: Size of InterpolatedColorLegend

ztsai:
closes #1568
demo: http://jsfiddle.net/ztsai/mwowegms/11

```expands(expands?: boolean)``` gets/sets whether the legend will expand in the long direction
if `expands` is set to `true` then the legend would span either the width or height depending on the orientation

### DIFF
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -3,7 +3,7 @@
 module Plottable {
 export module Components {
   export class InterpolatedColorLegend extends Component {
-    private static _DEFAULT_NUM_SWATCHES = 10;
+    private static _DEFAULT_NUM_SWATCHES = 11;
 
     private _measurer: SVGTypewriter.Measurers.Measurer;
     private _wrapper: SVGTypewriter.Wrappers.Wrapper;
@@ -230,6 +230,9 @@ export module Components {
 
       if (this.expands()) {
         let textHeight = this._measurer.measure().height;
+        if (textHeight === 0) {
+          return this;
+        }
         let offset = this._isVertical() ? 2 * padding :  4 * padding - text0Width - text1Width;
         let fullLength = this._isVertical() ? this.height() : this.width();
         numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), numSwatches);
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -3,15 +3,16 @@
 module Plottable {
 export module Components {
   export class InterpolatedColorLegend extends Component {
+    private static _DEFAULT_NUM_SWATCHES = 10;
+
     private _measurer: SVGTypewriter.Measurers.Measurer;
     private _wrapper: SVGTypewriter.Wrappers.Wrapper;
     private _writer: SVGTypewriter.Writers.Writer;
     private _scale: Scales.InterpolatedColor;
     private _orientation: String ;
     private _padding = 5;
-    private _numSwatches = 10;
     private _formatter: Formatter;
-    private _expand: boolean;
+    private _expands: boolean;
 
     private _swatchContainer: d3.Selection<void>;
     private _swatchBoundingBox: d3.Selection<void>;
@@ -44,7 +45,7 @@ export module Components {
       this._scale.onUpdate(this._redrawCallback);
       this._formatter = Formatters.general();
       this._orientation = "horizontal";
-      this._expand = false;
+      this._expands = false;
 
       this.addClass("legend");
       this.addClass("interpolated-color-legend");
@@ -76,21 +77,21 @@ export module Components {
     }
 
     /**
-     * Gets whether InterpolatedColorLegend will expand in the long direction
+     * Gets whether the InterpolatedColorLegend expands to occupy all offered space in the long direction
      */
-    public expand(): boolean;
+    public expands(): boolean;
     /**
-     * Sets whether InterpolatedColorLegend will expand in the long direction
+     * Sets whether the InterpolatedColorLegend expands to occupy all offered space in the long direction
      *
-     * @param {expand} boolean
+     * @param {expands} boolean
      * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
      */
-    public expand(expand: boolean): InterpolatedColorLegend;
-    public expand(expand?: boolean): any {
-      if (expand == null) {
-        return this._expand;
+    public expands(expands: boolean): InterpolatedColorLegend;
+    public expands(expands?: boolean): any {
+      if (expands == null) {
+        return this._expands;
       }
-      this._expand = expand;
+      this._expands = expands;
       this.redraw();
       return this;
     }
@@ -126,14 +127,14 @@ export module Components {
     }
 
     public fixedWidth() {
-      return !this.expand() || this._isVertical();
+      return !this.expands() || this._isVertical();
     }
 
     public fixedHeight() {
-      return !this.expand() || !this._isVertical();
+      return !this.expands() || !this._isVertical();
     }
 
-    private _generateTicks(numSwatches = this._numSwatches) {
+    private _generateTicks(numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES) {
       let domain = this._scale.domain();
       let slope = (domain[1] - domain[0]) / (numSwatches - 1);
       let ticks: number[] = [];
@@ -164,14 +165,15 @@ export module Components {
 
       let desiredHeight: number;
       let desiredWidth: number;
+      let numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES;
       if (this._isVertical()) {
         let longestWidth = Utils.Math.max(labelWidths, 0);
         desiredWidth = this._padding + textHeight + this._padding + longestWidth + this._padding;
-        desiredHeight = this._padding + this._numSwatches * textHeight + this._padding;
+        desiredHeight = this._padding + numSwatches * textHeight + this._padding;
       } else {
         desiredHeight = this._padding + textHeight + this._padding;
         desiredWidth = this._padding + labelWidths[0] + this._padding
-                        + this._numSwatches * textHeight
+                        + numSwatches * textHeight
                         + this._padding + labelWidths[1] + this._padding;
       }
 
@@ -224,13 +226,13 @@ export module Components {
         height: 0
       };
 
-      let numSwatches = this._numSwatches;
+      let numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES;
 
-      if (this.expand()) {
+      if (this.expands()) {
         let textHeight = this._measurer.measure().height;
         let offset = this._isVertical() ? 2 * padding :  4 * padding - text0Width - text1Width;
         let fullLength = this._isVertical() ? this.height() : this.width();
-        numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), this._numSwatches);
+        numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), numSwatches);
       }
 
       if (this._isVertical()) {
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -11,7 +11,7 @@ export module Components {
     private _padding = 5;
     private _numSwatches = 10;
     private _formatter: Formatter;
-    private _fixedSize: boolean;
+    private _expand: boolean;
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/components/interpolatedColorLegend.ts [hunk 1]
```diff
@@ -73,6 +75,26 @@ export module Components {
...
+     * Gets whether InterpolatedColorLegend is fixed size.
+     */
+    public fixedSize(): boolean;
+    /**
+     * Sets whether InterpolatedColorLegend is fixed size.
+     *
+     * @param {fixed} boolean
+     * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+     */
+    public fixedSize(fixed: boolean): InterpolatedColorLegend;
```
jtlan: Param name should be "`fixedSize`" in our style.

jtlan:
It's a little odd to have `fixedSize()`, `fixedWidth()`, and `fixedSize()`. Maybe a different name would make more sense?

We should also get some input on if we want to expand the cells, or use more cells.

ztsai:
talked to Ben offline and he said he is fine with either way; 
decided to go with the more swatches based on replies in the RFC thread

src/components/interpolatedColorLegend.ts
```diff
@@ -73,6 +75,26 @@ export module Components {
...
+     * Gets whether InterpolatedColorLegend will expand in the long direction
+     */
+    public expand(): boolean;
+    /**
+     * Sets whether InterpolatedColorLegend will expand in the long direction
+     *
+     * @param {expand} boolean
+     * @returns {InterpolatedColorLegend} The calling InterpolatedColorLegend.
+     */
+    public expand(expand: boolean): InterpolatedColorLegend;
```
jtlan: "expand**s**" would probably be a better name, since it's an adjective.

src/components/interpolatedColorLegend.ts
```diff
@@ -73,6 +75,26 @@ export module Components {
       return this;
     }
 
+    /**
+     * Gets whether InterpolatedColorLegend will expand in the long direction
```
jtlan: "will expand to occupy all offered space in the long direction."

src/components/interpolatedColorLegend.ts
```diff
@@ -73,6 +75,26 @@ export module Components {
...
+     */
+    public expand(): boolean;
+    /**
+     * Sets whether InterpolatedColorLegend will expand in the long direction
```
jtlan: Same doc change here.

test/components/interpolatedColorLegendTests.ts
```diff
@@ -187,4 +188,102 @@ describe("InterpolatedColorLegend", () => {
...
+
+    svg.remove();
+  });
+
+  it("spams the entire height if oriented vertically and expand is set to true", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.orientation("left");
+    legend.expand(true);
+    legend.renderTo(svg);
+    assert.closeTo(legend.height(), SVG_HEIGNT, 0.01, "legend height is the same as svg height");
```
jtlan: `assert.strictEqual()` should work here, right?

test/components/interpolatedColorLegendTests.ts
```diff
@@ -187,4 +188,102 @@ describe("InterpolatedColorLegend", () => {
...
+    svg.remove();
+  });
+
+  it("spams the entire width if oriented horizontally and expand is set to true", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.expand(true);
+    legend.renderTo(svg);
+    assert.closeTo(legend.width(), SVG_WIDTH, 0.01, "legend width is the same as svg width");
```
jtlan: `assert.strictEqual()` should work here, right?

test/components/interpolatedColorLegendTests.ts
```diff
@@ -187,4 +188,102 @@ describe("InterpolatedColorLegend", () => {
...
+    let horzontalSwatches = legend.content().selectAll(".swatch");
+    assert.isTrue(horzontalSwatches.size() > numSwatches, "there are more than 10 swatches when expanded");
+
+    legend.orientation("left");
+    let leftSwatches = legend.content().selectAll(".swatch");
+    assert.isTrue(leftSwatches.size() > numSwatches, "there are more than 10 swatches when expanded");
+
+    legend.orientation("right");
+    let rightSwatches = legend.content().selectAll(".swatch");
+    assert.isTrue(rightSwatches.size() > numSwatches, "there are more than 10 swatches when expanded");
```
jtlan: Probably clarify these assert messages with the orientation they're testing, ex:
```
"there are more than 10 swatches when expanded (orientation: right)"
```

test/components/interpolatedColorLegendTests.ts
```diff
@@ -187,4 +188,102 @@ describe("InterpolatedColorLegend", () => {
...
+    legend.renderTo(svg);
+    assert.closeTo(legend.width(), SVG_WIDTH, 0.01, "legend width is the same as svg width");
+    svg.remove();
+  });
+
+  it("has default number of swatches by default", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+
+    let numSwatches = (<any> legend)._numSwatches;
```
jtlan: Since this isn't a constant value, I'm not sure it's useful to test against it. For that matter, I'm not sure it's useful to assert the default number of swatches either...

test/components/interpolatedColorLegendTests.ts
```diff
@@ -187,4 +188,102 @@ describe("InterpolatedColorLegend", () => {
...
+    assert.strictEqual(rightSwatches.size(), numSwatches, "there are 10 swatches by default");
+
+    svg.remove();
+  });
+
+  it("has more swatches than default when expand is true", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.expand(true).renderTo(svg);
+
+    let numSwatches = (<any> legend)._numSwatches;
```
jtlan: Ditto here. A more interesting test might be one that compares the number of swatches for a given orientation when the value of `expands()` is changed:
* render `InterpolatedColorLegend`
* count number of swatches
* `expands(true)`
* assert number of swatches has increased

test/components/interpolatedColorLegendTests.ts
ztsai: it is used as a constant; I guess a better name would be `_DEFAULT_NUM_SWATCHES`
also, why is it not useful to assert there are default number of swatches? shouldn't we catch it in unit test if a legend doesn't have reasonable amount of swatches?

src/components/interpolatedColorLegend.ts
```diff
@@ -104,18 +127,18 @@ export module Components {
...
     public fixedHeight() {
-      return true;
+      return !this.expands() || !this._isVertical();
     }
 
-    private _generateTicks() {
+    private _generateTicks(numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES) {
       let domain = this._scale.domain();
-      let slope = (domain[1] - domain[0]) / this._numSwatches;
+      let slope = (domain[1] - domain[0]) / (numSwatches - 1);
```
jtlan: Doesn't this change the number of swatches generated by default?

Before:
screen shot 2015-08-25 at 9 08 18 pm

After:
screen shot 2015-08-25 at 9 08 09 pm

On the other hand, seeing as the previous logic generated `this._numSwatches + 1` swatches, this is probably fine...

src/components/interpolatedColorLegend.ts
```diff
@@ -208,6 +226,15 @@ export module Components {
         height: 0
       };
 
+      let numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES;
+
+      if (this.expands()) {
+        let textHeight = this._measurer.measure().height;
+        let offset = this._isVertical() ? 2 * padding :  4 * padding - text0Width - text1Width;
+        let fullLength = this._isVertical() ? this.height() : this.width();
+        numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), numSwatches);
```
jtlan: This can potentially cause a crash if `textHeight` evaluates to 0, as demonstrated here: http://jsfiddle.net/nr0mz1bm/ (comment the line back in)

Let's bail out if the font-size is 0, and add a test for that?

test/components/interpolatedColorLegendTests.ts
```diff
@@ -3,9 +3,10 @@
 describe("InterpolatedColorLegend", () => {
   let svg: d3.Selection<void>;
   let colorScale: Plottable.Scales.InterpolatedColor;
-
+  let SVG_HEIGNT = 400;
```
jtlan: `SVG_HEIGHT`?

test/components/interpolatedColorLegendTests.ts
```diff
@@ -20,6 +21,8 @@ describe("InterpolatedColorLegend", () => {
     assert.strictEqual(d3.select(swatches[0][swatches[0].length - 1]).attr("fill"),
                        colorScale.scale(scaleDomain[1]),
                        "last swatch's color corresponds with second domain value");
+    let defaultNumSwatches = (<any> Plottable.Components.InterpolatedColorLegend)._DEFAULT_NUM_SWATCHES;
+    assert.isTrue(swatches.size() >= defaultNumSwatches, "there are at least 10 swatches");
```
jtlan: `assert.operator()` is the preferred way to do this.

test/components/interpolatedColorLegendTests.ts
```diff
@@ -99,6 +102,18 @@ describe("InterpolatedColorLegend", () => {
     svg.remove();
   });
 
+  it("does not crash when font-size is 0px", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+    let style = (<any> legend)._element.append("style");
+    style.attr("type", "text/css");
+    style.text(".plottable .legend { font-size: 0px; }");
```
jtlan: This doesn't actually produce the desired effect; you need
```CSS
.plottable .legend text { font-size: 0px; }
```
To be specific enough.

test/components/interpolatedColorLegendTests.ts
```diff
@@ -99,6 +102,18 @@ describe("InterpolatedColorLegend", () => {
     svg.remove();
   });
 
+  it("does not crash when font-size is 0px", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+    let style = (<any> legend)._element.append("style");
```
jtlan: Append to `background()` or `content()` instead.

test/components/interpolatedColorLegendTests.ts
```diff
@@ -99,6 +102,21 @@ describe("InterpolatedColorLegend", () => {
     svg.remove();
   });
 
+  it("does not crash when font-size is 0px", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+    let style = legend.content().append("style");
+    style.attr("type", "text/css");
+    style.text(".plottable .interpolated-color-legend text { font-size: 0px; }" +
+               ".plottable .interpolated-color-legend { display: none; }");
```
jtlan: Why not "`.plottable .interpolated-color-legend text`" here?

ztsai: because it's IE :p

crmorford:
squished http://jsfiddle.net/mwowegms/12/
expanded http://jsfiddle.net/mwowegms/13/

crmorford:
http://jsfiddle.net/mwowegms/15/ looks much worse on firefox than safari or chrome. Do we know why / is it fixable?
firefox:
screen shot 2015-08-31 at 10 08 38 am
safari:
screen shot 2015-08-31 at 10 08 20 am
chrome:
screen shot 2015-08-31 at 10 08 12 am

ztsai:
@crmorford, it looks different in firefox because different browsers calculate text width differently.  
it's the same on develop without expanding: http://jsfiddle.net/89bwycby/
I don't think it's fixable unless perhaps adding a new api letting user set mixWidth/minHeight for each swatch

crmorford:
:imp:
darn ok

PR #2657, opened 2015-08-20
Add DragLineLayer

### LINKED ISSUE #1640: Slider (FR)

jtlan:
A subclass of `GuideLineLayer` that can be dragged.

It has the following API points in addition to the API points on `GuideLineLayer`:
```Typescript
/**
 * Get/set the detection radius in pixels.
 */
public detectionRadius(): number;
public detectionRadius(detectionRadius: number): DragLineLayer<D>;

/**
 * Get/set the enabled status.
 */
public enabled(): boolean;
public enabled(enabled: boolean): DragLineLayer<D>;

/**
 * Add or remove callbacks.
 */
public onDragStart(callback: DragLineCallback<D>): DragLineLayer<D>;
public offDragStart(callback: DragLineCallback<D>): DragLineLayer<D>;

public onDrag(callback: DragLineCallback<D>): DragLineLayer<D>;
public offDrag(callback: DragLineCallback<D>): DragLineLayer<D>;

public onDragEnd(callback: DragLineCallback<D>): DragLineLayer<D>;
public offDragEnd(callback: DragLineCallback<D>): DragLineLayer<D>;
```

**DEMO**: http://jsfiddle.net/ma9z844v/

Close #1640.


plottable.css
```diff
@@ -164,6 +164,14 @@ svg.plottable {
   stroke-width: 1px;
 }
 
+.plottable .drag-line-layer.enabled.vertical line.drag-edge {
+  cursor: ew-resize;
+}
+
+.plottable .drag-line-layer.enabled.horizontal line.drag-edge {
+  cursor: ns-resize;
+}
```
bluong: Do we want to use the resizing cursor or the hand grabbing cursor?

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    constructor(orientation: string) {
+      super(orientation);
+      this.addClass("drag-line-layer");
+      this.addClass("enabled");
+      if (this._isVertical()) {
+        this.addClass("vertical");
+      } else {
+        this.addClass("horizontal");
+      }
```
bluong: Why is the above vertical / horizontal class information not at the `GuideLineLayer` level?

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    private _dragStartCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _dragCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _dragEndCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _disconnectInteractionCallbacks: () => void;
+
+    constructor(orientation: string) {
+      super(orientation);
+      this.addClass("drag-line-layer");
+      this.addClass("enabled");
```
bluong: Why does "enabled" class have to exist?

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+        this.addClass("horizontal");
+      }
+
+      this._dragInteraction = new Plottable.Interactions.Drag();
+      this._dragInteraction.attachTo(this);
+
+      let grabbedLine = (p: Point) => {
+        return (this._isVertical() &&
+                 p.x >= this.pixelPosition() - this.detectionRadius() &&
+                 p.x <= this.pixelPosition() + this.detectionRadius()
```
bluong: Can we order the inequality so that it looks similar to "a <= b <= c"?

bluong: Also using `Utils.Math.inRange()` might make this more readable.  Consider that instead.

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+               ) ||
+               (
+                 !this._isVertical() &&
+                 p.y >= this.pixelPosition() - this.detectionRadius() &&
+                 p.y <= this.pixelPosition() + this.detectionRadius()
```
bluong: Above statements about the inequality being more readable

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+  export class DragLineLayer<D> extends GuideLineLayer<D> {
+    private _dragInteraction: Interactions.Drag;
+    private _detectionRadius = 3;
+    private _detectionEdge: d3.Selection<void>;
+    private _enabled = true;
+
+    private _dragStartCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _dragCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _dragEndCallbacks: Utils.CallbackSet<DragLineCallback<D>>;
+    private _disconnectInteractionCallbacks: () => void;
```
bluong: variable name is confusing.  Based on the plurality of the word, I expected the variable to point to a collection, but that is not the case...

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+      };
+      this._dragInteraction.onDragEnd(dragEndCallback);
+
+      this._disconnectInteractionCallbacks = () => {
+        this._dragInteraction.offDragStart(dragStartCallback);
+        this._dragInteraction.offDrag(dragCallback);
+        this._dragInteraction.offDragEnd(dragEndCallback);
+        delete this._disconnectInteractionCallbacks;
+      };
```
bluong: Why does the callback have to be deleted?

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+      };
+      this._dragInteraction.onDrag(dragCallback);
+      let dragEndCallback = (start: Point, end: Point) => {
+        if (dragging) {
+          dragging = false;
+          this._dragEndCallbacks.callCallbacks(this);
+        }
+      };
+      this._dragInteraction.onDragEnd(dragEndCallback);
```
bluong: Above variables should probably be named differently.  The `_dragStartCallback` calls the `_dragStartCallbacks` seems really odd.

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+      this._dragCallbacks = new Utils.CallbackSet<DragLineCallback<D>>();
+      this._dragEndCallbacks = new Utils.CallbackSet<DragLineCallback<D>>();
+    }
+
+    protected _setup() {
+      super._setup();
+      this._detectionEdge = this.content().append("line").style({
+                              opacity: 0,
+                              fill: "pink",
+                              "pointer-events": "visibleStroke"
```
bluong: It seems pointer-events does not exist in IE9? https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+        y2: this._isVertical() ? this.height() : this.pixelPosition(),
+        "stroke-width": this._detectionRadius * 2
+      });
+
+      return this;
+    }
+
+    public detectionRadius(): number;
+    public detectionRadius(detectionRadius: number): DragLineLayer<D>;
+    public detectionRadius(detectionRadius?: number): any {
```
bluong: Maybe name endpoint to `detectionWidth()`?  Radius sounds misleading since it's a line not a cylinder or anything circular

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,170 @@
...
+    public enabled(enabled?: boolean): any {
+      if (enabled == null) {
+        return this._enabled;
+      }
+      this._enabled = enabled;
+      if (enabled) {
+        this.addClass("enabled");
+      } else {
+        this.removeClass("enabled");
+      }
```
bluong: Still unsure about why the class needs to be added

src/components/dragLineLayer.ts
bluong: Why can't we make this a function?

src/components/guideLineLayer.ts
```diff
@@ -89,6 +89,14 @@ export module Components {
       }
     }
 
+    protected _setPixelPositionWithoutChangingMode(pixelPosition: number) {
```
bluong: Why does the mode have to stay the same?

bluong:
Documentation still needs to be added.

plottable.css
jtlan: Resizing, since an edge is being moved.

src/components/dragLineLayer.ts
jtlan: No particular reason. I'll move it up.

src/components/dragLineLayer.ts
jtlan: So that the cursors can be removed when not enabled.

src/components/dragLineLayer.ts
jtlan: I think it wouldn't be as clear as `a <= b <= c` in this case.

src/components/dragLineLayer.ts
jtlan: Goal was to keep it from being invoked multiple times, although if anyone invokes an underscored private method, you could argue that that was their fault...

src/components/dragLineLayer.ts
jtlan: No, it makes sense because there's only one callback being added to the interaction. I could call it `interactionDragStartCallback`, if that made things clearer.

src/components/dragLineLayer.ts
jtlan: Interesting. This is mostly for the benefit of non-IE browsers that care about it (Safari specifically) -- `DragBoxLayer` does this too.

src/components/guideLineLayer.ts
jtlan: If I just called `pixelPosition()` on dragging, that would change the mode, meaning the line wouldn't keep the same `value()` as the scale updated.

plottable.css
bluong: When I think of moving though I personally expect the hand cursor.  We can we `DragBoxLayer` as an example

src/components/dragLineLayer.ts
bluong: The thing that really sticks out to me is that this logic does not exist on `DragBoxlayer` and I do think `DragBoxLayer` has the so far cleaner implementation.  Why not look back to that implementation?

src/components/dragLineLayer.ts
bluong: Depends on how many usecases we have (we actually probably have a lot) but maybe we should have a `Utils.Math.withinError(value, compareTo, error)` or similar.  The code caught be slightly off guard at first glance even though it looks fine.

src/components/dragLineLayer.ts
bluong: Hmm...  why does this variable have to be deleted?

src/components/dragLineLayer.ts
jtlan: But `DragBoxLayer` *does* have a similar implementation:
```CSS
.plottable .drag-box-layer.x-resizable .drag-edge-lr {
  cursor: ew-resize;
}
```
It also sets the CSS classes in a similar way, but in helper methods which make it somewhat less obvious.

src/components/dragLineLayer.ts
jtlan: It doesn't. But since it's not being invoked again anyway, should it matter?

src/components/dragLineLayer.ts
bluong: Would it be more sensible to have a `movable()` endpoint, and then class under movable?  Or do we want to keep `enabled` and if we run into a situation where we want to do more than just `enabling` then we deprecate the declaration?

src/components/dragLineLayer.ts
bluong: 2 reasons why it could matter:
* By explicitly saying that this function has to be deleted at the end of the call, it signals to any reader of the code that there "must be a reason for the author to write this".  If the reason is, "it doesn't matter one way or the other" then readers would be hunting down a reason that doesn't really exist
* Since you delete it immediately on call, further invocations will crash.  Is this a wanted behavior?  If we run into a code path where a developer runs more than one call of this function, should it error?  This seems a bit weird to me.

src/components/dragLineLayer.ts
jtlan: Unlike `DragBoxLayer`, there aren't really other modes for `DragLineLayer`, so it seems sufficient to have enabled control the whole thing. I could use the class `movable` instead if that seems more flexible.

plottable.css
bluong: Using a hand cursor would also be general for non-90 degree lines, but I can see the argument for being more specific on 90-degree lines.  Can we get designer input on this?

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    constructor(orientation: string) {
+      super(orientation);
+      this.addClass("drag-line-layer");
+      this.addClass("enabled");
+
+      this._dragInteraction = new Plottable.Interactions.Drag();
+      this._dragInteraction.attachTo(this);
+
+      let grabbedLine = (p: Point) => {
```
bluong: Can we rename the function to be `onLine` or similar?  `grabbedLine` seems to be really specific to `Interactions.Drag` even though the function itself does not necessarily have to do with dragging

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,224 @@
...
+        }
+      };
+      this._dragInteraction.onDragEnd(interactionDragEndCallback);
+
+      this._disconnectInteraction = () => {
+        this._dragInteraction.offDragStart(interactionDragStartCallback);
+        this._dragInteraction.offDrag(interactionDragCallback);
+        this._dragInteraction.offDragEnd(interactionDragEndCallback);
+        this._dragInteraction.detachFrom(this);
+        delete this._disconnectInteraction;
```
bluong: I still am struggling to see the value of deleting this value for 2 reasons.
* Do we want to error if a user were to call this function twice?
* An explicit deletion here tells future readers that there is a specific reason for deleting.  What is that specific reason?

jtlan: Nothing bad comes of calling it more than once, I suppose.

crmorford:
Just playing devil's advocate:
with a large detection radius (http://jsfiddle.net/ma9z844v/2/), how did we decide to snap-to the mouse, versus leaving the mouse and DragLineLayer a constant distance apart?

crmorford:
On a "crosshairs" setup, the bottom dragLineLayer doesn't get CSS Styling for the mouse cursor :(
http://jsfiddle.net/ma9z844v/4/

jtlan:
>how did we decide to snap-to the mouse, versus leaving the mouse and DragLineLayer a constant distance apart?

The reason is that it gives the user better control, since the cursor and the thing being moved will stay closer together (`DragBoxLayer` uses similar logic for edge resizing). It'd be particularly weird if say the cursor wasn't on the layer, but the line was...

>On a "crosshairs" setup, the bottom dragLineLayer doesn't get CSS Styling for the mouse cursor :(

I'll see if I can fix this.

src/components/dragLineLayer.ts
```diff
@@ -0,0 +1,223 @@
...
+      this._dragCallbacks = new Utils.CallbackSet<DragLineCallback<D>>();
+      this._dragEndCallbacks = new Utils.CallbackSet<DragLineCallback<D>>();
+    }
+
+    protected _setup() {
+      super._setup();
+      this._detectionEdge = this.content().append("line").style({
+                              "opacity": 0,
+                              "stroke": "pink",
+                              "pointer-events": "visibleStroke"
```
bluong: Curious on why all strings have been surrounding by double quotes.  NBD though.

jtlan: For consistency; `"pointer-events"` has to be quoted, in this case. Also matches the style used in `DragBoxLayer`.

crmorford:
detectionRadius(0) http://jsfiddle.net/ma9z844v/1/
detectionRadius(large) http://jsfiddle.net/ma9z844v/2/
overlapping detectionRadius http://jsfiddle.net/ma9z844v/3/
crosshairs http://jsfiddle.net/ma9z844v/5/
.enabled() http://jsfiddle.net/ma9z844v/6/
resize selectionBox with DLL http://jsfiddle.net/ma9z844v/7/
selectionBoxLayer follow crosshairs http://jsfiddle.net/ma9z844v/9/
range moving together http://jsfiddle.net/ma9z844v/10/
css http://jsfiddle.net/ma9z844v/11/

crmorford:
cool, looks good! 
let's just fix css in the other PR.

crmorford:
the more i think about what detach/destroy should do, the more confused i get http://jsfiddle.net/ma9z844v/14/

jtlan:
Some methods on `destroy()`-ed objects still work; I'm not sure what should be done about this.

bluong:
This sounds like a global issue that is not just on DLL's but on every component.  We can file an issue and revisit

aicioara:
@jtlan 
Can I go ahead and refactor the test file?

jtlan:
> Can I go ahead and refactor the test file?

Yeah, that's fine.

PR #2659, opened 2015-08-20
renderTo() now accepts DOM nodes

### LINKED ISSUE #1819: [Usability] Be able to renderTo HTML DOM Element
When playing with the library, it made a lot of sense for me to do this

```typescript
var svg = document.getElementById('demo');
chartTable.renderTo(svg);
```

Should this not be supported?

crmorford:
Closes #1819.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1449,7 +1449,7 @@ declare module Plottable {
          * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
          * @returns {Component} The calling Component.
          */
-        renderTo(element: String | SVGSVGElement | d3.Selection<void>): Component;
+        renderTo(element: String | HTMLElement | d3.Selection<void>): Component;
         /**
          * Gets the x alignment of the Component.
          */
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1449,7 +1449,7 @@ declare module Plottable {
          * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
          * @returns {Component} The calling Component.
          */
-        renderTo(element: String | d3.Selection<void>): Component;
+        renderTo(element: String | SVGSVGElement | d3.Selection<void>): Component;
         /**
          * Gets the x alignment of the Component.
          */
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1449,7 +1449,7 @@ declare module Plottable {
          * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
          * @returns {Component} The calling Component.
          */
-        renderTo(element: String | HTMLElement | d3.Selection<void>): Component;
+        renderTo(element: String | Element | d3.Selection<void>): Component;
         /**
          * Gets the x alignment of the Component.
          */
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -262,14 +262,14 @@ export class Component {
    * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
    * @returns {Component} The calling Component.
    */
-  public renderTo(element: String | SVGSVGElement | d3.Selection<void>): Component {
+  public renderTo(element: String | HTMLElement | d3.Selection<void>): Component {
     this.detach();
     if (element != null) {
       let selection: d3.Selection<void>;
       if (typeof(element) === "string") {
         selection = d3.select(<string> element);
-      } else if (element instanceof SVGSVGElement) {
-        selection = d3.select(<SVGSVGElement> element);
+      } else if (element instanceof HTMLElement) {
+        selection = d3.select(<HTMLElement> element);
       } else {
         selection = <d3.Selection<void>> element;
       }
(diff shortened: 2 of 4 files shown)


src/components/component.ts
```diff
@@ -262,12 +262,14 @@ export class Component {
...
    */
-  public renderTo(element: String | d3.Selection<void>): Component {
+  public renderTo(element: String | SVGSVGElement | d3.Selection<void>): Component {
     this.detach();
     if (element != null) {
       let selection: d3.Selection<void>;
       if (typeof(element) === "string") {
         selection = d3.select(<string> element);
+      } else if (element instanceof SVGSVGElement) {
+        selection = d3.select(<SVGSVGElement> element);
```
bluong: You can collapse this case and the above case into a single case

```typescript
if (typeof(element) === "string || element instanceof SVGSVGElement)) {
  selection = d3.select(<string | SVGSVGElement> element);
}

File: src/components/component.ts [hunk 2]
```diff
@@ -262,12 +262,14 @@ export class Component {
    * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
    * @returns {Component} The calling Component.
    */
-  public renderTo(element: String | d3.Selection<void>): Component {
+  public renderTo(element: String | SVGSVGElement | d3.Selection<void>): Component {
```
aicioara: I prefer `HTMLElement` instead of `SVGSVGElement`

The reason for that is that in 99% of the cases the element is taken by using 
```javascript
document.getElementById("foo") // the type of this is not compatible with SVGSVGElement because it is too generic
```

test/components/componentTests.ts
```diff
@@ -445,6 +445,20 @@ describe("Component behavior", () => {
     svg.remove();
   });
 
+  it("renderTo() only accepts strings, selections containing svgs, and SVG elements", () => {
+    svg.attr("id", "render-to-test");
+    assert.doesNotThrow(() => c.renderTo("#render-to-test"), Error, "accepts strings that identify svgs");
+    assert.doesNotThrow(() => c.renderTo(svg), Error, "accepts selections that contain svgs");
+    assert.doesNotThrow(() => c.renderTo(<any> svg.node()), Error, "accepts svg elements");
```
aicioara: instead of svg.node(). use `document.getElementById("render-to-test")`. This + the suggestion above will allow you to not need the `<any>` cast anymore

test/components/componentTests.ts
```diff
@@ -445,6 +445,20 @@ describe("Component behavior", () => {
...
+    let parent = TestMethods.getSVGParent();
+    let div = parent.append("div");
+    // HACKHACK #2614: chai-assert.d.ts has the wrong signature
+    (<any> assert).throws(() => c.renderTo(div), Error, "", "rejects selections that don't contain svgs");
```
aicioara: I like this assert. Let's make another one where we check the same thing for `div[0][0]` (basically for the DOM element instead of the selectiob)

test/components/componentTests.ts
```diff
@@ -445,6 +445,20 @@ describe("Component behavior", () => {
...
+    (<any> assert).throws(() => c.renderTo("#not-a-element"), Error, "", "rejects strings that don't correspond to DOM elements");
+    (<any> assert).throws(() => c.renderTo(d3.select(null)), Error, "", "rejects empty d3 selections");
```
aicioara: The third argument corresponds to the error the assert is throwing. Let's use it to make sure the expected error is thrown

aicioara:
Rest looks really good. I really like the way you think about edge cases in tests. I can see the QE experience.

aicioara:
Looks like SauceLabs / Travis got flaky. I restarted the test for you.

test/components/componentTests.ts
crmorford: Does this mean I should find out what error is usually thrown when this happens and copy that error there verbatim?

aicioara: Yea. You can do that very easily by writing some garbage above, see what the test is expecting and just paste it there. It's better than leaving it empty

plottable.d.ts
```diff
@@ -1449,7 +1449,7 @@ declare module Plottable {
          * @param {String|d3.Selection} element A selector-string for the <svg>, or a d3 selection containing an <svg>.
          * @returns {Component} The calling Component.
          */
-        renderTo(element: String | d3.Selection<void>): Component;
+        renderTo(element: String | Element | d3.Selection<void>): Component;
```
aicioara: Is there a reason why we use Element instead of HTMLElement?

aicioara:
Apart from that it looks good

plottable.d.ts
crmorford: I had to use Element  instead of HTMLElement because of an issue surfaced by one of the tests you recommended.

document.getElementById("render-to-test") instanceof HTMLElement == false
document.getElementById("render-to-test") instanceof Element == true

aicioara: I learned something today. SVGSVGElement are not HTMLElement, but they are Element. Cool

jtlan:
Tested:
* `renderTo(selection containing SVG)`
* `renderTo(invalid selection)`
* `renderTo(document.getElementById("chart"))`
* creating an element from scratch:

  ```Javascript
var svg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
svg.setAttribute("width", "400px");
svg.setAttribute("height", "400px");
document.body.appendChild(svg);
chart.renderTo(svg);
```

aicioara:
@crmorford Please delete branch as well

Issue #2661, opened 2015-08-21
Chai definitions 


aicioara:
According to @jtlan 's experiments this code is valid 
```typescript
assert.throws(() => doWhatever(), Error, "error message", "message to the user");
```

However, DefinitelyTyped has a version of `chai.d.ts` that does not support it. Try to upgrade them.

jtlan:
Whoever takes this ticket should also fix any `HACKHACK`s that are `<any>`-casting around the wrong signature.

PR #2664, opened 2015-08-21
Updated Mocha and Chai typings and libraries

### LINKED ISSUE #2614: Update chai and its definitions
Currently our bower.json lists 1.9.0; the current version is **3.2.0**.

Also, signature for `assert.throws` is wrong in the definitions file. From experimenting, to include a message requires the following signature:
```Typescript
assert.throws(function, errorConstructor, message_error_throws, message);
```

But the .d.ts file doesn't seem to list the signature, nor do the docs?

aicioara:
Closes #2614

However, does not solve the original issue. 

Mocha -  `1.17.1` ->  `2.2.5` (latest)
Chai - `1.9.0` -> `2.0.0` | Latest would be `3.2.0` but there is no Definitely Typed `d.ts` file for it. This might link to #2661

To ease the Review Process, I did the following: 
- Check out Definitely Typed latest versions of the two libraries
- Updated the Commit hash in tsd.json to the latest master of Definitely Typed as per 21 August 2015 12:15 (PST)
- Ran `tsd reinstall -so`
- Updated bower.json with the correct libraries
- Ran `bower install`
- Noticed the change from `chai/chai-assert.d.ts` to `chai/chai.d.ts`
- Updated the reference files to reflect the file name change
- The `assert.doesNotThrow()` has changed signature so I had to <any> cast it now to keep the scope of this PR small
- For some reason there was no `let` specifier for the assert library so I had to add it for everything to compile. Not sure why it worked before. 
- However adding let makes tslint upset. Assert is not really used anywhere else in that file and tslint does not resolve global scopes. Oh well... Please let me know if you can think of anything.


aicioara:
Still fiddling with this. `tslint` and `tsc` have different opinions of how the code should look like

jtlan:
This should be [NOQE], right?

jtlan:
Were there other instances of `doesNotThrow()` that need to be fixed too?

aicioara:
Nope, everything compiled. Will make a followup ticket

aicioara:
[NOQE] As no JS changes. Only test, typings and libraries.

PR #2667, opened 2015-08-24
Refactor selection box layer - Pass 1

aicioara:
Part of #2623 

As part of this: 
- No nested `beforeEach()`es
screen shot 2015-08-24 at 12 59 14 pm

### DIFF
--- a/test/components/selectionBoxLayerTests.ts
+++ b/test/components/selectionBoxLayerTests.ts
@@ -15,7 +15,7 @@ describe("Interactive Components", () => {
         sbl = new Plottable.Components.SelectionBoxLayer();
       });
 
-      it("can get and set the boxVisible() property", () => {
+      it("can set the boxVisible() property", () => {
         assert.strictEqual(sbl.boxVisible(), false, "The box is not visible by default");
 
         sbl.renderTo(svg);
@@ -48,7 +48,7 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
-      it("can get and set the bounds() property", () => {
+      it("can set the bounds() property", () => {
         let defaultBounds = sbl.bounds();
         assert.strictEqual(defaultBounds.topLeft.x, 0, "top-left bound is correct (x)");
         assert.strictEqual(defaultBounds.topLeft.y, 0, "top-left bound is correct (y)");
@@ -108,7 +108,7 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
-      it("can get and set the xScale() property", () => {
+      it("can set the xScale() property", () => {
         let xScale = new Plottable.Scales.Linear();
         xScale.domain([0, 2000]);
         xScale.range([0, svgWidth]);
@@ -173,7 +173,7 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
-      it("can get and set the yScale() property", () => {
+      it("can set the yScale() property", () => {
         let yScale = new Plottable.Scales.Linear();
         yScale.domain([0, 2000]);
         yScale.range([0, svgHeight]);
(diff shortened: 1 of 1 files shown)


test/components/selectionBoxLayerTests.ts
```diff
@@ -1,193 +1,244 @@
...
+        sbl.renderTo(svg);
+        assert.strictEqual(sbl.boxVisible(), false, "The box is not visible by default after rendering to svg");
+
+        assert.strictEqual(sbl.boxVisible(true), sbl, "Setting the boxVisible attribute returns the selection box layer");
+        assert.strictEqual(sbl.boxVisible(), true, "Setting the boxVisible to true attribute works");
+        sbl.boxVisible(false);
+        assert.strictEqual(sbl.boxVisible(), false, "Setting the boxVisible attribute to false works");
+
+        svg.remove();
+      });
```
bluong: This crosses a very fine line of testing, but I have thinking about how to effectively test getters... only to end with the conclusion of not to test them directly (The best way to test them directly is to pull the corresponding variable backing the getter but then this leads into all sorts of trouble).

The best thing I can think of is to utilize the getters as our means of detecting correctness for the rest of the code (If we use a setter, then the getter should definitely reflect the set value) (If a state-modifying function is called, the getter should definitely reflect the modified state).

Sure we can use these unit tests to prove that the function exists, but I am unsure of its usefulness beyond that (Also this specific idea would be not useful for typescript consumers)

test/components/selectionBoxLayerTests.ts
```diff
@@ -1,193 +1,244 @@
...
+
+        sbl.renderTo(svg);
+        assert.strictEqual(sbl.boxVisible(), false, "The box is not visible by default after rendering to svg");
+
+        assert.strictEqual(sbl.boxVisible(true), sbl, "Setting the boxVisible attribute returns the selection box layer");
+        assert.strictEqual(sbl.boxVisible(), true, "Setting the boxVisible to true attribute works");
+        sbl.boxVisible(false);
+        assert.strictEqual(sbl.boxVisible(), false, "Setting the boxVisible attribute to false works");
+
+        svg.remove();
```
bluong: Suggestion - if we create a component, we destroy the component before removal.

test/components/selectionBoxLayerTests.ts
```diff
@@ -1,193 +1,244 @@
...
+      });
+
+      it("does not error on destroy() when scales are not inputted", () => {
+        sbl.renderTo(svg);
+        assert.doesNotThrow(() => sbl.destroy(), Error, "can destroy even with no scales");
+
+        svg.remove();
+      });
```
bluong: I really don't think this test should exist but should more be embedded into our unit tests or tested elsewhere?

test/components/selectionBoxLayerTests.ts
```diff
@@ -1,193 +1,244 @@
...
+
+      it("can get and set the bounds() property", () => {
```
bluong: Same description as above about getters

bluong:
A few suggestions

test/components/selectionBoxLayerTests.ts
aicioara: As @jtlan was saying. We have to build up our tests on axioms. This tests proves that the getter and the setter are having the expected behavior, so we can use them later assuming they are correct. 

I must agree that the usefulness of this test is below the average though.

test/components/selectionBoxLayerTests.ts
aicioara: I like that

test/components/selectionBoxLayerTests.ts
aicioara: ?

aicioara:
I liked the idea of destroying the components, hence I applied it. Not sure what you mean about the getters / setters

bluong:
I think having tests to test the setters is perfectly valid, but I do not think it makes sense for us to explicitly say that we are testing the getters.

Thus, in this scenario, I would say to change the test names to just saying that you can use the setters and then inside the test just use the getters to get at the set values.

Something like
`it("can set if the box is visible")`

aicioara:
Sounds good

PR #2669, opened 2015-08-24
Refactor x/y drag box layer tests - Pass 1

aicioara:
Part of #2623 

As part of this: 
- No nested `beforeEach()`es

### DIFF
--- a/test/components/xDragBoxLayerTests.ts
+++ b/test/components/xDragBoxLayerTests.ts
@@ -4,7 +4,7 @@ describe("Layer Components", () => {
   describe("XDragBoxLayer", () => {
     describe("Basic Usage", () => {
       let SVG_WIDTH = 400;
-      let SVG_HEIGHT = 400;
+      let SVG_HEIGHT = 300;
 
       let quarterTopLeftPoint = {
         x: SVG_WIDTH / 4,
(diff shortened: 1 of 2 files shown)


test/components/xDragBoxLayerTests.ts
```diff
@@ -1,178 +1,155 @@
...
 
-      let topLeft = {
-        x: SVG_WIDTH / 4,
-        y: SVG_HEIGHT / 4
-      };
-      let bottomRight = {
-        x: SVG_WIDTH * 3 / 4,
-        y: SVG_HEIGHT * 3 / 4
-      };
+      it("stays full height after resizing", () => {
```
jtlan: Probably rename this to distinguish it from the previous test -- maybe something like `"stays full-height when redrawn at a different size"`?

test/components/xDragBoxLayerTests.ts
```diff
@@ -1,178 +1,155 @@
...
-      assert.notStrictEqual(dbl.height(), heightBefore, "component changed size");
-
-      let boundsAfter = dbl.bounds();
-      assert.strictEqual(boundsAfter.topLeft.x, boundsBefore.topLeft.x, "box keeps same left edge");
-      assert.strictEqual(boundsAfter.topLeft.y, 0, "box still starts at top");
-      assert.strictEqual(boundsAfter.bottomRight.x, boundsBefore.bottomRight.x, "box keeps same right edge");
-      assert.strictEqual(boundsAfter.bottomRight.y, dbl.height(), "box still ends at bottom");
-      svg.remove();
-    });
+      it("has the correct bounds()", () => {
```
jtlan: Maybe something more specific, like "has bounds() that extend across its full width"?

PR #2672, opened 2015-08-24
add wheel plot

### LINKED ISSUE #2612: Timewheel support
Take a look at this old PR that never merged: https://github.com/palantir/plottable/pull/1347. Don't necessarily need to do this the same way - so consider the optimal implementation. An RFC would be a good idea.

ztsai:
part of #2612  (broke the task down to 4 steps)
demo: http://jsfiddle.net/ztsai/huzsoxpj/ http://jsfiddle.net/ztsai/huzsoxpj/4/

A sector is defined by `startAngle`, `endAngle`, `innerRadius`, and `outerRadius`:

```startAngle<T>(startAngle?: number | Accessor<number> | T | Accessor<T>, scale?: Scale<T, number>)```
```endAngle<T>(endAngle?: number | Accessor<number> | T | Accessor<T>, scale?: Scale<T, number>)```
```innerRadius<R>(innerRadius?: number | Accessor<number> | R | Accessor<R>, scale?: Scale<R, number>)```
```outerRadius<R>(outerRadius?: number | Accessor<number> | R | Accessor<R>, scale?: Scale<R, number>)```

### DIFF
--- a/src/plots/wheelPlot.ts
+++ b/src/plots/wheelPlot.ts
@@ -1,0 +1,219 @@
+///<reference path="../reference.ts" />
+
+module Plottable {
+export module Plots {
+  export class Wheel<R, T> extends Plot {
+
+    private static _INNER_RADIUS_KEY = "inner-radius";
+    private static _OUTER_RADIUS_KEY = "outer-radius";
+    private static _START_ANGLE_KEY = "start-angle";
+    private static _END_ANGLE_KEY = "end-angle";
+
+    /**
+     * @constructor
+     */
+    constructor() {
+      super();
+      this.addClass("wheel-plot");
+      this.attr("fill", (d, i) => String(i), new Scales.Color());
+    }
+
+    public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
+      super.computeLayout(origin, availableWidth, availableHeight);
+      this._renderArea.attr("transform", "translate(" + this.width() / 2 + "," + this.height() / 2 + ")");
+      let radiusLimit = Math.min(this.width(), this.height()) / 2;
+      if (this.innerRadius() != null && this.innerRadius().scale != null) {
+        this.innerRadius().scale.range([0, radiusLimit]);
+      }
+      if (this.outerRadius() != null && this.outerRadius().scale != null) {
+        this.outerRadius().scale.range([0, radiusLimit]);
+      }
+      return this;
+    }
+
+    protected _createDrawer(dataset: Dataset) {
+      return new Plottable.Drawers.Arc(dataset);
+    }
+
+    public entities(datasets = this.datasets()): PlotEntity[] {
+      let entities = super.entities(datasets);
+      entities.forEach((entity) => {
+        entity.position.x += this.width() / 2;
+        entity.position.y += this.height() / 2;
+      });
+      return entities;
+    }
+
+    protected _getDataToDraw() {
+      let dataToDraw = super._getDataToDraw();
+      if (this.datasets().length === 0) { return dataToDraw; }
+      let startAngleAccessor = Plot._scaledAccessor(this.startAngle());
+      let endAngleAccessor = Plot._scaledAccessor(this.endAngle());
+      let innerRadiusAccessor = Plot._scaledAccessor(this.innerRadius());
+      let outerRadiusAccessor = Plot._scaledAccessor(this.outerRadius());
+      let ds = this.datasets()[0];
+      let data = dataToDraw.get(ds);
+      let filteredData = data.filter((d, i) =>
+        Plottable.Utils.Math.isValidNumber(startAngleAccessor(d, i, ds)) &&
+        Plottable.Utils.Math.isValidNumber(endAngleAccessor(d, i, ds)) &&
+        Plottable.Utils.Math.isValidNumber(innerRadiusAccessor(d, i, ds)) &&
+        Plottable.Utils.Math.isValidNumber(outerRadiusAccessor(d, i, ds)));
+      dataToDraw.set(ds, filteredData);
+      return dataToDraw;
+    }
+
+    protected _propertyProjectors(): AttributeToProjector {
+      let attrToProjector = super._propertyProjectors();
+      let innerRadiusAccessor = Plot._scaledAccessor(this.innerRadius());
+      let outerRadiusAccessor = Plot._scaledAccessor(this.outerRadius());
+      let startAngleAccessor = Plot._scaledAccessor(this.startAngle());
+      let endAngleAccessor = Plot._scaledAccessor(this.endAngle());
+      attrToProjector["d"] = (datum: any, index: number, ds: Dataset) => {
+        return d3.svg.arc().innerRadius(innerRadiusAccessor(datum, index, ds))
+                           .outerRadius(outerRadiusAccessor(datum, index, ds))
+                           .startAngle(startAngleAccessor(datum, index, ds))
+                           .endAngle(endAngleAccessor(datum, index, ds))(datum, index);
+      };
+      return attrToProjector;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the start angle.
+     */
+    public startAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the start angle to a constant number or the result of an Accessor<number>.
+     *
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {S|Accessor<S>} startAngle
+     * @param {Scale<S, number>} scale
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle<T>(startAngle: T | Accessor<T>, scale: Scale<T, number>): Plots.Wheel<R, T>;
+    public startAngle<T>(startAngle?: number | Accessor<number> | T | Accessor<T>, scale?: Scale<T, number>): any {
+      if (startAngle == null) {
+        return this._propertyBindings.get(Wheel._START_ANGLE_KEY);
+      }
+
+      if (scale != null) {
+        scale.range([0, Math.PI * 2]);
+      }
+
+      this._bindProperty(Wheel._START_ANGLE_KEY, startAngle, scale);
+      this.render();
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the end angle.
+     */
+    public endAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the end angle to a constant number or the result of an Accessor<number>.
+     *
+     * @param {number|Accessor<number>} endAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public endAngle(endAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the end angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by jtlan** [comment]
Fixed fiddles:
http://jsfiddle.net/71xu10uz/
http://jsfiddle.net/597x450n/

**Comment by jtlan** [comment]
Since this is somewhat difficult to read before labeling, does it make sense to do all the development on a side branch, then merge the side branch to **`develop`** once everything is implemented?

File: src/plots/wheelPlot.ts [hunk 1]
```diff
@@ -0,0 +1,219 @@
...
+    private static _START_ANGLE_KEY = "start-angle";
+    private static _END_ANGLE_KEY = "end-angle";
+
+    /**
+     * @constructor
+     */
+    constructor() {
+      super();
+      this.addClass("wheel-plot");
+      this.attr("fill", (d, i) => String(i), new Scales.Color());
```
jtlan: I don't think this makes sense for `Plots.Wheel`; consider a parallel with `Plots.Rectangle`, which just has a single-color fallback if the user doesn't set `fill`.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,219 @@
...
+     * @param {number|Accessor<number>} endAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public endAngle(endAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the end angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {S|Accessor<S>} endAngle
+     * @param {Scale<S, number>} scale
```
jtlan: Docs don't match the signature

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,219 @@
...
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {S|Accessor<S>} startAngle
+     * @param {Scale<S, number>} scale
```
jtlan: Docs don't match signature.

test/testMethods.ts
```diff
@@ -253,23 +253,29 @@ module TestMethods {
...
+    let expectedAreaPathNumbers = tokenizePathString(expectedAreaPathStrings);
+
+    assert.lengthOf(actualAreaPathNumbers, expectedAreaPathNumbers.length, "number of numbers in each path should be equal");
+    actualAreaPathNumbers.forEach((actualAreaNumber, i) => {
+      let expectedAreaNumber = expectedAreaPathNumbers[i];
+        assert.closeTo(+actualAreaNumber, +expectedAreaNumber, 0.1, msg);
     });
   }
 
+  function tokenizePathString(pathStrings: string[]) {
```
jtlan: Should be "`tokenizePathStrings`", based on the input, or it should take only one `string`. Is there a reason it takes an array?

jtlan:
What's the value in allowing different scales for `innerRadius`/`outerRadius`, as well as different scales for `startAngle`/`endAngle`? On `Plots.Rectangle`, `x`/`x2` share a `Scale` because logically, they need to be displayed at the same scale.

jtlan:
From playing around, it seems like there isn't a distinction between `startAngle` and `endAngle`. Although this parallels roughly the effect of `x`/`x2` on `Plots.Rectangle`, it does produce some unexpected behavior. For example, I would expect the following two datums to produce different arcs:
```Typescript
{
  startAngle: -60,
  endAngle: 60
},
{
  startAngle: 60,
  endAngle: -60
}
```
but they don't (both datums draw the same arc). To get an arc that wraps around more than 180 degrees, I have to do:
```Typescript
{
  startAngle: 60,
  endAngle: 300
}
```

@crmorford, what's your opinion on this.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        wheelPlot.addDataset(dataset);
+        wheelPlot.renderTo(svg);
+
+        let slices = wheelPlot.selections();
+        let path1 = d3.select(slices[0][0]).attr("d");
+        let path2 = d3.select(slices[0][1]).attr("d");
+
+        let arc1 = d3.svg.arc().innerRadius(rScale.scale(0)).outerRadius(rScale.scale(1))
+                               .startAngle(tScale.scale(0)).endAngle(tScale.scale(180));
+        let arc1Path = (<any> arc1)();
```
jtlan: Is the cast needed because the signature expects a `d`? You can pass in a dummy datum in that case, or `null`.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+          {r1: 1, r2: 2, t1: 180, t2: 360 }];
+        let dataset = new Plottable.Dataset(data);
+        wheelPlot.addDataset(dataset);
+        wheelPlot.renderTo(svg);
+
+        assert.strictEqual(wheelPlot.content().selectAll("path").size(), 2, "there are 2 sectors for 2 data entries");
+
+        data = [{r1: 2, r2: 3, t1: 0, t2: 120 }];
+        dataset.data(data);
+        assert.strictEqual(wheelPlot.content().selectAll("path").size(), 1, "there are 2 sectors for 1 data entry");
```
jtlan: "there is 1 sector for 1 data entry"?

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        let arc2 = d3.svg.arc().innerRadius(rScale.scale(1)).outerRadius(rScale.scale(2))
+                               .startAngle(tScale.scale(180)).endAngle(tScale.scale(360));
+        let arc2Path = (<any> arc2)();
+        TestMethods.assertAreaPathCloseTo(path1, arc1Path, 0.1, "arc is drawn as represented by data");
+        TestMethods.assertAreaPathCloseTo(path2, arc2Path, 0.1, "arc is drawn as represented by data");
+
+        svg.remove();
+      });
+
+      it("updates slices when data changes", () => {
```
jtlan: This effectively checks that there's a re-render when a `Dataset` updates right? Theoretically that should have already been checked at the parent `Plot` level, so it seems a bit redundant to check it again here.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        let transform = d3.transform(renderArea.attr("transform")).translate;
+        assert.closeTo(transform[0], SVG_WIDTH / 2, 0.01, "origin is set to the middle of the svg");
+        assert.closeTo(transform[1], SVG_HEIGHT / 2, 0.01, "origin is set to the middle of the svg");
+        let radiusLimit = Math.min(SVG_WIDTH, SVG_HEIGHT) / 2;
+        assert.deepEqual(rScale.range(), [0, radiusLimit] , "the radius scale range is set properly");
+
+        svg.remove();
+      });
+
+      it("undefined, NaN and non-numeric strings not be represented in a Pie Chart", () => {
```
jtlan: "are not represented in a Wheel Plot"

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+          { r1: 1, r2: 2, t1: 180, t2: NaN },
+          { r1: 1, r2: 2, t1: 180, t2: "Bad String" },
+          { r1: 2, r2: 3, t1: 90, t2: 180 }
+        ];
+
+        let dataset = new Plottable.Dataset(data);
+        wheelPlot.addDataset(dataset);
+        wheelPlot.renderTo(svg);
+
+        let elementsDrawn = (<any> wheelPlot)._element.selectAll(".arc");
```
jtlan: `wheelPlot.content().selectAll(".arc")` should be used instead.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        wheelPlot.addDataset(dataset);
+      });
+
+      it("can set and get innerRadius", () => {
+        wheelPlot.outerRadius((d) => d.r2, rScale);
+        assert.isUndefined(wheelPlot.innerRadius(), "innerRadius is initiized to undefined");
+
+        wheelPlot.innerRadius(0);
+        wheelPlot.renderTo(svg);
+        let constInnerRadiusAccessor = (<any> Plottable.Plot)._scaledAccessor(wheelPlot.innerRadius());
```
jtlan: Rather than `<any>`-casting in to the `Plot`, shouldn't you call `innerRadius()` as a getter and check that it returns the correct value? In this case, it should return an `AccessorScaleBinding` with no `Scale` and an `Accessor` that returns a constant value.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        svg.remove();
+      });
+
+      it("can set and get outerRadius", () => {
+        wheelPlot.innerRadius((d) => d.r1, rScale);
+        assert.isUndefined(wheelPlot.outerRadius(), "outerRadius is initiized to undefined");
+
+        wheelPlot.outerRadius(0);
+        wheelPlot.renderTo(svg);
+        let constEndAngleAccessor = (<any> Plottable.Plot)._scaledAccessor(wheelPlot.outerRadius());
```
jtlan: Same feedback here -- this test doesn't actually check the getter mode of `outerRadius()`

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        wheelPlot.addDataset(dataset);
+      });
+
+      it("can set and get startAngle", () => {
+        wheelPlot.endAngle((d) => d.t2, tScale);
+        assert.isUndefined(wheelPlot.startAngle(), "startAngle is initiized to undefined");
+
+        wheelPlot.startAngle(0);
+        wheelPlot.renderTo(svg);
+        let constStartAngleAccessor = (<any> Plottable.Plot)._scaledAccessor(wheelPlot.startAngle());
```
jtlan: Same comment about actually checking the getter functionality.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        svg.remove();
+      });
+
+      it("can set and get endAngle", () => {
+        wheelPlot.startAngle((d) => d.t1, tScale);
+        assert.isUndefined(wheelPlot.endAngle(), "endAngle is initiized to undefined");
+
+        wheelPlot.endAngle(0);
+        wheelPlot.renderTo(svg);
+        let constOuterRadiusAccessor = (<any> Plottable.Plot)._scaledAccessor(wheelPlot.endAngle());
```
jtlan: Same feedback about actually checking the getter functionality.

test/plots/wheelPlotTests.ts
jtlan: Same feedback here about checking the return value.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,304 @@
...
+        svg.remove();
+      });
+
+      it("can set innerRadius and outerRadius to different scale", () => {
+        let rScale2 = new Plottable.Scales.ModifiedLog;
+        wheelPlot.innerRadius((d) => d.r1, rScale);
+        wheelPlot.outerRadius((d) => d.r2, rScale2);
+        wheelPlot.renderTo(svg);
+
+        let innerRadiusAccessor = (<any> Plottable.Plot)._scaledAccessor(wheelPlot.innerRadius());
```
jtlan: Same feedback here about checking the return value.

test/plots/wheelPlotTests.ts
jtlan: Same feedback here about checking the return value.

test/plots/wheelPlotTests.ts
jtlan: Same feedback here about checking the return value.

ztsai:
will create a side branch

Issue #2640, opened 2015-08-17
[Testing Audit] Scales


crmorford:

aicioara:
Note for self, this means
- [ ] `categoryScale.ts`
- [ ] `colorScale.ts`
- [ ] `interpolatedColorScale.ts`
- [ ] `linearScale.ts`
- [ ] `modifiedLogScale.ts`
- [ ] `quantitativeScale.ts`
- [ ] `scale.ts`
- [ ] `timeScale.ts`
- [ ] `tickGenerators.ts`

aicioara:
Blocking on scale and quantitative scale until children are merged.

jtlan:
`Scale` and `QuantitativeScale` done in #2974; closing this out.

PR #2674, opened 2015-08-24
Refactor color scale tests - Pass 1

aicioara:
Part of #2640 

As part of this: 
- No nested `beforeEach()`es
screen shot 2015-08-24 at 4 16 12 pm


test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,129 @@
...
+describe("Scales", () => {
+  describe("Color Scale", () => {
+
+    describe("Basic Usage", () => {
+      let defaultColors = [
+        "#5279c7", "#fd373e", "#63c261", "#fad419", "#2c2b6f",
+        "#ff7939", "#db2e65", "#99ce50", "#962565", "#06cccc"
+      ];
+
+      it("default colors are generated", () => {
```
ztsai: "it generates default colors"?

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,129 @@
...
+      });
+
+      it("uses altered colors if size of domain exceeds size of range", () => {
+        let scale = new Plottable.Scales.Color();
+        let colorRange = ["#5279c7", "#fd373e"];
+        scale.range(colorRange);
+        scale.domain(["a", "b", "c"]);
+        assert.strictEqual(scale.scale("a"), colorRange[0], "first color is used");
+        assert.strictEqual(scale.scale("b"), colorRange[1], "second color is used");
+        assert.notStrictEqual(scale.scale("c"), colorRange[0], "first color is slightly modified and used");
```
ztsai: can you add an assertion that scale.scale("c") is not equal to colorRange[1] as well?
also, the assertion message is confusing as you only tested those 2 colors are not equal, you didn't test that one is a slight modification of another

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,129 @@
...
+  describe("Color Scale", () => {
+
+    describe("Basic Usage", () => {
+      let defaultColors = [
+        "#5279c7", "#fd373e", "#63c261", "#fad419", "#2c2b6f",
+        "#ff7939", "#db2e65", "#99ce50", "#962565", "#06cccc"
+      ];
+
+      it("default colors are generated", () => {
+        let scale = new Plottable.Scales.Color();
```
ztsai: not sure what's the best practice here. it seems like you call this same line first time in all the tests, is it worth to create a beforeEach? (this is a question rather than comment)

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,177 @@
...
+        let alteredColor1 = scale.scale("c");
+        assert.notStrictEqual(alteredColor1, colorRange[0], "first color has not been reused");
+        assert.notStrictEqual(alteredColor1, colorRange[1], "second color has not been reused");
+        assert.notStrictEqual(alteredColor1, "#000000", "the color does not fallback to black when running out of colors");
+        assert.notStrictEqual(alteredColor1, "#ffffff", "the color does not fallback to white when running out of colors");
+
+        assert.operator(parseInt(alteredColor1.substr(1, 2), 16), ">", parseInt(colorRange[0].substr(1, 2), 16),
+          "The resulting color should be lighter in the red component");
+        assert.operator(parseInt(alteredColor1.substr(1, 3), 16), ">", parseInt(colorRange[0].substr(3, 2), 16),
+          "The resulting color should be lighter in the red component");
```
ztsai: green?

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,177 @@
...
+        assert.operator(parseInt(alteredColor1.substr(1, 5), 16), ">", parseInt(colorRange[0].substr(5, 2), 16),
+          "The resulting color should be lighter in the red component");
```
ztsai: blue?

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    describe("CSS integration", () => {
+      let defaultColors = [
+        "#5279c7", "#fd373e", "#63c261", "#fad419", "#2c2b6f",
+        "#ff7939", "#db2e65", "#99ce50", "#962565", "#06cccc"
+      ];
+
+      it("accepts CSS specified colors", () => {
+        let style = d3.select("body").append("style");
+        style.html(".plottable-colors-0 {background-color: #ff0000 !important; }");
```
ztsai: I am not comfortable with modifying the dom to add css when I first see this, but I guess that's needed for css testing.  Can you assign another reviewer just to be sure tho?

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,177 @@
...
+      });
+
+       it("caches CSS specified colors unless the cache is invalidated", () => {
+        let style = d3.select("body").append("style");
+        style.html(".plottable-colors-0 {background-color: #ff0000 !important; }");
+        Plottable.Scales.Color.invalidateColorCache();
+        let scale = new Plottable.Scales.Color();
+        style.remove();
+
+        assert.strictEqual(scale.range()[0], "#ff0000", "User has specified red color for first color scale color");
```
ztsai: why is this after style.remove()

test/scales/colorScaleTests.ts
bluong: We have done things similar to this style before: "scales icon sizes properly with font size (textHeight / 2 < symbolHeight < textHeight)" in "legendTests.ts".

The difference is that this style is being applied at the "body" level whereas the style in the legend tests is applied at the specific element level.

The bad thing about the body level is that it unfortunately makes the tests much less independent.  `style.remove()` must be called eventually before this test complete or else the styles in the rest of the tests are affected.

However, the color tester is applied at the "body" level, so I think this is unavoidable.

Also, there is a comment that the body level is most emulating what a user will do since they will usually just use a css file and list it appropriately.  I think ideally we would have a separate css file for testing, but since the color tester is so specific, this idea isn't possible I think.

All inall, I think this is fine, but I do think it can get dangerous easily.

ztsai: Do we need hackhack tag for things like this? 
just so that (1) if someone try to copy and paste this in the future they know to be cautious and (2) we can improve this if technology advances

aicioara: I am not really sure if it is a hackhack. I agree with (1), but I can see no way to improve this for (2)

bluong: I usually see hackhacks as known not-optimal workarounds that have a possible improvement through an issue fix or similar.  In this regard, I don't think this is a hackhack, but this does warrant a comment possibly.

If there is a more optimal solution, let's quickly articulate it, file and issue and then reference it through the hackhack.

ztsai: yeah, a comment would work, doesn't have to be hackhack

test/scales/colorScaleTests.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+
+    describe("CSS integration", () => {
+      let defaultColors = [
+        "#5279c7", "#fd373e", "#63c261", "#fad419", "#2c2b6f",
+        "#ff7939", "#db2e65", "#99ce50", "#962565", "#06cccc"
+      ];
+
+      it("accepts CSS specified colors", () => {
+        // Make sure you delete this at the end of the test
```
ztsai: it sounds a bit odd to me, maybe reasoning why it should be removed
talked to @bluong, and he suggested something along the line of "style should be removed before the test ends so it does not affect other tests"

aicioara: You mean have that as a comment?

aicioara:
[-1] because I want to enforce an `afterAll` to check no style was forgotten attached (just like the svg check)

aicioara:
Done

PR #2675, opened 2015-08-25
Refactor interpolated color scale tests - Pass 1

aicioara:
Part of #2640 

As part of this: 
- No nested `beforeEach()`es
screen shot 2015-08-25 at 3 31 56 pm

### DIFF
--- a/test/scales/interpolatedColorScaleTests.ts
+++ b/test/scales/interpolatedColorScaleTests.ts
@@ -19,12 +19,10 @@ describe("Scales", () => {
 
         assert.deepEqual(scale.range(), Plottable.Scales.InterpolatedColor.REDS,
           "the range of the default scale is made of shades of red");
-      });
 
-      it("linearly interpolates colors in L*a*b color space", () => {
         scale.domain([0, 1]);
-        assert.strictEqual(scale.scale(1), "#b10026", "domain maximum maps to red");
-        assert.strictEqual(scale.scale(0.5), "#feb24c", "domain median maps in between red and white");
+        assert.strictEqual(scale.scale(1), "#b10026", "new domain maximum maps to red");
+        assert.strictEqual(scale.scale(0.5), "#feb24c", "new domain median maps in between red and white");
         assert.strictEqual(scale.scale(0.9), "#d9151f", "different shades of red are obtained for different values");
       });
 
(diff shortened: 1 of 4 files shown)


test/scales/interpolatedColorScaleTests.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        assert.deepEqual(scale.range(), Plottable.Scales.InterpolatedColor.REDS,
+          "the range of the default scale is made of shades of red");
+      });
+
+      it("linearly interpolates colors in L*a*b color space", () => {
+        scale.domain([0, 1]);
+        assert.strictEqual(scale.scale(1), "#b10026", "domain maximum maps to red");
+        assert.strictEqual(scale.scale(0.5), "#feb24c", "domain median maps in between red and white");
+        assert.strictEqual(scale.scale(0.9), "#d9151f", "different shades of red are obtained for different values");
```
jtlan: Could this be moved to the previous test?

test/scales/interpolatedColorScaleTests.ts
```diff
@@ -0,0 +1,66 @@
...
+      it("clamps overflow and underflow values to range", () => {
+        scale.range(["black", "white"]);
+        scale.domain([0, 16]);
+        assert.strictEqual(scale.scale(0), "#000000", "domain minimum maps to black");
+        assert.strictEqual(scale.scale(16), "#ffffff", "domain maximum maps to white");
+        assert.strictEqual(scale.scale(-100), "#000000", "values smaller than the domain minimum clamp to black");
+        assert.strictEqual(scale.scale(100), "#ffffff", "values larger than the domain maximum clamp to white");
+      });
+
+      it("can be converted to a different range", () => {
```
jtlan: Isn't this effectively tested in any of the other tests where `range()` has been called?

aicioara: You mean tests for other scales?

jtlan:
Is there a better header than "Basic Usage"? Since there's only one tier of tests here, we don't really need a `describe()` for that at all, do we?

aicioara:
I would like to keep the structure symmetric for all the test files. In particular 3 degrees of nesting if possible. We may also argue that the `describe("Scales", () => {` is redundant. 

I don't have any strong preference except for being consistent across test files.

PR #2677, opened 2015-08-26
add wheel plot

ztsai:
part of #2612  (broke the task down to 4 steps)
demo:
http://jsfiddle.net/ztsai/71xu10uz/10/
http://jsfiddle.net/ztsai/597x450n/2/
http://jsfiddle.net/ztsai/tfsLo1y3/2/
A sector is defined by `t`, `t2`, `r`, and `r2`:

```t<T>(startAngle?: number | Accessor<number> | T | Accessor<T>, scale?: QuantitativeScale<T>)```
```t2<T>(endAngle?: number | Accessor<number> | T | Accessor<T>)```
```r<R>(innerRadius?: number | Accessor<number> | R | Accessor<R>, scale?: QuantitativeScale<R>)```
```r2<R>(outerRadius?: number | Accessor<number> | R | Accessor<R>)```

original PR: https://github.com/palantir/plottable/pull/2672

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3590,66 +3590,66 @@ declare module Plottable {
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
             protected _propertyProjectors(): AttributeToProjector;
             /**
-             * Gets the AccessorScaleBinding for t1 in degrees.
+             * Gets the AccessorScaleBinding for t in degrees.
              */
-            t1<T>(): AccessorScaleBinding<T, number>;
+            t<T>(): AccessorScaleBinding<T, number>;
             /**
-             * Sets t1 to a constant number or the result of an Accessor<number> in degrees.
+             * Sets t to a constant number or the result of an Accessor<number> in degrees.
              *
-             * @param {number|Accessor<number>} t1
+             * @param {number|Accessor<number>} t
              * @returns {Wheel} The calling Wheel Plot.
              */
-            t1(t1: number | Accessor<number>): Plots.Wheel<R, T>;
+            t(t: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Sets t1 to a scaled constant value or scaled result of an Accessor in degrees.
+             * Sets t to a scaled constant value or scaled result of an Accessor in degrees.
              * The supplied Scale will also be used for t2().
              * The provided Scale will account for the values when autoDomain()-ing.
              *
-             * @param {T|Accessor<T>} t1
+             * @param {T|Accessor<T>} t
              * @param {QuantitativeScale<T>} scale
              * @returns {Wheel} The calling Wheel Plot.
              */
-            t1<T>(t1: T | Accessor<T>, scale: QuantitativeScale<T>): Plots.Wheel<R, T>;
+            t<T>(t: T | Accessor<T>, scale: QuantitativeScale<T>): Plots.Wheel<R, T>;
             /**
              * Gets the AccessorScaleBinding for t2 in degrees.
              */
             t2<T>(): AccessorScaleBinding<T, number>;
             /**
              * Sets t2 to a constant number or the result of an Accessor<number> in degrees.
-             * If a Scale has been set for t1, it will also be used to scale t2.
+             * If a Scale has been set for t, it will also be used to scale t2.
              *
              * @param {number|Accessor<number>} t2
              * @returns {Wheel} The calling Wheel Plot.
              */
             t2(t2: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Gets the AccessorScaleBinding for r1.
+             * Gets the AccessorScaleBinding for r.
              */
-            r1<R>(): AccessorScaleBinding<R, number>;
+            r<R>(): AccessorScaleBinding<R, number>;
             /**
-             * Sets r1 to a constant number or the result of an Accessor<number>.
+             * Sets r to a constant number or the result of an Accessor<number>.
              *
-             * @param {number|Accessor<number>} r1
+             * @param {number|Accessor<number>} r
              * @returns {Wheel} The calling Wheel Plot.
              */
-            r1(r1: number | Accessor<number>): Plots.Wheel<R, T>;
+            r(r: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Sets r1 to a scaled constant value or scaled result of an Accessor.
+             * Sets r to a scaled constant value or scaled result of an Accessor.
              * The supplied Scale will also be used for r2().
              * The provided Scale will account for the values when autoDomain()-ing.
              *
-             * @param {R|Accessor<R>} r1
+             * @param {R|Accessor<R>} r
              * @param {QuantitativeScale<R>} scale
              * @returns {Wheel} The calling Wheel Plot.
              */
-            r1<R>(r1: R | Accessor<R>, scale: QuantitativeScale<R>): Plots.Wheel<R, T>;
+            r<R>(r: R | Accessor<R>, scale: QuantitativeScale<R>): Plots.Wheel<R, T>;
             /**
              * Gets the AccessorScaleBinding for r2.
              */
             r2<R>(): AccessorScaleBinding<R, number>;
             /**
              * Sets r2 to a constant number or the result of an Accessor<number>.
-             * If a Scale has been set for r1, it will also be used to scale r2.
+             * If a Scale has been set for r, it will also be used to scale r2.
              *
              * @param {number|Accessor<number>} r2
              * @returns {Wheel} The calling Wheel Plot.
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -54,10 +54,7 @@ declare module Plottable {
              * @return {number} dist(p1, p2)^2
              */
             function distanceSquared(p1: Point, p2: Point): number;
-            /**
-             * Converts degree to radian
-             */
-            function degreeToRadian(degree: number): number;
+            function degreesToRadians(degree: number): number;
         }
     }
 }
@@ -3593,18 +3590,18 @@ declare module Plottable {
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
             protected _propertyProjectors(): AttributeToProjector;
             /**
-             * Gets the AccessorScaleBinding for the start angle in degree.
+             * Gets the AccessorScaleBinding for the start angle in degrees.
              */
             startAngle<T>(): AccessorScaleBinding<T, number>;
             /**
-             * Sets the start angle to a constant number or the result of an Accessor<number> in degree.
+             * Sets the start angle to a constant number or the result of an Accessor<number> in degrees.
              *
              * @param {number|Accessor<number>} startAngle
              * @returns {Wheel} The calling Wheel Plot.
              */
             startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Sets the start angle to a scaled constant value or scaled result of an Accessor in degree.
+             * Sets the start angle to a scaled constant value or scaled result of an Accessor in degrees.
              * The supplied Scale will also be used for endAngle().
              * The provided Scale will account for the values when autoDomain()-ing.
              *
@@ -3614,11 +3611,11 @@ declare module Plottable {
              */
             startAngle<T>(startAngle: T | Accessor<T>, scale: Scale<T, number>): Plots.Wheel<R, T>;
             /**
-             * Gets the AccessorScaleBinding for the end angle in degree.
+             * Gets the AccessorScaleBinding for the end angle in degrees.
              */
             endAngle<T>(): AccessorScaleBinding<T, number>;
             /**
-             * Sets the end angle to a constant number or the result of an Accessor<number> in degree.
+             * Sets the end angle to a constant number or the result of an Accessor<number> in degrees.
              * If a Scale has been set for startAngle, it will also be used to scale endAngle.
              *
              * @param {number|Accessor<number>} endAngle
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3590,71 +3590,71 @@ declare module Plottable {
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
             protected _propertyProjectors(): AttributeToProjector;
             /**
-             * Gets the AccessorScaleBinding for the start angle in degrees.
+             * Gets the AccessorScaleBinding for t1 in degrees.
              */
-            startAngle<T>(): AccessorScaleBinding<T, number>;
+            t1<T>(): AccessorScaleBinding<T, number>;
             /**
-             * Sets the start angle to a constant number or the result of an Accessor<number> in degrees.
+             * Sets t1 to a constant number or the result of an Accessor<number> in degrees.
              *
-             * @param {number|Accessor<number>} startAngle
+             * @param {number|Accessor<number>} t1
              * @returns {Wheel} The calling Wheel Plot.
              */
-            startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+            t1(t1: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Sets the start angle to a scaled constant value or scaled result of an Accessor in degrees.
-             * The supplied Scale will also be used for endAngle().
+             * Sets t1 to a scaled constant value or scaled result of an Accessor in degrees.
+             * The supplied Scale will also be used for t2().
              * The provided Scale will account for the values when autoDomain()-ing.
              *
-             * @param {T|Accessor<T>} startAngle
-             * @param {Scale<T, number>} scale
+             * @param {T|Accessor<T>} t1
+             * @param {QuantitativeScale<T>} scale
              * @returns {Wheel} The calling Wheel Plot.
              */
-            startAngle<T>(startAngle: T | Accessor<T>, scale: Scale<T, number>): Plots.Wheel<R, T>;
+            t1<T>(t1: T | Accessor<T>, scale: QuantitativeScale<T>): Plots.Wheel<R, T>;
             /**
-             * Gets the AccessorScaleBinding for the end angle in degrees.
+             * Gets the AccessorScaleBinding for t2 in degrees.
              */
-            endAngle<T>(): AccessorScaleBinding<T, number>;
+            t2<T>(): AccessorScaleBinding<T, number>;
             /**
-             * Sets the end angle to a constant number or the result of an Accessor<number> in degrees.
-             * If a Scale has been set for startAngle, it will also be used to scale endAngle.
+             * Sets t2 to a constant number or the result of an Accessor<number> in degrees.
+             * If a Scale has been set for t1, it will also be used to scale t2.
              *
-             * @param {number|Accessor<number>} endAngle
+             * @param {number|Accessor<number>} t2
              * @returns {Wheel} The calling Wheel Plot.
              */
-            endAngle(endAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+            t2(t2: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Gets the AccessorScaleBinding for the inner radius.
+             * Gets the AccessorScaleBinding for r1.
              */
-            innerRadius<R>(): AccessorScaleBinding<R, number>;
+            r1<R>(): AccessorScaleBinding<R, number>;
             /**
-             * Sets the inner radius to a constant number or the result of an Accessor<number>.
+             * Sets r1 to a constant number or the result of an Accessor<number>.
              *
-             * @param {number|Accessor<number>} innerRadius
+             * @param {number|Accessor<number>} r1
              * @returns {Wheel} The calling Wheel Plot.
              */
-            innerRadius(innerRadius: number | Accessor<number>): Plots.Wheel<R, T>;
+            r1(r1: number | Accessor<number>): Plots.Wheel<R, T>;
             /**
-             * Sets the inner radius to a scaled constant value or scaled result of an Accessor.
-             * The supplied Scale will also be used for outerRadius().
+             * Sets r1 to a scaled constant value or scaled result of an Accessor.
+             * The supplied Scale will also be used for r2().
              * The provided Scale will account for the values when autoDomain()-ing.
              *
-             * @param {R|Accessor<R>} innerRadius
-             * @param {Scale<R, number>} scale
+             * @param {R|Accessor<R>} r1
+             * @param {QuantitativeScale<R>} scale
              * @returns {Wheel} The calling Wheel Plot.
              */
-            innerRadius<R>(innerRadius: R | Accessor<R>, scale: Scale<R, number>): Plots.Wheel<R, T>;
+            r1<R>(r1: R | Accessor<R>, scale: QuantitativeScale<R>): Plots.Wheel<R, T>;
             /**
-             * Gets the AccessorScaleBinding for the outer radius.
+             * Gets the AccessorScaleBinding for r2.
              */
-            outerRadius<R>(): AccessorScaleBinding<R, number>;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/plots/wheelPlot.ts [hunk 1]
```diff
@@ -0,0 +1,222 @@
...
+     */
+    public endAngle(endAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    public endAngle<T>(endAngle?: number | Accessor<number> | T | Accessor<T>): any {
+      if (endAngle == null) {
+        return this._propertyBindings.get(Wheel._END_ANGLE_KEY);
+      }
+
+      let startAngleBinding = this.startAngle();
+      let angleScale = startAngleBinding && startAngleBinding.scale;
+      this._bindProperty(Wheel._END_ANGLE_KEY, endAngle, angleScale);
```
jtlan: I notice that this logic isn't symmetric with the logic in `startAngle()`, which causes some odd behavior. For example, this works:
```Typescript
plot.endAngle(function(d){ return d.t2; });
plot.startAngle(function(d){ return d.t1; }, tScale);
```
but this doesn't:
```Typescript
plot.startAngle(function(d){ return d.t1; });
plot.endAngle(function(d){ return d.t2; }, tScale);
```

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+     */
+    public outerRadius(outerRadius: number | Accessor<number>): Plots.Wheel<R, T>;
+    public outerRadius<R>(outerRadius?: number | Accessor<number> | R | Accessor<R>): any {
+      if (outerRadius == null) {
+        return this._propertyBindings.get(Wheel._OUTER_RADIUS_KEY);
+      }
+
+      let innerRadiusBinding = this.innerRadius();
+      let radiusScale = innerRadiusBinding && innerRadiusBinding.scale;
+      this._bindProperty(Wheel._OUTER_RADIUS_KEY, outerRadius, radiusScale);
```
jtlan: Same "binding symmetry" problem as the angles here.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {S|Accessor<number>} startAngle
+     * @param {Scale<T, number>} scale
```
jtlan: Inconsistency in docs here.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the end angle.
+     */
+    public endAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the end angle to a constant number or the result of an Accessor<number>.
+     *
```
jtlan: Can we add some documentation explaining that it uses the same `Scale` as `startAngle()`?

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the outer radius.
+     */
+    public outerRadius<R>(): AccessorScaleBinding<R, number>;
+    /**
+     * Sets the outer radius to a constant number or the result of an Accessor<number>.
+     *
```
jtlan: Can we add some documentation explaining that it uses the same `Scale` as `innerRadius()`?

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor.
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {S|Accessor<number>} startAngle
```
jtlan: Should be `{T|Accessor<T>}`.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,222 @@
...
+    protected _propertyProjectors(): AttributeToProjector {
+      let attrToProjector = super._propertyProjectors();
+      let innerRadiusAccessor = Plot._scaledAccessor(this.innerRadius());
+      let outerRadiusAccessor = Plot._scaledAccessor(this.outerRadius());
+      let startAngleAccessor = Plot._scaledAccessor(this.startAngle());
+      let endAngleAccessor = Plot._scaledAccessor(this.endAngle());
+      attrToProjector["d"] = (datum: any, index: number, ds: Dataset) => {
+        let startAngle = startAngleAccessor(datum, index, ds) % (Math.PI * 2);
+        let endAngle = endAngleAccessor(datum, index, ds) % (Math.PI * 2);
+        while (endAngle < startAngle) { endAngle += Math.PI * 2; }
```
jtlan: I think the logic here is incorrect. I would expect something with a `startAngle()` of 0 and an `endAngle()` of 2π to draw a full disk but instead nothing is drawn: http://jsfiddle.net/71xu10uz/

plottable.d.ts
```diff
@@ -3577,6 +3577,93 @@ declare module Plottable {
...
+            startAngle<T>(): AccessorScaleBinding<T, number>;
+            /**
+             * Sets the start angle to a constant number or the result of an Accessor<number>.
+             *
+             * @param {number|Accessor<number>} startAngle
+             * @returns {Wheel} The calling Wheel Plot.
+             */
+            startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+            /**
+             * Sets the start angle to a scaled constant value or scaled result of an Accessor.
```
jtlan: Currently this method expects radians, but I wonder if we should expect degrees instead? `angle()` on `Label` and `tickLabelAngle()` on `Axes.Category` take degrees, for example (but not all possible values). Maybe worth an RFC? Or follow up on the existing thread.

Either way, we should specify the expected units in the docs.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+      };
+      return attrToProjector;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the start angle.
+     */
+    public startAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the start angle to a constant number or the result of an Accessor<number>.
```
jtlan: Currently this method expects radians, but I wonder if we should expect degrees instead? `angle()` on `Label` and `tickLabelAngle()` on `Axes.Category` take degrees, for example (but not all possible values). Maybe worth an RFC? Or follow up on the existing thread.

Either way, we should specify the expected units in the docs.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the end angle.
+     */
+    public endAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the end angle to a constant number or the result of an Accessor<number>.
+     * The scale of end angle is set to be the scale of start angle.
```
jtlan: We should probably be consistent with the phrasing in `Plots.Rectangle`, which reads
```
If a Scale has been set for X, it will also be used to scale X2.
```

So in this case, "`If a Scale has been set for startAngle, it will also be used to scale endAngle.`".

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Sets the start angle to a constant number or the result of an Accessor<number>.
+     *
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor.
+     * This also sets the scale for end angle.
```
jtlan: "The supplied scale will also be used for endAngle()."

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Sets the inner radius to a constant number or the result of an Accessor<number>.
+     *
+     * @param {number|Accessor<number>} innerRadius
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public innerRadius(innerRadius: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the inner radius to a scaled constant value or scaled result of an Accessor.
+     * This also sets the scale for outer radius.
```
jtlan: "The supplied scale will also be used for innerAngle()."

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+      return this;
+    }
+
+    /**
+     * Gets the AccessorScaleBinding for the outer radius.
+     */
+    public outerRadius<R>(): AccessorScaleBinding<R, number>;
+    /**
+     * Sets the outer radius to a constant number or the result of an Accessor<number>.
+     * The scale of outer radius is set to be the scale of inner radius.
```
jtlan: "If a Scale has been set for innerRadius, it will also be used to scale outerRadius."

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+        let transform = d3.transform(renderArea.attr("transform")).translate;
+        assert.closeTo(transform[0], SVG_WIDTH / 2, 0.01, "origin is set to the middle of the svg");
+        assert.closeTo(transform[1], SVG_HEIGHT / 2, 0.01, "origin is set to the middle of the svg");
+        let radiusLimit = Math.min(SVG_WIDTH, SVG_HEIGHT) / 2;
+        assert.deepEqual(rScale.range(), [0, radiusLimit] , "the radius scale range is set properly");
+
+        svg.remove();
+      });
+
+      it("draws arc clockwise from startAngle toendAngle", () => {
```
jtlan: Missing a space here.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+        let slices = wheelPlot.selections();
+        data.forEach((datum, i) => {
+          let startAngle = datum.t1;
+          let endAngle = datum.t2;
+          let expectedEndAngle = datum.expected;
+          let path = d3.select(slices[0][i]).attr("d");
+          let arc = d3.svg.arc().innerRadius(rScale.scale(datum.r1)).outerRadius(rScale.scale(datum.r2))
+                                .startAngle(tScale.scale(startAngle)).endAngle(tScale.scale(expectedEndAngle));
+          let expectedPath = arc(null);
+          TestMethods.assertAreaPathCloseTo(path, expectedPath, 0.1, `arc is drawn from ${startAngle} to ${endAngle}`);
```
jtlan: I forgot about template strings...!

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+            { r1: 3, r2: 4, t1: -60, t2: -180, expected: 180 }
+        ];
+        let dataset = new Plottable.Dataset(data);
+        wheelPlot.addDataset(dataset);
+        wheelPlot.renderTo(svg);
+        let slices = wheelPlot.selections();
+        data.forEach((datum, i) => {
+          let startAngle = datum.t1;
+          let endAngle = datum.t2;
+          let expectedEndAngle = datum.expected;
```
jtlan: Maybe name the property `expectedEndAngle` then?

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+        assert.strictEqual(wheelPlot.outerRadius().accessor(data[0], 0, dataset), 1, "access outerRadius correctly without a scale");
+        assert.strictEqual(wheelPlot.outerRadius().accessor(data[1], 1, dataset), 2, "access outerRadius correctly without a scale");
+        assert.isUndefined(wheelPlot.outerRadius().scale, "scale of outerRadius is undefined");
+
+        svg.remove();
+      });
+
+      it("sets scale of outerRadius when scale of innerRadius is set", () => {
+        wheelPlot.outerRadius((d) => d.r2);
+        wheelPlot.renderTo(svg);
```
jtlan: shouldn't be required, right?

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+        wheelPlot.endAngle((d) => d.t2);
+        assert.strictEqual(wheelPlot.endAngle().accessor(data[0], 0, dataset), 180, "access endAngle correctly without a scale");
+        assert.strictEqual(wheelPlot.endAngle().accessor(data[1], 1, dataset), 360, "access endAngle correctly without a scale");
+        assert.isUndefined(wheelPlot.endAngle().scale, "scale of endAngle is undefined");
+        svg.remove();
+      });
+
+      it("sets scale of endAngle when scale of startAngle is set", () => {
+        wheelPlot.endAngle((d) => d.t2);
+        wheelPlot.renderTo(svg);
```
jtlan: Also not required for this test.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+      });
+
+      it("sets scale of outerRadius when scale of innerRadius is set", () => {
+        wheelPlot.outerRadius((d) => d.r2);
+        wheelPlot.renderTo(svg);
+
+        assert.isUndefined(wheelPlot.outerRadius().scale, "scale of outerRadius is undefined initially");
+
+        wheelPlot.innerRadius((d) => d.r1, rScale);
+        assert.deepEqual(wheelPlot.outerRadius().scale, rScale, "scale of outerRadius is set to be the same scale as innerRadius");
```
jtlan: Might be worth confirming that setting `outerRadius()` after setting `innerRadius()` still picks up the `Scale`.

test/plots/wheelPlotTests.ts
```diff
@@ -0,0 +1,296 @@
...
+      });
+
+      it("sets scale of endAngle when scale of startAngle is set", () => {
+        wheelPlot.endAngle((d) => d.t2);
+        wheelPlot.renderTo(svg);
+
+        assert.isUndefined(wheelPlot.endAngle().scale, "scale of endAngle is undefined initially");
+
+        wheelPlot.startAngle((d) => d.t1, tScale);
+        assert.deepEqual(wheelPlot.endAngle().scale, tScale, "scale of endAngle is set to be the same scale of startAngle");
```
jtlan: Might be worth confirming that setting `endAngle()` after setting `startAngle()` still picks up the `Scale`.

src/utils/mathUtils.ts
```diff
@@ -103,6 +103,13 @@ export module Utils {
     export function distanceSquared(p1: Point, p2: Point) {
       return nativeMath.pow(p2.y - p1.y, 2) + nativeMath.pow(p2.x - p1.x, 2);
     }
+
+    /**
+     * Converts degree to radian
+     */
+     export function degreeToRadian(degree: number) {
```
jtlan: I think `degreesToRadians()` would be a better name. Since the name encapsulates the entire functionality, you don't need docs.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,228 @@
...
+    public startAngle<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets the start angle to a constant number or the result of an Accessor<number> in degree.
+     *
+     * @param {number|Accessor<number>} startAngle
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public startAngle(startAngle: number | Accessor<number>): Plots.Wheel<R, T>;
+    /**
+     * Sets the start angle to a scaled constant value or scaled result of an Accessor in degree.
```
jtlan: "[...] in degrees"

jtlan:
In multiple locations "in degrees" should be used instead of "in degree".

crmorford:
what are the units of inner and outer radius?
I expected they would be pixels, and i can't figure out what they are here.

crmorford:
going from 0 degrees to 500 degrees wraps all the way around, but 500 degrees to 0 degrees doesn't get a full rotation in. I'd prefer if these were consistent.
http://jsfiddle.net/71xu10uz/2/

crmorford:
I don't think wheel plot needs padding on the t-axis. I can't think of a case someone would want it, if the point of timewheels is to show cyclical data. http://jsfiddle.net/71xu10uz/3/
screen shot 2015-08-27 at 3 09 05 pm

crmorford:
Looks like right now this only supports Quantitative scales (let me know if i'm wrong). Slate is going to need the ability to do category (for both scales). The use case I know of is Days of Week on r-axis, hour of day on t-axis.

ztsai:
> going from 0 degrees to 500 degrees wraps all the way around, but 500 degrees to 0 degrees doesn't get a full rotation in. I'd prefer if these were consistent.
> http://jsfiddle.net/71xu10uz/2/

it is by design, we always go from first angle to the second angle clockwise
@jtlan commented about this in the original PR (https://github.com/palantir/plottable/pull/2672)
but if you feel strongly about it, we can work on the expected behavior for this as well

crmorford:
> it is by design, we always go from first angle to the second angle clockwise

ah ok, that seems logical. i'm ok with it now.

crmorford:
This fiddle has a wheel plot and 2 rectangle plots. http://jsfiddle.net/71xu10uz/4/
Even though two x-y plots can happily share a scale, it causes big problems here

Wheel Plot alone:
screen shot 2015-08-27 at 5 10 25 pm

Wheel Plot, with renderRect_1 uncommented:
screen shot 2015-08-27 at 5 10 34 pm

Is there anything we can do about this?

crmorford:
I can't figure out how to get a stroke.
Attempt here:
http://jsfiddle.net/71xu10uz/6/

ztsai:
for scale interference, I think what happened here is that rectangle did a auto domain for tScale and I doubt if we can do anything about it on the wheel side to lock the domain/range

as for stroke, that's due to the stroke fix we made for pie selection. right now wheel plot doesn't support stroke, and there is a follow up issue to fix it (https://github.com/palantir/plottable/issues/2685)

crmorford:
for symmetry with rectangle and segment plots, which have the accessors:
.x(), .x2(), .y, .y2(), 
can we change .r1() --> .r() and .t1() --> .t()?

crmorford:
I think  it makes sense to error on or throw out data that returns a negative radius.
@jtlan, do you have an opinion on this?

ztsai:
do we want to restrict the radius domain to be non-negative?

jtlan:
@crmorford wrote:
>I think it makes sense to error on or throw out data that returns a negative radius.

@ztsai wrote:
>do we want to restrict the radius domain to be non-negative?

What if we just reject things that result in a negative *pixel* value?

crmorford:
> What if we just reject things that result in a negative pixel value?
works for me

jtlan:
To some extent, negative values on the r *domain* don't make much sense either, but that's more of a user-supplied-data issue I think...

crmorford:
ok negative pixel values rejected looks good http://jsfiddle.net/71xu10uz/9/,
I like the API points now
I think we should prioritize putting stroke on this, it's hard to tell what's happening without it, when you have sparse data.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+     */
+    public t2<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets t2 to a constant number or the result of an Accessor<number> in degrees.
+     * If a Scale has been set for t, it will also be used to scale t2.
+     *
+     * @param {number|Accessor<number>} t2
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public t2(t2: number | Accessor<number>): Plots.Wheel<R, T>;
```
jtlan: Wrong signature? Should be `number | Accessor<number> | T | Accessor<T>`, as below.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+     */
+    public r2<R>(): AccessorScaleBinding<R, number>;
+    /**
+     * Sets r2 to a constant number or the result of an Accessor<number>.
+     * If a Scale has been set for r, it will also be used to scale r2.
+     *
+     * @param {number|Accessor<number>} r2
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public r2(r2: number | Accessor<number>): Plots.Wheel<R, T>;
```
jtlan: Signature should be `number | Accessor<number> | R | Accessor<R>`.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+     * Sets t to a scaled constant value or scaled result of an Accessor in degrees.
+     * The supplied Scale will also be used for t2().
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {T|Accessor<T>} t
+     * @param {QuantitativeScale<T>} scale
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public t<T>(t: T | Accessor<T>, scale: QuantitativeScale<T>): Plots.Wheel<R, T>;
+    public t<T>(t?: number | Accessor<number> | T | Accessor<T>, scale?: QuantitativeScale<T>): any {
```
jtlan: The method itself shouldn't need to be generic on `<T>`

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+    public t2<T>(): AccessorScaleBinding<T, number>;
+    /**
+     * Sets t2 to a constant number or the result of an Accessor<number> in degrees.
+     * If a Scale has been set for t, it will also be used to scale t2.
+     *
+     * @param {number|Accessor<number>} t2
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public t2(t2: number | Accessor<number>): Plots.Wheel<R, T>;
+    public t2<T>(t2?: number | Accessor<number> | T | Accessor<T>): any {
```
jtlan: Same thing here -- the method itself doesn't need to be generic on `<T>`.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+     * Sets r to a scaled constant value or scaled result of an Accessor.
+     * The supplied Scale will also be used for r2().
+     * The provided Scale will account for the values when autoDomain()-ing.
+     *
+     * @param {R|Accessor<R>} r
+     * @param {QuantitativeScale<R>} scale
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public r<R>(r: R | Accessor<R>, scale: QuantitativeScale<R>): Plots.Wheel<R, T>;
+    public r<R>(r?: number | Accessor<number> | R | Accessor<R>, scale?: QuantitativeScale<R>): any {
```
jtlan: Doesn't need to be generic on `<R>`.

src/plots/wheelPlot.ts
```diff
@@ -0,0 +1,233 @@
...
+    public r2<R>(): AccessorScaleBinding<R, number>;
+    /**
+     * Sets r2 to a constant number or the result of an Accessor<number>.
+     * If a Scale has been set for r, it will also be used to scale r2.
+     *
+     * @param {number|Accessor<number>} r2
+     * @returns {Wheel} The calling Wheel Plot.
+     */
+    public r2(r2: number | Accessor<number>): Plots.Wheel<R, T>;
+    public r2<R>(r2?: number | Accessor<number> | R | Accessor<R>): any {
```
jtlan: Doesn't need to be generic on `<R>`.

PR #2679, opened 2015-08-26
Refactor linear scale tests - Pass 1

aicioara:
Part of #2640 

As part of this: 
- No nested `beforeEach()`es
- Removed annoying console warnings
screen shot 2015-08-26 at 11 56 41 am

### DIFF
--- a/test/scales/linearScaleTests.ts
+++ b/test/scales/linearScaleTests.ts
@@ -313,80 +313,101 @@ describe("Scales", () => {
     });
 
     describe("Plot interaction", () => {
-      let data = [
-        {foo: 2, bar: 1},
-        {foo: 5, bar: -20},
-        {foo: 0, bar: 0}
-      ];
+      let data: any[];
       let dataset: Plottable.Dataset;
       let scale: Plottable.Scales.Linear;
 
       beforeEach(() => {
+        data = [{x: 2}, {x: 5}, {x: 0}];
         dataset = new Plottable.Dataset(data);
         scale = new Plottable.Scales.Linear();
         scale.padProportion(0);
       });
 
       it("scale autorange works as expected with single dataset", () => {
-        let svg = TestMethods.generateSVG(100, 100);
+        let svg = TestMethods.generateSVG();
         let plot = new Plottable.Plot();
         (<any> plot)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
-        plot.addDataset(dataset)
-            .attr("x", (d) => d.foo, scale)
-            .renderTo(svg);
+        plot.addDataset(dataset);
+        plot.attr("x", (d) => d.x, scale);
+        plot.renderTo(svg);
         assert.deepEqual(scale.domain(), [0, 5], "scale domain was autoranged properly");
-        data.push({foo: 100, bar: 200});
-        dataset.data(data);
-        assert.deepEqual(scale.domain(), [0, 100], "scale domain was autoranged properly");
+
+        dataset.data([{x: 100}]);
+        assert.deepEqual(scale.domain(), [99, 101], "scale domain was updated properly");
+
         svg.remove();
       });
 
       it("scale reference counting works as expected", () => {
-        let svg1 = TestMethods.generateSVG(100, 100);
-        let svg2 = TestMethods.generateSVG(100, 100);
-        let renderer1 = new Plottable.Plot();
-        (<any> renderer1)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
-        renderer1.addDataset(dataset).attr("x", (d) => d.foo, scale);
-        renderer1.renderTo(svg1);
-        let renderer2 = new Plottable.Plot();
-        (<any> renderer2)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
-        renderer2.addDataset(dataset).attr("x", (d) => d.foo, scale);
-        renderer2.renderTo(svg2);
+        let svg1 = TestMethods.generateSVG();
+        let svg2 = TestMethods.generateSVG();
+
+        let plot1 = new Plottable.Plot();
+        (<any> plot1)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
+        plot1.addDataset(dataset).attr("x", (d) => d.x, scale);
+        plot1.renderTo(svg1);
+
+        let plot2 = new Plottable.Plot();
+        (<any> plot2)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
+        plot2.addDataset(dataset).attr("x", (d) => d.x, scale);
+        plot2.renderTo(svg2);
+
+        assert.deepEqual(scale.domain(), [0, 5], "correct domain is initially set for the scale");
+
         let otherScale = new Plottable.Scales.Linear();
-        renderer1.attr("x", (d) => d.foo, otherScale);
-        dataset.data([{foo: 10}, {foo: 11}]);
+        plot1.attr("x", (d) => d.x, otherScale);
+
+        assert.deepEqual(scale.domain(), [0, 5], "the domain for the scale did not change as it is still attached to plot 2");
+        dataset.data([{x: 10}, {x: 11}]);
         assert.deepEqual(scale.domain(), [10, 11], "scale was still listening to dataset after one perspective deregistered");
-        renderer2.attr("x", (d) => d.foo, otherScale);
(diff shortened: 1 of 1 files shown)


test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,267 @@
...
+        scale.range([400, 500]);
+        assert.strictEqual(scale.scale(10), 400, "first value in domain maps to first value in range");
+        assert.strictEqual(scale.scale(20), 500, "last value in domain maps to last value in range");
+        assert.strictEqual(scale.scale(15), 450, "middle value in domain maps to middle value in range");
+
+        assert.strictEqual(scale.invert(400), 10, "first value in range maps to first value in domain");
+        assert.strictEqual(scale.invert(500), 20, "last value in range maps to last value in domain");
+        assert.strictEqual(scale.invert(450), 15, "middle value in range maps to middle value in domain");
+
+        scale.domain([20, 10]);
```
jtlan: Do you think it's worth pulling the reversed-domain tests into a different `it()`?

test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,267 @@
...
+        let arrayWithBadValues: any[] = [null, NaN, undefined, Infinity, -Infinity, "a string", 1, 1.2];
+        let expectedExtent = [1, 1.2];
+        let extent = scale.extentOfValues(arrayWithBadValues);
+        assert.deepEqual(extent, expectedExtent, "invalid values were filtered out");
+      });
+
+      it("does not accept NaN or Infinity as domain", () => {
+        let warningCalled = false;
+        let oldWarn = Plottable.Utils.Window.warn;
+        Plottable.Utils.Window.warn = (msg: string) => {
```
jtlan: If this is common enough in tests, maybe we should write `TestMethod.assert.warns()`?

test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,267 @@
...
+        scale.range([-Infinity, 5]);
+        assert.deepEqual(scale.range(), [-Infinity, 5], "-Infinity accepted as part of range");
+        scale.range([NaN, 7]);
+        assert.deepEqual(scale.range(), [NaN, 7], "NaN accepted as part of range");
+      });
+
+      it("can receive custom tick generator", () => {
+        scale.domain([0, 10]);
+        let defaultTicks = scale.ticks();
+        assert.strictEqual(defaultTicks.length, 11, "ticks were generated correctly with default generator");
```
jtlan: Do we want to assert that specific values were generated, as in the later `assert()`?

test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,267 @@
...
-      scale.domain([-1, 5]);
-      assert.deepEqual(scale.domain(), [-1, 5], "Regular domains still accepted");
+    describe("Auto Domaining", () => {
+      let scale: Plottable.Scales.Linear;
+
+      beforeEach(() => {
+        scale = new Plottable.Scales.Linear();
+      });
+
+      it("auto domains by default", () => {
```
jtlan: Technically this tests the default domain with no extents, rather than that it auto domains by default...

test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,267 @@
...
+        assert.deepEqual(scale.domain(), [0, 1], "the default domain is [0, 1]");
+
+        scale.autoDomain();
+        assert.deepEqual(scale.domain(), [0, 1], "the domain is still [0, 1]");
+      });
+
+      it("modififes the autodomain flag accordingly", () => {
+        scale.addIncludedValuesProvider((scale) => [0, 5]);
+        assert.isTrue((<any> scale)._autoDomainAutomatically,
```
jtlan: I'm not really comfortable with `<any>`-casting in to check a flag, since really we want to check that the behavior is correct, but that flag's state isn't a guarantee of that. Can we create a test that adds/removes `IncludedValueProvider`s when the scale is in either mode, and uses that to infer if the autodomaining is working correctly?

An analogous case is the value/pixel mode on `GuideLineLayer`, which is tested via behavior instead of checking the flag.

test/scales/linearScaleTests.ts
```diff
@@ -204,109 +330,109 @@ describe("Scales", () => {
...
-        scale.addIncludedValuesProvider((scale: Plottable.Scale<number, number>) => d3.extent(data, (e) => e.foo));
-        assert.isTrue((<any> scale)._autoDomainAutomatically,
-                            "the autoDomain flag is still set after autoranginging and padding and nice-ing");
-        scale.domain([0, 5]);
-        assert.isFalse((<any> scale)._autoDomainAutomatically, "the autoDomain flag is false after domain explicitly set");
-      });
-
-      it("scale autorange works as expected with single dataset", () => {
-        let svg = TestMethods.generateSVG(100, 100);
+      it("receives updates from the plot and autodomains in accordange", () => {
```
jtlan: `"and autodomains accordingly"`

Does this make more sense as a `Plot` test?

test/scales/linearScaleTests.ts
```diff
@@ -204,109 +330,109 @@ describe("Scales", () => {
...
+
         (<any> plot)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
-        plot.addDataset(dataset)
-            .attr("x", (d) => d.foo, scale)
-            .renderTo(svg);
+        plot.addDataset(dataset);
+        plot.attr("x", (d) => d.x, scale);
+        plot.renderTo(svg);
+
         assert.deepEqual(scale.domain(), [0, 5], "scale domain was autoranged properly");
```
jtlan: `"was autodomained correctly"`

test/scales/linearScaleTests.ts
```diff
@@ -204,109 +330,109 @@ describe("Scales", () => {
...
+
+        assert.deepEqual(scale.domain(), [0, 5], "the domain for the scale did not change as it is still attached to plot 2");
+        dataset.data([{x: 10}, {x: 11}]);
         assert.deepEqual(scale.domain(), [10, 11], "scale was still listening to dataset after one perspective deregistered");
-        renderer2.attr("x", (d) => d.foo, otherScale);
-        // "scale not listening to the dataset after all perspectives removed"
-        dataset.data([{foo: 99}, {foo: 100}]);
+
+        plot2.attr("x", (d) => d.x, otherScale);
+        assert.deepEqual(scale.domain(), [0, 1], "scale resetted to the default domain as it is no longer attached to plots");
```
jtlan: `"[...] reset to the default domain [...]"`

test/scales/linearScaleTests.ts
aicioara: yup

test/scales/linearScaleTests.ts
aicioara: Well, we do that later anyway. Shall I just delete this test?

test/scales/linearScaleTests.ts
aicioara: It is not really a common scenario, but I am fine with setting up the method. What exactly do you mean?

test/scales/linearScaleTests.ts
aicioara: Possibly, this is a RFC all the scale + plot integration tests

test/scales/linearScaleTests.ts
aicioara: Done

test/scales/linearScaleTests.ts
```diff
@@ -1,134 +1,289 @@
...
+        assert.deepEqual(scale.range(), [-Infinity, 5], "-Infinity accepted as part of range");
+        scale.range([NaN, 7]);
+        assert.deepEqual(scale.range(), [NaN, 7], "NaN accepted as part of range");
+      });
+
+      it("can receive custom tick generator", () => {
+        scale.domain([0, 10]);
+        let defaultTicks = scale.ticks();
+        assert.strictEqual(defaultTicks.length, 11, "ticks were generated correctly with default generator");
+        assert.deepEqual(defaultTicks, Array.apply(null, Array(11)).map((d: any, i: number) => i),
```
jtlan: I think it would be clearer if the array from 0 to 10 was written out explicitly:
```Javascript
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

test/scales/linearScaleTests.ts
jtlan: Yeah, if that's already being tested, then this seems redundant.

jtlan:
[NOQE], test code only.

PR #2680, opened 2015-08-26
[SelectionBoxLayer] - bounds now explicitly sets pixels, xExtent / yExtent setter added

bluong:
There adds separate setter for setting the data edges of the `SelectionBoxLayer` through the setter version of `xExtent` and `yExtent`.

Mirroring the path taken with `GuideLineLayer`, there is such a separate setter for setting the data bounds of the `SelectionBoxLayer`.  In this regard, there is an idea of modes, where if `bounds()` is used, then the box will obey the pixel values strictly, whereas these new setters will have the box obey the data values strictly (scaling the data values accordingly to pixels).  To reduce API complexity, the data setters have been added through `xExtent` / `yExtent` setters

JSFiddle: http://jsfiddle.net/bluong63/ewga1des/

API Additions:
```typescript
/**
 * Sets the data values backing the left and right edges of the box.
 */
xExtent(xExtent: (number | { valueOf(): number; })[]): SelectionBoxLayer;
/**
 * Sets the data values backing the top and bottom edges of the box.
 */
yExtent(yExtent: (number | { valueOf(): number; })[]): SelectionBoxLayer;
```


aicioara:
Blocking on the absence of the Fiddle. Ping somewhere when done

aicioara:
TL;DR (AFAIK)
before
```typescript
dragBoxLayer.bounds({topLeft: {x: xScale.scale(0.1), y: yScale.scale(0.1) }, bottomRight: {x: xScale.scale(0.4), y: yScale.scale(0.5) }})
```
after
```typescript
dragboxLayer.xExtent([0.1, 0.4]).yExtent([0.1, 0.6])
```

src/components/selectionBoxLayer.ts
```diff
@@ -2,6 +2,7 @@
 
 module Plottable {
 export module Components {
+  enum PropertyMode { VALUE, PIXEL };
```
aicioara: This has already been defined in `guideLineLayer`, hence can be removed from here. 
Even better, extract it from there, put it into `interfaces` or `component` so we can find it easier now.

src/components/selectionBoxLayer.ts
```diff
@@ -83,11 +81,53 @@ export module Components {
         topLeft: topLeft,
         bottomRight: bottomRight
       };
-      this._bindBoxDataValues();
+    }
+
+    // sets pixelPosition() or value() based on the other, depending on which was the last one set
```
aicioara: JSDocs

src/components/selectionBoxLayer.ts
```diff
@@ -83,11 +81,53 @@ export module Components {
...
+          let bottomRight = {
+            x: this._xScale.scale(this._xExtent[1]),
+            y: this.bounds().bottomRight.y
+          };
+          this._setBounds({
+            topLeft: topLeft,
+            bottomRight: bottomRight
+          });
+        } else if (this._xBoundsMode === PropertyMode.PIXEL) {
+          this._xExtent = [this._xScale.invert(this.bounds().topLeft.x), this._xScale.invert(this.bounds().bottomRight.x)];
```
aicioara: I don't really like the idea of using both `_boxBounds` and `_xExtent` / `yExtent`. Now we have to keep those two in sync, which opens room for a lot of subtle bugs. They can be easily deduced from each other using the scales.

src/components/selectionBoxLayer.ts
```diff
@@ -83,11 +81,53 @@ export module Components {
...
+        }
+      }
+      if (this._yScale != null) {
```
aicioara: Scenario (not sure if potential bug but):
- Set bounds to pixel equivalent of [a, b]
- Set yExtent to [c, d]
- Ask for bounds -> they will return pixel equivalent of [c, d]
- Remove yScale
- Ask for bounds -> they will return pixel equivalent of [a, b]

src/components/selectionBoxLayer.ts
```diff
@@ -151,7 +191,8 @@ export module Components {
       }
       this._xScale = xScale;
       this._xScale.onUpdate(this._adjustBoundsCallback);
-      this._bindBoxDataValues();
+      this._syncPixelPositionAndValue();
+      this.render();
```
aicioara: Now you will call `_syncPixelPositionAndValue()` twice. Once here and once inside the `renderImmediately()`.

src/components/selectionBoxLayer.ts
```diff
@@ -174,7 +215,8 @@ export module Components {
       }
       this._yScale = yScale;
       this._yScale.onUpdate(this._adjustBoundsCallback);
-      this._bindBoxDataValues();
+      this._syncPixelPositionAndValue();
+      this.render();
```
aicioara: Same as above

src/components/selectionBoxLayer.ts
```diff
@@ -183,9 +225,21 @@ export module Components {
      *
      * Returns an undefined array if the edges are not backed by a scale.
```
aicioara: How about if edges are backed by a scale?

src/components/selectionBoxLayer.ts
```diff
@@ -193,16 +247,21 @@ export module Components {
...
```
aicioara: Same

src/components/selectionBoxLayer.ts
```diff
@@ -183,9 +225,21 @@ export module Components {
...
+    public xExtent(xExtent?: (number | { valueOf(): number })[]): any {
       // Explicit typing for Typescript 1.4
-      return [this._boxLeftDataValue, this._boxRightDataValue];
+      if (xExtent == null) {
+        return this._xExtent;
+      }
+      this._xExtent = [xExtent[0], xExtent[1]];
+      this._xBoundsMode = PropertyMode.VALUE;
+      this._syncPixelPositionAndValue();
+      this.render();
```
aicioara: Same comment about double calling sync method

src/components/selectionBoxLayer.ts
```diff
@@ -214,6 +273,17 @@ export module Components {
         this.yScale().offUpdate(this._adjustBoundsCallback);
       }
     }
+
+    protected _setBoundsWithoutChangingMode(bounds: Bounds) {
```
aicioara: just calling `_setBounds()` does not change the mode either. The name of this method smells. (actually the existance of three different methods for setting bounds smells)

src/components/xDragBoxLayer.ts
```diff
@@ -17,7 +17,7 @@ export module Components {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.bounds(this.bounds()); // set correct bounds when width/height changes
+      this._setBoundsWithoutChangingMode(this.bounds()); // set correct bounds when width/height changes
```
aicioara: Code like the one that was previously here smells `this.bounds(this.bounds())`. Can we use the appropriate protected / private method for the invalidation of old bounds? 

Current change is just a rewriting of what was previously there (and wrong)

src/components/xDragBoxLayer.ts
```diff
@@ -45,7 +45,9 @@ export module Components {
       throw new Error("yScales cannot be set on an XDragBoxLayer");
     }
 
-    public yExtent(): (number | { valueOf(): number })[] {
+    public yExtent(): (number | { valueOf(): number })[];
+    public yExtent(yExtent: (number | { valueOf(): number })[]): SelectionBoxLayer;
+    public yExtent(yExtent?: (number | { valueOf(): number })[]): any {
       throw new Error("XDragBoxLayer has no yExtent");
```
aicioara: Amend Message. `yExtents do not make sense for xDragBoxLayer`. The idea is that the message does not make sense when you are trying to set the yExtent

src/components/yDragBoxLayer.ts
```diff
@@ -45,7 +45,9 @@ export module Components {
       throw new Error("xScales cannot be set on an YDragBoxLayer");
     }
 
-    public xExtent(): (number | { valueOf(): number })[] {
+    public xExtent(): (number | { valueOf(): number })[];
+    public xExtent(xExtent: (number | { valueOf(): number })[]): SelectionBoxLayer;
+    public xExtent(xExtent?: (number | { valueOf(): number })[]): any {
       throw new Error("YDragBoxLayer has no xExtent");
```
aicioara: Same

src/components/yDragBoxLayer.ts
```diff
@@ -17,7 +17,7 @@ export module Components {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.bounds(this.bounds()); // set correct bounds when width/height changes
+      this._setBoundsWithoutChangingMode(this.bounds()); // set correct bounds when width/height changes
```
aicioara: Same

src/components/xDragBoxLayer.ts
aicioara: Whatever is the fix, can we not have end of the line comments? Instead move them on the line above.

test/components/selectionBoxLayerTests.ts
```diff
@@ -118,18 +118,7 @@ describe("Interactive Components", () => {
...
-        };
-        let bottomRight: Plottable.Point = {
-          x: 250,
-          y: 300
-        };
-        sbl.bounds({
-          topLeft: topLeft,
-          bottomRight: bottomRight
-        });
+        sbl.xExtent([0, 1000]);
```
aicioara: This looks like a change in behavior. Previously we were setting the bounds, right?

test/components/selectionBoxLayerTests.ts
```diff
@@ -184,18 +173,7 @@ describe("Interactive Components", () => {
...
-        };
-        let bottomRight: Plottable.Point = {
-          x: 250,
-          y: 300
-        };
-        sbl.bounds({
-          topLeft: topLeft,
-          bottomRight: bottomRight
-        });
+        sbl.yExtent([0, 1000]);
```
aicioara: Same

aicioara:
I would feel more comfortable if we had some test for this as well. I don't think it is merge-able to develop without.

aicioara:
Lots of the things above are guesses based on what it seems the code is doing, please correct me if I am wrong, did not checkout on the branch to see for myself

bluong:
> TL;DR (AFAIK)
> before
> 
> dragBoxLayer.bounds({topLeft: {x: xScale.scale(0.1), y: yScale.scale(0.1) }, bottomRight: {x: xScale.scale(0.4), y: yScale.scale(0.5) }})
> 
> after
> 
> dragboxLayer.xExtent([0.1, 0.4]).yExtent([0.1, 0.6])
> 
> That is all this PR should be about, right?

Not necessarily.  Similar to the path of `GuideLineLayer`, there are 2 situations that the user may be under:
* The user wants the bounds to be in a static pixel position
* The user wants the bounds to be in a position in accordance with the scale

Based on the above, again similar to `GuideLineLayer`, we need to think about this component being in one of two modes (data-backed or pixel-backed).

src/components/selectionBoxLayer.ts
bluong: I think that since the `PropertyMode` is not exported, `GuideLineLayer` has no access to it.  We'll need to think about it again if we want `PropertyMode` to be public.

src/components/selectionBoxLayer.ts
bluong: The problem is that the two values can easily not be in sync.  Imagine the scenario when the two values are set and a scale does not exist on the component.

src/components/selectionBoxLayer.ts
bluong: I would think [c, d].  The thing that is most important is that [c, d] is the latest set itme.

src/components/selectionBoxLayer.ts
bluong: Then it returns "the data values backing the left and right edges of the box."

The statement is mainly to satisfy the edge case.

src/components/xDragBoxLayer.ts
bluong: Note that I did not write the former code.  I can fix that code, but that is an outside concern.

src/components/xDragBoxLayer.ts
bluong: Note that this message has not been added in this pull request, and thus this is an outside concern.

test/components/selectionBoxLayerTests.ts
bluong: The main difference is that the unit test specifically stated that we're essentially setting the data extent of the box, not the pixel bounds of the box.

src/components/selectionBoxLayer.ts
bluong: This is untrue if the component is not already anchored to an svg.

src/components/xDragBoxLayer.ts
aicioara: Can we get an issue on this?

src/components/selectionBoxLayer.ts
aicioara: fair enough

src/components/selectionBoxLayer.ts
aicioara: But the extent can always be calculated by inverting the box bounds, as long as there is a scale. Keeping things in sync is bad.

src/components/selectionBoxLayer.ts
aicioara: I think the same, but will that happen? Did you test this scenario?

src/components/selectionBoxLayer.ts
aicioara: What I mean is that the code smells. Does the sync method have to be called inside the renderImmediately()?

src/components/selectionBoxLayer.ts
aicioara: Since this is a public method, we need more thorough documentation. What happens in general, how to use this method, what are the edge cases. We just have edge cases above

src/components/selectionBoxLayer.ts
```diff
@@ -193,16 +249,21 @@ export module Components {
      *
      * Returns an undefined array if the edges are not backed by a scale.
```
aicioara: Same

src/components/xDragBoxLayer.ts
```diff
@@ -17,7 +17,7 @@ export module Components {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.bounds(this.bounds()); // set correct bounds when width/height changes
+      this._setBoundsAndExtents(this.bounds()); // set correct bounds when width/height changes
```
aicioara: Separate code and comment on 2 separate lines

src/components/xDragBoxLayer.ts
aicioara: Please create an issue and reference it.

test/components/selectionBoxLayerTests.ts
aicioara: fair enough

aicioara:
Can we test the integration of this feature with the previous environment?

For example 
- What happens when we sett the bounds() and then the x/yExtent, does the latter take over?
- How about the reverse?
- How about removing scales and adding new ones? Will the change be picked up?
- How about updating the scale, will the change be dispatched?

src/components/xDragBoxLayer.ts
bluong: Issue made at #2741

src/components/selectionBoxLayer.ts
```diff
@@ -2,6 +2,7 @@
 
 module Plottable {
 export module Components {
+  export enum PropertyMode { VALUE, PIXEL };
```
aicioara: We are exporting this now?

aicioara: Oh, I see it is needed by subclasses.

src/components/selectionBoxLayer.ts
```diff
@@ -62,10 +56,12 @@ export module Components {
     public bounds(newBounds: Bounds): SelectionBoxLayer;
     public bounds(newBounds?: Bounds): any {
       if (newBounds == null) {
-        return this._boxBounds;
+        return this._getBounds();
```
aicioara: Can we blow up either here or in the _getBounds when the `propertyMode.VALUE` and no scale.

Thoughts on this?

src/components/selectionBoxLayer.ts
```diff
@@ -150,8 +175,9 @@ export module Components {
         this._xScale.offUpdate(this._adjustBoundsCallback);
       }
       this._xScale = xScale;
+      this._xBoundsMode = PropertyMode.VALUE;
```
aicioara: I thought only `.value()` and `.bounds()` change the state

src/components/selectionBoxLayer.ts
```diff
@@ -183,9 +210,33 @@ export module Components {
      *
      * Returns an undefined array if the edges are not backed by a scale.
      */
```
aicioara: Since this is a public method, we need more thorough documentation. What happens in general, how to use this method, what are the edge cases. We just have edge cases above

src/components/selectionBoxLayer.ts
```diff
@@ -183,9 +210,33 @@ export module Components {
...
+      }
+      this._setXExtent(xExtent);
+      this._xBoundsMode = PropertyMode.VALUE;
+      this.render();
+      return this;
+    }
+
+    private _getXExtent(): (number | { valueOf(): number })[] {
+      return this._xBoundsMode === PropertyMode.VALUE ?
+               this._xExtent :
```
aicioara: I don't really like this sort of indentation because it can easily get off sync when global search replacing for var renames

src/components/selectionBoxLayer.ts
```diff
@@ -10,29 +11,22 @@ export module Components {
       topLeft: { x: 0, y: 0 },
       bottomRight: { x: 0, y: 0 }
     };
-    private _boxLeftDataValue: number | { valueOf(): number };
-    private _boxRightDataValue: number | { valueOf(): number };
-    private _boxTopDataValue: number | { valueOf(): number };
-    private _boxBottomDataValue: number | { valueOf(): number };
+    private _xExtent: [number | { valueOf(): number }, number | { valueOf(): number }];
```
aicioara: I suggest to type this as 
```typescript
_xExtent: number | { valueOf(): number }[]
```

I agree the current version is better and safer, but it is most definitely going to cause problems on the long run because of incompatibilities. 

For example a linearScale, `_defaultExtent()` returns `number[]` instead of `[number, number]`.

Another example is the hack below
```typescript
protected _setXExtent(xExtent: (number | { valueOf(): number })[]) {
  this._xExtent = [xExtent[0], xExtent[1]];
}
```

src/components/selectionBoxLayer.ts
```diff
@@ -193,16 +244,33 @@ export module Components {
...
+      this._yBoundsMode = PropertyMode.VALUE;
+      this.render();
+      return this;
+    }
+
+    private _getYExtent(): (number | { valueOf(): number })[] {
+      return this._yBoundsMode === PropertyMode.VALUE ?
+               this._yExtent :
+               (this._yScale == null ?
+                 [undefined, undefined] :
```
aicioara: Again, don't we want to blow up here?

src/components/xDragBoxLayer.ts
aicioara: cool

aicioara:
I like this version so much better.

aicioara:
Can we test the integration of this feature with the previous environment?

For example

What happens when we sett the bounds() and then the x/yExtent, does the latter take over?
How about the reverse?
How about removing scales and adding new ones? Will the change be picked up?
How about updating the scale, will the change be dispatched?

src/components/selectionBoxLayer.ts
bluong: As experienced with `XDragBoxLayer` / `YDragBoxLayer`, it is incredibly disruptive to have a getter error unless there is a good exception architecture surrounding the language.  In which case, I would much rather return something similar to `null` than error.

src/components/selectionBoxLayer.ts
bluong: I needed to put this here to satisfy the case where a user only attaches a scale to a `SelectionBoxLayer`

src/components/selectionBoxLayer.ts
bluong: What type of indentation would you prefer?

src/components/selectionBoxLayer.ts
bluong: Same as above.

src/components/selectionBoxLayer.ts
aicioara: 2 spaces compared to the previous line

src/components/selectionBoxLayer.ts
aicioara: You mean they do not specify any x/yExtent?

src/components/selectionBoxLayer.ts
aicioara: Okay

src/components/selectionBoxLayer.ts
bluong: yes.  this is specifically important for drag boxes

src/components/selectionBoxLayer.ts
bluong: The first line tells what happens in the general case

src/components/selectionBoxLayer.ts
```diff
@@ -10,29 +11,22 @@ export module Components {
...
-          },
-          bottomRight: {
-            x: this._xScale ? this._xScale.scale(this._boxRightDataValue) : this._boxBounds.bottomRight.x,
-            y: this._yScale ? this._yScale.scale(this._boxBottomDataValue) : this._boxBounds.bottomRight.y
-          }
-        });
+        this.render();
       };
+      this._xExtent = [undefined, undefined];
+      this._yExtent = [undefined, undefined];
```
aicioara: Now that this is no longer required, let's initialize it in the declaration as `[]` (instead of `[undefined, undefined]`)

src/components/selectionBoxLayer.ts
```diff
@@ -83,15 +79,44 @@ export module Components {
...
         bottomRight: bottomRight
       };
-      this._bindBoxDataValues();
+    }
+
+    private _getBounds(): Bounds {
+      return {
+        topLeft: {
+          x: this._xBoundsMode === PropertyMode.PIXEL ?
+               this._boxBounds.topLeft.x :
```
aicioara: can you fix the indentation by deleting exactly 3 spaces

src/components/selectionBoxLayer.ts
aicioara: My bad. Github did not show those lines

src/components/selectionBoxLayer.ts
```diff
@@ -183,9 +210,33 @@ export module Components {
...
+      return this._xBoundsMode === PropertyMode.VALUE ?
+        this._xExtent :
+        (this._xScale == null ?
+          [undefined, undefined] :
+          [this._xScale.invert(this._boxBounds.topLeft.x),
+          this._xScale.invert(this._boxBounds.bottomRight.x)]);
+    }
+
+    protected _setXExtent(xExtent: (number | { valueOf(): number })[]) {
+      this._xExtent = [xExtent[0], xExtent[1]];
```
aicioara: `this._xExtent = xExtent`

src/components/selectionBoxLayer.ts
```diff
@@ -193,16 +244,33 @@ export module Components {
...
+          this._yScale.invert(this._boxBounds.bottomRight.y)]);
     }
 
-    private _bindBoxDataValues() {
-      this._boxLeftDataValue = this._xScale ? this._xScale.invert(this._boxBounds.topLeft.x) : null;
-      this._boxTopDataValue = this._yScale ? this._yScale.invert(this._boxBounds.topLeft.y) : null;
-      this._boxRightDataValue = this._xScale ? this._xScale.invert(this._boxBounds.bottomRight.x) : null;
-      this._boxBottomDataValue = this._yScale ? this._yScale.invert(this._boxBounds.bottomRight.y) : null;
+    protected _setYExtent(yExtent: (number | { valueOf(): number })[]) {
+      this._yExtent = [yExtent[0], yExtent[1]];
```
aicioara: `this._yExtent = yExtent`

Also, why is this method needed?

src/components/xDragBoxLayer.ts
```diff
@@ -17,7 +17,7 @@ export module Components {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.bounds(this.bounds()); // set correct bounds when width/height changes
+      this._setBounds(this.bounds()); // set correct bounds when width/height changes
```
aicioara: Can we have the comment on a separate line to the code? Please? :D

src/components/yDragBoxLayer.ts
```diff
@@ -17,7 +17,7 @@ export module Components {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this.bounds(this.bounds()); // set correct bounds when width/height changes
+      this._setBounds(this.bounds()); // set correct bounds when width/height changes
```
aicioara: Can we have the comment on a separate line to the code? Please? :D
 Add a line note

test/components/selectionBoxLayerTests.ts
```diff
@@ -144,30 +162,101 @@ describe("Interactive Components", () => {
...
+      it("can set the data values of the left and right sides directly", () => {
         let xScale = new Plottable.Scales.Linear();
         xScale.domain([0, 2000]);
         xScale.range([0, svgWidth]);
 
-        assert.deepEqual(sbl.xExtent(), [undefined, undefined], "xExtent is not set unless an yScale is set");
+        sbl.boxVisible(true);
         sbl.xScale(xScale);
+
+        assert.deepEqual(sbl.xExtent(), [undefined, undefined], "xExtent is not set unless an xScale is set");
```
aicioara: but the xScale is set on the line above...

test/components/selectionBoxLayerTests.ts
```diff
@@ -144,30 +162,101 @@ describe("Interactive Components", () => {
...
+            x: 200,
+            y: 200
+          }
         };
-        sbl.bounds({
-          topLeft: topLeft,
-          bottomRight: bottomRight
-        });
+        sbl.bounds(bounds);
+        assert.deepEqual(sbl.bounds(), bounds, "latest bounds set");
```
aicioara: there is no verb in this sentence

test/components/selectionBoxLayerTests.ts
```diff
@@ -144,30 +162,101 @@ describe("Interactive Components", () => {
...
+        xScale.domain([0, 2000]);
+        xScale.range([0, svgHeight]);
+        sbl.xScale(xScale);
+
+        let xExtent = [100, 250];
+        sbl.xExtent(xExtent);
+        assert.strictEqual(sbl.bounds().topLeft.x, xScale.scale(xExtent[0]), "left pixel position adjusts accordingly");
+        assert.strictEqual(sbl.bounds().bottomRight.x, xScale.scale(xExtent[1]), "right pixel position adjusts accordingly");
+
+        let xScale2 = new Plottable.Scales.Linear();
```
aicioara: Let's use a ModifedLog scale for coverage

test/components/selectionBoxLayerTests.ts
```diff
@@ -144,30 +162,101 @@ describe("Interactive Components", () => {
...
+        sbl.xScale(xScale);
+
+        let xExtent = [100, 250];
+        sbl.xExtent(xExtent);
+        assert.strictEqual(sbl.bounds().topLeft.x, xScale.scale(xExtent[0]), "left pixel position adjusts accordingly");
+        assert.strictEqual(sbl.bounds().bottomRight.x, xScale.scale(xExtent[1]), "right pixel position adjusts accordingly");
+
+        xScale.domain([0, 1000]);
+
+        assert.strictEqual(sbl.bounds().topLeft.x, xScale.scale(xExtent[0]), "left pixel position adjusts accordingly");
```
aicioara: I don't really like these blind tests. Can you also assert that the xExtent[0] now is different than xExtent[0] before the scale update?

test/components/selectionBoxLayerTests.ts
```diff
@@ -144,30 +162,101 @@ describe("Interactive Components", () => {
...
+
+      it("updates left and right edge pixel positions if in VALUE mode and scale is switched", () => {
+        let xScale = new Plottable.Scales.Linear();
+        xScale.domain([0, 2000]);
+        xScale.range([0, svgHeight]);
+        sbl.xScale(xScale);
+
+        let xExtent = [100, 250];
+        sbl.xExtent(xExtent);
+        assert.strictEqual(sbl.bounds().topLeft.x, xScale.scale(xExtent[0]), "left pixel position adjusts accordingly");
```
aicioara: I don't really like these blind tests. Can you also assert that the xExtent[0] now is different than xExtent[0] before the scale changes?

test/components/selectionBoxLayerTests.ts
```diff
@@ -210,31 +288,101 @@ describe("Interactive Components", () => {
...
+            x: 200,
+            y: 200
+          }
         };
-        sbl.bounds({
-          topLeft: topLeft,
-          bottomRight: bottomRight
-        });
+        sbl.bounds(bounds);
+        assert.deepEqual(sbl.bounds(), bounds, "latest bounds set");
```
aicioara: same

test/components/selectionBoxLayerTests.ts
```diff
@@ -210,31 +288,101 @@ describe("Interactive Components", () => {
...
+        yScale.domain([0, 2000]);
+        yScale.range([0, svgHeight]);
+        sbl.yScale(yScale);
+
+        let yExtent = [100, 250];
+        sbl.yExtent(yExtent);
+        assert.strictEqual(sbl.bounds().topLeft.y, yScale.scale(yExtent[0]), "top pixel position adjusts accordingly");
+        assert.strictEqual(sbl.bounds().bottomRight.y, yScale.scale(yExtent[1]), "bottom pixel position adjusts accordingly");
+
+        let yScale2 = new Plottable.Scales.Linear();
```
aicioara: same

test/components/selectionBoxLayerTests.ts
```diff
@@ -210,31 +288,101 @@ describe("Interactive Components", () => {
...
+        yScale2.domain([0, 1000]);
+        yScale2.range([0, svgHeight]);
+
+        sbl.yScale(yScale2);
```
aicioara: same

test/components/selectionBoxLayerTests.ts
```diff
@@ -210,31 +288,101 @@ describe("Interactive Components", () => {
...
+
+      it("updates top and bottom edge pixel positions if in VALUE mode and scale updates", () => {
+        let yScale = new Plottable.Scales.Linear();
+        yScale.domain([0, 2000]);
+        yScale.range([0, svgHeight]);
+        sbl.yScale(yScale);
+
+        let yExtent = [100, 250];
+        sbl.yExtent(yExtent);
+        assert.strictEqual(sbl.bounds().topLeft.y, yScale.scale(yExtent[0]), "top pixel position adjusts accordingly");
```
aicioara: same

test/components/selectionBoxLayerTests.ts
```diff
@@ -210,31 +288,101 @@ describe("Interactive Components", () => {
...
+        assert.strictEqual(sbl.bounds().bottomRight.y, yScale.scale(yExtent[1]), "bottom pixel position adjusts accordingly");
+
+        let yScale2 = new Plottable.Scales.Linear();
+        yScale2.domain([0, 1000]);
+        yScale2.range([0, svgHeight]);
+
+        sbl.yScale(yScale2);
+        assert.strictEqual(sbl.bounds().topLeft.y, yScale2.scale(yExtent[0]), "top pixel position adjusts accordingly");
```
aicioara: same

src/components/selectionBoxLayer.ts
bluong: Children classes will want to set the extent without affecting the mode

src/components/selectionBoxLayer.ts
bluong: Previous behavior had the extent return an undefined array, though.  This would introduce a change in behavior

src/components/selectionBoxLayer.ts
bluong: Children classes will want to set the extent without affecting the mode

crmorford:
testing fiddlez
[x] selectionBoxLayer http://jsfiddle.net/ewga1des/5/
[x] dragBoxLayer http://jsfiddle.net/ewga1des/4/
[x] xDragBox http://jsfiddle.net/ewga1des/2/ yes! good errors for calling yExtent on xDragBox
[x] yDragBox http://jsfiddle.net/ewga1des/3/ not good errors :( can I get the xDragBox error here too?
[ ] timeScales
[ ] linear scales
[ ] log scales
[ ] category scales
[ ] destroy
[ ] detach
[ ] nonsense inputs
[ ] layer 2 dragBoxLayers

crmorford:
develop: http://jsfiddle.net/ewga1des/15/
this branch: http://jsfiddle.net/ewga1des/16/

I can't drag a yDragBoxLayer out anymore without getting errors

bluong:
For the `xDragBox` / `yDragBox` case, the `xDragBox` case runs into an error on the `yExtent` setter which occurs after the `xExtent` setter.  Thus, the error is met, the `xExtent` is set, and then a render with the correct information occurs.

However for the `yDragBox` case, the error occurs on the `xExtent` setter which occurs before the `yExtent` setter.  This means that the error is met, the `yExtent` is not set, and then a render occurs but since the `yExtent` is not set, errors pop up.

bluong:
For the `nonsense` inputs, the error messages essentially come from D3 I believe.  There can be some error checking at the render step if that is preferable.  Note that plots have a similar problem to this, but we have no current answer to this.

crmorford:
Is it possible to get errors for those cases? I want to tend towards having more errors when possible.

Both cases are gonna be hard to figure out from the current errors, if someone messes up.

bluong:
I'm looking into it how we can get the error messages but interaction with the scale makes things a bit difficult to check for the right type.  I can hack a little thing but it is a hack at best.  I would need to work with the scale a bit to get validation to actually work correctly.

crmorford:
the selection box isn't getting hidden by the clip path (pan zoom on the plot)
http://jsfiddle.net/ewga1des/27/
screen shot 2015-09-15 at 10 39 05 am

bluong:
@crmorford I believe this is outside scope as this is an existing bug on SelectionBoxLayer as issued at #2769

crmorford:
There must be a way to create a better error for calling xExtent on yDragBox.
This is too confusing, and the asymmetry with the opposite case bothers me http://jsfiddle.net/ewga1des/3/

crmorford:
For these 2 cases, the errors are also unhelpful.  
set just yExtent on dragbox http://jsfiddle.net/ewga1des/17/ needs clearer error
set just xExtent on dragbox http://jsfiddle.net/ewga1des/18/ needs clearer error

can we make sure both extents are set before trying to draw the dragBox at least?

bluong:
> There must be a way to create a better error for calling xExtent on yDragBox.
> This is too confusing, and the asymmetry with the opposite case bothers me http://jsfiddle.net/ewga1des/3/

The asymmetry occurs because once an error happens in javascript, all of the code following the error does not get executed.  Thus, if you just switch the type of `DragBoxLayer` in the call, that's not perfectly symmetric, since `YDragBoxLayer` will error on the `xExtent` call and never call `yExtent` whereas `XDragBoxLayer` will call `xExtent` before erroring on `yExtent`.

crmorford:
> The asymmetry occurs because once an error happens in javascript, all of the code following the error does not get executed.  Thus, if you just switch the type of `DragBoxLayer` in the call, that's not perfectly symmetric, since `YDragBoxLayer` will error on the `xExtent` call and never call `yExtent` whereas `XDragBoxLayer` will call `xExtent` before erroring on `yExtent`.

I understand this. I think we should check earlier to make sure that neiYDragBoxLayer doesn't call xExtent if it will cause gross errors.

aicioara:
Rest looks good

PR #2681, opened 2015-08-26
Refactor modified log scale tests - Pass 1

aicioara:
Part of #2640 

As part of this: 
- No nested `beforeEach()`es
- Cleaned the console
screen shot 2015-08-26 at 2 55 02 pm

### DIFF
--- a/test/scales/modifiedLogScaleTests.ts
+++ b/test/scales/modifiedLogScaleTests.ts
@@ -231,11 +231,11 @@ describe("Scales", () => {
         assert.operator(betweenPivots.length, ">", 0, "should be ticks between -base and base");
       });
 
-      it("ticks() always has more than 2 ticks", () => {
+      it("always has more than 2 ticks", () => {
         [null, [2, 9], [0, 1], [1, 2], [0.001, 0.01], [-0.1, 0.1], [-3, -2]].forEach((domain) => {
           scale.domain(domain);
           let ticks = scale.ticks();
-          assert.operator(ticks.length, ">", 2);
+          assert.operator(ticks.length, ">", 2, "there should be at least 2 ticks in domain " + domain);
         });
       });
     });
--- a/test/scales/modifiedLogScaleTests.ts
+++ b/test/scales/modifiedLogScaleTests.ts
@@ -6,6 +6,7 @@ describe("Scales", () => {
     describe("Basic Usage", () => {
       let scale: Plottable.Scales.ModifiedLog;
       let base = 10;
+      let delta = 0.00001;
       let epsilon = 0.00001;
 
       beforeEach(() => {
@@ -15,18 +16,18 @@ describe("Scales", () => {
       it("is an increasing, continuous function that can go negative", () => {
         d3.range(-base * 2, base * 2, base / 20).forEach((x: number) => {
           // increasing
-          assert.operator(scale.scale(x - epsilon), "<", scale.scale(x));
-          assert.operator(scale.scale(x), "<", scale.scale(x + epsilon));
+          assert.operator(scale.scale(x - delta), "<", scale.scale(x));
+          assert.operator(scale.scale(x), "<", scale.scale(x + delta));
           // continuous
-          assert.closeTo(scale.scale(x - epsilon), scale.scale(x), epsilon);
-          assert.closeTo(scale.scale(x), scale.scale(x + epsilon), epsilon);
+          assert.closeTo(scale.scale(x - delta), scale.scale(x), epsilon);
+          assert.closeTo(scale.scale(x), scale.scale(x + delta), epsilon);
         });
         assert.closeTo(scale.scale(0), 0, epsilon);
       });
 
       it("has log() behavior at values > base", () => {
-        [10, 100, 23103.4, 5].forEach((x) => {
-          assert.closeTo(scale.scale(x), Math.log(x) / Math.log(10), 0.1);
+        [10, 100, 23103.4, 1e+45].forEach((x) => {
+          assert.closeTo(scale.scale(x), Math.log(x) / Math.log(10), epsilon);
         });
       });
 
@@ -42,7 +43,7 @@ describe("Scales", () => {
       });
 
       it("can be padded", () => {
-        scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => [0, base]);
+        scale.addIncludedValuesProvider(() => [0, base]);
         scale.padProportion(0);
         let unpaddedDomain = scale.domain();
         scale.padProportion(0.1);
@@ -63,22 +64,23 @@ describe("Scales", () => {
 
     });
 
-    describe("Other bases", () => {
+    describe("Scale bases", () => {
       it("uses 10 as the default base", () => {
         let scale = new Plottable.Scales.ModifiedLog();
-        assert.strictEqual((<any>scale)._base, 10, "10 is the default base");
+        scale.range([0, 1]);
+        assert.strictEqual(scale.scale(10), 1, "10 is base");
+        assert.strictEqual(scale.scale(100), 2, "10^2 will result in a double value compared to the base");
       });
 
-      it("works with base 2", () => {
+      it("can scale values using base 2", () => {
         let scale = new Plottable.Scales.ModifiedLog(2);
-        assert.strictEqual((<any>scale)._base, 2, "base 2 is selected");
         scale.domain([0, 16]);
         scale.range([0, 1]);
 
         assert.strictEqual(scale.scale(-2), -0.25, "scales negative values");
         assert.strictEqual(scale.scale(0), 0, "scales 0");
         assert.strictEqual(scale.scale(2), 0.25, "scales base");
-        assert.strictEqual(scale.scale(4), 0.5, "scales random values");
+        assert.strictEqual(scale.scale(4), 0.5, "scales other values");
         assert.strictEqual(scale.scale(16), 1, "scales maximum value");
         assert.strictEqual(scale.scale(256), 2, "scales values outside the domain");
 
@@ -98,7 +100,7 @@ describe("Scales", () => {
 
       it("expands single value domains to [value / base, value * base].sort()", () => {
         let singleValue = 15;
-        scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => [singleValue]);
+        scale.addIncludedValuesProvider(() => [singleValue]);
         assert.deepEqual(scale.domain(), [singleValue / base, singleValue * base],
           "positive single-value extent was expanded to [value / base, value * base]");
         singleValue = -15;
@@ -113,7 +115,7 @@ describe("Scales", () => {
 
       it("can force the minimum of the domain with domainMin()", () => {
         let requestedDomain = [-5, 5];
-        scale.addIncludedValuesProvider((scale) => requestedDomain);
+        scale.addIncludedValuesProvider(() => requestedDomain);
 
         let minBelowBottom = -10;
         assert.strictEqual(scale.domainMin(minBelowBottom), scale, "the scale is returned by the setter");
@@ -136,13 +138,13 @@ describe("Scales", () => {
 
         scale.domainMin(minInMiddle);
         let requestedDomain2 = [-10, 10];
-        scale.addIncludedValuesProvider((scale) => requestedDomain2);
+        scale.addIncludedValuesProvider(() => requestedDomain2);
         assert.deepEqual(scale.domain(), [minInMiddle, requestedDomain2[1]], "adding another ExtentsProvider doesn't change domainMin()");
       });
 
       it("can force the maximum of the domain with domainMax()", () => {
         let requestedDomain = [-5, 5];
-        scale.addIncludedValuesProvider((scale) => requestedDomain);
+        scale.addIncludedValuesProvider(() => requestedDomain);
 
         let maxAboveTop = 10;
         assert.strictEqual(scale.domainMax(maxAboveTop), scale, "the scale is returned by the setter");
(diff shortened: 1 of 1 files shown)


test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-        // continuous
-        assert.closeTo(scale.scale(x - epsilon), scale.scale(x), epsilon);
-        assert.closeTo(scale.scale(x), scale.scale(x + epsilon), epsilon);
-      });
-      assert.closeTo(scale.scale(0), 0, epsilon);
-    });
+    describe("Basic Usage", () => {
+      let scale: Plottable.Scales.ModifiedLog;
+      let base = 10;
+      let epsilon = 0.00001;
```
bluong: Hmm, while the old tests do this, currently the `epsilon` variable holds 2 semantics. It is, for one, the delta for choosing close data values, and it is also the error to use for close to.  Should we split this up into two variables?

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      scale.padProportion(0);
-      let unpaddedDomain = scale.domain();
-      scale.padProportion(0.1);
-      assert.operator(scale.domain()[0], "<", unpaddedDomain[0], "left side of domain has been padded");
-      assert.operator(unpaddedDomain[1], "<", scale.domain()[1], "right side of domain has been padded");
-    });
+      it("has log() behavior at values > base", () => {
+        [10, 100, 23103.4, 5].forEach((x) => {
+          assert.closeTo(scale.scale(x), Math.log(x) / Math.log(10), 0.1);
+        });
```
bluong: Isn't the base 10 by default? It's a bit weird that 5 works here

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      scale.autoDomain();
-      assert.deepEqual(scale.domain(), [-base, base],
-        "zero single-value extent was expanded to [base, -base]");
-    });
+      it("ensures x = invert(scale(x))", () => {
+        [0, 1, base, 100, 0.001, -1, -0.3, -base, base - 0.001].forEach((x) => {
+          assert.closeTo(x, scale.invert(scale.scale(x)), epsilon);
+          assert.closeTo(x, scale.scale(scale.invert(x)), epsilon);
+        });
+      });
```
bluong: I see the usefulness of this test, but why should it be here in the subclass?

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      scale.domainMin(minBelowBottom);
-      assert.deepEqual(scale.domain(), [minBelowBottom, requestedDomain[1]], "lower end of domain was set by domainMin()");
+      it("can be padded", () => {
+        scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => [0, base]);
+        scale.padProportion(0);
+        let unpaddedDomain = scale.domain();
+        scale.padProportion(0.1);
+        assert.operator(scale.domain()[0], "<", unpaddedDomain[0], "left side of domain has been padded");
+        assert.operator(unpaddedDomain[1], "<", scale.domain()[1], "right side of domain has been padded");
+      });
```
bluong: Curious on why this needs to be tested since padding works on the QuantitativeScale level

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+      it("works on inverted domain", () => {
+        scale.domain([200, -100]);
+        scale.range([10, 20]);
+        let range = scale.range();
+        assert.strictEqual(scale.scale(-100), range[1], "minimum value in domain maps to maximum value in range");
+        assert.strictEqual(scale.scale(200), range[0], "maximum value in domain maps to minimum value in range");
+        let a = [-100, -10, -3, 0, 1, 3.64, 50, 60, 200];
+        let b = a.map((x) => scale.scale(x));
+        assert.deepEqual(b, b.slice().sort().reverse(), "should be decreasing function; reverse is sorted");
+      });
```
bluong: Same as above.

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      scale.autoDomain();
-      assert.deepEqual(scale.domain(), requestedDomain, "calling autoDomain() overrides domainMin()");
+    });
 
-      let minEqualTop = scale.domain()[1];
-      scale.domainMin(minEqualTop);
-      assert.deepEqual(scale.domain(), [minEqualTop, minEqualTop * base],
-        "domain is set to [min, min * base] if the requested value is >= autoDomain()-ed max value");
+    describe("Other bases", () => {
+      it("uses 10 as the default base", () => {
```
bluong: Description of "Other bases" is quite counter to the test of "uses 10 as the default base"

Maybe change the describe to just be about bases

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        let scale = new Plottable.Scales.ModifiedLog();
+        assert.strictEqual((<any>scale)._base, 10, "10 is the default base");
```
bluong: While is technically correct, is there a more public way to prove that 10 is the default base?  Maybe scale a few values to extrapolate the fact.

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+      it("works with base 2", () => {
+        let scale = new Plottable.Scales.ModifiedLog(2);
+        assert.strictEqual((<any>scale)._base, 2, "base 2 is selected");
+        scale.domain([0, 16]);
+        scale.range([0, 1]);
+
+        assert.strictEqual(scale.scale(-2), -0.25, "scales negative values");
+        assert.strictEqual(scale.scale(0), 0, "scales 0");
+        assert.strictEqual(scale.scale(2), 0.25, "scales base");
+        assert.strictEqual(scale.scale(4), 0.5, "scales random values");
```
bluong: This value is not random.  The description should not state as such

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      assert.deepEqual(scale.domain(), [minInMiddle, requestedDomain2[1]], "adding another ExtentsProvider doesn't change domainMin()");
+      it("works with base 2", () => {
+        let scale = new Plottable.Scales.ModifiedLog(2);
+        assert.strictEqual((<any>scale)._base, 2, "base 2 is selected");
+        scale.domain([0, 16]);
+        scale.range([0, 1]);
+
+        assert.strictEqual(scale.scale(-2), -0.25, "scales negative values");
+        assert.strictEqual(scale.scale(0), 0, "scales 0");
+        assert.strictEqual(scale.scale(2), 0.25, "scales base");
```
bluong: Why is scaling the base value 0.25

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+      it("uses 10 as the default base", () => {
+        let scale = new Plottable.Scales.ModifiedLog();
+        assert.strictEqual((<any>scale)._base, 10, "10 is the default base");
+      });
 
-      scale.domainMin(minInMiddle);
-      let requestedDomain2 = [-10, 10];
-      scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => requestedDomain2);
-      assert.deepEqual(scale.domain(), [minInMiddle, requestedDomain2[1]], "adding another ExtentsProvider doesn't change domainMin()");
+      it("works with base 2", () => {
```
bluong: "works with base 2" is quite a bit vague

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        scale = new Plottable.Scales.ModifiedLog(base);
+        scale.padProportion(0);
+      });
 
-      let maxInMiddle = 0;
-      scale.domainMax(maxInMiddle);
-      assert.deepEqual(scale.domain(), [requestedDomain[0], maxInMiddle], "upper end was set even if requested value cuts off some data");
+      it("expands single value domains to [value / base, value * base].sort()", () => {
+        let singleValue = 15;
+        scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => [singleValue]);
```
bluong: I personally would prefer to not show the parameters here, but no big deal.

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
-      assert.deepEqual(scale.domain(), [requestedDomain[0], maxAboveTop], "upper end of domain was set by domainMax()");
+      beforeEach(() => {
+        scale = new Plottable.Scales.ModifiedLog(base);
+        scale.padProportion(0);
+      });
 
-      let maxInMiddle = 0;
-      scale.domainMax(maxInMiddle);
-      assert.deepEqual(scale.domain(), [requestedDomain[0], maxInMiddle], "upper end was set even if requested value cuts off some data");
+      it("expands single value domains to [value / base, value * base].sort()", () => {
```
bluong: Can we have a less codified way to word this?  Why do we want to expand the single value to this?

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        let singleValue = 15;
+        scale.addIncludedValuesProvider((scale: Plottable.Scales.ModifiedLog) => [singleValue]);
+        assert.deepEqual(scale.domain(), [singleValue / base, singleValue * base],
+          "positive single-value extent was expanded to [value / base, value * base]");
```
bluong: Same as above

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        singleValue = -15;
+        scale.autoDomain();
+        assert.deepEqual(scale.domain(), [singleValue * base, singleValue / base],
+          "negative single-value extent was expanded to [value * base, value / base]");
```
bluong: Same as above.

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        let minEqualTop = scale.domain()[1];
+        scale.domainMin(minEqualTop);
+        assert.deepEqual(scale.domain(), [minEqualTop, minEqualTop * base],
+          "domain is set to [min, min * base] if the requested value is >= autoDomain()-ed max value");
+
+        scale.domainMin(minInMiddle);
+        let requestedDomain2 = [-10, 10];
+        scale.addIncludedValuesProvider((scale) => requestedDomain2);
+        assert.deepEqual(scale.domain(), [minInMiddle, requestedDomain2[1]], "adding another ExtentsProvider doesn't change domainMin()");
+      });
```
bluong: Is this not technically tested on the QuantativeScale level?

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        let maxEqualBottom = scale.domain()[0];
+        scale.domainMax(maxEqualBottom);
+        assert.deepEqual(scale.domain(), [maxEqualBottom * base, maxEqualBottom],
+          "domain is set to [max * base, max] if the requested value is <= autoDomain()-ed min value and negative");
+
+        scale.domainMax(maxInMiddle);
+        let requestedDomain2 = [-10, 10];
+        scale.addIncludedValuesProvider((scale) => requestedDomain2);
+        assert.deepEqual(scale.domain(), [requestedDomain2[0], maxInMiddle], "adding another ExtentsProvider doesn't change domainMax()");
+      });
```
bluong: Is this not technically tested on the QuantativeScale level?

test/scales/modifiedLogScaleTests.ts
```diff
@@ -2,187 +2,241 @@
...
+        scale.domainMax(desiredMax);
+        assert.deepEqual(scale.domain(), [desiredMin, desiredMax], "setting domainMin() and domainMax() sets the domain");
+
+        scale.autoDomain();
+        let bigMin = 10;
+        let smallMax = -10;
+        scale.domainMin(bigMin);
+        scale.domainMax(smallMax);
+        assert.deepEqual(scale.domain(), [bigMin, smallMax], "setting both is allowed even if it reverse the domain");
+      });
```
bluong: Is this not technically tested on the QuantativeScale level?

bluong:
A good many of questions mainly regard to usefulness of the tests as well as more precise test semantics

test/scales/modifiedLogScaleTests.ts
aicioara: Nice catch, that was a bug because we were not using `epsilon`

test/scales/modifiedLogScaleTests.ts
aicioara: Why not? We need to test that the `modifedLogScale` satisfies the contract of a `quantitativeScale`. If we have a mistake in either our `scale()` or `invert()` method, then this test will catch that.

test/scales/modifiedLogScaleTests.ts
aicioara: Not sure how I can test this on the quantitative Scale. That is an abstract scale. Will see when I get there and get to pass 2.

test/scales/modifiedLogScaleTests.ts
aicioara: That is a function of the scale itself. Tests that the scale does not blow up when the domain is reversed

test/scales/modifiedLogScaleTests.ts
aicioara: Because of the way I chose the domain.

test/scales/modifiedLogScaleTests.ts
aicioara: But that is what I am testing

test/scales/modifiedLogScaleTests.ts
aicioara: Because they represent a unit in the range. How do I say that? :D

test/scales/modifiedLogScaleTests.ts
aicioara: I don't know how to phrase it

test/scales/modifiedLogScaleTests.ts
aicioara: .

test/scales/modifiedLogScaleTests.ts
aicioara: That is probably an excellent idea, as I've seen this code so many times already. But I started bottom-up, and no idea what happens on the QuantitativeScale level. Will make a note of it

test/scales/modifiedLogScaleTests.ts
aicioara: same

test/scales/modifiedLogScaleTests.ts
aicioara: re

aicioara:
That is for Pass 2

PR #2686, opened 2015-08-27
add label() and labelsEnable() for Plots.Wheel

### LINKED ISSUE #2683: enable labels in wheel plot

ztsai:
closes #2683
demo: http://jsfiddle.net/ztsai/use8b6ek/

`labelsEnabled(enabled?: boolean)`: gets/sets whether labels are enabled.
`label(label?: Accessor<string>)`: gets/sets the text of labels to the result of an Accessor.

It does not show labels that are too big to be contained in the sector or cut off by edges
However, it will show labels that overlap with other sectors (since the computation is too complicated and it seems like a lower priority, I decided to leave this as a future enhancement. Will open an issue for it after this is closed)

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -54,10 +54,7 @@ declare module Plottable {
              * @return {number} dist(p1, p2)^2
              */
             function distanceSquared(p1: Point, p2: Point): number;
-            /**
-             * Converts degree to radian
-             */
-            function degreeToRadian(degree: number): number;
+            function degreesToRadians(degree: number): number;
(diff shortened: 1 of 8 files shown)


jtlan:
While this is neat, I think it might be more helpful to get R and Theta axes first. @bluong once tried to implement this as "radial axis group"; I'm not sure if it makes sense to implement it as a radial axis, or just a feature of `Plots.Wheel`.

ztsai:
close PR till further decision is made

PR #2690, opened 2015-08-28
Refactor tick generator tests - Pass 1

aicioara:
Part of #2640 

As part of this: 
- No nested `beforeEach()`es
- Cleaned the console
screen shot 2015-08-27 at 5 19 57 pm

### DIFF
--- a/test/scales/tickGeneratorsTests.ts
+++ b/test/scales/tickGeneratorsTests.ts
@@ -18,7 +18,7 @@ describe("Scales", () => {
         ticks.forEach((tick) => {
           assert.operator(start, "<=", tick, "tick " + tick + " should be greater than the lower bound of the domain");
           assert.operator(tick, "<=", end, "tick " + tick + " should be less than the upper bound of the domain");
-        })
+        });
       });
 
       it("generates ticks for a domain crossing 0", () => {
--- a/test/scales/tickGeneratorsTests.ts
+++ b/test/scales/tickGeneratorsTests.ts
@@ -13,7 +13,12 @@ describe("Scales", () => {
         let start = 0.5, end = 4.01, interval = 1;
         scale.domain([start, end]);
         let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
-        assert.deepEqual(ticks, [0.5, 1, 2, 3, 4, 4.01], "generated ticks contains all possible ticks within range");
+
+        assert.strictEqual(ticks.length, 6, "ticks are generated");
+        ticks.forEach((tick) => {
+          assert.operator(start, "<=", tick, "tick " + tick + " should be greater than the lower bound of the domain");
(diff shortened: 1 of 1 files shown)


test/scales/tickGeneratorsTests.ts
```diff
@@ -1,72 +1,79 @@
...
-      let scale = new Plottable.Scales.Linear();
-      scale.domain([start, end]);
-      let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
-      assert.deepEqual(ticks, [-7.6, -7.5, -5, -2.5, -2.2], "generated all ticks between lower and higher value");
-    });
+      it("generates ticks within domain", () => {
+        let start = 0.5, end = 4.01, interval = 1;
+        scale.domain([start, end]);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
+        assert.deepEqual(ticks, [0.5, 1, 2, 3, 4, 4.01], "generated ticks contains all possible ticks within range");
```
jtlan: Given the name of the test, wouldn't it make more sense to assert that each value is inside the domain?

Alternatively, we could change the message.

test/scales/tickGeneratorsTests.ts
```diff
@@ -1,72 +1,79 @@
...
+        scale.domain([0, 4]);
+        assert.deepEqual(integerTickGenerator(scale), [0, 1, 2, 3, 4], "only the integers are returned");
+      });
+
+      it("works across negative numbers", () => {
+        scale.domain([-2, 1]);
+        assert.deepEqual(integerTickGenerator(scale), [-2, -1, 0, 1], "only the integers are returned");
+      });
+
+      it("includes endticks", () => {
```
jtlan: "includes end ticks"

test/scales/tickGeneratorsTests.ts
```diff
@@ -1,72 +1,84 @@
...
-      let start = 0.5, end = 10.01, interval = 11;
-      let scale = new Plottable.Scales.Linear();
-      scale.domain([start, end]);
-      let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
-      assert.deepEqual(ticks, [0.5, 10.01], "no middle ticks were added");
-    });
+        assert.strictEqual(ticks.length, 6, "ticks are generated");
+        ticks.forEach((tick) => {
+          assert.operator(start, "<=", tick, "tick " + tick + " should be greater than the lower bound of the domain");
+          assert.operator(tick, "<=", end, "tick " + tick + " should be less than the upper bound of the domain");
```
jtlan: Optionally, you can use a template string here, but I think the tick value doesn't even need to be in the assert message, since it's already passed in as one of the arguments.

aicioara: It's cleaner, but I agree with the template. Will probably be part of the second pass if it is a common case

PR #2697, opened 2015-08-28
Wheel plot

ztsai:


jtlan:
Identical changes to #2677, so I think this is good to go.

Issue #2628, opened 2015-08-17
[Testing Audit] Interactions and Dispatchers


crmorford:

aicioara:
Note for self. This means:

Interactions
- [ ] `clickInteractionTests.ts`
- [ ] `doubleClickInteractionTests.ts`
- [ ] `dragInteractionTests.ts`
- [ ] `keyInteractionTests.ts`
- [ ] `panZoomInteractionTests.ts`
- [ ] `pointerInteractionTests.ts`
- [ ] `interactionTests.ts`

Dispatchers
- [ ] `keyDispatcherTests.ts`
- [ ] `mouseDispatcherTests.ts`
- [ ] `touchDispatcherTests.ts`
- [ ] `dispatcherTests.ts`

aicioara:
Blocking on `interactionTests.ts` upon the rest merging in

nmiyake:
Taking another look at this.

Going to separate dispatchers and interactions to further sub-divide work. Focusing on interactions first.

For each interaction, test:

##### Registration
* Register listener, fire, callback called
* Register listener, unregister listener, fire, callback not called
* Register multiple listeners, fire, all callbacks called
* Register multiple listeners, unregister one, fire, one callback called
* Verify register/unregister returns self

##### Callbacks
* Verify callbacks called properly
  * Cover Mouse and Touch for callbacks that should be equivalent
* Verify any special cases

##### Public methods
* Test all other public methods

jtlan:
Assuming direct control.

jtlan:
Completed by #2893.

ztsai:
Interactions that need second pass:
- [x] `clickInteractionTests.ts`
- [x] `doubleClickInteractionTests.ts`
- [x] `dragInteractionTests.ts`
- [ ] `keyInteractionTests.ts`
- [ ] `panZoomInteractionTests.ts`
- [ ] `pointerInteractionTests.ts`
- [ ] `interactionTests.ts`

PR #2698, opened 2015-08-28
Refactor double click interaction tests - Pass 1

aicioara:
Part of #2628

As part of this: 
- No nested `beforeEach()`es
- Cleaned the console
screen shot 2015-08-28 at 2 21 32 pm

### DIFF
--- a/test/interactions/doubleClickInteractionTests.ts
+++ b/test/interactions/doubleClickInteractionTests.ts
@@ -84,7 +84,7 @@ describe("Interactions", () => {
         svg.remove();
       });
 
-      it("works with touch events as well", () => {
+      it("works with touch events", () => {
         let doubleClickedPoint: Plottable.Point = null;
         let dblClickCallback = (point: Plottable.Point) => doubleClickedPoint = point;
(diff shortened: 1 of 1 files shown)


test/interactions/doubleClickInteractionTests.ts
```diff
@@ -1,35 +1,56 @@
...
 
 describe("Interactions", () => {
-  describe("DoubleClick", () => {
-    let SVG_WIDTH = 400;
-    let SVG_HEIGHT = 400;
+  describe("DoubleClick Interaction", () => {
 
-    describe("onDblClick generic callback", () => {
+    describe("Basic Usage", () => {
+      let userClickPoint: Plottable.Point;
```
bluong: Unsure of the name `userClickPoint` as the `user` prefix is not useful.  I think we should retain the previous variable name or just `point`

bluong: Hmm I see I think now.  Can we rename to like `clickedPoint` or `userClickedPoint`?  I think I prefer the former.

PR #2701, opened 2015-08-29
Setting to add title elements to Plottable's legend

alyssaq:
This pull request allows specifying whether to create title elements on a Legend component.

It addresses an issue where the title element appears when hovering over the chart area on safari (http://jsfiddle.net/alyssa/q319g8ww). Was traced by @jtlan as an [open bug on safari](https://bugs.webkit.org/show_bug.cgi?id=139690).

Pull request also removes a line where 2 title elements were created per legend entry (screenshot below).
screen shot 2015-08-29 at 6 49 18 pm

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -899,6 +899,10 @@ var Plottable;
          * Specifies if Plottable should show warnings.
          */
         Configs.SHOW_WARNINGS = true;
+        /**
+         * Specifies if Plottable should add title elements.
+         */
+        Configs.ADD_TITLE_ELEMENTS = true;
     })(Configs = Plottable.Configs || (Plottable.Configs = {}));
 })(Plottable || (Plottable = {}));
 
@@ -5174,7 +5178,7 @@ var Plottable;
                 fakeLegendEntry.append("text");
                 this._measurer = new SVGTypewriter.Measurers.Measurer(fakeLegendRow);
                 this._wrapper = new SVGTypewriter.Wrappers.Wrapper().maxLines(1);
-                this._writer = new SVGTypewriter.Writers.Writer(this._measurer, this._wrapper).addTitleElement(true);
+                this._writer = new SVGTypewriter.Writers.Writer(this._measurer, this._wrapper).addTitleElement(Plottable.Configs.ADD_TITLE_ELEMENTS);
             };
             Legend.prototype.maxEntriesPerRow = function (maxEntriesPerRow) {
                 if (maxEntriesPerRow == null) {
@@ -5342,7 +5346,6 @@ var Plottable;
                 var padding = this._padding;
                 var textContainers = entries.select("g.text-container");
                 textContainers.text(""); // clear out previous results
-                textContainers.append("title").text(function (value) { return value; });
                 var self = this;
                 textContainers.attr("transform", "translate(" + layout.textHeight + ", 0)")
                     .each(function (value) {
--- a/plottable.js
+++ b/plottable.js
@@ -900,7 +900,7 @@ var Plottable;
          */
         Configs.SHOW_WARNINGS = true;
         /**
-         * Specifies if Plottable should add title elements.
+         * Specifies if Plottable should add <title> elements to text.
          */
         Configs.ADD_TITLE_ELEMENTS = true;
     })(Configs = Plottable.Configs || (Plottable.Configs = {}));
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -6,5 +6,10 @@ export module Configs {
    * Specifies if Plottable should show warnings.
    */
   export var SHOW_WARNINGS = true;
+
+  /**
+   * Specifies if Plottable should add title elements.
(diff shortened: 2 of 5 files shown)


plottable.js
```diff
@@ -5342,7 +5354,6 @@ var Plottable;
                 var padding = this._padding;
                 var textContainers = entries.select("g.text-container");
                 textContainers.text(""); // clear out previous results
-                textContainers.append("title").text(function (value) { return value; });
```
alyssaq: This is where the 2nd title element is created.

jtlan: I suspect this was left over from before the introduction of SVGTypewriter.

jtlan:
@alyssaq: Currently Plottable doesn't use `<title>`s elsewhere, but I could see them being used on `Axes.Category` for example. Instead of having an API point on `Legend`, can we implement this as a configuration option under `Plottable.Configs`? That would allow users to apply the show/hiding of titles globally, rather than having to do it every time they create a Legend.

alyssaq:
Ah yes, that is much cleaner. Sure, I'll move it to `Plottable.Configs`

src/core/config.ts
```diff
@@ -6,5 +6,10 @@ export module Configs {
    * Specifies if Plottable should show warnings.
    */
   export var SHOW_WARNINGS = true;
+
+  /**
+   * Specifies if Plottable should add title elements.
```
jtlan: Probably expand this to "`Specifies if Plottable should add <title> elements to text.`" for clarity.

test/components/legendTests.ts
```diff
@@ -413,4 +413,33 @@ describe("Legend", () => {
...
     svg.remove();
   });
+
+  it("Title elements are created by default", () => {
+    color.domain(["foo", "bar", "baz"]);
+    legend.renderTo(svg);
+
+    let entries = (<any> legend)._element.selectAll(entrySelector);
+    let titles = entries.selectAll("title");
+    assert.strictEqual(titles[0].length, 1, "only one title node per legend entry should be present");
```
jtlan: `titles.size()` is [preferred](https://github.com/palantir/plottable/wiki/Testing-best-practices#checking-the-size-of-a-d3-selection).

test/components/legendTests.ts
```diff
@@ -413,4 +413,33 @@ describe("Legend", () => {
...
+      let text = d3this.select("text").text();
+      let titleText = d3this.select("title").text();
+      assert.strictEqual(text, titleText, "the text and title node have the same text");
+    });
+    svg.remove();
+  });
+
+  it("No title elements are created if configuration is set to false", () => {
+    color.domain(["foo", "bar", "baz"]);
+    Plottable.Configs.ADD_TITLE_ELEMENTS = false;
```
jtlan: You'll want to set it back to its original setting to avoid polluting the test environment:
```Typescript
let originalSetting = Plottable.Configs.ADD_TITLE_ELEMENTS;
Plottable.Configs.ADD_TITLE_ELEMENTS = false;
legend.renderTo(svg);
Plottable.Configs.ADD_TITLE_ELEMENTS = originalSetting; // do this before asserts so it will happen even if the test fails

... asserts ...

```

test/components/legendTests.ts
```diff
@@ -413,4 +413,33 @@ describe("Legend", () => {
...
+    legend.renderTo(svg);
+
+    let entries = (<any> legend)._element.selectAll(entrySelector);
+    let titles = entries.selectAll("title");
+    assert.strictEqual(titles[0].length, 0, "no titles should be rendered");
```
jtlan: `titles.size()` is [preferred](https://github.com/palantir/plottable/wiki/Testing-best-practices#checking-the-size-of-a-d3-selection).

jtlan:
@alyssaq: some minor issues in the tests and the documentation. Otherwise, this looks good!

jtlan:
@alyssaq One more spot in the tests ... :/

jtlan:
Context for @crmorford:

By default, text on `Legend`s shows a tooltip if hovered-over:
legend tooltip
This is most useful when the full text gets truncated:
legend tooltip truncated
http://jsfiddle.net/a972s9of/

On Safari, however, the tooltip would show up everywhere in the `<svg>`, so long as there was a `Component` behind the `Legend` (try hovering in the blank space on the `Legend`). A new API point has been added to switch titles on and off globally:

```Typescript
module Plottable {
  export module Configs {
    /**
     * Specifies if Plottable should add <title> elements to text.
     */
     var ADD_TITLE_ELEMENTS: boolean;
  }
}
```

Usage:
```Javascript
Plottable.Configs.ADD_TITLE_ELEMENTS = false;
legend.renderTo(svg);
```

The option has to be set before the `Legend` is rendered.

jtlan:
Thanks @alyssaq !

alyssaq:
Thanks for reviewing and merging @jtlan!

PR #2706, opened 2015-08-31
Higher retry rate for saucelabs

aicioara:
This is to account for the tests being slow on iPad. Increased the upper bound from 3 min to 5 min. 

Settings taken from here https://github.com/axemclion/grunt-saucelabs

I also played around with other environments taken from here https://docs.saucelabs.com/reference/platforms-configurator/ . It seems like all of them need between 2:03 and 3:00 + to run all the tests so a higher upper bound make sense.

The reason why we did not have this issue before was less tests. Right now the tests need 60% more time compared to what we had when we released `1.0.0`, due to new features.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -282,6 +282,8 @@ module.exports = function(grunt) {
       options: {
         urls: ["http://127.0.0.1:9999/test/tests.html"],
         testname: "Plottable Sauce Unit Tests",
+        pollInterval: 5000,
+        statusCheckAttempts: 60,
         browsers: [{
           browserName: "firefox",
           platform: "linux"


Gruntfile.js
```diff
@@ -282,6 +282,8 @@ module.exports = function(grunt) {
       options: {
         urls: ["http://127.0.0.1:9999/test/tests.html"],
         testname: "Plottable Sauce Unit Tests",
+        pollInterval: 5000,
+        statusCheckAttempts: 60,
```
bluong: why is statusCheckAttempts lowered?

aicioara: Total time is = pollInterval * statusCheckAttempts. Since our test suite is a single page, it does not make much difference. 

Another reasoning I went through was that polling less often means disturbing the test less often, hence the system might run faster. 

The downside of that is that tests will finish in a second that is multiple of 5, so if a test finishes in 41 seconds, it will only be acknowledged at second 45

aicioara: All the above are wild guesses I thought it is worth experimenting with

bluong: Let's see if this leads to less failures.  I don't think the increases in time outweigh the pain of the tests flaking.

PR #2715, opened 2015-09-02
refactor Plots.Scatter unit test

### LINKED ISSUE #2631: [Testing Audit] Scatter

ztsai:
#2631 

screen shot 2015-09-02 at 3 51 33 pm

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -489,7 +489,7 @@ export class Plot extends Component {
 
       dataset.data().forEach((datum: any, datasetIndex: number) => {
         let position = this._pixelPoint(datum, datasetIndex, dataset);
-        if (!Utils.Math.isValidNumber(position.x) || !Utils.Math.isValidNumber(position.y)) {
+        if (Utils.Math.isNaN(position.x) || Utils.Math.isNaN(position.y)) {
           return;
         }
         lightweightEntities.push({
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -489,7 +489,7 @@ export class Plot extends Component {
 
       dataset.data().forEach((datum: any, datasetIndex: number) => {
         let position = this._pixelPoint(datum, datasetIndex, dataset);
-        if (Utils.Math.isNaN(position.x) || Utils.Math.isNaN(position.y)) {
+        if (!Utils.Math.isValidNumber(position.x) || !Utils.Math.isValidNumber(position.y)) {
           return;
         }
         lightweightEntities.push({
--- a/test/plots/scatterPlotTests.ts
+++ b/test/plots/scatterPlotTests.ts
@@ -2,395 +2,386 @@
 
 describe("Plots", () => {
   describe("ScatterPlot", () => {
-    it("renders correctly with no data", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale);
-      plot.y((d: any) => d.y, yScale);
-      assert.doesNotThrow(() => plot.renderTo(svg), Error);
-      assert.strictEqual(plot.width(), 400, "was allocated width");
-      assert.strictEqual(plot.height(), 400, "was allocated height");
-      svg.remove();
-    });
-
-    it("the accessors properly access data, index and Dataset", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-      xScale.domain([0, 400]);
-      yScale.domain([400, 0]);
-      let data = [{x: 0, y: 0}, {x: 1, y: 1}];
-      let metadata = {foo: 10, bar: 20};
-      let xAccessor = (d: any, i: number, dataset: Plottable.Dataset) => d.x + i * dataset.metadata().foo;
-      let yAccessor = (d: any, i: number, dataset: Plottable.Dataset) => dataset.metadata().bar;
-      let dataset = new Plottable.Dataset(data, metadata);
-      let plot = new Plottable.Plots.Scatter()
-                                  .x(xAccessor)
-                                  .y(yAccessor);
-      plot.addDataset(dataset);
-      plot.renderTo(svg);
-      let symbols = plot.selections();
-      let c1 = d3.select(symbols[0][0]);
-      let c2 = d3.select(symbols[0][1]);
-      let c1Position = d3.transform(c1.attr("transform")).translate;
-      let c2Position = d3.transform(c2.attr("transform")).translate;
-      assert.closeTo(c1Position[0], 0, 0.01, "first symbol cx is correct");
-      assert.closeTo(c1Position[1], 20, 0.01, "first symbol cy is correct");
-      assert.closeTo(c2Position[0], 11, 0.01, "second symbol cx is correct");
-      assert.closeTo(c2Position[1], 20, 0.01, "second symbol cy is correct");
-
-      let changedData = [{x: 2, y: 2}, {x: 4, y: 4}];
-      dataset.data(changedData);
-      c1Position = d3.transform(c1.attr("transform")).translate;
-      c2Position = d3.transform(c2.attr("transform")).translate;
-      assert.closeTo(c1Position[0], 2, 0.01, "first symbol cx is correct after data change");
-      assert.closeTo(c1Position[1], 20, 0.01, "first symbol cy is correct after data change");
-      assert.closeTo(c2Position[0], 14, 0.01, "second symbol cx is correct after data change");
-      assert.closeTo(c2Position[1], 20, 0.01, "second symbol cy is correct after data change");
-
-      let changedMetadata = {foo: 0, bar: 0};
-      dataset.metadata(changedMetadata);
-      c1Position = d3.transform(c1.attr("transform")).translate;
-      c2Position = d3.transform(c2.attr("transform")).translate;
-
-      assert.closeTo(c1Position[0], 2, 0.01, "first symbol cx is correct after metadata change");
-      assert.closeTo(c1Position[1], 0, 0.01, "first symbol cy is correct after metadata change");
-      assert.closeTo(c2Position[0], 4, 0.01, "second symbol cx is correct after metadata change");
-      assert.closeTo(c2Position[1], 0, 0.01, "second symbol cy is correct after metadata change");
-
-      svg.remove();
-    });
-
-    it("selections()", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-      let data = [{x: 0, y: 0}, {x: 1, y: 1}];
-      let data2 = [{x: 1, y: 2}, {x: 3, y: 4}];
-      let plot = new Plottable.Plots.Scatter()
-                                   .x((d: any) => d.x, xScale)
-                                   .y((d: any) => d.y, yScale)
-                                   .addDataset(new Plottable.Dataset(data))
-                                   .addDataset(new Plottable.Dataset(data2));
-      plot.renderTo(svg);
-      let allCircles = plot.selections();
-      assert.strictEqual(allCircles.size(), 4, "all circles retrieved");
-      let selectionData = allCircles.data();
-      assert.includeMembers(selectionData, data, "first dataset data in selection data");
-      assert.includeMembers(selectionData, data2, "second dataset data in selection data");
-
-      svg.remove();
-    });
-
-    it("entityNearest()", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
-      let plot = new Plottable.Plots.Scatter()
-                                   .x((d: any) => d.x, xScale)
-                                   .y((d: any) => d.y, yScale)
-                                   .addDataset(dataset)
-                                   .addDataset(dataset2);
-      plot.renderTo(svg);
-
-      let points = d3.selectAll(".scatter-plot path");
-      let d0 = dataset.data()[0];
-      let d0Px = {
-        x: xScale.scale(d0.x),
-        y: yScale.scale(d0.y)
-      };
-
-      let expected: Plottable.Plots.PlotEntity = {
-        datum: d0,
-        index: 0,
-        dataset: dataset,
-        position: d0Px,
-        selection: d3.selectAll([points[0][0]]),
-        component: plot
-      };
-
-      let closest = plot.entityNearest({ x: d0Px.x + 1, y: d0Px.y + 1 });
-      TestMethods.assertPlotEntitiesEqual(closest, expected, "it selects the closest data point");
-
-      yScale.domain([0, 1.9]);
-
-      let d1 = dataset.data()[1];
-      let d1Px = {
-        x: xScale.scale(d1.x),
-        y: yScale.scale(d1.y)
-      };
-
-      expected = {
-        datum: d1,
-        index: 1,
-        dataset: dataset,
-        position: d1Px,
-        selection: d3.selectAll([points[0][1]]),
-        component: plot
-      };
-
-      closest = plot.entityNearest({ x: d1Px.x, y: 0 });
-      TestMethods.assertPlotEntitiesEqual(closest, expected, "it ignores off-plot data points");
-
-      svg.remove();
-    });
-
-    it("can retrieve entities in a certain range", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesIn({ min: xScale.scale(1), max: xScale.scale(1) },
-                                     { min: yScale.scale(1), max: yScale.scale(1) });
-
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      svg.remove();
-    });
-
-    it("entities are not returned if their center lies outside the range", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesIn({ min: xScale.scale(1.001), max: xScale.scale(1.001) },
-                                     { min: yScale.scale(1.001), max: yScale.scale(1.001) });
-
-      assert.lengthOf(entities, 0, "no entities retrieved");
-
-      svg.remove();
-    });
-
-    it("can retrieve entities in a certain bounds", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesIn({ topLeft: {
-                                         x: xScale.scale(1),
-                                         y: yScale.scale(1)
-                                       },
-                                       bottomRight: {
-                                         x: xScale.scale(1),
-                                         y: yScale.scale(1)
-                                       }});
-
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      svg.remove();
-    });
-
-    it("can retrieve entities centered at a given point", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      svg.remove();
-    });
-
-    it("determines whether an entity contains a given point by its size", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.size(10)
-          .x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      plot.size(6);
-      entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
-      assert.lengthOf(entities, 0, "none of the entities is retrieved");
-      svg.remove();
-    });
-
-    it("returns all entities containing a given point across all datasets", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 200, y: 200}]);
-      let dataset2 = new Plottable.Dataset([{x: 0, y: 1}, {x: 1, y: 0}]);
-      let plot = new Plottable.Plots.Scatter();
-
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
-
-      let entities = plot.entitiesAt({ x: xScale.scale(0.5), y: yScale.scale(0.5)});
-      assert.lengthOf(entities, 3, "all 3 entities containing the point have been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 0, y: 0}, "correct datum has been retrieved");
-      assert.deepEqual(entities[1].datum, {x: 0, y: 1}, "correct datum has been retrieved");
-      assert.deepEqual(entities[2].datum, {x: 1, y: 0}, "correct datum has been retrieved");
-      svg.remove();
-    });
-
-    it("correctly handles NaN and undefined x and y values", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let data = [
-        { foo: 0.0, bar: 0.0 },
-        { foo: 0.2, bar: 0.2 },
-        { foo: 0.4, bar: 0.4 },
-        { foo: 0.6, bar: 0.6 },
-        { foo: 0.8, bar: 0.8 }
-      ];
-      let dataset = new Plottable.Dataset(data);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
-      let plot = new Plottable.Plots.Scatter();
-      plot.addDataset(dataset);
-      plot.x((d: any) => d.foo, xScale)
-          .y((d: any) => d.bar, yScale);
-      plot.renderTo(svg);
-
-      let dataWithNaN = data.slice();
-      dataWithNaN[2] = { foo: 0.4, bar: NaN };
-      dataset.data(dataWithNaN);
-      assert.strictEqual(plot.selections().size(), 4, "does not draw NaN point");
-
-      let dataWithUndefined = data.slice();
-      dataWithUndefined[2] = { foo: 0.4, bar: undefined };
-      dataset.data(dataWithUndefined);
-      assert.strictEqual(plot.selections().size(), 4, "does not draw undefined point");
-      dataWithUndefined[2] = { foo: undefined, bar: 0.4 };
-      dataset.data(dataWithUndefined);
-      assert.strictEqual(plot.selections().size(), 4, "does not draw undefined point");
-
-      svg.remove();
-    });
-
-    describe("Example ScatterPlot with quadratic series", () => {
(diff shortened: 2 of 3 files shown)


aicioara:
Branch name. It's not er, it's refactor

src/plots/plot.ts
```diff
@@ -489,7 +489,7 @@ export class Plot extends Component {
 
       dataset.data().forEach((datum: any, datasetIndex: number) => {
         let position = this._pixelPoint(datum, datasetIndex, dataset);
-        if (Utils.Math.isNaN(position.x) || Utils.Math.isNaN(position.y)) {
+        if (!Utils.Math.isValidNumber(position.x) || !Utils.Math.isValidNumber(position.y)) {
```
aicioara: Outside the scope of this refactor. Skip test (if failing), and make separate PR please.

aicioara:
Add a SS with the current test structure

aicioara:
Are you doing more passes or this is final?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+
+        svg.remove();
+      });
     });
 
-    it("can retrieve entities in a certain range", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
+    describe("the Accessors properly access data, index and Dataset", () => {
```
aicioara: Capitalize for consistency

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      assert.strictEqual(plot.selections().size(), 4, "does not draw undefined point");
-      dataWithUndefined[2] = { foo: undefined, bar: 0.4 };
-      dataset.data(dataWithUndefined);
-      assert.strictEqual(plot.selections().size(), 4, "does not draw undefined point");
-
-      svg.remove();
     });
 
-    describe("Example ScatterPlot with quadratic series", () => {
+    describe("selections", () => {
```
aicioara: Capitalize for consistency

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        assert.lengthOf(entities, 3, "all 3 Entities containing the Point have been retrieved");
+        assert.deepEqual(entities[0].datum, { x: 0, y: 0 }, "correct datum has been retrieved");
+        assert.deepEqual(entities[1].datum, { x: 0, y: 1 }, "correct datum has been retrieved");
+        assert.deepEqual(entities[2].datum, { x: 1, y: 0 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+    });
+
+    describe("invalid data and deprecated method", () => {
```
aicioara: Capitalize for consistency

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-        component: plot
-      };
-
-      closest = plot.entityNearest({ x: d1Px.x, y: 0 });
-      TestMethods.assertPlotEntitiesEqual(closest, expected, "it ignores off-plot data points");
-
-      svg.remove();
+      it("is initialized correctly", () => {
+        let data = [{ x: 0, y: 0 }];
+        let dataset = new Plottable.Dataset(data);
```
aicioara: inline

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
+      beforeEach(() => {
+        svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+        let xScale = new Plottable.Scales.Linear().domain([0, SVG_WIDTH]);
+        let yScale = new Plottable.Scales.Linear().domain([SVG_HEIGHT, 0]);
```
aicioara: Is there a reason why this is a reversed domain?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        let metadata = { foo: 10, bar: 20 };
+        let xAccessor = (d: any, i: number, dataset: Plottable.Dataset) => d.x + i * dataset.metadata().foo;
```
aicioara: Inline and don't type it anymore

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        let yAccessor = (d: any, i: number, dataset: Plottable.Dataset) => dataset.metadata().bar;
```
aicioara: Same

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        svg.remove();
+      });
     });
 
-    it("can retrieve entities in a certain range", () => {
-      let svg = TestMethods.generateSVG(400, 400);
-      let xScale = new Plottable.Scales.Linear();
-      let yScale = new Plottable.Scales.Linear();
+    describe("the Accessors properly access data, index and Dataset", () => {
+      const SVG_WIDTH = 400;
```
aicioara: This is never used. Inline.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      let dataset = new Plottable.Dataset([{x: 0, y: 0}, {x: 1, y: 1}]);
-      let dataset2 = new Plottable.Dataset([{x: 1, y: 2}, {x: 3, y: 4}]);
-      let plot = new Plottable.Plots.Scatter();
-      plot.x((d: any) => d.x, xScale)
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
+      beforeEach(() => {
+        svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
```
aicioara: Is there a reason why we need a specific SVG_WIDTH and height? If not, let's use the default method.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
+        let c1 = d3.select(symbols[0][0]);
+        let c2 = d3.select(symbols[0][1]);
+        let c1Position = d3.transform(c1.attr("transform")).translate;
+        let c2Position = d3.transform(c2.attr("transform")).translate;
+        assert.closeTo(c1Position[0], 0, 0.01, "first symbol cx is correct");
+        assert.closeTo(c1Position[1], 20, 0.01, "first symbol cy is correct");
+        assert.closeTo(c2Position[0], 11, 0.01, "second symbol cx is correct");
+        assert.closeTo(c2Position[1], 20, 0.01, "second symbol cy is correct");
```
aicioara: If these values are easy to compute let's try to compute them instead of magically dumping them? 

Please comment as we have not reached a sound conclusion regarding this best practice.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset);
-      plot.renderTo(svg);
+        svg.remove();
+      });
 
-      let entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
+      it("sets and gets size() correctly", () => {
```
aicioara: I was suggested that testing the getter is useless as we do that everywhere anyway. Amend the test message to `sets the size() correctly`

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      assert.lengthOf(entities, 3, "all 3 entities containing the point have been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 0, y: 0}, "correct datum has been retrieved");
-      assert.deepEqual(entities[1].datum, {x: 0, y: 1}, "correct datum has been retrieved");
-      assert.deepEqual(entities[2].datum, {x: 1, y: 0}, "correct datum has been retrieved");
-      svg.remove();
-    });
+        svg.remove();
+      });
+
+      it("sets and gets symbol() correctly", () => {
```
aicioara: re

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
+      it("sets and gets size() correctly", () => {
+        assert.strictEqual(plot.size().accessor(data[0], 0, dataset), 6, "first symbol size is initialized.");
+        assert.strictEqual(plot.size().accessor(data[1], 1, dataset), 6, "second symbol size is initialized.");
 
-      svg.remove();
-    });
+        plot.size(10);
+        assert.strictEqual(plot.size().accessor(data[0], 0, dataset), 10, "first symbol size is set to a constant.");
```
aicioara: no full stop. Same for the rest of the asserts

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      plot.size(6);
-      entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
-      assert.lengthOf(entities, 0, "none of the entities is retrieved");
-      svg.remove();
-    });
+        plot.size((d: any, i: number) => i + d.x * d.y);
+        assert.strictEqual(plot.size().accessor(data[0], 0, dataset), 0, "first symbol size is calculated correctly.");
```
aicioara: re

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      let entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
+      it("sets and gets size() correctly", () => {
+        assert.strictEqual(plot.size().accessor(data[0], 0, dataset), 6, "first symbol size is initialized.");
+        assert.strictEqual(plot.size().accessor(data[1], 1, dataset), 6, "second symbol size is initialized.");
 
-      svg.remove();
-    });
+        plot.size(10);
```
aicioara: Assert that it returns the plot

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      let entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
-      assert.lengthOf(entities, 1, "only one entity has been retrieved");
-      assert.deepEqual(entities[0].datum, {x: 1, y: 1}, "correct datum has been retrieved");
-
-      plot.size(6);
-      entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
-      assert.lengthOf(entities, 0, "none of the entities is retrieved");
-      svg.remove();
-    });
+        plot.size((d: any, i: number) => i + d.x * d.y);
```
aicioara: No need to explicitly type

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
-      it("rendering is idempotent", () => {
-        circlePlot.render();
-        circlePlot.render();
-        circlePlot.selections().each(getCirclePlotVerifier());
-        assert.strictEqual(circlesInArea, 10, "10 circles were drawn");
+      it("selects the closest data point", () => {
+        plot.renderTo(svg);
+        let closest = plot.entityNearest({
+          x: xScale.scale(0) + 1,
+          y: yScale.scale(0) + 1 });
```
aicioara: \}\ on new line

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
 
-      it("rendering is idempotent", () => {
-        circlePlot.render();
-        circlePlot.render();
-        circlePlot.selections().each(getCirclePlotVerifier());
-        assert.strictEqual(circlesInArea, 10, "10 circles were drawn");
+      it("selects the closest data point", () => {
+        plot.renderTo(svg);
+        let closest = plot.entityNearest({
+          x: xScale.scale(0) + 1,
```
aicioara: Does it work without + 1 ?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+          y: xScale.scale(2) });
+        assert.deepEqual(closest.datum, dataset.data()[1], "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("can retrieve Entities in a certain range", () => {
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesIn({ min: xScale.scale(1), max: xScale.scale(1) },
```
aicioara: Indent.

Ideally \n after `{` and `,`

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("does not return Entities whose center lies outside the range", () => {
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesIn({ min: xScale.scale(1.001), max: xScale.scale(1.001) },
+                                      { min: yScale.scale(1.001), max: yScale.scale(1.001) });
```
aicioara: indent

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+
+        assert.lengthOf(entities, 0, "no Entities retrieved");
+
+        svg.remove();
+      });
+
+      it("can retrieve Entities in a certain bounds", () => {
+        plot.renderTo(svg);
+        let entities = plot.entitiesIn({ topLeft: {
```
aicioara: \n and indent with 2 ws compared to the parent. Looks neater

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+                                          y: yScale.scale(1)
+                                        }});
+
+        assert.lengthOf(entities, 1, "only one Entity has been retrieved");
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("can retrieve Entities centered at a given Point", () => {
```
aicioara: How is this different than the one on line 179 `selects all points in all datasets` and 190 `selects the closest data point`

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        let entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
+        assert.lengthOf(entities, 1, "only one Entity has been retrieved");
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("determines whether an Entity contains a given Point by its position and size", () => {
+        plot.size(10);
+        plot.removeDataset(dataset2);
```
aicioara: Code smell. Do not add the dataset 2 in `beforeEach`, instead add it only in the test which tests for 2 datasets.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        plot.size(6);
+        entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
+        assert.lengthOf(entities, 0, "none of the Entities is retrieved");
+        svg.remove();
+      });
+
+      it("returns all Entities containing a given Point across all Datasets", () => {
+        dataset.data([{ x: 0, y: 0 }, { x: 200, y: 200 }]);
```
aicioara: Don't change the data of the dataset, instead add a new dataset with the new data. This stacks with the "remove dataset2 from beforeEach"

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+
+        dataWithString = data.slice();
+        dataWithString[2] = { x: "abc", y: 0.4 };
+        dataset.data(dataWithString);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the x values is \"abc\"");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the x values is \"abc\"");
+        svg.remove();
+      });
+
+      it("shows warning when deprecated method is called", () => {
```
aicioara: I suggest delete this test. We are testing a deprecated private method. This cannot be right.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+      let plot: Plottable.Plots.Scatter<number, number>;
+      beforeEach(() => {
+        svg = TestMethods.generateSVG(400, 400);
+        plot = new Plottable.Plots.Scatter<number, number>();
+        plot.x((d: any) => d.x)
+            .y((d: any) => d.y);
+      });
+
+      it("correctly handles NaN, undefined, Infinity, and non-number x and y values", () => {
+        let data: { [key: string]: any }[] = [
```
aicioara: do not type it

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+
+        assert.strictEqual(plot.selections().size(), 5, "all 5 data points are drawn");
+
+        let dataWithNaN = data.slice();
+        dataWithNaN[2] = { x: 0.4, y: NaN };
+        dataset.data(dataWithNaN);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the y values is NaN");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned  when one of the y values is NaN");
+
+        dataWithNaN = data.slice();
```
aicioara: Just building on top of the previous test and modify element 3

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the x values is NaN");
+
+        let dataWithUndefined = data.slice();
+        dataWithUndefined[2] = { x: 0.4, y: undefined };
+        dataset.data(dataWithUndefined);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the y values is undefined");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the y values is undefined");
+
+        dataWithUndefined = data.slice();
+        dataWithUndefined[2] = { x: undefined, y: 0.4 };
```
aicioara: re

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,386 @@
...
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the x values is undefined");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the x values is undefined");
+
+        let dataWithInfinity = data.slice();
+        dataWithInfinity[2] = { x: 0.4, y: Infinity };
+        dataset.data(dataWithInfinity);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the y values is Infinity");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the y values is Infinity");
+
+        dataWithInfinity = data.slice();
```
aicioara: re

test/plots/scatterPlotTests.ts
ztsai: so that 0 is at the bottom which is the common visualization of x-y plane

test/plots/scatterPlotTests.ts
ztsai: it's used to set domain below

test/plots/scatterPlotTests.ts
ztsai: below we tested for the position of rendered symbol.
setting SVG_WIDTH and SVG_HEIGHT(and domains for Scales) prevents making assumption for expected positions

test/plots/scatterPlotTests.ts
aicioara: I am not sure I see what you mean. 0 is always at the bottom on the y-axis

test/plots/scatterPlotTests.ts
aicioara: But that happens inside the beforeEach block and nowhere else

test/plots/scatterPlotTests.ts
ztsai: well, unit tests require more or less some hard coded values
I feel like hard coding expected values may prevent error not being caught by copying and pasting code, but recomputing the values might be more reusable/readable?
I don't really feel that strongly either way

test/plots/scatterPlotTests.ts
ztsai: it will work, but it's different testing against the exact center position versus a point near it

test/plots/scatterPlotTests.ts
ztsai: `selects all points in all datasets` tests for `selections()`
`selects the closest data point` tests for `entityNearest()`
and `can retrieve Entities centered at a given Point`  tests for `entitiesAt()`

test/plots/scatterPlotTests.ts
ztsai: I will move it inside then

test/plots/scatterPlotTests.ts
ztsai: 0 is at the top in pixel space

test/plots/scatterPlotTests.ts
ztsai: in the compiled js, it's just protected by `private` declaration and this test is not for the functionality of the method but for it to show deprecation warning message

ztsai:
I will do another pass only if anyone adds anything to the testing best practice list

test/plots/scatterPlotTests.ts
aicioara: This practice is not scalable. We can't do that for all deprecated methods. I feel strong against testing this. If we tested that the method is still doing the right thing then sure, but testing that deprecated methods throw a warning is already tested in the `windowUtilsTests.ts`

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
-          .y((d: any) => d.y, yScale)
-          .addDataset(dataset)
-          .addDataset(dataset2);
-      plot.renderTo(svg);
+      beforeEach(() => {
+        const SVG_WIDTH = 400;
+        const SVG_HEIGHT = 400;
+        svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+        let xScale = new Plottable.Scales.Linear().domain([0, SVG_WIDTH]);
+        let yScale = new Plottable.Scales.Linear().domain([SVG_HEIGHT, 0]);
```
aicioara: Why not use normal domain.

Currently looks like this
screen shot 2015-09-02 at 5 34 31 pm

Unusual behavior should mentioned either in the describe or in the it or in the assert and should be used for testing purposes only

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+        plot.y((d, i, dataset) => dataset.metadata().bar, yScale);
+        plot.addDataset(dataset);
+      });
 
-      let entities = plot.entitiesIn({ min: xScale.scale(1), max: xScale.scale(1) },
-                                     { min: yScale.scale(1), max: yScale.scale(1) });
+      it("renders symbols correctly", () => {
+        plot.renderTo(svg);
+        let symbols = plot.selections();
+        assert.strictEqual(symbols.size(), 2, "exactly 2 symbols are rendered");
```
aicioara: replace 2 with data.length

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
-        circlePlot.render();
-        circlePlot.render();
-        circlePlot.selections().each(getCirclePlotVerifier());
-        assert.strictEqual(circlesInArea, 10, "10 circles were drawn");
+      it("selects the closest data point", () => {
+        plot.addDataset(dataset2);
+        plot.renderTo(svg);
+        let closest = plot.entityNearest({
+          x: xScale.scale(0) + 1,
+          y: yScale.scale(0) + 1
```
aicioara: Why the `+1`s?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
         svg.remove();
       });
 
-      it("rendering is idempotent", () => {
-        circlePlot.render();
-        circlePlot.render();
-        circlePlot.selections().each(getCirclePlotVerifier());
-        assert.strictEqual(circlesInArea, 10, "10 circles were drawn");
+      it("selects the closest data point", () => {
+        plot.addDataset(dataset2);
```
aicioara: Can we make the first dataset more comprehensive such that we do not need a second one? This is only to be tested in the first test.

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+
         svg.remove();
       });
 
-      describe("after the scale has changed", () => {
-        beforeEach(() => {
-          xScale.domain([0, 3]);
-          yScale.domain([0, 9]);
+      it("ignores off-plot data points when retrieving the nearest Entity", () => {
+        plot.addDataset(dataset2);
```
aicioara: Can we make the first dataset more comprehensive such that we do not need a second one?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
-        it("the circles re-rendered properly", () => {
-          let circles = circlePlot.selections();
-          circles.each(getCirclePlotVerifier());
-          assert.strictEqual(circlesInArea, 4, "four circles were found in the render area");
-          svg.remove();
+        svg.remove();
+      });
+
+      it("can retrieve Entities in a certain range", () => {
+        plot.addDataset(dataset2);
```
aicioara: Can we make the first dataset more comprehensive such that we do not need a second one?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+        });
+
+        assert.lengthOf(entities, 1, "only one Entity has been retrieved");
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("does not return Entities whose center lies outside the range", () => {
+        plot.addDataset(dataset2);
```
aicioara: Can we make the first dataset more comprehensive such that we do not need a second one?

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+          max: yScale.scale(1.001)
         });
+
+        assert.lengthOf(entities, 0, "no Entities retrieved");
+
+        svg.remove();
+      });
+
+      it("can retrieve Entities in a certain bounds", () => {
+        plot.addDataset(dataset2);
```
aicioara: Can we make the first dataset more comprehensive such that we do not need a second one?

test/plots/scatterPlotTests.ts
aicioara: makes sense

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+        assert.lengthOf(entities, 1, "only one Entity has been retrieved");
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("can retrieve Entities centered at a given Point", () => {
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesAt({ x: xScale.scale(1), y: yScale.scale(1) });
```
aicioara: Separate on multiple lines for consistency with the above. Either that or amend above to match this

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+        assert.deepEqual(entities[0].datum, { x: 1, y: 1 }, "correct datum has been retrieved");
+
+        svg.remove();
+      });
+
+      it("determines whether an Entity contains a given Point by its position and size", () => {
+        plot.size(10);
+        plot.renderTo(svg);
+
+        let entities = plot.entitiesAt({ x: xScale.scale(1) + 5, y: yScale.scale(1) - 5});
```
aicioara: Extract `5` to a variable. 
```typescript
let diameter = 10;
let radius = diameter / 2;
plot.size(diameter);
```

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the y values is Infinity");
+
+        dataWithError = data.slice();
+        dataWithError[2] = { x: -Infinity, y: 0.4 };
+        dataset.data(dataWithError);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the x values is -Infinity");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned when one of the x values is -Infinity");
+
+        dataWithError = data.slice();
+        (<any> dataWithError)[2] = { x: 0.4, y: "12" };
```
aicioara: <any> cast it at the very beginning so you don't have to do it here

```typescript
let dataWithError: any = data.slice();
```

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,399 @@
...
+
+        assert.strictEqual(plot.selections().size(), 5, "all 5 data points are drawn");
+
+        let dataWithError = data.slice();
+        dataWithError[2] = { x: 0.4, y: NaN };
+        dataset.data(dataWithError);
+        assert.strictEqual(plot.selections().size(), 4, "only 4 data points are drawn when one of the y values is NaN");
+        assert.lengthOf(plot.entities(), 4, "only 4 Entities returned  when one of the y values is NaN");
+
+        dataWithError = data.slice();
```
aicioara: Instead of creating so many objects, can we keep amending one item at a time and see the `selections().size()` go down all the time.

Don't have a strong preference here though.

test/plots/scatterPlotTests.ts
ztsai: it's testing for the nearest, so +1 to avoid testing against the exact position to reduce bias

test/plots/scatterPlotTests.ts
ztsai: you asked for not setting the type in previous CR

test/plots/scatterPlotTests.ts
ztsai: `dataset.data()` is a direct assignment, so if you modify `dataWithError` directly it will modify `dataset._data` as well
I think it's safer to create new array every time

test/plots/scatterPlotTests.ts
ztsai: we could, but I think the original intention was to make sure those functions work with multiple datasets

test/plots/scatterPlotTests.ts
aicioara: That assumes a radius of 1, shall we try to catch the closest outside the radius?

I don't have a strong preference, but in either way, extract `1` into a variable so it is clearer what is going on. From just reading it, it was not clear to me

test/plots/scatterPlotTests.ts
aicioara: We have a dedicated `it()` test that does that. This test is just `"can retrieve Entities in a certain bounds"` There is no mention of multiple datasets in the title.

test/plots/scatterPlotTests.ts
aicioara: I think in general we prefer not to type the objects where the type can be inferred.

For example 
```typescript
let x: number = 2; // It is obvious it is a number
```

However, in this case, the array is of anys (that is why you need to cast it on line 368 and below)
```typescript
let x = [1, 2, 3]; // x will be of type number[]
x.push("foo"); // compiler complains, x was of type number[]. <any>-casting it is a hack
let y = [1, "foo"]; // this will still not be of type any, it is of type (number | string)[]
```

Hence the typing is needed to be any

test/plots/scatterPlotTests.ts
aicioara: agree

test/plots/scatterPlotTests.ts
ztsai: well, in the original commit I typed it to `{ [key: string]: any }` because it will be inferred to {x: number, y: number}[] which will cause an error on line 368, when `y` is no longer a number
I don't have a strong opinion about how to approach it, but I would like to know what's the plottable convention as you seem not liking either way

aicioara: Sorry, now that I look back I believe my comments were confusing and indeed contradictory. Let me try again.

Initially I was suggesting to reduce the test size by just creating an initial object with
```typescript
x = [badValue, badValue, badValue, badValue, goodValue].
assert.lengthOf(plot.entities(), 1) // just the goodValue
```

That meant that the array does not need to be typed.
But then I deleted that comment as I realized this is such a better way of testing it. However, I did not realize that now typing the object is required.
My mistake, I apologize for the contradictory comments. My current stand right now is: we should either type it as `any[]` or `{[key: string]: any }[] `. This is because <any> casting it does not make sense on this line, we want to avoid that.

Convention wise:
- Wherever the type can be inferred, don't type it explicitly 
- Avoid <any> casting as much as possible

test/plots/scatterPlotTests.ts
```diff
@@ -2,395 +2,379 @@
...
+
         svg.remove();
       });
 
-      describe("after the scale has changed", () => {
-        beforeEach(() => {
-          xScale.domain([0, 3]);
-          yScale.domain([0, 9]);
+      it("determines whether an Entity contains a given Point by its position and size", () => {
+        let originalSize = 10;
```
aicioara: why the prefix `original`?

ztsai: to suggest size will change and `originalRadius` won't work when the size changes

aicioara: Makes sense

aicioara: call it `initialRadius` ?

Issue #2629, opened 2015-08-17
[Testing Audit] Factory Methods and Utilities


crmorford:
- [x] `Formatters`
- [x] `TickGenerators`
- [ ] `SymbolFactories`

PR #2719, opened 2015-09-03
Formatter Unit test refactor

ztsai:
part of #2629
screen shot 2015-09-03 at 3 17 08 pm

### DIFF
--- a/test/core/formattersTests.ts
+++ b/test/core/formattersTests.ts
@@ -1,193 +1,280 @@
 ///<reference path="../testReference.ts" />
 
 describe("Formatters", () => {
-  describe("fixed", () => {
-    it("shows exactly [precision] digits", () => {
+  describe("fixed()", () => {
+    it("shows correct amount of digits according to precision", () => {
       let fixed3 = Plottable.Formatters.fixed();
-      let result = fixed3(1);
-      assert.strictEqual(result, "1.000", "defaults to three decimal places");
-      result = fixed3(1.234);
-      assert.strictEqual(result, "1.234", "shows three decimal places");
-      result = fixed3(1.2346);
-      assert.strictEqual(result, "1.235", "changed values are not shown (get turned into empty strings)");
-    });
+      assert.strictEqual(fixed3(1), "1.000", "shows three decimal places");
+      assert.strictEqual(fixed3(-1), "-1.000", "shows three decimal places");
+      assert.strictEqual(fixed3(1.234), "1.234", "shows three decimal places");
+      assert.strictEqual(fixed3(-1.234), "-1.234", "shows three decimal places");
+      assert.strictEqual(fixed3(1.2346111111), "1.235", "shows three decimal places");
+      assert.strictEqual(fixed3(-1.2346111111), "-1.235", "shows three decimal places");
 
-    it("precision can be changed", () => {
+      assert.strictEqual(fixed3(Infinity), "Infinity", "formats non-numeric number correctly");
+      assert.strictEqual(fixed3(-Infinity), "-Infinity", "formats non-numeric number correctly");
+      assert.strictEqual(fixed3(NaN), "NaN", "formats non-numeric number correctly");
+
       let fixed2 = Plottable.Formatters.fixed(2);
-      let result = fixed2(1);
-      assert.strictEqual(result, "1.00", "formatter was changed to show only two decimal places");
+      assert.strictEqual(fixed2(1), "1.00", "shows two decimal places");
+      assert.strictEqual(fixed2(-1), "-1.00", "shows two decimal places");
+      assert.strictEqual(fixed2(1.23), "1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(-1.23), "-1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(1.2346111111), "1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(-1.2346111111), "-1.23", "shows two decimal places");
+
+      let fixed0 = Plottable.Formatters.fixed(0);
+      assert.strictEqual(fixed0(1), "1", "shows no decimal places");
+      assert.strictEqual(fixed0(-1), "-1", "shows no decimal places");
+      assert.strictEqual(fixed0(1.23), "1", "shows no decimal places");
+      assert.strictEqual(fixed0(-1.23), "-1", "shows no decimal places");
     });
 
-    it("can be set to show rounded values", () => {
-      let fixed3 = Plottable.Formatters.fixed(3);
-      let result = fixed3(1.2349);
-      assert.strictEqual(result, "1.235", "long values are rounded correctly");
+    it.skip("throws exception for non-numeric values", () => {
+      let nonNumbricValues = [null, undefined, "123", "abc", ""];
+      let fixed = Plottable.Formatters.fixed();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => fixed(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.fixed`)
+      );
     });
 
-    it("non-integer precision throws exception", () => {
-      assert.throws(() => Plottable.Formatters.fixed(2.1)(123.313), "Formatter precision must be an integer");
-    });
-  });
+    it("throws exception for invalid precision", () => {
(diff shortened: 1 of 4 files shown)


test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
 ///<reference path="../testReference.ts" />
 
 describe("Formatters", () => {
-  describe("fixed", () => {
-    it("shows exactly [precision] digits", () => {
+  describe("fixed()", () => {
+    it("shows correct amount of digits according to precision", () => {
       let fixed3 = Plottable.Formatters.fixed();
```
aicioara: This is actually testing that the default formatter has precision 3. Either mention 3 or extract in a different it

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
-      let result = fixed3(1);
-      assert.strictEqual(result, "1.000", "defaults to three decimal places");
-      result = fixed3(1.234);
-      assert.strictEqual(result, "1.234", "shows three decimal places");
-      result = fixed3(1.2346);
-      assert.strictEqual(result, "1.235", "changed values are not shown (get turned into empty strings)");
-    });
+      assert.strictEqual(fixed3(1), "1.000", "shows three decimal places");
```
aicioara: Single digit number is an edge case, try something >=10

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
+      assert.strictEqual(fixed2(1.23), "1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(-1.23), "-1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(1.2346111111), "1.23", "shows two decimal places");
+      assert.strictEqual(fixed2(-1.2346111111), "-1.23", "shows two decimal places");
+
+      let fixed0 = Plottable.Formatters.fixed(0);
+      assert.strictEqual(fixed0(1), "1", "shows no decimal places");
+      assert.strictEqual(fixed0(-1), "-1", "shows no decimal places");
+      assert.strictEqual(fixed0(1.23), "1", "shows no decimal places");
+      assert.strictEqual(fixed0(-1.23), "-1", "shows no decimal places");
```
aicioara: The description of this test and line 28 is identical. Same for 27 & 30. If the test fails, there is no way to tell which of the asserts failed.

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
     });
 
-    it("can be set to show rounded values", () => {
-      let fixed3 = Plottable.Formatters.fixed(3);
-      let result = fixed3(1.2349);
-      assert.strictEqual(result, "1.235", "long values are rounded correctly");
+    it.skip("throws exception for non-numeric values", () => {
+      let nonNumbricValues = [null, undefined, "123", "abc", ""];
```
aicioara: Numeric

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
+      let fixed = Plottable.Formatters.fixed();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => fixed(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.fixed`)
```
aicioara: Best practice, 3rd argument should not be empty string but the error message

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
+      );
+    });
+
+    it.skip("throws exception for non-numeric values", () => {
+      let nonNumbricValues: any[] = [null, undefined, "123", "abc", ""];
+      // "$0.00", "$NaN", "$123.00", "$NaN", "$0.00"
+      let defaultFormatter = Plottable.Formatters.currency();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => defaultFormatter(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.currency`)
```
aicioara: Error message

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
+      assert.strictEqual(timeFormatter(new Date(2000, 2, 1, 20, 34, 53, 0)), ":53", "seconds was displayed");
+      assert.strictEqual(timeFormatter(new Date(2000, 0, 1, 0, 0, 0, 950)), ".950", "milliseconds was displayed");
+    });
+
+    it.skip("throws exception for non-Date values", () => {
+      let nonNumbricValues = [null, undefined, NaN, "123", "abc", "", 0];
+      let timeFormatter = Plottable.Formatters.multiTime();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => timeFormatter(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.multiTime`)
```
aicioara: Error message

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
-      result = timeFormatter(new Date(2000, 2, 1, 20, 34, 53, 0));
-      assert.strictEqual(result, ":53", "seconds was displayed");
-      result = timeFormatter(new Date(2000, 0, 1, 0, 0, 0, 950));
-      assert.strictEqual(result, ".950", "milliseconds was displayed");
+    it.skip("throws exception for non-number values", () => {
+      let nonNumbricValues = [null, undefined, "123", "abc", ""];
+      let percentFormatter = Plottable.Formatters.percentage();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => percentFormatter(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.percentage`)
```
aicioara: Message

test/core/formattersTests.ts
```diff
@@ -1,193 +1,280 @@
...
+          "Formatter precision must be between 0 and 20", `${value} is not a valid precision value`)
+      );
+    });
+
+    it.skip("throws exception for non-number values", () => {
+      let nonNumbricValues = [null, undefined, "123", "abc", ""];
+      let lnFormatter = Plottable.Formatters.siSuffix();
+      nonNumbricValues.forEach((value) =>
+        (<any> assert).throws(() => lnFormatter(<any> value), Error,
+          "", `${value} is not a valid value for Formatter.siSuffix`)
```
aicioara: Message

test/core/formattersTests.ts
ztsai: it's not set because the formatter doesn't handle these special cases.
I feel like determining how error handling should work is outside the scope of this CR
same for below

test/core/formattersTests.ts
ztsai: it's not set because the formatter doesn't handle these special cases.
(note the test is skipped)
I feel like determining how error handling should work is outside the scope of this CR
same for below

aicioara: Makes sense. In that case add a random string there. If the pattern match will fail you will remember to update it upon unskipping, otherwise you won't notice.

test/core/formattersTests.ts
aicioara: This?

test/core/formattersTests.ts
```diff
@@ -1,193 +1,276 @@
...
+  describe("fixed()", () => {
+    it("shows correct amount of digits according to precision", () => {
       let fixed3 = Plottable.Formatters.fixed();
-      let result = fixed3(1);
-      assert.strictEqual(result, "1.000", "defaults to three decimal places");
-      result = fixed3(1.234);
-      assert.strictEqual(result, "1.234", "shows three decimal places");
-      result = fixed3(1.2346);
-      assert.strictEqual(result, "1.235", "changed values are not shown (get turned into empty strings)");
+      assert.strictEqual(fixed3(1), "1.000", "shows three decimal places by defalut");
```
aicioara: Single digit number is an edge case, try something >=10. Here and below

aicioara:
Apart from the 2 comments, looks good

test/core/formattersTests.ts
ztsai: the assertion message is changed to reflect that

test/core/formattersTests.ts
ztsai: it's tested in line 13

PR #2720, opened 2015-09-03
Rectangle fill and Wheel fill to Plottable-Blue

crmorford:
http://jsfiddle.net/jmnwbrzq/

It was bothering me that rectangle and wheel rendered black, so I added the line to fix it.

I didn't add tests because I didn't see tests for this anywhere else. But I can if I should.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -209,7 +209,6 @@ module.exports = function(grunt) {
     server: {
       options: {
         port: 9999,
-        protocol: "https",
         hostname: "*",
         base: "",
         livereload: true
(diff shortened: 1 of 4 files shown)


jtlan:
You'll probably want to remove [this line](https://github.com/palantir/plottable/blob/develop/src/plots/rectanglePlot.ts#L453) From `Rectangle` (it was what was causing the fill to be black).

Oddly enough, `Wheel`'s arcs are black because that's the default fill color, rather than being explicitly set.

bluong:
Can you check the default color is the desired blue?

bluong:
We should probably have tests to test default attributes on the plot, but we don't have a really good stance on this yet.  Holding off on having tests on default attributes until we encounter this in the test refactor.

PR #2721, opened 2015-09-04
TickGeneratorsTests refactor

ztsai:
part of #2629 

screen shot 2015-09-09 at 1 46 59 pm

### DIFF
--- a/test/scales/tickGeneratorsTests.ts
+++ b/test/scales/tickGeneratorsTests.ts
@@ -12,8 +12,9 @@ describe("Scales", () => {
       it("generates ticks within domain", () => {
         let start = 0.5;
         let end = 4.01;
+        let interval = 1;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(1)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
 
         assert.strictEqual(ticks.length, 6, "ticks are generated");
         ticks.forEach((tick) => {
@@ -25,24 +26,27 @@ describe("Scales", () => {
       it("generates ticks for a domain crossing 0", () => {
         let start = -1.5;
         let end = 1;
+        let interval = 0.5;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(0.5)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
         assert.deepEqual(ticks, [-1.5, -1, -0.5, 0, 0.5, 1], "generated all number divisible by 0.5 in domain");
       });
 
       it("generates ticks with reversed domain", () => {
         let start = -2.2;
         let end = -7.6;
+        let interval = 2.5;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(2.5)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
         assert.deepEqual(ticks, [-7.6, -7.5, -5, -2.5, -2.2], "generated all ticks between lower and higher value");
       });
 
       it("returns the ends of the domain if interval is NaN or bigger than the domain", () => {
         let start = 0.5;
         let end = 10.01;
+        let interval = 11;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(11)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
         assert.deepEqual(ticks, [start, end], "no middle ticks were added when interval is 11");
 
         ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(Infinity)(scale);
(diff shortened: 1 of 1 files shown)


test/scales/tickGeneratorsTests.ts
```diff
@@ -79,6 +107,23 @@ describe("Scales", () => {
...
+
+      it.skip("rejects non-QuantitativeScale<number> Scales", () => {
+        let categoryScale: any = new Plottable.Scales.Category();
+        (<any>assert).throws(() => integerTickGenerator(categoryScale), Error,
+          "scale needs to inherit from Scale.QuantitativeScale<number>", "Scales.Category is not a valid parameter for TickGenerators");
+
+        let timeScale: any = new Plottable.Scales.Time();
+        (<any>assert).throws(() => integerTickGenerator(timeScale), Error,
+          "scale needs to inherit from Scale.QuantitativeScale<number>", "Scales.Time is not a valid parameter for TickGenerators");
+      });
```
bluong: Why must we skip these tests?  This should definitely be listed in a comment

test/scales/tickGeneratorsTests.ts
```diff
@@ -2,7 +2,7 @@
 
 describe("Scales", () => {
   describe("Tick generators", () => {
-    describe("Interval", () => {
+    describe("intervalTickGenerator()", () => {
```
bluong: can we more semantically describe what this suite is for?

Maybe something like
"generating ticks with a given interval"?

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
       });
 
       it("generates ticks for a domain crossing 0", () => {
-        let start = -1.5, end = 1, interval = 0.5;
+        let start = -1.5;
+        let end = 1;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(0.5)(scale);
```
bluong: Why not factor out the interval value?

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
...
         assert.deepEqual(ticks, [-1.5, -1, -0.5, 0, 0.5, 1], "generated all number divisible by 0.5 in domain");
       });
 
       it("generates ticks with reversed domain", () => {
-        let start = -2.2, end = -7.6, interval = 2.5;
+        let start = -2.2;
+        let end = -7.6;
         scale.domain([start, end]);
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(2.5)(scale);
```
bluong: Above

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
...
-        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(interval)(scale);
-        assert.deepEqual(ticks, [0.5, 10.01], "no middle ticks were added");
+        let ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(11)(scale);
+        assert.deepEqual(ticks, [0.5, 10.01], "no middle ticks were added when interval is 11");
+
+        ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(Infinity)(scale);
+        assert.deepEqual(ticks, [0.5, 10.01], "no middle ticks were added when interval is Infinity");
+
+        ticks = Plottable.Scales.TickGenerators.intervalTickGenerator(NaN)(scale);
+        assert.deepEqual(ticks, [0.5, 10.01], "no middle ticks were added when interval is NaN");
```
bluong: Seems like the above is
`[start, end]`

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
...
+        let logTicks = Plottable.Scales.TickGenerators.intervalTickGenerator(40)(logScale);
+        assert.deepEqual(logTicks, [0, 40, 80, 100], "generates ticks for Scales.ModifiedLog");
       });
 
-      it("rejects negative values", () => {
+      it("rejects non-positive values", () => {
         (<any>assert).throws(() => Plottable.Scales.TickGenerators.intervalTickGenerator(0), Error, "interval must be positive number",
-          "interval must be positive number");
+          "interval cannot be 0");
         (<any>assert).throws(() => Plottable.Scales.TickGenerators.intervalTickGenerator(-2), Error, "interval must be positive number",
```
bluong: Let's put a HACKHACK for the any-cast

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
...
-          "interval must be positive number");
+          "interval cannot be negatvie");
+      });
+
+      it.skip("rejects non-QuantitativeScale<number> Scales", () => {
```
bluong: Please explain the skip

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,62 @@ describe("Scales", () => {
...
+
+        let timeScale: any = new Plottable.Scales.Time();
+        (<any>assert).throws(() => Plottable.Scales.TickGenerators.intervalTickGenerator(1)(timeScale), Error,
+          "scale needs to inherit from Scale.QuantitativeScale<number>", "Scales.Time is not a valid parameter for TickGenerators");
       });
+
     });
 
-    describe("Integer", () => {
+    describe("integerTickGenerator()", () => {
```
bluong: Same as above about semantically talking about the test suite rather than through endpoints

test/scales/tickGeneratorsTests.ts
ztsai: I am not sure if we want to put HACKHACK for this as this is how we should do error assertion according to the testing best practice.  It would be everywhere

test/scales/tickGeneratorsTests.ts
ztsai: Do you mean using one interval value for the entire describe block?  
this aligns with our conversation regarding "randomness" of the environment setting, so that we kinda test for integer interval, interval less than one, interval greater than one, etc...

test/scales/tickGeneratorsTests.ts
bluong: This still applies

test/scales/tickGeneratorsTests.ts
bluong: Not necessarily one interval value for the entire block, but I would like to place these magic numbers outside in a variable in case we want to tweak these numbers when we modify the tests later.

test/scales/tickGeneratorsTests.ts
bluong: HACKHACK's are mainly used if there is a decently awkward workaround for something that should be inherent.  In this case, we should be very rarely any-casting and especially in this case, we are any-casting since our d.ts file does not have the proper typing definition.

I think I do agree that using this signature is a best practice, but usage of it is a HACKHACK until we fix the typings file.  In this case, we should definitely have a HACKHACK comment everywhere we see this.

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,64 @@ describe("Scales", () => {
...
       });
 
-      it("rejects negative values", () => {
+      it("rejects non-positive values", () => {
         (<any>assert).throws(() => Plottable.Scales.TickGenerators.intervalTickGenerator(0), Error, "interval must be positive number",
-          "interval must be positive number");
+          "interval cannot be 0");
         (<any>assert).throws(() => Plottable.Scales.TickGenerators.intervalTickGenerator(-2), Error, "interval must be positive number",
-          "interval must be positive number");
+          "interval cannot be negatvie");
```
bluong: "interval cannot be negative"

test/scales/tickGeneratorsTests.ts
```diff
@@ -22,35 +23,64 @@ describe("Scales", () => {
...
+      });
+
+      // HACKHACK: skipping failing test
+      // intervalTickGenerator() does not detect invalids scale
```
bluong: "does not detect invalid scale"

bluong: If we have a HACKHACK we should definitely mark it with an issue.  Otherwise, we will not be able to tell when we can remove the `skip`

PR #2723, opened 2015-09-04
[RenderControllerTests] - revamping

### LINKED ISSUE #2639: [Testing Audit] RenderController and RenderPolicies

bluong:
Mostly addition of non-existent tests.

Note that it is infeasible to test RenderPolicies since that would require very precise timing, which could itself be troublesome but it would also require a timeout in our tests.  Timeouts in our tests can be bad since test runners can take excruciatingly long if those tests fail.

Closes #2639

### DIFF
--- a/test/core/renderControllerTests.ts
+++ b/test/core/renderControllerTests.ts
@@ -1,8 +1,80 @@
 ///<reference path="../testReference.ts" />
 
 describe("RenderController", () => {
-  // HACKHACK: #2083
-  it.skip("Components whose render() is triggered by another Component's render() will be drawn", () => {
+  let SVG_WIDTH = 400;
+  let SVG_HEIGHT = 300;
+
+  describe("configuring the render policy", () => {
+    let oldWarn: (msg: string) => void;
+    let warned: boolean;
+
+    before(() => {
+      oldWarn = Plottable.Utils.Window.warn;
+      Plottable.Utils.Window.warn = () => {
+        warned = true;
+      };
+    });
+
+    beforeEach(() => {
+      warned = false;
+    });
+
+    after(() => {
+      Plottable.RenderController.renderPolicy(Plottable.RenderController.Policy.IMMEDIATE);
+      Plottable.Utils.Window.warn = oldWarn;
+    });
+
+    it("can set a render policy", () => {
+      let renderPolicy = Plottable.RenderController.Policy.TIMEOUT;
+      Plottable.RenderController.renderPolicy(renderPolicy);
+
+      assert.strictEqual(Object.getPrototypeOf(Plottable.RenderController.renderPolicy()),
+        Plottable.RenderPolicies.Timeout.prototype, "render policy is of the same type");
+    });
+
+    it("throws a warning for unrecognized render policies", () => {
+      let unrecognizedRenderPolicy = "foo";
+      Plottable.RenderController.renderPolicy(unrecognizedRenderPolicy);
+
+      assert.isTrue(warned, "warning sent for unrecognized render policy");
+    });
+  });
+
+  it("can queue a component to render", () => {
+    let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+    let component = new Plottable.Component();
+    let rendered = false;
+    component.renderImmediately = () => {
+      rendered = true;
+      return component;
+    };
+    component.anchor(svg);
+    Plottable.RenderController.registerToRender(component);
+    assert.isTrue(rendered, "component has rendered");
+    component.destroy();
+    svg.remove();
+  });
+
+  it("can queue a component to undergo layout computation and render", () => {
+    let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+    let component = new Plottable.Component();
+    let rendered = false;
+    component.renderImmediately = () => {
+      rendered = true;
+      return component;
+    };
+    component.anchor(svg);
+    Plottable.RenderController.registerToComputeLayout(component);
+    assert.isTrue(rendered, "component has rendered");
+    assert.deepEqual(component.origin(), {x: 0, y: 0}, "origin set");
+    assert.strictEqual(component.width(), SVG_WIDTH, "width set");
+    assert.strictEqual(component.height(), SVG_HEIGHT, "height set");
+    component.destroy();
+    svg.remove();
+  });
+
+  // HACKHACK: https://github.com/palantir/plottable/issues/2083
+  it.skip("can render components that are triggered by another component's render", () => {
     let link1 = new Plottable.Component();
     let svg1 = TestMethods.generateSVG();
     link1.anchor(svg1).computeLayout();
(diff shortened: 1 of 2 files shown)


jtlan:
Is there some way we could test `flush()`? Is it possible to set a `RenderPolicy` that doesn't render at all until `flush()` is called?

bluong:
Apparently, you can't set a `RenderPolicy` that is not one of the three known policies.

bluong:
Just noting that as a result of the above, we cannot set a `RenderPolicy` that does nothing.

test/core/renderControllerTests.ts
```diff
@@ -1,8 +1,80 @@
...
+    assert.isTrue(rendered, "component has rendered");
+    component.destroy();
+    svg.remove();
+  });
+
+  it("can queue a component to undergo layout computation and render", () => {
+    let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+    let component = new Plottable.Component();
+    let rendered = false;
+    component.renderImmediately = () => {
```
jtlan: Instead of replacing this method, can we assert that the `Component` has drawn something in the DOM?

bluong: Less performant, but probably more safe.  I'm thinking of making an element and then checking that the element exists

test/core/renderControllerTests.ts
```diff
@@ -1,8 +1,80 @@
...
+
+      assert.strictEqual(Object.getPrototypeOf(Plottable.RenderController.renderPolicy()),
+        Plottable.RenderPolicies.Timeout.prototype, "render policy is of the same type");
+    });
+
+    it("throws a warning for unrecognized render policies", () => {
+      let unrecognizedRenderPolicy = "foo";
+      Plottable.RenderController.renderPolicy(unrecognizedRenderPolicy);
+
+      assert.isTrue(warned, "warning sent for unrecognized render policy");
```
jtlan: Do we want to assert anything about the content of the warning message? That would require storing off the message.

bluong: Which part of the warning message do you think is important?  Whenever it comes to this specific portion, I get a bit nervous since if we decide to change the warning message, tests like this will fail.

jtlan: We have a similar best-practice in place [for error messages](https://github.com/palantir/plottable/wiki/Testing-best-practices#assert-error-throwing-should-include-the-error-message), so I feel like it's reasonable to do so for warnings too.

test/core/renderControllerTests.ts
```diff
@@ -1,8 +1,80 @@
...
+  let SVG_WIDTH = 400;
+  let SVG_HEIGHT = 300;
+
+  describe("configuring the render policy", () => {
+    let oldWarn: (msg: string) => void;
+    let storedWarningMsg: string;
+
+    before(() => {
+      oldWarn = Plottable.Utils.Window.warn;
+      Plottable.Utils.Window.warn = (warningMsg: string) => {
```
jtlan: This is only used in one test, but I see you're using the `before()`/`after()` block. While the `after()` block is good because you won't have a case where `warn` doesn't get restored (due to the test failing), it's odd because other tests in the same `describe()` are affected by this. @aicioara, do you have an opinion on this?

aicioara:
Sorry, reading through this. It is possible to test both flush() and timeouts.

```typescript
let oldFlush = RendererController.flush();
RendererController.flush = function() {
  // whatever assert
  oldFlush.call(this)
}
//whatever
RendererController.flush = oldFlush
```

```typescript
// Replacing timeout with instant invocation to avoid waiting for the deferred rendering
nativeTimeout = (<any>window).setTimeout;
(<any>window).setTimeout = function(fn: Function) {
  fn.call(this);
};
```

jtlan:
It sounds like we should maybe look into some sort of "spy" library for our tests?

test/core/renderControllerTests.ts
aicioara: This is getting scarier and scarier. On top of the above, can we do the following. 

In the `globalInitialization.ts`, in the `before()` make a copy of all the methods which we hack like this (warn, setTimeout, etc) and in the `after()` assert that we restored all the methods to their original state.

Regarding the other question, what exactly is the warning we are catching here?

bluong:
Slate used to use `Sinon`.  We can look into this, but I don't think it's in scope.

test/core/renderControllerTests.ts
aicioara: For completeness: Hacking warnings like that should be avoidable at all costs. We did that for deprecated methods, why is it needed here?

jtlan: The idea is to be able to assert something about the warning message, similar to [this best-practice](https://github.com/palantir/plottable/wiki/Testing-best-practices#assert-error-throwing-should-include-the-error-message) for `assert.throws()`.

bluong:
I think that testing flush in that manner is probably suboptimal compared to having a `RenderPolicy` that does nothing.  There might be some things that I want to test inbetween the render step and the registration step.  In this regard, testing should be more complete after #2729 is closed.

test/core/renderControllerTests.ts
jtlan: Can we make something like `TestMethods.assertWarns(function, warningMessage, assertMessage)`, similar to `assert.throws()`? That way we can replace `warn()` within the scope of that method only.

test/testMethods.ts
```diff
@@ -333,4 +333,13 @@ module TestMethods {
       return "#" + redHex + greenHex + blueHex;
   }
 
+  export function assertWarns(funct: Function, warningMessage: string, assertMessage: string) {
+    let receivedWarning = "";
+    let oldWarn = Plottable.Utils.Window.warn;
+    Plottable.Utils.Window.warn = (msg: string) => receivedWarning = msg;
+    funct();
+    assert.include(receivedWarning, warningMessage, assertMessage);
+    Plottable.Utils.Window.warn = oldWarn;
```
jtlan: This line needs to happen before the `assert`; otherwise, if the `assert` fails, `warn` is never restored.

test/testMethods.ts
```diff
@@ -333,4 +333,13 @@ module TestMethods {
       return "#" + redHex + greenHex + blueHex;
   }
 
+  export function assertWarns(funct: Function, warningMessage: string, assertMessage: string) {
```
aicioara: I prefer `function` or `fn` instead of `funct`. And then, if possible, invoke it with `funct.call(this)` instead of `funct()`. But this is just a nitpick

aicioara:
The `asssertWarns` method seems sound. Just double checking, is it thoroughly tested? Have you tried to call it on 1. an assert failing method 2. a JS failing method and see if the Utils.warn gets back to the original value? By looking at it should be fine, but we should be sure

bluong:
> The asssertWarns method seems sound. Just double checking, is it thoroughly tested? Have you tried to call it on 1. an assert failing method 2. a JS failing method and see if the Utils.warn gets back to the original value? By looking at it should be fine, but we should be sure

Should I open a file `testMethodsTests.ts`?

jtlan:
>Should I open a file testMethodsTests.ts?

Not sure if this winds up being excessively deep introspection.

bluong:
I just tested it myself offline and those cases seem to be handled.

jtlan:
Let's make a note to maybe have `TestMethods.Assert` as a sub-module later.

PR #2724, opened 2015-09-04
timeAxis size now includes annotation space

bluong:
`Axis.Time` has a bug where if annotations are enabled, then the axis would not allocate the space required for those annotations, resulting in a few weird bugs where margin is forced in order to show annotations but then the margin also appears, leaving the Axis in a mess...

### DIFF
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -305,8 +305,8 @@ describe("TimeAxis", () => {
     });
 
     it("includes the annotation space in the final size calculation", () => {
-      const SVG_WIDTH = 400;
-      const SVG_HEIGHT = 400;
+      let SVG_WIDTH = 400;
+      let SVG_HEIGHT = 400;
       let svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
       let xScale = new Plottable.Scales.Time();
       let xAxis = new Plottable.Axes.Time(xScale, "bottom");
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -305,7 +305,9 @@ describe("TimeAxis", () => {
     });
 
     it("includes the annotation space in the final size calculation", () => {
-      let svg = TestMethods.generateSVG(400, 400);
+      let svgWidth = 400;
+      let svgHeight = 400;
+      let svg = TestMethods.generateSVG(svgWidth, svgHeight);
       let xScale = new Plottable.Scales.Time();
       let xAxis = new Plottable.Axes.Time(xScale, "bottom");
       xAxis.margin(100);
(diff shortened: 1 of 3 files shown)


test/axes/timeAxisTests.ts
```diff
@@ -303,6 +303,24 @@ describe("TimeAxis", () => {
...
+      xAxis.annotationsEnabled(true);
+      xAxis.annotationTierCount(3);
+
+      xAxis.anchor(svg);
+      xAxis.computeLayout({ x: 0, y: 0}, 400, 400);
+      let coreHeight = xAxis.tickLabelPadding() + xAxis.innerTickLength();
+      let annotationHeight = xAxis.annotationTierCount() * (<any> xAxis)._annotationTierHeight();
+      let minimumHeight = coreHeight + xAxis.margin() + annotationHeight;
+      assert.operator(xAxis.height(), ">=", minimumHeight, "height includes all relevant pieces");
+      xAxis.destroy();
```
jtlan: Should we be doing this going forward? That'd certainly get more testing of `destroy()`.

test/axes/timeAxisTests.ts
```diff
@@ -303,6 +303,24 @@ describe("TimeAxis", () => {
...
+      let xScale = new Plottable.Scales.Time();
+      let xAxis = new Plottable.Axes.Time(xScale, "bottom");
+      xAxis.margin(100);
+      xAxis.annotationsEnabled(true);
+      xAxis.annotationTierCount(3);
+
+      xAxis.anchor(svg);
+      xAxis.computeLayout({ x: 0, y: 0}, 400, 400);
+      let coreHeight = xAxis.tickLabelPadding() + xAxis.innerTickLength();
+      let annotationHeight = xAxis.annotationTierCount() * (<any> xAxis)._annotationTierHeight();
```
jtlan: Not a huge fan of the `<any>`-cast here, but not sure how else to test this other than redoing the measurement. Could we draw three annotations and assert that they don't overlap?

test/axes/timeAxisTests.ts
bluong: I'm making it a habit to destroy all components that I create in a test.

test/axes/timeAxisTests.ts
```diff
@@ -303,6 +303,24 @@ describe("TimeAxis", () => {
...
+    it("includes the annotation space in the final size calculation", () => {
+      let svg = TestMethods.generateSVG(400, 400);
+      let xScale = new Plottable.Scales.Time();
+      let xAxis = new Plottable.Axes.Time(xScale, "bottom");
+      xAxis.margin(100);
+      xAxis.annotationsEnabled(true);
+      xAxis.annotationTierCount(3);
+
+      xAxis.anchor(svg);
+      xAxis.computeLayout({ x: 0, y: 0}, 400, 400);
```
jtlan: Factor out the `400`s here and on the line where `svg` is declared.

test/axes/timeAxisTests.ts
bluong: Not sure how seeing if the three annotations overlap really does much.  The main issue is that the visibility of the rectangles is hidden due to bad space calculations.

crmorford:
this isn't working at all in ff or safari for me :'( http://jsfiddle.net/86trca4u/30/

PR #2735, opened 2015-09-08
Horizontal bar plots no longer autoDomain wrongly

### LINKED ISSUE #2704: Horizontal Bar Plots + dataset.data() bad behavior
http://jsfiddle.net/ubwgm9cn/
repro:
1. open jsfiddle, note yScale domain
2. click on plot
3. yScale domain is now smaller

expected: yScale domain doesn't change when the callback is called
actual: yScale domain becomes smaller

also happens on stackedBar, clusteredBar

aicioara:
Closes #2704
The issue is the following. 

When calculating the extent for a scale that backs up the major scale of a bar plot, that scale needs to take into account the width of the bars. This is done in pixel space to account for log scales. Now, the issue is sometime the scales are having inverted domain (most common scenario, all yScales have the lowest number in the domain corresponding to the highest number in pixel such that the origin is bottom - left).

Solved this in the most generic / readable possible way.

**QE:** Please let me know if more information is required. I should not have changed anything except auto domain calculation code for bar plot (and clustered bar and stacked bar) 

Fiddle tagged on this.
Before: http://jsfiddle.net/o3rzuqur/5/
After: http://jsfiddle.net/o3rzuqur/4/

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -438,13 +438,13 @@ export module Plots {
       }
 
       let scale = <QuantitativeScale<any>>accScaleBinding.scale;
-      // console.log("aici", extents, scale.scale(extents[0][0]), scale.scale(extents[0][1]));
-
-      extents = extents.map((extent) => [
+      // To account for inverted domains
+      extents = extents.map((extent) => d3.extent([
         scale.invert(scale.scale(extent[0]) - this._barPixelWidth / 2),
-        scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2),
-      ]);
-      // console.log("acolo", extents);
+        scale.invert(scale.scale(extent[0]) + this._barPixelWidth / 2),
+        scale.invert(scale.scale(extent[1]) - this._barPixelWidth / 2),
+        scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2)
+      ]));
 
       return extents;
     }
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -438,6 +438,7 @@ export module Plots {
       }
 
       let scale = <QuantitativeScale<any>>accScaleBinding.scale;
+
       // To account for inverted domains
       extents = extents.map((extent) => d3.extent([
         scale.invert(scale.scale(extent[0]) - this._barPixelWidth / 2),
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -438,13 +438,13 @@ export module Plots {
       }
 
       let scale = <QuantitativeScale<any>>accScaleBinding.scale;
-      console.log("aici", extents, scale.scale(extents[0][0]), scale.scale(extents[0][1]));
+      // console.log("aici", extents, scale.scale(extents[0][0]), scale.scale(extents[0][1]));
 
       extents = extents.map((extent) => [
         scale.invert(scale.scale(extent[0]) - this._barPixelWidth / 2),
         scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2),
       ]);
-      console.log("acolo", extents);
+      // console.log("acolo", extents);
 
       return extents;
     }
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -438,11 +438,13 @@ export module Plots {
       }
 
       let scale = <QuantitativeScale<any>>accScaleBinding.scale;
+      console.log("aici", extents, scale.scale(extents[0][0]), scale.scale(extents[0][1]));
 
       extents = extents.map((extent) => [
         scale.invert(scale.scale(extent[0]) - this._barPixelWidth / 2),
         scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2),
       ]);
+      console.log("acolo", extents);
 
       return extents;
     }
--- a/test/plots/barPlotTests.ts
+++ b/test/plots/barPlotTests.ts
@@ -794,12 +794,8 @@ describe("Plots", () => {
       });
 
       it("pads the domain in the correct direction", () => {
-
-        let data = Array.apply(null, Array(10)).map(function(d: any, i: any) {
-          return {
-            x: i + 1,
-            y: i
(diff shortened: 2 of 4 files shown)


src/plots/barPlot.ts
```diff
@@ -439,10 +439,13 @@ export module Plots {
...
 
-      extents = extents.map((extent) => [
+      // To account for inverted domains
+      extents = extents.map((extent) => d3.extent([
         scale.invert(scale.scale(extent[0]) - this._barPixelWidth / 2),
-        scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2),
-      ]);
+        scale.invert(scale.scale(extent[0]) + this._barPixelWidth / 2),
+        scale.invert(scale.scale(extent[1]) - this._barPixelWidth / 2),
+        scale.invert(scale.scale(extent[1]) + this._barPixelWidth / 2)
```
bluong: Would a better write of the code be
```typescript
let maxPixelValue = Math.max(scale.scale(extent[0]), scale.scale(extent[1]));
let minPixelValue = Math.min(scale.scale(extent[0]), scale.scale(extent[1]));
return [minPixelValue - this._barPixelWidth / 2, maxPixelValue + this._barPixelWidth / 2];
```

test/plots/barPlotTests.ts
```diff
@@ -775,6 +775,61 @@ describe("Plots", () => {
...
+        xScale = new Plottable.Scales.Linear();
+        yScale = new Plottable.Scales.Linear();
+
+        plot = new Plottable.Plots.Bar<number, number>("horizontal");
+        plot.x((d) => d.x, xScale);
+        plot.y((d) => d.y, yScale);
+      });
+
+      it("pads the domain in the correct direction", () => {
+
```
bluong: Don't think the newline here is necessary.  NBD though.

test/plots/barPlotTests.ts
```diff
@@ -775,6 +775,61 @@ describe("Plots", () => {
...
+        let data = Array.apply(null, Array(10)).map(function(d: any, i: any) {
```
bluong: The function declaration here isn't necessarily, let's stick with the "=>" notation

test/plots/barPlotTests.ts
```diff
@@ -775,6 +775,61 @@ describe("Plots", () => {
...
+        plot.addDataset(new Plottable.Dataset(data));
+        plot.renderTo(svg);
+
+        assert.operator(yScale.domain()[0], "<", data[0].y, "lower end of the domain is padded");
+        assert.operator(yScale.domain()[1], ">", data[data.length - 1].y, "higher end of the domain is padded");
+
+        svg.remove();
+      });
+
+      it("has the same size before and after autoDomain()-ing", () => {
```
bluong: Would a more appropriate title be that calling autoDomain right after render does not change the domain?

src/plots/barPlot.ts
aicioara: The issue with yours is that you have to `Scale.invert()` your values and re-sort them again.

Hence I prefer my way.

bluong: Yup, your write I think is better.

test/plots/barPlotTests.ts
```diff
@@ -775,6 +775,53 @@ describe("Plots", () => {
...
+        svg.remove();
+      });
+
+      it("computes the correct extent when autoDomain()-ing right after render", () => {
+        let data = Array.apply(null, Array(10)).map((d: any, i: number) => {
+          return { x: i + 1, y: i + 1 };
+        });
+        plot.addDataset(new Plottable.Dataset(data));
+
+        plot.renderTo(svg);
```
bluong: Very minor, in the above, `renderTo` is called without the newline but here there is a newline.  Please make consistent?  I think no newline is ok here?

test/plots/barPlotTests.ts
```diff
@@ -775,6 +775,53 @@ describe("Plots", () => {
...
+      let yScale: Plottable.Scales.Linear;
+      let plot: Plottable.Plots.Bar<number, number>;
+
+      beforeEach(() => {
+        svg = TestMethods.generateSVG();
+
+        xScale = new Plottable.Scales.Linear();
+        yScale = new Plottable.Scales.Linear();
+
+        plot = new Plottable.Plots.Bar<number, number>("horizontal");
```
bluong: Let's use the string enum instead of the raw string.

crmorford:
tested on this too http://jsfiddle.net/o3rzuqur/6/, http://jsfiddle.net/o3rzuqur/7/, http://jsfiddle.net/o3rzuqur/8/
(toggle right side plots with '1' and '2' keys)

src/plots/barPlot.ts
crmorford: Checking to see if I understand.
1. Extent is what you would get by autodomaining on any other type of plot.
2. It used to worked/works on vertical plots because extent[0] is always <= extent[1] (assumption), so you could subtract 1/2 * barPixelWidth from extent[0], and add it to extent[1], and get a strictly wider range
4. So on horizontal, can extent[1] be less than extent[0]? So you account for that by always subtracting and adding from both extent[0] and extent[1]?

bluong: > 1. Extent is what you would get by autodomaining on any other type of plot.

Similar enough I think.  A plot gathers information about extents just in general, but this is most useful for autodomaining, since autodomaining will look at extents to determine the final domain.

> 2. It used to worked/works on vertical plots because extent[0] is always <= extent[1] --(assumption), so you could subtract 1/2 * barPixelWidth from extent[0], and add it to extent[1], and get a strictly wider range

I think the extents themselves were not really the problem but the scaled version of the extents.  For example, let's take Andrei's Before JSFiddle.  The yAxis domain is increasing starting from the bottom.  This means the scaled version of extent[0] is actually greater in value than the scaled version of extent[1] (since y counts increasing downwards)

> 3. So on horizontal, can extent[1] be less than extent[0]? So you account for that by always subtracting and adding from both extent[0] and extent[1]?

This was not necessarily the problem.  It's mainly that the assumptions about the scale were incorrect.

PR #2736, opened 2015-09-09
Horizontal Bar Plot hides labels that are partially cut off in y

### LINKED ISSUE #2658: Faulty label hiding on horizontal bars
`Plots.Bar` in `"horizontal"` mode hides bar labels even though it would still be possible to draw them:
screen shot 2015-08-20 at 10 43 18 am
**repro**: http://jsfiddle.net/ejzaqdcq/

In other cases, it fails to hide the label when it should:
screen shot 2015-08-20 at 11 08 30 am
**repro**: http://jsfiddle.net/ejzaqdcq/1/

yunhaolucky:
**Issue**
In the horizontal bar plot, a label should be hidden if the bounding box is inside the chart area. The issue is sometimes a label is visible when there's space to show it while sometime it is partially cut off in y. 

**Fix**
The `labelPosition.y` is not set correctly in the `_drawLabel` function. I add one line to set  `labelPosition.y` to the center of the bar.

**JSFiddle**
*Case1*: 
Before: http://jsfiddle.net/ejzaqdcq/
screen shot 2015-09-10 at 11 14 25 am

After: http://jsfiddle.net/pnqm5hrb/1/
screen shot 2015-09-10 at 11 14 34 am

*Case2*:
Before: http://jsfiddle.net/ejzaqdcq/1/
screen shot 2015-09-10 at 11 13 55 am

After:http://jsfiddle.net/54rf5to1/1/
screen shot 2015-09-10 at 11 14 07 am

Close #2658.

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -452,7 +452,7 @@ export module Plots {
       let labelsTooWide = false;
       this.datasets().forEach((dataset) => labelsTooWide = labelsTooWide || this._drawLabel(dataToDraw.get(dataset), dataset));
       if (this._hideBarsIfAnyAreTooWide && labelsTooWide) {
-         this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
+        this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
       }
     }
 
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -452,7 +452,7 @@ export module Plots {
       let labelsTooWide = false;
       this.datasets().forEach((dataset) => labelsTooWide = labelsTooWide || this._drawLabel(dataToDraw.get(dataset), dataset));
       if (this._hideBarsIfAnyAreTooWide && labelsTooWide) {
-        this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
+         this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
       }
     }
 
@@ -489,7 +489,6 @@ export module Plots {
           } else {
             x += offset;
           }
-
           let showLabel = true;
           let labelPosition = {
             x: x,
@@ -499,6 +498,7 @@ export module Plots {
           if (this._isVertical) {
             labelPosition.x = x + w / 2 - measurement.width / 2;
           } else {
+            labelPosition.y = y + h / 2 - measurement.height / 2;
             if (!positive) {
               labelPosition.x = x + w - measurement.width;
             } else {
--- a/test/plots/barPlotTests.ts
+++ b/test/plots/barPlotTests.ts
@@ -782,7 +782,7 @@ describe("Plots", () => {
 
         texts.each(function(d, i) {
           let textBounding = (<Element> this).getBoundingClientRect();
-          let svgBounding = (<Element> svg[0][0]).getBoundingClientRect();
+          let svgBounding = (<Element> barPlot.background().node()).getBoundingClientRect();
           let isLabelCutOff = (textBounding.top < svgBounding.top && textBounding.bottom > svgBounding.top)
             || (textBounding.top < svgBounding.bottom && textBounding.bottom > svgBounding.bottom);
           assert.isTrue(isLabelCutOff, `label ${i} is partially cut off`);
--- a/test/plots/barPlotTests.ts
+++ b/test/plots/barPlotTests.ts
@@ -724,20 +724,20 @@ describe("Plots", () => {
 
     describe("Horizontal Bar Plot label visibility", () => {
       let svg: d3.Selection<void>;
-      let yScale: Plottable.Scales.Category;
+      let yScale: Plottable.Scales.Linear;
       let xScale: Plottable.Scales.Linear;
-      let barPlot: Plottable.Plots.Bar<number, string>;
+      let barPlot: Plottable.Plots.Bar<number, number>;
       beforeEach(() => {
         svg = TestMethods.generateSVG(600, 400);
-        yScale = new Plottable.Scales.Category().domain(["A", "B"]);
+        yScale = new Plottable.Scales.Linear();
         xScale = new Plottable.Scales.Linear();
 
         let data = [
-          {y: "A", x: -1.5},
-          {y: "B", x: 1},
+          {y: 1, x: -1.5},
+          {y: 2, x: 1},
         ];
 
-        barPlot = new Plottable.Plots.Bar<number, string>(Plottable.Plots.Bar.ORIENTATION_HORIZONTAL);
+        barPlot = new Plottable.Plots.Bar<number, number>(Plottable.Plots.Bar.ORIENTATION_HORIZONTAL);
         barPlot.addDataset(new Plottable.Dataset(data));
         barPlot.x((d) => d.x, xScale);
         barPlot.y((d) => d.y, yScale);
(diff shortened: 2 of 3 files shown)


src/plots/barPlot.ts
```diff
@@ -452,7 +452,7 @@ export module Plots {
       let labelsTooWide = false;
       this.datasets().forEach((dataset) => labelsTooWide = labelsTooWide || this._drawLabel(dataToDraw.get(dataset), dataset));
       if (this._hideBarsIfAnyAreTooWide && labelsTooWide) {
-        this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
+         this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
```
bluong: Extra space here it seems.  Please remove.

As a note, indentation standards currently are 2 spaces per indent.

bluong:
Please write a pull request description that would be useful to people that glance over the pull request.

Things that are likely to be useful include:
 * API Changes (Usually additions)
 * A description of what was broken
 * A description of the fix
 * A Proof-Of-Concept JSFiddle
 * A Proof-Of-Concept Picture

As a general note of the description, I personally like to try to keep things starting from most general to more detailed as the pull request goes on.  If this requires a TL;DR, then I would put that at the top I think (I am mostly thinking about QE here since sometimes a description can get to be really code-wordy which is not their concern)

test/plots/barPlotTests.ts
```diff
@@ -724,20 +724,20 @@ describe("Plots", () => {
...
         xScale = new Plottable.Scales.Linear();
 
         let data = [
-          {y: "A", x: -1.5},
-          {y: "B", x: 1},
+          {y: 1, x: -1.5},
+          {y: 2, x: 1},
         ];
 
-        barPlot = new Plottable.Plots.Bar<number, string>(Plottable.Plots.Bar.ORIENTATION_HORIZONTAL);
```
bluong: Why were the above changes needed?

test/plots/barPlotTests.ts
```diff
@@ -773,6 +773,20 @@ describe("Plots", () => {
         assert.include(["visible", "inherit"], label2.style("visibility"), "label 1 is visible");
         svg.remove();
       });
+
+      it("hides labels that are partially cut off in y", () => {
+        yScale.domain([1, 2]);
+        let texts = svg.selectAll("text");
```
bluong: Instead of selecting inside the `svg`, please select inside of `barPlot.content()`.  It's just a lot more precise I feel.

test/plots/barPlotTests.ts
```diff
@@ -773,6 +773,20 @@ describe("Plots", () => {
...
+
+        assert.strictEqual(texts.size(), 2, "There should be two labels rendered");
+
+        let label1 = d3.select(texts[0][0]);
+        let label2 = d3.select(texts[0][1]);
+
+        assert.strictEqual(label1.style("visibility"), "hidden", "label 1 is not visible");
+        assert.strictEqual(label2.style("visibility"), "hidden", "label 2 is not visible");
```
bluong: Can we have more information on why these labels are invisible?  From what I gather the labels are not visible because the y scale places them outside the top and bottom of the area.  Can we instead use a `forEach` loop, check through the DOM attributes, and then assert the visibility style based on the DOM attributes?

bluong:
The change mostly looks good.  We just need to do some documentation cleanup so that we have an idea what's going on in the future :smiley:

test/plots/barPlotTests.ts
yunhaolucky: We change the test because we need to set the domain so that the label is partially cut off.

test/plots/barPlotTests.ts
```diff
@@ -773,6 +773,23 @@ describe("Plots", () => {
         assert.include(["visible", "inherit"], label2.style("visibility"), "label 1 is visible");
         svg.remove();
       });
+
+      it("hides labels that are partially cut off in y", () => {
+        yScale.domain([1, 2]);
+        let texts = barPlot.content().selectAll("text");
+
+        assert.strictEqual(texts.size(), 2, "There should be two labels rendered");
```
bluong: Can we replace the 2 to be aligned with our input?

Maybe count the number of datum we have.

@jtlan However this does mean that if the number of texts is 0, everything would pass.  Should we also assert that `texts.size()` is greater than 0?

test/plots/barPlotTests.ts
```diff
@@ -773,6 +773,23 @@ describe("Plots", () => {
...
+
+        texts.each(function(d, i) {
+          let textBounding = (<Element> this).getBoundingClientRect();
+          let svgBounding = (<Element> barPlot.background().node()).getBoundingClientRect();
+          let isLabelCutOff = (textBounding.top < svgBounding.top && textBounding.bottom > svgBounding.top)
+            || (textBounding.top < svgBounding.bottom && textBounding.bottom > svgBounding.bottom);
+          assert.isTrue(isLabelCutOff, `label ${i} is partially cut off`);
+          assert.strictEqual(d3.select(this).style("visibility"), "hidden", `label ${i} is not visible`);
```
bluong: This idea seems a big fragile?  if the data now contains even one label that should be visible, the test will now fail.  I do agree that we definitely want to scout out cases where the label is partially cut off, but I am unsure if we can assume that the input data will contain only cut off labels.

This also makes it a bit awkward if we wanted to add a new test, especially to test for visible labels.

bluong: Followup comment since I missed the `domain` setter call above which is a significant part of this test.

Now I currently see why these labels would be cut off.  My main concern is that if `data` were to change in the `beforeEach` loop then this test would be guaranteed to fail since some labels would be visible.  Would it be possible to select only selections that are partially cut off?

test/plots/barPlotTests.ts
```diff
@@ -773,6 +773,23 @@ describe("Plots", () => {
         assert.include(["visible", "inherit"], label2.style("visibility"), "label 1 is visible");
         svg.remove();
       });
+
+      it("hides labels that are partially cut off in y", () => {
+        yScale.domain([1, 2]);
```
bluong: Can we have this domain call based on the data?

bluong:
Mainly issues with trying to make the test code less fragile while still making it relevant.  @jtlan feel free to chime in if you think these changes are necessary

test/plots/barPlotTests.ts
jtlan: What if there were no selections that were partially cut off? Here we assert that the initial conditions for the test are valid, which I think is fair because that's part of the "correctness" of the test.

test/plots/barPlotTests.ts
jtlan: I think this can be done with `data.length`, but really a bigger fix should be done when the test file is refactored.

bluong:
Looking at this again, this seems like code that would be acceptable before we finish all the refactors over our test code.  Given this, as long as this code gets refactored at the later date, I think we should be fine.

crmorford:
Such a good summary! Thank you :)
Starting QE

crmorford:
testing summary (so that Yun knows what kind of stuff I look at):
dragged the bars off the top/bottom/right/left to make sure panning/zooming off all edges still looks good on vertical plots:
http://jsfiddle.net/54rf5to1/2/

dragged the bars off the top/bottom/right/left to make sure panning/zooming off all edges now looks good on horizontal plots:
http://jsfiddle.net/54rf5to1/3/

stackedBar, and clusteredBar are descended from bar (waterfall is also, but doesn't have a horizontal orientation), so I want to make sure everything looks good there:
stacked: http://jsfiddle.net/54rf5to1/4/

**this comment is a work in progress**

aicioara:
Congrats @yunhaolucky

PR #2745, opened 2015-09-10
detach DragInteraction on DragBoxLayer.destroy()

### LINKED ISSUE #2553: .destroy doesn't .destroy dragBoxLayers
http://jsfiddle.net/fee6p9w0/6/
after clicking the destroy button, you can still drag and call the callback, even though the box doesn't show up any more.

ztsai:
closes #2553 
detach all dragInteractions on a DragBoxLayer when it's destroyed or detached

destroyed:
original: http://jsfiddle.net/fee6p9w0/6/
fixed: http://jsfiddle.net/fee6p9w0/16/
note that after clicking destroy button, there's still console log indicating interaction callback being fired in the original fiddle when drag events happen.

detached:
original: http://jsfiddle.net/fee6p9w0/24/
fixed: http://jsfiddle.net/fee6p9w0/25/
after clicking `detach` button there shouldn't be any console log while dragging, clicking `attach` will bring interaction and dbl back.

### DIFF
--- a/src/components/dragBoxLayer.ts
+++ b/src/components/dragBoxLayer.ts
@@ -437,6 +437,19 @@ export module Components {
       this._disconnectInteraction();
     }
 
+    public detach() {
+      this.bounds({
+        topLeft: { x: 0, y: 0 },
+        bottomRight: { x: 0, y: 0 }
+      });
+      this._dragInteraction.detachFrom(this);
+      return super.detach();
+    }
+
+    public anchor(selection: d3.Selection<void>) {
+      this._dragInteraction.attachTo(this);
+      return super.anchor(selection);
+    }
   }
 }
 }
--- a/src/components/dragBoxLayer.ts
+++ b/src/components/dragBoxLayer.ts
@@ -438,7 +438,7 @@ export module Components {
     }
 
     public detach() {
-      this._resetBound();
+      this._resetState();
       this._dragInteraction.detachFrom(this);
       return super.detach();
     }
@@ -448,7 +448,7 @@ export module Components {
       return super.anchor(selection);
     }
 
-    private _resetBound() {
+    private _resetState() {
       this.bounds({
         topLeft: { x: 0, y: 0 },
         bottomRight: { x: 0, y: 0 }
--- a/src/components/dragBoxLayer.ts
+++ b/src/components/dragBoxLayer.ts
@@ -438,10 +438,7 @@ export module Components {
     }
 
     public detach() {
-      this.bounds({
-        topLeft: { x: 0, y: 0 },
-        bottomRight: { x: 0, y: 0 }
-      });
+      this._resetBound();
       this._dragInteraction.detachFrom(this);
       return super.detach();
     }
@@ -449,6 +446,13 @@ export module Components {
     public anchor(selection: d3.Selection<void>) {
       this._dragInteraction.attachTo(this);
       return super.anchor(selection);
+    }
+
+    private _resetBound() {
+      this.bounds({
+        topLeft: { x: 0, y: 0 },
+        bottomRight: { x: 0, y: 0 }
+      });
     }
   }
 }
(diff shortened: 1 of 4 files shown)


aicioara:
Short description of what has been done is always welcome :)

src/components/dragBoxLayer.ts
```diff
@@ -416,6 +416,12 @@ export module Components {
       this._setMovableClass();
       return this;
     }
+
+    public destroy() {
```
aicioara: Can we quickly check that this bug does not affect the `guideLineLayer` as well?

src/components/dragBoxLayer.ts
```diff
@@ -416,6 +416,12 @@ export module Components {
...
+      super.destroy();
+      this._dragInteraction.detachFrom(this);
+    }
+
```
aicioara: Let's do exactly what the `dragLineLayer` is doing for completeness and consistency.

aicioara:
```typescript
        this._dragInteraction.offDragStart(interactionDragStartCallback);
        this._dragInteraction.offDrag(interactionDragCallback);
        this._dragInteraction.offDragEnd(interactionDragEndCallback);
        this._dragInteraction.detachFrom(this);

        this._dragStartCallbacks.forEach((callback) => this._dragStartCallbacks.delete(callback));
        this._dragCallbacks.forEach((callback) => this._dragCallbacks.delete(callback));
        this._dragEndCallbacks.forEach((callback) => this._dragEndCallbacks.delete(callback));
```

test/components/dragBoxLayerTests.ts
```diff
@@ -106,6 +106,26 @@ describe("Interactive Components", () => {
...
+        let group = new Plottable.Components.Group([dbl]).renderTo(svg);
+        let target = group.background();
+        let onDragStartCallbackCalled = false;
+        let onDragCallbackCalled = false;
+        let onDragEndcallbackCalled = false;
+        dbl.onDragStart(() => onDragStartCallbackCalled = true);
+        dbl.onDrag(() => onDragCallbackCalled = true);
+        dbl.onDragEnd(() => onDragEndcallbackCalled = true);
+
+        dbl.destroy();
```
aicioara: assert that the callbacks were initially called

test/components/dragBoxLayerTests.ts
```diff
@@ -106,6 +106,26 @@ describe("Interactive Components", () => {
         svg.remove();
       });
 
+      it("does not call callbacks after destroy() is called", () => {
+        let group = new Plottable.Components.Group([dbl]).renderTo(svg);
```
aicioara: why do we need the group?

ztsai: so that we can fake the mouse event somewhere (dbl.content() will be removed after destroy())

aicioara: Can we add a comment which explains this? It was not clear to me why is the group needed

aicioara:
Apart from that it looks good.

aicioara:
Oh, the fiddle is gone :(

ztsai:
added comment and fixed fiddle

crmorford:
Test Destroy() for Dragging/Resizing/Moving on DragBox/XDragBox/YDragBox
x: http://jsfiddle.net/fee6p9w0/17/
y: http://jsfiddle.net/fee6p9w0/18/
normal: http://jsfiddle.net/fee6p9w0/19/

Can we fix this for Detach(), too? Let me know if this is actually out of scope. http://jsfiddle.net/fee6p9w0/20/

ztsai:
it's probably outside of the scope of the original issue, but I think it will make sense to fix it together
it seems like Detach() doesn't work on YDragBox as well

ztsai:
QE notes above

aicioara:
Apart from that and the bounds question it looks good.

src/components/dragBoxLayer.ts
```diff
@@ -416,6 +428,32 @@ export module Components {
...
+      this._dragInteraction.detachFrom(this);
+      return super.detach();
+    }
+
+    public anchor(selection: d3.Selection<void>) {
+      this._dragInteraction.attachTo(this);
+      return super.anchor(selection);
+    }
+
+    private _resetBound() {
```
aicioara: Can we call it `_resetState()`, as ideally, we can make more resets here. 

By doing so, we make it more obvious that reseting the bounds is done for this purpose (resetting the state). Otherwise setting both bounds to 0 and the method `_resetBound()` are synonyms, so we don't get any benefit.

PR #2749, opened 2015-09-10
change the background-fill pointer-events

### LINKED ISSUE #2647: Cursor Styles only show on top dragbox layer
I'm not sure if there's actually a case for this monstrosity...
but if you layer multiple dragbox layers, the little cursor indicators for grabbing only show up on the 'top' layer. 
http://jsfiddle.net/fa3mjzew/5/
1. make a drag box
2. hover over green section and see the grabby hand cursor
3. hover anywhere else -- no grabby hand

yunhaolucky:
**Issue** 
When there are several components in a group, the cursor style only show on the top component.

**Fix**
This issue happens because the `background-fill` in the `background` of the top component is blocking the `pointer-events` to lower components. I fix this by adding `pointer-events:none` to `background-fill` elements in css.

There still might be problems that `content` instead of `background`  is blocking the pointer-events, i.e. if we put a `label` on top of a `dragLineLayer`, the cursor style of `dragLineLayer` will not show correctly. But this is beyond the scope of this issue.

**JSFiddle**
A component is in front of the dragLineLayer.
Before:http://jsfiddle.net/7Lxzacoz/
After:http://jsfiddle.net/tm2oot96/1/

close #2647

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -46,6 +46,7 @@ svg.plottable {
 
 .plottable .background-fill {
   fill: none;
+  pointer-events: none;
 }
 
 .plottable .bounding-box {
--- a/test/components/componentTests.ts
+++ b/test/components/componentTests.ts
@@ -344,7 +344,7 @@ describe("Component behavior", () => {
     c.renderTo(svg);
     let backgroundFill = c.background().select(".background-fill").node();
     let pointerEvent = window.getComputedStyle(<Element>backgroundFill).pointerEvents;
-    assert.strictEqual(pointerEvent, "none", "background-fill's pointer-event is not set to none");
+    assert.strictEqual(pointerEvent, "none", "background-fill's pointer-event is set to none");
     svg.remove();
   });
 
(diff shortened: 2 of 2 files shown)


bluong:
Do you know if there's a way to test the current status of a cursor?  If so, can we have a small test regarding this?

If finding the status of a cursor is too annoying, this looks fine.

bluong:
Can we add a small unit test to assert that the computed style of a `background-fill`does not respond to pointer-events?

Essentially that (backgroundFillSelection.node().getComputedStyle("pointer-event") is "none")

crmorford:
tests:
http://jsfiddle.net/tm2oot96/2/
http://jsfiddle.net/tm2oot96/3/
http://jsfiddle.net/tm2oot96/4/
http://jsfiddle.net/tm2oot96/5/

Looks good on chrome, ff, safari

I don't have access to IE now. @yunhaolucky, can you verify that on this test, you can see the cursor over all the lines and the dragbox?  http://jsfiddle.net/tm2oot96/5/
@jtlan can show you where the IE computer is.

jtlan:
@yunhaolucky and I tested on IE9, it works.

PR #2752, opened 2015-09-10
Deprecation documentation in JSDocs

### LINKED ISSUE #2750: Formatters.RelativeDate - Warns deprecation, API doesnt say so
When you instantiate a Formatters.RelativeDate, it prints a warning to console saying that it's deprecated. However, the API docs don't say this, so someone reading http://plottablejs.org/docs/modules/plottable.formatters.html may think it's fine to depend on.

As an aside, a relative date formatter seems useful to have, does it really make sense to deprecate it due to a bug? As opposed to just fixing the bug? My impression is you deprecate things because they are the wrong api, not because they have a bad implementation. Or is the API for RelativeDate formatter fundamentally unworkable because they don't get the TZ info from js date objects? I never really grokked time zones.

aicioara:
Closes #2750

- Added better deprecated documentation for deprecated methods
- Synced the versioning number between existing comments and the `deprecate()` method
- Added missing deprecation messages

### DIFF
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -244,6 +244,8 @@ export module Formatters {
   }
 
   /**
+   * @deprecated As of release 1.3.0, not safe for use with time zones.
+   *
    * Creates a formatter for relative dates.
    *
    * @param {number} baseValue The start date (as epoch time) used in computing relative dates (default 0)
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -244,7 +244,7 @@ export module Formatters {
   }
 
   /**
-   * @deprecated As of release 1.3.0, not safe for use with time zones.
+   * @deprecated As of release v1.3.0, not safe for use with time zones.
    *
    * Creates a formatter for relative dates.
    *
@@ -255,7 +255,7 @@ export module Formatters {
    * @returns {Formatter} A formatter for time/date values.
    */
   export function relativeDate(baseValue: number = 0, increment: number = MILLISECONDS_IN_ONE_DAY, label: string = "") {
-    Plottable.Utils.Window.deprecated("relativeDate()", "1.3", "Not safe for use with time zones.");
+    Plottable.Utils.Window.deprecated("relativeDate()", "v1.3.0", "Not safe for use with time zones.");
     return (d: any) => {
       let relativeDate = Math.round((d.valueOf() - baseValue) / increment);
       return relativeDate.toString() + label;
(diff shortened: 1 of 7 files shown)


src/core/formatters.ts
```diff
@@ -244,6 +244,8 @@ export module Formatters {
   }
 
   /**
+   * @deprecated As of release 1.3.0, not safe for use with time zones.
+   *
```
bluong: Let's make the deprecated call take "1.3.0"?

src/plots/barPlot.ts
```diff
@@ -313,8 +313,11 @@ export module Plots {
       return closest;
     }
 
+    /**
+     * @deprecated As of release 1.1.0, replaced by _entityVisibleOnPlot()
+     */
     protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
-      Utils.Window.deprecated("Bar._visibleOnPlot()", "v1.1.0");
+      Utils.Window.deprecated("Bar._visibleOnPlot()", "v1.1.0", "replaced by _entityVisibleOnPlot()");
```
bluong: Should there be a "v" in front of the version number?

src/plots/plot.ts
```diff
@@ -544,8 +544,11 @@ export class Plot extends Component {
     return this._lightweightPlotEntityToPlotEntity(closestPointEntity);
   }
 
+  /**
+   * @deprecated As of release 1.1.0, replaced by _entityVisibleOnPlot()
+   */
   protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
-    Utils.Window.deprecated("Plot._visibleOnPlot()", "v1.1.0");
+    Utils.Window.deprecated("Plot._visibleOnPlot()", "v1.1.0", "replaced by _entityVisibleOnPlot()");
```
bluong: Same as above

src/plots/scatterPlot.ts
```diff
@@ -93,8 +93,11 @@ export module Plots {
       return drawSteps;
     }
 
+    /**
+     * @deprecated As of release 1.1.0, replaced by _entityVisibleOnPlot()
+     */
     protected _visibleOnPlot(datum: any, pixelPoint: Point, selection: d3.Selection<void>): boolean {
-      Utils.Window.deprecated("Scatter._visibleOnPlot()", "v1.1.0");
+      Utils.Window.deprecated("Scatter._visibleOnPlot()", "v1.1.0", "replaced by _entityVisibleOnPlot()");
```
bluong: Same as above

aicioara:
[NOQE] Only comments and warning messages have changed.

@bluong thoughts?

PR #2753, opened 2015-09-10
update ignore file to ignore changed vsc filename

bluong:
As listed here
https://code.visualstudio.com/updates ("Renamed VS Code Settings Folder" section)

Visual Studio Code has renamed its configuration directory to be .vscode

Ignoring this to not see annoying git messages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,5 @@ quicktests/overlaying/list_of_quicktests.json
 *.project
 
 # Visual Studio Code
+.settings/
 .vscode/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -30,5 +30,4 @@ quicktests/overlaying/list_of_quicktests.json
 *.project
 
 # Visual Studio Code
-.settings/
-
+.vscode/
```
aicioara: Let's leave both for a while. Does vsc update automatically? Maybe not everyone wants to have the latest version.

bluong: vsc lets you know at the top if a new version comes out.  That said, I don't see the harm in keeping both.

bluong: Also noting that vsc will change the name of the file automatically at the upgrade step.

PR #2755, opened 2015-09-11
Refactor key interaction tests - Pass 1

aicioara:
Part of #2628

As part of this: 
- No nested `beforeEach()`es
- Cleaned the console
  screen shot 2015-09-11 at 12 35 35 pm

### DIFF
--- a/test/interactions/keyInteractionTests.ts
+++ b/test/interactions/keyInteractionTests.ts
@@ -107,7 +107,7 @@ describe("Interactions", () => {
         svg.remove();
       });
 
-      it("can remove only the specified keypress callbacks", () => {
+      it("removes only the specified callback", () => {
         let aKeyCodeCallback1Called = false;
         let aKeyCodeCallback1 = () => aKeyCodeCallback1Called = true;
         let aKeyCodeCallback2Called = false;
@@ -145,7 +145,7 @@ describe("Interactions", () => {
         svg.remove();
       });
 
-      it("does not fire the callback if the key is pressed and held down outside, then the mouse moved inside", () => {
+      it("does not fire the callback when component is initially out of focus", () => {
         keyInteraction.onKeyPress(aKeyCode, aKeyCodeCallback);
         keyInteraction.attachTo(component);
 
@@ -179,6 +179,20 @@ describe("Interactions", () => {
         component.renderTo(svg);
       });
 
+      it("fires the callback upon releasing the key", () => {
+        keyInteraction.onKeyRelease(aKeyCode, aKeyCodeCallback);
+        keyInteraction.attachTo(component);
+
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode);
+        assert.isFalse(aKeyCodeCallbackCalled, "callback for \"a\" was not called before releasing the key");
+        TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aKeyCode);
+        assert.isTrue(aKeyCodeCallbackCalled, "callback for \"a\" was called when \"a\" key was released");
+
+        keyInteraction.offKeyRelease(aKeyCode, aKeyCodeCallback);
+        svg.remove();
+      });
+
       it("doesn't fire callback if key was released without being pressed", () => {
         assert.strictEqual(keyInteraction.onKeyRelease(aKeyCode, aKeyCodeCallback), keyInteraction,
           "setting the keyRelease callback returns the interaction");
@@ -191,18 +205,6 @@ describe("Interactions", () => {
         svg.remove();
       });
 
-      it("fires callback if key was released after being pressed", () => {
-        keyInteraction.onKeyRelease(aKeyCode, aKeyCodeCallback);
-        keyInteraction.attachTo(component);
-
-        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
-        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode);
-        TestMethods.triggerFakeKeyboardEvent("keyup", component.background(), aKeyCode);
-        assert.isTrue(aKeyCodeCallbackCalled, "callback for \"a\" was called when \"a\" key was released");
-
-        keyInteraction.offKeyRelease(aKeyCode, aKeyCodeCallback);
-        svg.remove();
-      });
 
       it("only fires callback for key that has been released", () => {
         let bKeyCode = 66;
--- a/test/interactions/keyInteractionTests.ts
+++ b/test/interactions/keyInteractionTests.ts
@@ -1,32 +1,32 @@
 ///<reference path="../testReference.ts" />
 
 describe("Interactions", () => {
-  describe("KeyInteraction", () => {
-    let svg: d3.Selection<void>;
-    let component: Plottable.Component;
-    let keyInteraction: Plottable.Interactions.Key;
-    let aCode: number;
-    let bCode: number;
-
-    let aCallbackCalled: boolean;
-    let bCallbackCalled: boolean;
-    let aCallback: Plottable.KeyCallback;
-    let bCallback: Plottable.KeyCallback;
-
-    beforeEach(() => {
-      aCode = 65;
-      bCode = 66;
-      component = new Plottable.Component();
-      svg = TestMethods.generateSVG(400, 400);
-      keyInteraction = new Plottable.Interactions.Key();
-      aCallback = () => aCallbackCalled = true;
-      bCallback = () => bCallbackCalled = true;
-      component.renderTo(svg);
-      aCallbackCalled = false;
-      bCallbackCalled = false;
-    });
-
+  describe("Key Interaction", () => {
     describe("onKeyPress", () => {
+      let svg: d3.Selection<void>;
+      let component: Plottable.Component;
+      let keyInteraction: Plottable.Interactions.Key;
+      let aCode: number;
+      let bCode: number;
+
+      let aCallbackCalled: boolean;
+      let bCallbackCalled: boolean;
+      let aCallback: Plottable.KeyCallback;
+      let bCallback: Plottable.KeyCallback;
+
+      beforeEach(() => {
+        aCode = 65;
+        bCode = 66;
+        component = new Plottable.Component();
+        svg = TestMethods.generateSVG(400, 400);
+        keyInteraction = new Plottable.Interactions.Key();
+        aCallback = () => aCallbackCalled = true;
+        bCallback = () => bCallbackCalled = true;
+        component.renderTo(svg);
+        aCallbackCalled = false;
+        bCallbackCalled = false;
+      });
+
       it("only fires callback for \"a\" when \"a\" key is pressed", () => {
         keyInteraction.onKeyPress(aCode, aCallback);
         keyInteraction.onKeyPress(bCode, bCallback);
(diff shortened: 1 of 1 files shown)


test/interactions/keyInteractionTests.ts
```diff
@@ -1,34 +1,35 @@
...
+      let svg: d3.Selection<void>;
+      let component: Plottable.Component;
+      let keyInteraction: Plottable.Interactions.Key;
+      let aCode: number;
+      let bCode: number;
+
+      let aCallbackCalled: boolean;
+      let bCallbackCalled: boolean;
+      let aCallback: Plottable.KeyCallback;
+      let bCallback: Plottable.KeyCallback;
```
bluong: Why we need the `b` version of these version

test/interactions/keyInteractionTests.ts
```diff
@@ -75,7 +77,7 @@ describe("Interactions", () => {
         svg.remove();
       });
```
bluong: The "can remove keyPress callbacks" should assert that the function returns the calling object.

test/interactions/keyInteractionTests.ts
```diff
@@ -75,7 +77,7 @@ describe("Interactions", () => {
...
 
-      it("multiple keyPress callbacks are possible", () => {
+      it("can attach multiple multiple keyPress callbacks", () => {
```
bluong: double "multiple"?  Remove one of them

bluong: If we have this, what's the use of `bCallback`?

test/interactions/keyInteractionTests.ts
bluong: Can we have better names for these variables instead of a/b?

test/interactions/keyInteractionTests.ts
```diff
@@ -149,14 +151,40 @@ describe("Interactions", () => {
...
+      let svg: d3.Selection<void>;
+      let component: Plottable.Component;
+      let keyInteraction: Plottable.Interactions.Key;
+      let aCode: number;
+      let bCode: number;
+
+      let aCallbackCalled: boolean;
+      let bCallbackCalled: boolean;
+      let aCallback: Plottable.KeyCallback;
+      let bCallback: Plottable.KeyCallback;
```
bluong: Same as above concerns.

test/interactions/keyInteractionTests.ts
```diff
@@ -230,7 +258,7 @@ describe("Interactions", () => {
         svg.remove();
       });
 
-      it("removing keyRelease callbacks is possible", () => {
+      it("can remove removing keyRelease callbacks", () => {
         keyInteraction.onKeyRelease(aCode, aCallback);
```
bluong: We should assert that calling this function returns the calling object.

bluong:
Main concern being the usage of the `aCallback` and `bCallback` variables.  I think it would make more sense to have a single callback at the describe level and then declare the other variables as needed in the unit tests.

test/interactions/keyInteractionTests.ts
aicioara: I do that already in the first `it`

test/interactions/keyInteractionTests.ts
aicioara: `bCallback` is for key "b"

test/interactions/keyInteractionTests.ts
aicioara: That is being asserted in the previous "it". Does it make sense to only assert it here?

aicioara:
Fixed the `aCallback` `bCallback` issue

test/interactions/keyInteractionTests.ts
aicioara: I actually moved it here now as it makes more sense.

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
+        component = new Plottable.Component();
+        keyInteraction = new Plottable.Interactions.Key();
+        aKeyCode = 65;
+        aKeyCodeCallbackCalled = false;
+        aKeyCodeCallback = () => aKeyCodeCallbackCalled = true;
+        svg = TestMethods.generateSVG();
+        component.renderTo(svg);
+      });
+
       it("only fires callback for \"a\" when \"a\" key is pressed", () => {
```
bluong: This does not sounds like the type of test that should be at the front of this test suite.

Instead, can we have something like
it("fires callback for the a key code when that key is pressed)
as the first test and then have
it("does not fire callbacks for key codes other than the pressed key")
as an edge case second test

bluong: As a result we can move the setter assertion up to this new first test here.

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
-        keyInteraction.onKeyPress(aCode, aCallback);
-        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode);
-        assert.isTrue(aCallbackCalled, "callback for \"a\" was properly connected back to the interaction");
+        keyInteraction.onKeyPress(aKeyCode, aKeyCodeCallback);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode);
+        assert.isTrue(aKeyCodeCallbackCalled, "callback for \"a\" was properly connected back to the interaction");
 
-        keyInteraction.offKeyPress(aCode, aCallback);
+        assert.strictEqual(keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback), keyInteraction,
+          "unsetting the keyPress callback returns the interaction");
```
bluong: It seems that this test is essentially divided into two parts...  What is the second part trying to test?

Also it seems like the last line here is testing that the setter is chainable; shouldn't this assert be made higher up?

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
         svg.remove();
       });
 
-      it("multiple keyPress callbacks are possible", () => {
-        let aCallback1Called = false;
-        let aCallback1 = () => aCallback1Called = true;
-        let aCallback2Called = false;
-        let aCallback2 = () => aCallback2Called = true;
+      it("can attach multiple keyPress callbacks", () => {
```
bluong: let's be more descriptive and append
" to the same key code" or " to the same key"

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
+        assert.isTrue(aKeyCodeCallback2Called, "callback 2 for \"a\" was called when \"a\" key was pressed");
 
-        keyInteraction.offKeyPress(aCode, aCallback1);
-        keyInteraction.offKeyPress(aCode, aCallback2);
+        keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback1);
+        keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback2);
         svg.remove();
       });
 
       it("can remove only one of the registered keypress callbacks", () => {
```
bluong: Maybe rename to
it("removes only the specified callback")?

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
+        aKeyCodeCallbackCalled = false;
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode, { repeat : true });
+        assert.isFalse(aKeyCodeCallbackCalled, "callback for \"a\" was not called when keydown was fired the second time");
 
-        keyInteraction.offKeyPress(aCode, aCallback);
+        keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback);
         svg.remove();
       });
 
       it("does not fire the callback if the key is pressed and held down outside, then the mouse moved inside", () => {
```
bluong: This sentence is quite a bit long and a bit nondescriptive ("outside... what?")

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,327 @@
...
+        aKeyCode = 65;
+        component = new Plottable.Component();
+        keyInteraction = new Plottable.Interactions.Key();
+        aKeyCodeCallbackCalled = false;
+        aKeyCodeCallback = () => aKeyCodeCallbackCalled = true;
+        svg = TestMethods.generateSVG();
+        component.renderTo(svg);
+      });
+
       it("doesn't fire callback if key was released without being pressed", () => {
```
bluong: Same as above how we should just have a base unit test about setting the `onKeyRelease` callback

bluong:
Few test naming issues and we should add in the base unit tests.

test/interactions/keyInteractionTests.ts
aicioara: Moved it higher up. It was a mistake to make this assert here. The test was basically checking that the callback can be removed and reattahed

bluong: Can we put the reattach in a different test?

test/interactions/keyInteractionTests.ts
aicioara: Fixed

test/interactions/keyInteractionTests.ts
aicioara: I think that is a bit overkill. The reattach test will basically copy + paste this entire testsuite. I totally agree with the modular approach, but the benefits of having this vs the extra test + time to run the testsuite is not worth it, hence I suggest we leave it like this

test/interactions/keyInteractionTests.ts
aicioara: FIxed

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,340 @@
...
-      bCallbackCalled = false;
-    });
-
+  describe("Key Interaction", () => {
     describe("onKeyPress", () => {
+      let svg: d3.Selection<void>;
+      let component: Plottable.Component;
+      let keyInteraction: Plottable.Interactions.Key;
+
+      let aKeyCode: number;
```
bluong: Technically this should be `const`ed, or we can even capitalize case it.  What do you think?

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,340 @@
...
+      it("fires callback for key press", () => {
+        assert.strictEqual(keyInteraction.onKeyPress(aKeyCode, aKeyCodeCallback), keyInteraction,
+          "setting the keyPress callback returns the interaction");
+        keyInteraction.attachTo(component);
+
+        TestMethods.triggerFakeMouseEvent("mouseover", component.background(), 100, 100);
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode);
+        assert.isTrue(aKeyCodeCallbackCalled, "callback for \"a\" was called when \"a\" key was pressed");
+
+        keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback);
```
bluong: Why this line?

test/interactions/keyInteractionTests.ts
bluong: Do you think it would be a simpler yet sufficient test if we instead just test if "bCallback" is not called if "a" is pressed?  What I mean by this is not to touch the setup for "a" at all.

test/interactions/keyInteractionTests.ts
```diff
@@ -1,302 +1,340 @@
...
         keyInteraction.attachTo(component);
 
         TestMethods.triggerFakeMouseEvent("mouseout", component.background(), -100, -100);
-        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aCode);
-        assert.isFalse(aCallbackCalled, "callback for \"a\" was not called when not moused over the Component");
+        TestMethods.triggerFakeKeyboardEvent("keydown", component.background(), aKeyCode);
+        assert.isFalse(aKeyCodeCallbackCalled, "callback for \"a\" was not called when not moused over the Component");
 
-        keyInteraction.offKeyPress(aCode, aCallback);
+        keyInteraction.offKeyPress(aKeyCode, aKeyCodeCallback);
```
bluong: Why this line?

bluong:
The above new comments pervade through the file in multiple locations so I jsut commented on the first occurrence.

test/interactions/keyInteractionTests.ts
aicioara: Fixed

aicioara:
Since can't continue on this branch anymore, moving to the fork.

PR #2756, opened 2015-09-11
[SelectionBoxLayer] - remove errors on getters

### LINKED ISSUE #2734: [XDragBoxLayer / YDragBoxLayer] - xScale / yScale getters should not error
In dealing with these classes, whenever a developer is using the parent class `DragBoxLayer`, they would have to tiptoe very carefully in that using either of these getters would error depending on which subclass a user is using.

This is incredibly disruptive, and should return null for the appropriate circumstance.

Mainly the fault of the poor exception handling of Javascript and a somewhat awkward class structure, but errors I think are definitely not the way to go here.

bluong:
Having the getters on a class error can be incredibly disruptive, and so this pull request seeks to remove these from the appropriate getters.  This is disruptive since the variables in association with the getters are private, which means that if one just wants to query if the variable exists, then javascript would error.

This is highly relevant on classes like `SelectionBoxLayer`, which depending on the situation might want to query the `xScale()` or `yScale()` and then operate on it if it is not null.

Closes #2734

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -4640,9 +4640,6 @@ declare module Plottable {
             yScale<D extends number | {
                 valueOf(): number;
             }>(yScale: QuantitativeScale<D>): SelectionBoxLayer;
-            yExtent(): (number | {
-                valueOf(): number;
-            })[];
         }
     }
 }
@@ -4667,9 +4664,6 @@ declare module Plottable {
             xScale<D extends number | {
                 valueOf(): number;
             }>(xScale: QuantitativeScale<D>): SelectionBoxLayer;
-            xExtent(): (number | {
-                valueOf(): number;
-            })[];
         }
     }
 }
--- a/plottable.js
+++ b/plottable.js
@@ -12255,12 +12255,9 @@ var Plottable;
             };
             XDragBoxLayer.prototype.yScale = function (yScale) {
                 if (yScale == null) {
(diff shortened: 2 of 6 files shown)


jtlan:
What do we expect to be returned for `xExtent()` and `yExtent()`?

bluong:
The same as what we expect on `SelectionBoxLayer` if no scale is set and if no extent is set, which currently should be an undefined array - `[undefined, undefined]`

crmorford:
The dragbox isn't appearing anymore after i've set the extent here: http://jsfiddle.net/ewga1des/19/
It used to: http://jsfiddle.net/ewga1des/11/
I'm using the values .1 and .4

bluong:
@crmorford Note that you are targeting 2 sibling branches.  This pull request was based off of develop and thus does not have the setter for the extents yet.

crmorford:
Why doesn't it error when I call .xExtent and .yExtent, then?

crmorford:
tested on x, y, normal, selection by tweaking this http://jsfiddle.net/ewga1des/22/
no scales set, called after destroy() 
all applicable scales set, called after destroy() 
all applicable scales set, called after detach()
no scales set, called after detach()
all applicable scales set, no detach() or destroy()
no scales set, no detach() or destroy()

crmorford:
for the unit tests, can you change the test to make sure that now there is **no** error thrown on getting xScale, yScale, instead of deleting the test?

PR #2765, opened 2015-09-12
Refactor dispatcher tests - Pass 1

aicioara:
Part of #2628

As part of this: 
- No nested `beforeEach()`es
- Cleaned the console
  screen shot 2015-09-11 at 6 08 33 pm


bluong:
The base `Dispatcher` class is incredibly odd in that the public constructor is not meant to be used, resulting in a class with no public API.

Thus, we have 2 ways that we can properly test this.  One is to make a child test class from `Dispatcher` that has a public API that hits the appropriate API from the parent class.

Another way is to use an already existing class and then test the protected endpoints by hitting the public endpoints in that existing class.

In essence, let's avoid the anycasting.

aicioara:
I'll go for the mock idea.

test/dispatchers/dispatcherTests.ts
```diff
@@ -2,72 +2,100 @@
...
-      (<any> dispatcher)._connect();
-      callbackCalls = 0;
-      TestMethods.triggerFakeUIEvent("click", d3document);
-      assert.strictEqual(callbackCalls, 1, "can't double-connect (callback only called once)");
-
-      (<any> dispatcher)._disconnect();
+    let callbackCalls = 0;
+    let eventCallbackCalls = 0;
+    let callback: () => void;
+    let d3document: d3.Selection<void>;
```
bluong: `d3Document`?

test/dispatchers/dispatcherTests.ts
```diff
@@ -2,72 +2,100 @@
...
+      TestMethods.triggerFakeUIEvent("click", d3document);
+            assert.strictEqual(eventCallbackCalls, 1, "eventCallback is still called");
+      assert.strictEqual(callbackCalls, 1, "callback is called exactly once");
+      dispatcher.unsetCallback(callbackSet, callback2);
     });
   });
 });
+
+module Mocks {
+  export class BasicDispatcher extends Plottable.Dispatcher {
```
bluong: The main thing I worry about is that this might be disconnected from the actual child class implementations.  Do you think we should work the the mock here or do you think we should use a different method?

bluong: If `Dispatcher` was more of an `interface` or an `abstract class`, we would experience less of this problem...

test/dispatchers/dispatcherTests.ts
```diff
@@ -2,72 +2,100 @@
...
-      (<any> dispatcher)._callbacks = [callbackSet];
+      dispatcher.unsetCallback(callbackSet, callback);
+      callbackCalls = 0;
+      callbackSet.callCallbacks();
+      assert.strictEqual(callbackCalls, 0, "callback is not after removed by calling _unsetCallback()");
+    });
 
-      let d3document = d3.select(document);
-      (<any> dispatcher)._connect();
+    it("connects dispatcher to the Dom when callback is set", () => {
```
bluong: DOM or dom

Dom looks like a name when it really is an acronym

bluong:
We need to figure out the `Mocks` answer.  Otherwise we can iterate on this pretty quickly.

ztsai:
talked to @bluong offline, decided to go with black box approach
it's up to subclasses of `Plottable.Dispatcher` to test for all these cases
will open a follow up issue for things to check

PR #2766, opened 2015-09-14
Show labels that overlaps with tick marks when `tier label position` is set to `center`. 

### LINKED ISSUE #2725: [Axes.Time] tierLabelPositions() does not work correctly
Setting any tier label position to `"center"` using `tierLabelPositions()` causes the labels on that tier to disappear.

Repro: http://jsfiddle.net/beph2ohn/

**Actual**: No labels are shown (they get hidden). Extra space does not appear to be allocated for the labels.
**Expected**: Labels are centered underneath the tick marks.

Note that this only happens for tiers that are set to `"center"`; other tiers are unaffected. (ex: http://jsfiddle.net/beph2ohn/1/, http://jsfiddle.net/beph2ohn/2/).

yunhaolucky:
**Issue**
Setting any tier label position to `center` using tierLabelPositions() causes the labels on that tier to disappear.

**fix**
This happens because in function hideOverlappingAndCutOffLabels, labels that overlap with tick marks in horizontal dimension are hidden. Now the label is hidden only if it overlaps with tick marks in two dimensions.

:exclamation: This fix will cause problems that all labels will be hidden when axis is in "top" orientation.  

**JSFiddle**
Before: http://jsfiddle.net/beph2ohn/
screen shot 2015-09-14 at 9 25 18 am

After: http://jsfiddle.net/beph2ohn/3/ 
screen shot 2015-09-14 at 9 25 08 am
This does not work on firefox in my computer. 

Close #2725

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -1203,18 +1203,18 @@ var Plottable;
         Formatters.fixed = fixed;
         /**
          * Creates a formatter that formats numbers to show no more than
-         * [maxNumberOfDecimalPlaces] decimal places. All other values are stringified.
+         * [precision] decimal places. All other values are stringified.
          *
-         * @param {number} [maxNumberOfDecimalPlaces] The number of decimal places to show (default 3).
+         * @param {number} [precision] The number of decimal places to show (default 3).
          *
          * @returns {Formatter} A formatter for general values.
          */
-        function general(maxNumberOfDecimalPlaces) {
-            if (maxNumberOfDecimalPlaces === void 0) { maxNumberOfDecimalPlaces = 3; }
-            verifyPrecision(maxNumberOfDecimalPlaces);
+        function general(precision) {
+            if (precision === void 0) { precision = 3; }
+            verifyPrecision(precision);
             return function (d) {
                 if (typeof d === "number") {
-                    var multiplier = Math.pow(10, maxNumberOfDecimalPlaces);
+                    var multiplier = Math.pow(10, precision);
                     return String(Math.round(d * multiplier) / multiplier);
                 }
                 else {
@@ -1254,17 +1254,17 @@ var Plottable;
         }
         Formatters.percentage = percentage;
         /**
-         * Creates a formatter for values that displays [numberOfSignificantFigures] significant figures
+         * Creates a formatter for values that displays [precision] significant figures
          * and puts SI notation.
          *
-         * @param {number} [numberOfSignificantFigures] The number of significant figures to show (default 3).
+         * @param {number} [precision] The number of significant figures to show (default 3).
          *
          * @returns {Formatter} A formatter for SI values.
          */
-        function siSuffix(numberOfSignificantFigures) {
-            if (numberOfSignificantFigures === void 0) { numberOfSignificantFigures = 3; }
-            verifyPrecision(numberOfSignificantFigures);
-            return function (d) { return d3.format("." + numberOfSignificantFigures + "s")(d); };
+        function siSuffix(precision) {
+            if (precision === void 0) { precision = 3; }
+            verifyPrecision(precision);
+            return function (d) { return d3.format("." + precision + "s")(d); };
         }
         Formatters.siSuffix = siSuffix;
         /**
--- a/plottable.js
+++ b/plottable.js
@@ -4428,7 +4428,7 @@ var Plottable;
                 var xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
                 var yTranslate = this.orientation() === "bottom" ?
                     d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-                    this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+                    this.height() - this.innerTickLength() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
                 var textSelection = tickLabels.selectAll("text");
                 if (textSelection.size() > 0) {
                     Plottable.Utils.DOM.translate(textSelection, xTranslate, yTranslate);
@@ -4561,8 +4561,8 @@ var Plottable;
                     var leadingTickMark = visibleTickMarkRects[i];
                     var trailingTickMark = visibleTickMarkRects[i + 1];
                     var isOverlappingLastLabel = (lastLabelClientRect != null && Plottable.Utils.DOM.clientRectsOverlap(clientRect, lastLabelClientRect));
-                    var isOverlappingLeadingTickMark = (leadingTickMark && Plottable.Utils.DOM.clientRectsOverlap(clientRect, leadingTickMark));
-                    var isOverlappingTrailingTickMark = (trailingTickMark && Plottable.Utils.DOM.clientRectsOverlap(clientRect, trailingTickMark));
+                    var isOverlappingLeadingTickMark = (leadingTickMark != null && Plottable.Utils.DOM.clientRectsOverlap(clientRect, leadingTickMark));
+                    var isOverlappingTrailingTickMark = (trailingTickMark != null && Plottable.Utils.DOM.clientRectsOverlap(clientRect, trailingTickMark));
                     if (!isInsideBBox(clientRect) || isOverlappingLastLabel || isOverlappingLeadingTickMark || isOverlappingTrailingTickMark) {
                         tickLabel.style("visibility", "hidden");
                     }
--- a/plottable.js
+++ b/plottable.js
@@ -4428,7 +4428,7 @@ var Plottable;
                 var xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
                 var yTranslate = this.orientation() === "bottom" ?
                     d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-                    this.height() - this.innerTickLength() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+                    this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
                 var textSelection = tickLabels.selectAll("text");
                 if (textSelection.size() > 0) {
                     Plottable.Utils.DOM.translate(textSelection, xTranslate, yTranslate);
--- a/plottable.js
+++ b/plottable.js
@@ -1203,18 +1203,18 @@ var Plottable;
         Formatters.fixed = fixed;
         /**
          * Creates a formatter that formats numbers to show no more than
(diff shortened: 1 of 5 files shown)


bluong:
Few pull request description notes:
* Title is long and at the same time nondescriptive.  I personally like putting the relevant class in square brackets but there should be some identifier in the title about what object you are targeting.  Also, the word `hideOverlappingAndCutOffLabels` only has meaning to developers that are well versed in the `timeAxis` code.  I would avoid that word entirely and find a more semantic way of titling the pull request.
* "This does not work on firefox in my computer. " This sounds like something to investigate...

src/axes/timeAxis.ts
```diff
@@ -547,7 +548,8 @@ export module Axes {
         let leadingTickMark = visibleTickMarkRects[i];
         let trailingTickMark = visibleTickMarkRects[i + 1];
         if (!isInsideBBox(clientRect) || (lastLabelClientRect != null && Utils.DOM.clientRectsOverlap(clientRect, lastLabelClientRect))
-            || (leadingTickMark.right > clientRect.left || trailingTickMark.left < clientRect.right)) {
+            || (tierLabelPositions[i] === "between"
+            && (leadingTickMark.right > clientRect.left || trailingTickMark.left < clientRect.right))) {
```
bluong: Instead of this fix, would it be better to check for collision using `leadingTickMark.top` / `leadingTickMark.bottom` / `trailingTickMark.top` / `trailingTickMark.bottom`?

Also if this expression gets too large, feel free to extract into a variable.

src/axes/timeAxis.ts
```diff
@@ -546,8 +546,12 @@ export module Axes {
         let tickLabel = d3.select(this);
         let leadingTickMark = visibleTickMarkRects[i];
         let trailingTickMark = visibleTickMarkRects[i + 1];
-        if (!isInsideBBox(clientRect) || (lastLabelClientRect != null && Utils.DOM.clientRectsOverlap(clientRect, lastLabelClientRect))
-            || (leadingTickMark.right > clientRect.left || trailingTickMark.left < clientRect.right)) {
+
+        let isOverlappingLastLabel = (lastLabelClientRect != null && Utils.DOM.clientRectsOverlap(clientRect, lastLabelClientRect));
+        let isOverlappingLeadingTickMark = (leadingTickMark != null && Utils.DOM.clientRectsOverlap(clientRect, leadingTickMark));
+        let isOverlappingTrailingTickMark = (trailingTickMark != null && Utils.DOM.clientRectsOverlap(clientRect, trailingTickMark));
```
bluong: We should be consistent about using a value in an && statement or != null statement when checking existence. I think I prefer the latter since we have had trouble in the past with using the raw value.

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,34 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels show correctly when display format is set to 'center'", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    scale.domain([new Date("2015-09-02"), new Date("2015-09-03")]);
+    axis.tierLabelPositions(["center", "center"]);
+    axis.renderTo(svg);
+
+    let labels = axis.content().selectAll("." + Plottable.Axis.TICK_LABEL_CLASS);
```
bluong: Let's use the `${}` feature here

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,34 @@ describe("TimeAxis", () => {
...
+    assert.isFalse(labels.size() === 0, "More than one labels are selected in testing");
```
bluong: Let's use `assert.notStrictEquals`

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,34 @@ describe("TimeAxis", () => {
...
+          Math.floor(axisBoundingRect.left) <= Math.ceil(innerRect.left) &&
+          Math.floor(axisBoundingRect.top) <= Math.ceil(innerRect.top) &&
+          Math.floor(innerRect.right) <= Math.ceil(axisBoundingRect.right) &&
+          Math.floor(innerRect.bottom) <= Math.ceil(axisBoundingRect.bottom)
+        );
+    };
+
+    labels.each(function(d, i) {
+      let labelVisibility = window.getComputedStyle(this).visibility;
+      let bounding = this.getBoundingClientRect();
```
bluong: Let's change this variable name to like
`clientRect` or `boundingClientRect`?

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,34 @@ describe("TimeAxis", () => {
...
   });
 
+  it("tick labels show correctly when display format is set to 'center'", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    scale.domain([new Date("2015-09-02"), new Date("2015-09-03")]);
+    axis.tierLabelPositions(["center", "center"]);
+    axis.renderTo(svg);
+
+    let labels = axis.content().selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`);
+    assert.notStrictEqual(labels.size(), 0, "More than one labels are selected in testing");
```
bluong: I think I messed up in my review.  Especially since the message reports information about using "more than one labels", the assert should reflect as such.

Thus, can we change this to...
`assert.operator(labels.size(), ">", 1, msg)`

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,34 @@ describe("TimeAxis", () => {
...
+          Math.floor(innerRect.right) <= Math.ceil(axisBoundingRect.right) &&
+          Math.floor(innerRect.bottom) <= Math.ceil(axisBoundingRect.bottom)
+        );
+    };
+
+    labels.each(function(d, i) {
+      let labelVisibility = window.getComputedStyle(this).visibility;
+      let boundingClientRect = this.getBoundingClientRect();
+      let isInside = isInsideAxisBoundingRect(boundingClientRect);
+      assert.isFalse(isInside && (labelVisibility === "hidden"),  `Wrong labels ${i} are ${labelVisibility}`);
```
bluong: Can we have a better assert message here?  If the test were to fail, the test would result to something like
"Wrong labels 3 are hidden."

This isn't really that helpful.

Let's take a step similar to the previous pull request and check the relationship between `isInside` and `labelVisibility === "hidden"` to get a better assert.

As a general rule of thumb, `assert.isTrue` / `assert.isFalse` is generally not at all useful unless testing for the simple cases.

bluong:
Test issues

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,38 @@ describe("TimeAxis", () => {
...
   });
 
+  it("tick labels show correctly when display format is set to 'center'", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    scale.domain([new Date("2015-09-02"), new Date("2015-09-03")]);
+    axis.tierLabelPositions(["center", "center"]);
+    axis.renderTo(svg);
+
+    let labels = axis.content().selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`);
+    assert.operator(labels.size(), ">", 0, "At least one label is selected in testing");
```
bluong: If the message is "At least one label"..., let's change the assert to
`..., ">=", 1, ...`?

plottable.js
```diff
@@ -1,5 +1,5 @@
 /*!
-Plottable 1.10.0 (https://github.com/palantir/plottable)
+Plottable 1.11.0 (https://github.com/palantir/plottable)
```
bluong: This diff shouldn't be occurring.

src/axes/timeAxis.ts
```diff
@@ -394,7 +394,7 @@ export module Axes {
       let xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
       let yTranslate = this.orientation() === "bottom" ?
           d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-          this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+          this.height() - this.innerTickLength() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
```
bluong: Why was this change made?

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,38 @@ describe("TimeAxis", () => {
...
+        );
+    };
+
+    labels.each(function(d, i) {
+      let labelVisibility = window.getComputedStyle(this).visibility;
+      let boundingClientRect = this.getBoundingClientRect();
+      let isInside = isInsideAxisBoundingRect(boundingClientRect);
+      if (labelVisibility === "hidden") {
+        assert.isFalse(isInside,  `label ${i} "${this.textContent}" is hidden, should be visible as it is inside the axis bounding box`);
+      }else {
```
bluong: There should be a space before the `else` here

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,38 @@ describe("TimeAxis", () => {
...
+        assert.isTrue(isInside,  `label ${i} "${this.textContent}" is visible, should be hidden as it is outside the axis bounding box`);
```
bluong: Seems to be a double space in front of "`label "

src/axes/timeAxis.ts
yunhaolucky: Previously, when calculating the label position in `top` orientation, tick mark's height is not considered.  This error makes labels overlaps with tick marks, which make `hidelabel` function fails. Therefore, I change it here.

bluong: This sounds like a separate bug and the fix does not necessarily seem correct (Why not `_maxLabelTickLength()`?)  Let's make a separate pull request for this with the appropriate bugs?

test/axes/timeAxisTests.ts
```diff
@@ -294,6 +294,38 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels show correctly when display format is set to 'center'", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    scale.domain([new Date("2015-09-02"), new Date("2015-09-03")]);
```
aicioara: Sorry for jumping in so late. This dates are not formatted properly. That might be why Firefox was behaving differently before.

Chrome
screen shot 2015-09-15 at 2 41 28 pm

Firefox
screen shot 2015-09-15 at 2 41 34 pm

Proper way of doing it
```typescript
new Date(2015, 08, 02)
```

yunhaolucky:
Cassie said we should create a separate pull request for top orientation issue. I remove that part.

crmorford:
http://jsfiddle.net/beph2ohn/4/ (compares side by side)

Issue #2396, opened 2015-06-25
Updating scale does not auto domain


aicioara:
Expected: http://jsfiddle.net/peedyvze/2/
Actual: http://jsfiddle.net/peedyvze/1/

Explanation: Swapping the scale from a linear scale to a new linear scale should have no effect?

jtlan:
Not only that, it appears to not attach the new `Scale` at all: http://jsfiddle.net/x7x05L4k/

PR #2767, opened 2015-09-14
Updating scale should auto domain

ztsai:
fix auto-domaining of scale on bind for Plots.Plot
auto-range scale on set for Plots.XYPlot
#2396

before: http://jsfiddle.net/x7x05L4k/
after: http://jsfiddle.net/x7x05L4k/1/

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -155,19 +155,23 @@ export class Plot extends Component {
   }
 
   protected _bindProperty(property: string, value: any, scale: Scale<any, any>) {
-    this._bind(property, value, scale, this._propertyBindings, this._propertyExtents);
-    this._updateExtentsForProperty(property);
+    let updateExtentsForKey = () => this._updateExtentsForProperty(property);
+    this._bind(property, value, scale, this._propertyBindings, this._propertyExtents, updateExtentsForKey);
   }
 
   private _bindAttr(attr: string, value: any, scale: Scale<any, any>) {
-    this._bind(attr, value, scale, this._attrBindings, this._attrExtents);
-    this._updateExtentsForAttr(attr);
+    let updateExtentsForKey = () => this._updateExtentsForAttr(attr);
+    this._bind(attr, value, scale, this._attrBindings, this._attrExtents, updateExtentsForKey);
   }
 
   private _bind(key: string, value: any, scale: Scale<any, any>,
-                    bindings: d3.Map<Plots.AccessorScaleBinding<any, any>>, extents: d3.Map<any[]>) {
+                bindings: d3.Map<Plots.AccessorScaleBinding<any, any>>, extents: d3.Map<any[]>,
+                updateExtentsForKey: (key: string) => void) {
     let binding = bindings.get(key);
     let oldScale = binding != null ? binding.scale : null;
+
+    bindings.set(key, { accessor: d3.functor(value), scale: scale });
+    updateExtentsForKey(key);
 
     if (oldScale != null) {
       this._uninstallScaleForKey(oldScale, key);
@@ -175,8 +179,6 @@ export class Plot extends Component {
     if (scale != null) {
       this._installScaleForKey(scale, key);
     }
-
-    bindings.set(key, { accessor: d3.functor(value), scale: scale });
   }
 
   protected _generateAttrToProjector(): AttributeToProjector {
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -155,29 +155,32 @@ export class Plot extends Component {
   }
 
   protected _bindProperty(property: string, value: any, scale: Scale<any, any>) {
-    let updateExtentsForKey = () => this._updateExtentsForProperty(property);
-    this._bind(property, value, scale, this._propertyBindings, this._propertyExtents, updateExtentsForKey);
+    let binding = this._propertyBindings.get(property);
+    let oldScale = binding != null ? binding.scale : null;
+
+    this._propertyBindings.set(property, { accessor: d3.functor(value), scale: scale });
+    this._updateExtentsForProperty(property);
+
+    if (oldScale != null) {
+      this._uninstallScaleForKey(oldScale, property);
+    }
+    if (scale != null) {
+      this._installScaleForKey(scale, property);
+    }
   }
 
   private _bindAttr(attr: string, value: any, scale: Scale<any, any>) {
-    let updateExtentsForKey = () => this._updateExtentsForAttr(attr);
-    this._bind(attr, value, scale, this._attrBindings, this._attrExtents, updateExtentsForKey);
-  }
-
-  private _bind(key: string, value: any, scale: Scale<any, any>,
-                bindings: d3.Map<Plots.AccessorScaleBinding<any, any>>, extents: d3.Map<any[]>,
-                updateExtentsForKey: (key: string) => void) {
-    let binding = bindings.get(key);
+    let binding = this._attrBindings.get(attr);
     let oldScale = binding != null ? binding.scale : null;
 
-    bindings.set(key, { accessor: d3.functor(value), scale: scale });
-    updateExtentsForKey(key);
+    this._attrBindings.set(attr, { accessor: d3.functor(value), scale: scale });
+    this._updateExtentsForAttr(attr);
 
     if (oldScale != null) {
-      this._uninstallScaleForKey(oldScale, key);
+      this._uninstallScaleForKey(oldScale, attr);
     }
     if (scale != null) {
-      this._installScaleForKey(scale, key);
+      this._installScaleForKey(scale, attr);
     }
   }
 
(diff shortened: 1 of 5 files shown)


test/plots/plotTests.ts
```diff
@@ -133,6 +133,26 @@ describe("Plots", () => {
 
     });
 
+    it("sets the scale domain automatically", () => {
```
jtlan: Can we specify that this happens when `attr()` is called?

src/plots/plot.ts
```diff
@@ -155,28 +155,30 @@ export class Plot extends Component {
...
 
   private _bind(key: string, value: any, scale: Scale<any, any>,
-                    bindings: d3.Map<Plots.AccessorScaleBinding<any, any>>, extents: d3.Map<any[]>) {
+                bindings: d3.Map<Plots.AccessorScaleBinding<any, any>>, extents: d3.Map<any[]>,
+                updateExtentsForKey: (key: string) => void) {
     let binding = bindings.get(key);
     let oldScale = binding != null ? binding.scale : null;
 
+    bindings.set(key, { accessor: d3.functor(value), scale: scale });
+    updateExtentsForKey(key);
```
jtlan: Rather than passing in an extent-updating function, would it be better to factor out the `Scale`-setting code, and call that in `_bindProperty()` and `_bindAttr()`?
```Typescript
protected _bindProperty(property: string, value: any, scale: Scale<any, any>) {
  this._bind(property, value, scale, this._propertyBindings, this._propertyExtents);
  this._updateExtentsForProperty(property);
  this._updateScaleForKey(property, scale);
}
```
This means one fewer argument needs to be passed to `_bind()`.

ztsai: yeah, I tried that first, but it actually produces a lot more code and looks messier
because you need to retrieve and store `oldScale` at `_bindProperty` and `_bindAttr` level
I could do it that way to avoid API signature change if it matters for protected functions

jtlan: I see, because the `Scale` gets replaced during `_bind()`. We could get rid of `_bind()` entirely and have its code duplicated in `_bindProperty()`/`_bindAttr()`? Not sure if that's better.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
       });
     });
 
+    describe("Automatic domain and range adjustmnet", () => {
```
jtlan: Can we specify for X and Y scales?

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+      const SVG_HEIGHT = 400;
+      const SVG_WIDTH = 400;
```
jtlan: Should use `let` until we have a way to enforce `const`.

test/plots/plotTests.ts
```diff
@@ -133,6 +133,26 @@ describe("Plots", () => {
...
+      let scale = new Plottable.Scales.Linear();
+      let defaultDomain = [0, 1];
+
+      plot.attr("x", (d) => d.x);
+      plot.attr("y", (d) => 1);
+      plot.renderTo(svg);
+      assert.deepEqual(scale.domain(), defaultDomain, "scale domain scale is same as default");
+
+      plot.attr("x", (d) => d.x, scale);
+      assert.deepEqual(scale.domain(), [4.5, 10.5], "scale domain scale is auto updated");
```
jtlan: If you set `padProportion(0)` and `snappingDomainEnabled(false)` you can use the data values for the bounds, rather than these magic numbers.

test/plots/plotTests.ts
```diff
@@ -133,6 +133,26 @@ describe("Plots", () => {
...
+      (<any> plot)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
+      plot.addDataset(dataset);
+
+      let scale = new Plottable.Scales.Linear();
+      let defaultDomain = [0, 1];
+
+      plot.attr("x", (d) => d.x);
+      plot.attr("y", (d) => 1);
+      plot.renderTo(svg);
+      assert.deepEqual(scale.domain(), defaultDomain, "scale domain scale is same as default");
```
jtlan: This seems unnecessary -- default domain is the `Scale`s functionality, which shouldn't be tested here.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let plot: Plottable.XYPlot<number, number>;
+      let dataset: Plottable.Dataset;
+      let defalutInterval = [0, 1];
```
jtlan: "`defaultInterval`"

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+        ]);
+        plot = new Plottable.XYPlot<number, number>();
+        (<any> plot)._createDrawer = (dataset: Plottable.Dataset) => createMockDrawer(dataset);
+        plot.addDataset(dataset);
+      });
+
+      it("automatically adjusts domain and range after rendering to DOM", () => {
+        plot.x((d: any) => d.a, xScale)
+            .y((d: any) => d.b, yScale);
+        assert.deepEqual(xScale.domain(), defalutInterval, "X domain has not been set automatically");
```
jtlan: Might want to specify in the error message that it's not set automatically *before rendering*.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+        assert.deepEqual(xScale.range(), defalutInterval, "X range has not been set automatically");
+        assert.deepEqual(yScale.domain(), defalutInterval, "Y domain has not been set automatically");
+        assert.deepEqual(yScale.range(), defalutInterval, "Y range has not been set automatically");
+
+        plot.renderTo(svg);
+        assert.deepEqual(xScale.domain(), [-7, 7], "X domain has been set automatically");
```
jtlan: Might want to specify in the error message that it *is* set automatically *after rendering*.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+        svg.remove();
+      });
+
+      it("automatically adjusts range when a Scale is attached after rendering", () => {
+        (<any> window).scale1 = xScale;
+        (<any> window).scale2 = yScale;
+        plot.x((d: any) => d.a)
+            .y((d: any) => d.b);
+
+        assert.deepEqual(xScale.domain(), defalutInterval, "X domain has not been set automatically");
```
jtlan: Might want to specify in the error message that it's not set automatically *before rendering*.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+        assert.deepEqual(xScale.range(), defalutInterval, "X range has not been set automatically");
+        assert.deepEqual(yScale.domain(), defalutInterval, "Y domain has not been set automatically");
+        assert.deepEqual(yScale.range(), defalutInterval, "Y range has not been set automatically");
+
+        plot.renderTo(svg);
+        plot.x((d: any) => d.a, xScale)
+            .y((d: any) => d.b, yScale);
+
+        assert.deepEqual(xScale.domain(), [-7, 7], "X domain has been set automatically");
```
jtlan: Might want to specify in the error message that it *is* set automatically *after rendering*.

test/plots/xyPlotTests.ts
```diff
@@ -128,6 +128,70 @@ describe("Plots", () => {
...
+        assert.deepEqual(xScale.range(), [0, SVG_WIDTH], "X range has been set automatically");
+        assert.deepEqual(yScale.domain(), [-7, 7], "Y domain has been set automatically");
+        assert.deepEqual(yScale.range(), [SVG_HEIGHT, 0], "Y range has been set automatically");
+        svg.remove();
+      });
+
+      it("automatically adjusts range when a Scale is attached after rendering", () => {
+        (<any> window).scale1 = xScale;
+        (<any> window).scale2 = yScale;
```
jtlan: What's the point of these two lines? They pollute the test space...

test/plots/xyPlotTests.ts
jtlan: Again, this seems unnecessary -- the `assert`s here are only checking `Scale` functionality, which shouldn't be tested here.

crmorford:
the values on the xAxis aren't right, the domain should be approximately [0, 6], but it's [0, 1]
screen shot 2015-09-17 at 10 35 37 am
screen shot 2015-09-17 at 10 35 25 am

ztsai:
It's a bug in fiddle code, xAxis is hooded up with the original domain
this should work: http://jsfiddle.net/ztsai/x7x05L4k/2/

PR #2768, opened 2015-09-15
Change name of arguments in siSuffix and general function in Formatter

### LINKED ISSUE #2594: Formatter factory arguments should be named more specifically
In the `Formatters` module, we have these factory methods:
```Typescript
function general(precision?: number): (d: any) => string;
function siSuffix(precision?: number): (d: any) => string;
```

However, the two factories use their `precision` argument differently: on `general` it acts as a "max number of decimals" limiter, while on `siSuffix` it specifies the number of significant figures.

We should rename the parameters to better reflect their behavior. (Renaming a parameter doesn't count as an API break).

yunhaolucky:
**Issue**
The argument `precision` in `Formatter.siSuffix` and `Formatter.general` is not clear.

**Fix** 
In `siSuffix`, `precision` is change to `maxNumberOfDecimalPlaces`
In `general`, `precision` is change to `numberOfSignificantFigures`.

close #2594

### DIFF
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -64,17 +64,17 @@ export module Formatters {
 
   /**
    * Creates a formatter that formats numbers to show no more than
-   * [precision] decimal places. All other values are stringified.
+   * [maxNumberOfDecimalPlaces] decimal places. All other values are stringified.
    *
-   * @param {number} [precision] The number of decimal places to show (default 3).
+   * @param {number} [maxNumberOfDecimalPlaces] The number of decimal places to show (default 3).
    *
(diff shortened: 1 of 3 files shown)


aicioara:
[NOQE] Just parameter rename. No actual JavaScript changes.

src/core/formatters.ts
```diff
@@ -121,9 +121,9 @@ export module Formatters {
    *
```
aicioara: Rename the parameter name in the JSDocs as well

aicioara:
merge-able

Issue #2771, opened 2015-09-15
Multiple classes not calling super.renderImmediately() in renderImmediately()


jtlan:
This isn't currently causing problems, but I wanted to point it out. A number of classes don't call `super.renderImmediately()` in their implementations of `renderImmediately()`, which is an architectural mistake (even if it doesn't cause bugs right now). Classes affected:
- `Plot`
- `SelectionBoxLayer`
- `Axis`
- `Axes.Time`

jtlan:
Looks like the restructuring in #2904 might take care of this.

PR #2773, opened 2015-09-15
clipPath now updates on render

### LINKED ISSUE #2733: clipPath loses reference if URL changes
Contour found an issue with zooming and time series plots. See CONTOUR-654 for more details.

jtlan:
Previously, if the URL changed without reloading the page, any `Component` that was re-anchored would lose its reference to its `clipPath`, because Plottable uses an absolute reference to the `clipPath` (due to https://github.com/palantir/plottable/pull/865). Now, `Component`s with `clipPath`s will update their references on rendering.

Before: http://jsfiddle.net/mpqwxg4n/7/
After: http://jsfiddle.net/4mrzogsp/

**NOTE**: For some reason, the bug currently only manifests on Firefox, even though less than 24 hours ago I saw it on Chrome too...

Close #2733.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -6463,6 +6463,7 @@ var Plottable;
                 this._bindBoxDataValues();
             };
             SelectionBoxLayer.prototype.renderImmediately = function () {
+                _super.prototype.renderImmediately.call(this);
                 if (this._boxVisible) {
                     var t = this._boxBounds.topLeft.y;
                     var b = this._boxBounds.bottomRight.y;
--- a/plottable.js
+++ b/plottable.js
@@ -3165,7 +3165,7 @@ var Plottable;
          * Renders the Component without waiting for the next frame.
          */
         Component.prototype.renderImmediately = function () {
-            if (this._clipPathEnabled && this._updateClipPath != null) {
+            if (this._clipPathEnabled) {
                 this._updateClipPath();
             }
             return this;
@@ -3263,19 +3263,18 @@ var Plottable;
             return box;
         };
         Component.prototype._generateClipPath = function () {
-            var _this = this;
             // The clip path will prevent content from overflowing its Component space.
-            var clipPathId = Plottable.Utils.DOM.generateUniqueClipPathId();
-            var clipPathParent = this._boxContainer.append("clipPath").attr("id", clipPathId);
+            this._clipPathID = Plottable.Utils.DOM.generateUniqueClipPathId();
+            var clipPathParent = this._boxContainer.append("clipPath").attr("id", this._clipPathID);
             this._addBox("clip-rect", clipPathParent);
-            this._updateClipPath = function () {
-                // HACKHACK: IE <= 9 does not respect the HTML base element in SVG.
-                // They don't need the current URL in the clip path reference.
-                var prefix = /MSIE [5-9]/.test(navigator.userAgent) ? "" : document.location.href;
-                prefix = prefix.split("#")[0]; // To fix cases where an anchor tag was used
-                _this._element.attr("clip-path", "url(\"" + prefix + "#" + clipPathId + "\")");
-            };
             this._updateClipPath();
+        };
+        Component.prototype._updateClipPath = function () {
+            // HACKHACK: IE <= 9 does not respect the HTML base element in SVG.
+            // They don't need the current URL in the clip path reference.
+            var prefix = /MSIE [5-9]/.test(navigator.userAgent) ? "" : document.location.href;
+            prefix = prefix.split("#")[0]; // To fix cases where an anchor tag was used
+            this._element.attr("clip-path", "url(\"" + prefix + "#" + this._clipPathID + "\")");
         };
         /**
          * Checks if the Component has a given CSS class.
--- a/plottable.js
+++ b/plottable.js
@@ -6834,6 +6834,7 @@ var Plottable;
             return h;
         };
         Plot.prototype.renderImmediately = function () {
+            _super.prototype.renderImmediately.call(this);
             if (this._isAnchored) {
                 this._paint();
                 this._dataChanged = false;
--- a/plottable.js
+++ b/plottable.js
@@ -3165,6 +3165,9 @@ var Plottable;
          * Renders the Component without waiting for the next frame.
          */
         Component.prototype.renderImmediately = function () {
+            if (this._clipPathEnabled && this._updateClipPath != null) {
+                this._updateClipPath();
+            }
             return this;
         };
         /**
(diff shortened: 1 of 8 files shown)


plottable.js
```diff
@@ -909,7 +909,7 @@ var Plottable;
 ///<reference path="../reference.ts" />
 var Plottable;
 (function (Plottable) {
-    Plottable.version = "1.10.0";
+    Plottable.version = "1.11.0";
```
bluong: Looks like the built file got a version change... which seems incredibly off?

aicioara: we are on 1.11.0 why is it off?

bluong: It doesn't make sense for this diff to occur, since either we are updating the version number of Plottable ("highly unlikely") or there is a potential merge with the develop branch that is not tracked via the commit / commit messages.  It is likely an issue with the built files.

src/components/component.ts
```diff
@@ -356,15 +360,17 @@ export class Component {
...
-                                           .attr("id", clipPathId);
+    let clipPathParent = this._boxContainer.append("clipPath").attr("id", clipPathId);
     this._addBox("clip-rect", clipPathParent);
+    this._updateClipPath = () => {
+      // HACKHACK: IE <= 9 does not respect the HTML base element in SVG.
+      // They don't need the current URL in the clip path reference.
+      let prefix = /MSIE [5-9]/.test(navigator.userAgent) ? "" : document.location.href;
+      prefix = prefix.split("#")[0]; // To fix cases where an anchor tag was used
+      this._element.attr("clip-path", "url(\"" + prefix + "#" + clipPathId + "\")");
+    };
```
bluong: I am still personally unsure of using functions that are initialized on-the-fly like this rather than have a `clipPathId` instance variable and then a function called `_updateClipPath` that is private

test/components/componentTests.ts
```diff
@@ -794,20 +794,50 @@ describe("Component", () => {
...
   describe("restricting rendering through clipPath", () => {
 
-    let c: Plottable.Component;
+    let clippedComponent: Plottable.Component;
     let svg: d3.Selection<void>;
 
     beforeEach(() => {
-      c = new Plottable.Component();
+      clippedComponent = new Plottable.Component();
+      (<any> clippedComponent)._clipPathEnabled = true;
```
bluong: Should you any-cast or just extend the component?

test/components/componentTests.ts
```diff
@@ -794,20 +794,50 @@ describe("Component", () => {
...
+    });
+
+    it("uses the correct clipPath", () => {
+      clippedComponent.renderTo(svg);
+      TestMethods.verifyClipPath(clippedComponent);
+      svg.remove();
+    });
+
+    it("updates the clipPath reference when render()-ed", () => {
+      if (window.history && window.history.replaceState) { // not supported on IE9 (http://caniuse.com/#feat=history)
```
bluong: I prefer a return instead of a huge if block.  1 extra level of nesting isn't fun...

src/components/component.ts
jtlan: `DragBoxLayer` does this, in order to expose as few variables as possible.

test/components/componentTests.ts
jtlan: It's only being used in two tests here. But I concede that it would be cleaner. I'll see if I can do that.

src/components/component.ts
bluong: Is it ok for `this._updateClipPath()` to error if `generateClipPath()` has not been called so far?

jtlan: It will be `undefined` otherwise. Do you mean that I should make it throw an error, or asking if it's acceptable that it might be undefined?

bluong: The latter.

I guess the thing that I'm not used to is having functions declared that we can call on some-regular basis not being defined until well into runtime.

jtlan: There's only one place where it's used, and it's checked if it exists there.

bluong: I guess my main concern is that, since `_updateClipPath` is defined at runtime, developers will need to consider the case if that function is actually defined by the time it runs into their code path, meaning the need to have existence checks for this function.

That said, it seems that the clipPath code is so one-off that the scope of defining and checking for this functions is pretty narrow.  We can run with this for now and hopefully it doesn't bite us.

test/components/componentTests.ts
jtlan: I'd have to declare it in `Mocks`. That seems farther away from where it's used, so not sure if it's helpful?

bluong: I think declaring it in the `Mocks` module is fine.  That said, we should probably more consider about renaming that module since some of them could be normal extensions.

src/components/component.ts
crmorford: > We can run with this for now and hopefully it doesn't bite us. 
NOPE

crmorford:
veto based on brandon's concerns above.

src/components/component.ts
bluong: Note that the pain point that I am concerned about is mostly from the developer end (where developers may need to hold additional cost to trade for the fact that we get nice security around a variable that should essentially never change) and nothing from the user end (these endpoints are not exposed).

src/components/component.ts
```diff
@@ -234,6 +235,9 @@ export class Component {
    * Renders the Component without waiting for the next frame.
    */
   public renderImmediately() {
+    if (this._clipPathEnabled) {
+      this._updateClipPath();
+    }
```
crmorford: @jtlan this works by updating the clip path to match the current url every time it renders?

Why did clip path have to be dependent on the URL in the first place?

src/components/component.ts
```diff
@@ -356,15 +360,18 @@ export class Component {
...
   private _generateClipPath() {
     // The clip path will prevent content from overflowing its Component space.
-    // HACKHACK: IE <=9 does not respect the HTML base element in SVG.
+    this._clipPathID = Utils.DOM.generateUniqueClipPathId();
+    let clipPathParent = this._boxContainer.append("clipPath").attr("id", this._clipPathID);
+    this._addBox("clip-rect", clipPathParent);
+    this._updateClipPath();
+  }
+
+  private _updateClipPath() {
```
crmorford: I only see _updateClipPath() called by renderImmediately(). Are there other spots that should be calling it, or is that the only function that renders things?

src/components/component.ts
jtlan: Usually just doing `clip-path="url(#clipPathID)"` (a relative path) would work ... but if there is a [`<base>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) on the page, it changes the base URL used for all relative paths ... causing the relative reference to the `clipPath` to get lost. The only way to work under a `<base>` tag is to specify an absolute path, as was done in https://github.com/palantir/plottable/pull/865.

src/components/component.ts
jtlan: That's the only place where things are drawn. Subclasses should be calling `_renderImmediately()`, and then calling their own logic inside that -- I've filed https://github.com/palantir/plottable/issues/2771 for classes that aren't doing this properly. I fixed it for `Plot` and `DragBoxLayer`; the other two (`Axes` classes) aren't using `clipPath`.

crmorford:
ff is the only browser that will repro this for me
with time axis http://jsfiddle.net/mpqwxg4n/8/
with base tags http://jsfiddle.net/mpqwxg4n/9/

PR #2775, opened 2015-09-16
throw an error with invalid annotation tier count

### LINKED ISSUE #2730: [Axis] annotationTierCount() does not reject bad values
In particular, negative values produce bad results: http://jsfiddle.net/cmf3ut7g/

**Actual**: Setting to `annotationTierCount()` to a negative value works, and will cause the tick labels to be hidden. More negative values cause the entire axis to disappear and errors like `Error: Invalid negative value for <rect> attribute height="-176"` to appear in the console.

**Expected**: `annotationTierCount()` should reject negative values. Maybe it should reject `Infinity` as well?

yunhaolucky:
**Issue**
`annotationTierCount()` will not reject negative values and a negative annotation tier count will result in negative height. 

**Fix**
In `annotationTierCount(annotationTierCount:number)`, if `annotationTierCount` is negative, throw an error.

**JSFiddle**
Before: https://jsfiddle.net/nup2ya6w/6/
After: https://jsfiddle.net/drq123db/1/

Close #2730

### DIFF
--- a/src/axes/axis.ts
+++ b/src/axes/axis.ts
@@ -273,7 +273,7 @@ export class Axis<D> extends Component {
     if (annotationTierCount == null) {
       return this._annotationTierCount;
     }
-    if (annotationTierCount < 0){
+    if (annotationTierCount < 0) {
       throw new Error(`annotationTierCount ${annotationTierCount} is not valid, should be positive or 0`);
     }
     this._annotationTierCount = annotationTierCount;
--- a/src/axes/axis.ts
+++ b/src/axes/axis.ts
@@ -274,7 +274,7 @@ export class Axis<D> extends Component {
       return this._annotationTierCount;
     }
     if (annotationTierCount < 0) {
-      throw new Error(`annotationTierCount ${annotationTierCount} cannot be negative`);
+      throw new Error(`annotationTierCount cannot be negative`);
     }
     this._annotationTierCount = annotationTierCount;
     this.redraw();
--- a/src/axes/axis.ts
+++ b/src/axes/axis.ts
@@ -274,7 +274,7 @@ export class Axis<D> extends Component {
       return this._annotationTierCount;
     }
     if (annotationTierCount < 0) {
-      throw new Error(`annotationTierCount ${annotationTierCount} is not valid, should be positive or 0`);
+      throw new Error(`annotationTierCount ${annotationTierCount} cannot be negative`);
     }
     this._annotationTierCount = annotationTierCount;
     this.redraw();
--- a/src/axes/axis.ts
+++ b/src/axes/axis.ts
@@ -273,6 +273,9 @@ export class Axis<D> extends Component {
     if (annotationTierCount == null) {
       return this._annotationTierCount;
     }
+    if (annotationTierCount < 0){
+      throw new Error(`annotationTierCount ${annotationTierCount} is not valid, should be positive or 0`);
+    }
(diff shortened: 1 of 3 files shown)


src/axes/axis.ts
```diff
@@ -273,6 +273,9 @@ export class Axis<D> extends Component {
     if (annotationTierCount == null) {
       return this._annotationTierCount;
     }
+    if (annotationTierCount < 0) {
+      throw new Error(`annotationTierCount ${annotationTierCount} is not valid, should be positive or 0`);
```
bluong: I personally think this can be made more concise and it would be better with one of the following ways:
* "annotationTierCount must be positive or 0"
* "annotationTierCount cannot be negative"
* "annotationTierCount ${annotationTierCount}..." with any of the above.

test/axes/baseAxisTests.ts
```diff
@@ -266,6 +266,16 @@ describe("BaseAxis", () => {
         assert.strictEqual(axis.annotationTierCount(annotationTierCount), axis, "setting annotation tier count returns calling axis");
         assert.deepEqual(axis.annotationTierCount(), annotationTierCount, "can set the annotation tier count");
       });
+
+      it("throws an error when annotation tier count is not valid", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        let annotationTierCount = [-1, -100];
+        annotationTierCount.forEach(function(tierCount) {
+          let currentTierCount = axis.annotationTierCount();
+          assert.throw(() => axis.annotationTierCount(tierCount), `annotationTierCount ${tierCount} is not valid, should be positive or 0`);
```
bluong: We should keep consistent on which of `assert.throw` / `assert.throws` we use.  I see `assert.throws` over and over if I remember correctly.  We should consistently use one of the two and then ban the other.

test/axes/baseAxisTests.ts
```diff
@@ -266,6 +266,16 @@ describe("BaseAxis", () => {
         assert.strictEqual(axis.annotationTierCount(annotationTierCount), axis, "setting annotation tier count returns calling axis");
         assert.deepEqual(axis.annotationTierCount(), annotationTierCount, "can set the annotation tier count");
       });
+
+      it("throws an error when annotation tier count is not valid", () => {
+        let axis = new Plottable.Axis(new Plottable.Scale<{}, number>(), "bottom");
+        let annotationTierCount = [-1, -100];
```
bluong: Let's call this array `invalidAnnotationTierCounts`?

test/axes/baseAxisTests.ts
```diff
@@ -266,6 +266,16 @@ describe("BaseAxis", () => {
...
+        annotationTierCount.forEach(function(tierCount) {
+          let currentTierCount = axis.annotationTierCount();
+          assert.throw(() => axis.annotationTierCount(tierCount), `annotationTierCount ${tierCount} is not valid, should be positive or 0`);
+          assert.strictEqual(axis.annotationTierCount(), currentTierCount, "annotationTierCound should not be changed to invalid values");
```
bluong: "annotationTierCount..."

crmorford:
Can we change the error just a bit?
screen shot 2015-09-17 at 10 50 27 am
I don't think we need the value (-1, -3, etc) included.

yunhaolucky:
Change error message to "annotationTierCount cannot be negative"

PR #2778, opened 2015-09-17
Correct the label position with "center" tierlabelposition and "top" orientation

### LINKED ISSUE #2772: Wrong Label Tick labels positions in 'top' orientation in Time Axis
The position of tick labels are set correctly when the orientation of axis is set to `top`. 

Repro: http://jsfiddle.net/o87wsq9h/2/
**Actual**: In this case, tick labels are hidden because their position are wrong and overlap with tick marks.
**Expected**: Tick labels should be visible when `orientation` is `top` and `tierLabelPositions` is `center`.

yunhaolucky:
**Issue**:
In a time axis with `top` orientation, the `y` value of label positions are not calculated correctly when its `tierlabelposition` is set to `center` . This is an issue because it makes tick marks overlap with tick labels and hide all labels as a result. 

**Fix**:
When calculate the `y` value of label positions, we do not consider the height of tick marks. I fix this.

**JSFiddle**:
Before:http://jsfiddle.net/o87wsq9h/2/
After:http://jsfiddle.net/o87wsq9h/3/

close #2772

### DIFF
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -392,9 +392,17 @@ export module Axes {
       let tickLabelsEnter = tickLabels.enter().append("g").classed(Axis.TICK_LABEL_CLASS, true);
       tickLabelsEnter.append("text");
       let xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
-      let yTranslate = this.orientation() === "bottom" ?
-          d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-          this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+
+      let yTranslate = 0;
+      if (this.orientation() === "bottom") {
+        yTranslate =  d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
+      } else {
+          if (this._tierLabelPositions[index] === "center") {
+          yTranslate = this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding() - this.innerTickLength();
+        } else {
+          yTranslate = this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+        }
+      }
 
       let textSelection = tickLabels.selectAll("text");
       if (textSelection.size() > 0) {
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -395,7 +395,7 @@ export module Axes {
 
       let yTranslate = 0;
       if (this.orientation() === "bottom") {
-        yTranslate =  d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
+        yTranslate = d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
       } else {
           if (this._tierLabelPositions[index] === "center") {
           yTranslate = this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding() - this.innerTickLength();
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -396,12 +396,10 @@ export module Axes {
       let yTranslate = 0;
       if (this.orientation() === "bottom") {
         yTranslate = d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
-      } else {
-          if (this._tierLabelPositions[index] === "center") {
+      } else if (this._tierLabelPositions[index] === "center") {
           yTranslate = this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding() - this.innerTickLength();
-        } else {
+      } else {
           yTranslate = this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
-        }
       }
 
       let textSelection = tickLabels.selectAll("text");
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -393,7 +393,7 @@ export module Axes {
       tickLabelsEnter.append("text");
       let xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
 
-      let yTranslate = 0;
+      let yTranslate: number;
       if (this.orientation() === "bottom") {
         yTranslate = d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
       } else if (this._tierLabelPositions[index] === "center") {
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -152,7 +152,7 @@ describe("TimeAxis", () => {
   it("tick labels do not overlap with tick marks in top orientation", () => {
      let svg = TestMethods.generateSVG(400, 100);
       scale = new Plottable.Scales.Time();
-      scale.domain([new Date("2015-09-02"), new Date("2015-09-03")]);
+      scale.domain([new Date(2015, 8, 2), new Date(2015, 8, 3)]);
       axis = new Plottable.Axes.Time(scale, "top");
 
       function checkTierDisplayPosition (tierDisplayPositions: any[]) {
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -167,12 +167,28 @@ describe("TimeAxis", () => {
           return window.getComputedStyle(this).visibility !== "hidden";
         });
         assert.operator(tickLabels.size(), ">=", 1, "There are more than one tick marks in the test");
-
+        
+        function myclientRectsOverlap(clientRectA:ClientRect, clientRectB:ClientRect) {
+                if (clientRectA.right < clientRectB.left + window.Pixel_CloseTo_Requirement) {
+                    return false;
+                }
+                if ( window.Pixel_CloseTo_Requirement + clientRectA.left > clientRectB.right) {
+                    return false;
+                }
+                if (clientRectA.bottom < clientRectB.top + window.Pixel_CloseTo_Requirement) {
+                    return false;
+                }
+                if (window.Pixel_CloseTo_Requirement + clientRectA.top > clientRectB.bottom) {
+                    return false;
+                }
+                return true;
+            }
+            
         tickMarks.each(function(tickMark) {
           let tickMarkRect = this.getBoundingClientRect();
           tickLabels.each(function(tickLabel) {
             let tickLabelRect = this.getBoundingClientRect();
-            let isOverlap = Plottable.Utils.DOM.clientRectsOverlap(tickMarkRect, tickLabelRect);
+            let isOverlap = myclientRectsOverlap(tickMarkRect, tickLabelRect);
             assert.isFalse(isOverlap, `Tick marks "${tickMark}" should not overlap with tick labels "${this.textContent}" `);
           });
         });
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -167,14 +167,30 @@ describe("TimeAxis", () => {
           return window.getComputedStyle(this).visibility !== "hidden";
         });
       assert.operator(tickLabels.size(), ">=", 1, "There are at least one tick labels in the test");
+
+      function clientRectsOverlap(clientRectA: ClientRect, clientRectB: ClientRect) {
+        if (clientRectA.right < clientRectB.left + window.Pixel_CloseTo_Requirement) {
+          return false;
+        } else if (window.Pixel_CloseTo_Requirement + clientRectA.left > clientRectB.right) {
+          return false;
+        } else if (clientRectA.bottom < clientRectB.top + window.Pixel_CloseTo_Requirement) {
+          return false;
+        } else if (window.Pixel_CloseTo_Requirement + clientRectA.top > clientRectB.bottom) {
+          return false;
+        } else { 
+          return true;
+        }
+      }
+
       tickMarks.each(function(tickMark) {
         let tickMarkRect = this.getBoundingClientRect();
         tickLabels.each(function(tickLabel) {
           let tickLabelRect = this.getBoundingClientRect();
-          assert.isFalse(Plottable.Utils.DOM.clientRectsOverlap(tickMarkRect, tickLabelRect),
+          assert.isFalse(clientRectsOverlap(tickMarkRect, tickLabelRect), 
             `Tick marks "${tickMark}" should not overlap with tick labels "${this.textContent}" `);
         });
       });
+    style.remove();
     }
 
     checkTierDisplayPosition(["between", "between"]);
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -152,7 +152,7 @@ describe("TimeAxis", () => {
   it("tick labels do not overlap with tick marks in top orientation", () => {
      let svg = TestMethods.generateSVG(400, 100);
       scale = new Plottable.Scales.Time();
-      scale.domain([new Date("2015-09-01 12:00:00"), new Date("2015-09-03 12:00:00")]);
+      scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
       axis = new Plottable.Axes.Time(scale, "top");
 
       function checkTierDisplayPosition (tierDisplayPositions: any[]) {
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -155,13 +155,13 @@ describe("TimeAxis", () => {
       scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
       axis = new Plottable.Axes.Time(scale, "top");
 
-      let style = (<any> axis)._element.append("style");
-      style.attr("type", "text/css");
-      style.text(".plottable .axis.time-axis text { font-family: Arial; }");
-
       function checkTierDisplayPosition (tierDisplayPositions: any[]) {
         axis.tierLabelPositions(tierDisplayPositions);
         axis.renderTo(svg);
+
+        let style = (<any>axis)._element.append("style");
+        style.attr("type", "text/css");
+        style.text(".plottable .axis.time-axis text { font-family: Arial; }");
 
         let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
         assert.operator(tickMarks.size(), ">=", 1, "There are more than one tick marks in the test");
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -129,32 +129,28 @@ describe("TimeAxis", () => {
     assert.isTrue(lastTick.classed(Plottable.Axis.END_TICK_MARK_CLASS), "last end tick has the end-tick-mark class");
     svg.remove();
   });
-
-  it("tick labels do not overlap with tick marks", () => {
-    let svg = TestMethods.generateSVG(400, 100);
-    scale = new Plottable.Scales.Time();
-    scale.domain([new Date("2009-12-20"), new Date("2011-01-01")]);
-    axis = new Plottable.Axes.Time(scale, "bottom");
-    axis.renderTo(svg);
-    let tickRects = d3.selectAll("." + Plottable.Axis.TICK_MARK_CLASS)[0].map((mark: Element) => mark.getBoundingClientRect());
-    let labelRects = d3.selectAll("." + Plottable.Axis.TICK_LABEL_CLASS)
-        .filter(function(d: Element, i: number) {
-          return d3.select(this).style("visibility") === "visible";
-        })[0].map((label: Element) => label.getBoundingClientRect());
-    labelRects.forEach(function(labelRect: ClientRect) {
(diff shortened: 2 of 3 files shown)


test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+      scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
+      axis = new Plottable.Axes.Time(scale, "top");
+
+      function checkTierDisplayPosition (tierDisplayPositions: any[]) {
+        axis.tierLabelPositions(tierDisplayPositions);
+        axis.renderTo(svg);
+
+        let style = (<any>axis)._element.append("style");
+        style.attr("type", "text/css");
+        style.text(".plottable .axis.time-axis text { font-family: Arial; }");
```
yunhaolucky: This css part is for font problem in firefox.

src/axes/timeAxis.ts
```diff
@@ -392,9 +392,17 @@ export module Axes {
       let tickLabelsEnter = tickLabels.enter().append("g").classed(Axis.TICK_LABEL_CLASS, true);
       tickLabelsEnter.append("text");
       let xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
-      let yTranslate = this.orientation() === "bottom" ?
-          d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-          this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+
+      let yTranslate = 0;
+      if (this.orientation() === "bottom") {
+        yTranslate =  d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding();
```
aicioara: remove extra whitespace after `=`

src/axes/timeAxis.ts
```diff
@@ -392,9 +392,17 @@ export module Axes {
...
+      } else {
```
aicioara: Avoid nestedness by using `else if` instead of `else { if` (basically have a 3 branch if statement)

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels do not overlap with tick marks in top orientation", () => {
+     let svg = TestMethods.generateSVG(400, 100);
+      scale = new Plottable.Scales.Time();
```
aicioara: indentation broke here and below

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels do not overlap with tick marks in top orientation", () => {
+     let svg = TestMethods.generateSVG(400, 100);
```
aicioara: Indentation broke after this line

test/axes/timeAxisTests.ts
aicioara: The scale is already initialized in the `beforeEach()` block, on line 7. No need to do it here as well

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+      scale = new Plottable.Scales.Time();
+      scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
+      axis = new Plottable.Axes.Time(scale, "top");
+
+      function checkTierDisplayPosition (tierDisplayPositions: any[]) {
```
aicioara: remove space between function name and (

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels do not overlap with tick marks in top orientation", () => {
+     let svg = TestMethods.generateSVG(400, 100);
+      scale = new Plottable.Scales.Time();
+      scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
```
aicioara: Not a big deal, but I would prefer dates to be formatted as
```typescript
new Date(2010, 00, 01);
new Date(2014, 00, 01);
```

But up to you, as we don't have a standard about this yet

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+      axis = new Plottable.Axes.Time(scale, "top");
+
+      function checkTierDisplayPosition (tierDisplayPositions: any[]) {
+        axis.tierLabelPositions(tierDisplayPositions);
+        axis.renderTo(svg);
+
+        let style = (<any>axis)._element.append("style");
```
aicioara: can we append this to the `<body>` instead? 
```
d3.select("body").append("style")
```

Also, can we
```typescript
style.remove()
```
at the end

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+        style.attr("type", "text/css");
```
aicioara: Not needed

test/axes/timeAxisTests.ts
aicioara: Can we add `!important` just in case
https://css-tricks.com/when-using-important-is-the-right-choice/

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+        style.text(".plottable .axis.time-axis text { font-family: Arial; }");
+
+        let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
```
aicioara: Hey, I really like this way of selecting (the `${...}` bit). We should make it a testing standard

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+        assert.operator(tickMarks.size(), ">=", 1, "There are more than one tick marks in the test");
```
aicioara: The assert text and the operator do not match. 

Either use `>` or say `at least 1`

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+
+        let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`)
```
aicioara: move the .filter from the previous line here (or indent it there)

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+        .filter(function(d: Element, i: number){
+          return window.getComputedStyle(this).visibility !== "hidden";
```
aicioara: use 
```typescript 
d3.select(this).style("visibility") !== "hidden"
```

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+
+        let style = (<any>axis)._element.append("style");
+        style.attr("type", "text/css");
+        style.text(".plottable .axis.time-axis text { font-family: Arial; }");
+
+        let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
+        assert.operator(tickMarks.size(), ">=", 1, "There are more than one tick marks in the test");
+
+        let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`)
+        .filter(function(d: Element, i: number){
```
aicioara: No need to type them explicitly. Also no need to even put them in the parameter list

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+          return window.getComputedStyle(this).visibility !== "hidden";
+        });
+        assert.operator(tickLabels.size(), ">=", 1, "There are more than one tick labels in the test");
```
aicioara: Same as above

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+
+        function myclientRectsOverlap(clientRectA: ClientRect, clientRectB: ClientRect) {
```
aicioara: rename to `clientRectsOverlap`. Also can we use the use `Utils.DOM.clientRectsOverlap()` or there is a issue with the Pixel Close To?

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+                if (clientRectA.right < clientRectB.left + window.Pixel_CloseTo_Requirement) {
```
aicioara: Use 2 space indentation

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+                    return false;
+                }
+                if ( window.Pixel_CloseTo_Requirement + clientRectA.left > clientRectB.right) {
```
aicioara: No whitespace after `(`

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,60 @@ describe("TimeAxis", () => {
...
+                    return false;
+                }
+                return true;
+            }
+
+        tickMarks.each(function(tickMark) {
+          let tickMarkRect = this.getBoundingClientRect();
+          tickLabels.each(function(tickLabel) {
+            let tickLabelRect = this.getBoundingClientRect();
+            let isOverlap = myclientRectsOverlap(tickMarkRect, tickLabelRect);
```
aicioara: you can inline this in the assert below

aicioara:
Mainly style issues. Sorry we don't have `tslint` rules for all of the above.

test/axes/timeAxisTests.ts
yunhaolucky: It seems `<g style = "   ">` has the highest priority. What could break this?

src/axes/timeAxis.ts
yunhaolucky: Would nested loops be more easier to understand than the 3 branch if statement? Because in the first if statement, we try to pick from `orientation` and the in the nested if statement, we try to pick in `tierlabalpoistion`.

test/axes/timeAxisTests.ts
aicioara: This is not a `<g style = " "`. This is a 
```html
<style> 
  .plottable .axis.time-axis text {
    font-family: Arial;
  }
</style>

File: src/axes/timeAxis.ts (hunk 3 above)
Comment (aicioara) [inline src/axes/timeAxis.ts]: They are equivalent. We want to keep the code as flat as possible and avoid the arrow anti pattern as much as possible http://c2.com/cgi/wiki?ArrowAntiPattern

File: test/axes/timeAxisTests.ts (hunk 13 above)
Comment (yunhaolucky) [inline test/axes/timeAxisTests.ts]: Can I ask what is the difference between `d3.select` and `window.getComputedStyle'?

File: test/axes/timeAxisTests.ts (hunk 16 above)
Comment (yunhaolucky) [inline test/axes/timeAxisTests.ts]: The reason I do not use Utils.DOM.clientRectsOverlap() is that it does not take `pixel_close_to` into account and It cause some problems in firefox.

File: test/axes/timeAxisTests.ts (hunk 13 above)
Comment (aicioara) [inline test/axes/timeAxisTests.ts]: Sure. Internally, I believe d3 just calls `window.getComputedStyle`, but the `d3` version is more consistent with the rest of the codebase.

Also, if you check line 311, there is a leftover (I believe) HACKHACK which says that `window.getComputedStyle` behaves weirdly in IE9, so my suggestion is safer.

File: test/axes/timeAxisTests.ts (hunk 16 above)
Comment (aicioara) [inline test/axes/timeAxisTests.ts]: Is it still the case with the new font: Arial hack?

Comment (yunhaolucky) [inline test/axes/timeAxisTests.ts]: Still not working with Arial hack.

Comment (aicioara) [inline test/axes/timeAxisTests.ts]: Then it's fine

File: test/axes/timeAxisTests.ts [hunk 20]
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+      axis.tierLabelPositions(tierDisplayPositions);
+      axis.renderTo(svg);
+      let style = (<any>axis)._element.append("style");
+      style.attr("type", "text/css");
+      style.text(".plottable .axis.time-axis text { font-family: Arial; }");
+      let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
+      assert.operator(tickMarks.size(), ">=", 1, "There are at least one tick marks in the test");
+      let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`)
+      .filter(function(d, i){
+        return window.getComputedStyle(this).visibility !== "hidden";
```
yunhaolucky: `d3.style` also does not work on IE9. As the other test in this file use `window.getComputedStyle`, I might still want to use `window.getComputedStyle`.

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+            return false;
+        }
+        if (clientRectA.bottom < clientRectB.top + window.Pixel_CloseTo_Requirement) {
+            return false;
+        }
+        if (window.Pixel_CloseTo_Requirement + clientRectA.top > clientRectB.bottom) {
+            return false;
+        }
+        return true;
+      }
```
yunhaolucky: Firefox test cannot be passed without this part.

test/axes/timeAxisTests.ts
aicioara: Using this passes IE9. Am I missing something?
```typescript
return d3.select(this).style("visibility") !== "hidden";
```

src/axes/timeAxis.ts
```diff
@@ -392,9 +392,15 @@ export module Axes {
       let tickLabelsEnter = tickLabels.enter().append("g").classed(Axis.TICK_LABEL_CLASS, true);
       tickLabelsEnter.append("text");
       let xTranslate = (this._tierLabelPositions[index] === "center" || config.step === 1) ? 0 : this.tickLabelPadding();
-      let yTranslate = this.orientation() === "bottom" ?
-          d3.sum(this._tierHeights.slice(0, index + 1)) - this.tickLabelPadding() :
-          this.height() - d3.sum(this._tierHeights.slice(0, index)) - this.tickLabelPadding();
+
+      let yTranslate = 0;
```
aicioara:
```typescript
let yTranslate: number;
```

It is the same thing, but more semantically correct.

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("tick labels do not overlap with tick marks in top orientation", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
+    axis = new Plottable.Axes.Time(scale, "top");
```
aicioara: This line might be confusing. Let's declare a new variable called topAxis (and prefix it with `let`)

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+    function checkTierDisplayPosition (tierDisplayPositions: string[]) {
+      axis.tierLabelPositions(tierDisplayPositions);
+      axis.renderTo(svg);
+      let style = (<any>axis)._element.append("style");
+      style.attr("type", "text/css");
+      style.text(".plottable .axis.time-axis text { font-family: Arial; }");
+      let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
+      assert.operator(tickMarks.size(), ">=", 1, "There are at least one tick marks in the test");
+      let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`)
+      .filter(function(d, i){
```
aicioara: Can we move this `.filter` on the previous line? This way indentation does not get messed.

```typescript
let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`).filter(function(d, i){
```

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+    axis = new Plottable.Axes.Time(scale, "top");
+
+    function checkTierDisplayPosition (tierDisplayPositions: string[]) {
+      axis.tierLabelPositions(tierDisplayPositions);
+      axis.renderTo(svg);
+      let style = (<any>axis)._element.append("style");
+      style.attr("type", "text/css");
+      style.text(".plottable .axis.time-axis text { font-family: Arial; }");
+      let tickMarks = d3.selectAll(`.${Plottable.Axis.TICK_MARK_CLASS}:not(.${Plottable.Axis.END_TICK_MARK_CLASS})`);
+      assert.operator(tickMarks.size(), ">=", 1, "There are at least one tick marks in the test");
```
aicioara: "There **is** at least one..."

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+      let tickLabels = d3.selectAll(`.${Plottable.Axis.TICK_LABEL_CLASS}`)
+      .filter(function(d, i){
+        return window.getComputedStyle(this).visibility !== "hidden";
+      });
+      assert.operator(tickLabels.size(), ">=", 1, `There are at least one tick labels in the test`);
```
aicioara: "There **is** at least one..."

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+
+      function clientRectsOverlap(clientRectA: ClientRect, clientRectB: ClientRect) {
+        if (clientRectA.right < clientRectB.left + window.Pixel_CloseTo_Requirement) {
+            return false;
```
aicioara: two space indentation. Here and below

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+        if (window.Pixel_CloseTo_Requirement + clientRectA.left > clientRectB.right) {
+            return false;
+        }
+        if (clientRectA.bottom < clientRectB.top + window.Pixel_CloseTo_Requirement) {
+            return false;
+        }
+        if (window.Pixel_CloseTo_Requirement + clientRectA.top > clientRectB.bottom) {
+            return false;
+        }
+        return true;
```
aicioara: can we collapse this entire block into a single return statement?
```typescript
return !(clientRectA.right < clientRectB.left + window.Pixel_CloseTo_Requirement && window.Pixel_CloseTo_Requirement + clientRectA.left > clientRectB.right ...)
```

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+      }
+      tickMarks.each(function(tickMark) {
+        let tickMarkRect = this.getBoundingClientRect();
+        tickLabels.each(function(tickLabel) {
+          let tickLabelRect = this.getBoundingClientRect();
+          let isOverlap = clientRectsOverlap(tickMarkRect, tickLabelRect);
```
aicioara: <del> this variable has an awkward name. Let's call it `markOverlapsLabel`. </del>

Inline in the assert below. No need for new variable.

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+          assert.isFalse(isOverlap, `Tick marks "${tickMark}" should not overlap with tick labels "${this.textContent}"`);
```
aicioara: Use singular "Tick **mark** [...] with tick **label** [...]"

test/axes/timeAxisTests.ts
yunhaolucky: The hackhack you mentioned is because both window.getComputedStyle and d3.select return `inherit` instead of `visible`.

test/axes/timeAxisTests.ts
aicioara: Doing this 
```typescript
let isOverlap = Plottable.Utils.DOM.clientRectsOverlap(tickMarkRect, tickLabelRect);
```
Works for firefox (all tests passing). Am I missing something?

test/axes/timeAxisTests.ts
```diff
@@ -149,6 +149,54 @@ describe("TimeAxis", () => {
...
+        });
+      });
+    }
+    checkTierDisplayPosition(["between", "between"]);
+    checkTierDisplayPosition(["center", "center"]);
```
aicioara: If we do something like this, either:
1. Separate this logic into 2 tests (two separate its) or
2. Move everything inside that function: the `svg.remove()`, the svg creation and the axis creation. 

I prefer the former.

yunhaolucky: Can I just make the error message includes the tier label position setting?

yunhaolucky: It seems every time we change the tier label position, redraw() is called. Why do we need to create svg and remove svg in every iteration?

PR #2779, opened 2015-09-17
[Plots.Bar] Off-bar Labels (was: "Contextual Label Placement")

### LINKED ISSUE #2578: More display options for labels on Plots.Bar

jtlan:
Recreation of #2597.

Draws labels above bars rather than inside them if the bar is too small:
image

New CSS classes:
```CSS
.plottable .off-bar-label text {
}
```
On-bar labels now carry the `on-bar-label` class.

To be fixed:
 - <del>How does this work on `StackedBar`? Off-bar labels shouldn't be shown for bars lower in the stack, but does it even make sense to have the last off-bar label? Would it show the total of the stack, or just the value of the top item in the stack?</del> --> #2795.
 - [x] On horizontal `Plots.Bar`, sometimes off-bar labels aren't hidden if they're cut off:

  screen shot 2015-09-17 at 11 04 44 am
screen shot 2015-09-21 at 1 38 18 pm

 - [x] **(solved aside from corner case with reversed domains)** On horizontal `Plots.Bar`, zero isn't shown on the "positive" side of the line:

  screen shot 2015-09-17 at 11 04 52 am

Close #2578.


src/plots/barPlot.ts
```diff
@@ -468,78 +468,119 @@ export module Plots {
...
+      let drawLabel = (d: any, i: number) => {
+        let valueAccessor = this._isVertical ? this.y().accessor : this.x().accessor;
+        let value = valueAccessor(d, i, dataset);
+        let valueScale: Scale<any, number> = this._isVertical ? this.y().scale : this.x().scale;
+        let scaledValue = valueScale != null ? valueScale.scale(value) : value;
+        let scaledBaseline = valueScale != null ? valueScale.scale(this.baselineValue()) : this.baselineValue();
+
+        let barWidth = attrToProjector["width"](d, i, dataset);
+        let barHeight = attrToProjector["height"](d, i, dataset);
+        let text = this._labelFormatter(valueAccessor(d, i, dataset)).toString();
```
bluong: The `toString()` portion here seems extraneous

src/plots/barPlot.ts
```diff
@@ -468,78 +468,119 @@ export module Plots {
...
+            let offset = Math.min( (barHeight - measurement.height ) / 2, Bar._LABEL_VERTICAL_PADDING);
+            if (scaledValue < scaledBaseline) {
+              labelContainerOrigin.y += offset;
+              yAlignment = "top";
+              labelOrigin.y += offset;
+            } else {
+              labelContainerOrigin.y -= offset;
+              yAlignment = "bottom";
+              labelOrigin.y += containerHeight - offset - measurement.height;
+            }
```
bluong: Would the above be easier to read with ternary?

Factor out the `scaledValue < scaledBaseline` into a variable and then use that with the ternary statements

src/plots/barPlot.ts
```diff
@@ -468,78 +468,119 @@ export module Plots {
...
+            let offset = Bar._LABEL_VERTICAL_PADDING;
+            containerHeight = barHeight + offset + measurement.height;
+            if (scaledValue <= scaledBaseline) {
+              labelContainerOrigin.y -= offset + measurement.height;
+              yAlignment = "top";
+              labelOrigin.y -= offset + measurement.height;
+            } else {
+              yAlignment = "bottom";
+              labelOrigin.y += barHeight + offset;
+            }
```
bluong: Same as above

test/components/dragBoxLayerTests.ts
```diff
@@ -103,7 +103,7 @@ describe("Layer Components", () => {
       });
 
       it("applies the given detection radius property", () => {
-        dbl.renderTo("svg");
```
bluong: -_______-

test/plots/barPlotTests.ts
```diff
@@ -722,74 +722,127 @@ describe("Plots", () => {
...
+        let texts = barPlot.content().selectAll("text");
+        assert.strictEqual(texts.size(), dataset.data().length, "one label drawn per datum");
 
-        assert.include(["visible", "inherit"], label1.style("visibility"), "label 1 is visible");
-        assert.strictEqual(label2.style("visibility"), "hidden", "label 2 is not visible");
+        let offBarLabels = barPlot.content().selectAll(".off-bar-label");
+        assert.strictEqual(offBarLabels.size(), 3, "there are 3 off-bar labels");
 
+        let onBarLabels = barPlot.content().selectAll(".on-bar-label");
+        assert.strictEqual(onBarLabels.size(), 2, "there are 2 on-bar labels");
```
bluong: What are these magic numbers?

test/plots/barPlotTests.ts
```diff
@@ -839,138 +892,192 @@ describe("Plots", () => {
...
-        plot.labelsEnabled(true);
-        plot.renderTo(svg);
-        plot.labelFormatter((n: number) => n.toString() + (n === 12345 ? "looong" : ""));
-        let texts = svg.selectAll("text")[0].map((n: any) => d3.select(n).text());
-        assert.lengthOf(texts, 0, "no text drawn");
+        svg.attr("width", 100);
+        barPlot.labelsEnabled(true);
+        barPlot.renderTo(svg);
+        let texts = barPlot.content().selectAll("text");
+        assert.strictEqual(texts.size(), 0, "no labels drawn");
```
bluong: Consider `texts.empty()`

test/plots/barPlotTests.ts
```diff
@@ -839,138 +892,192 @@ describe("Plots", () => {
...
       it("formatters are used properly", () => {
-        plot.labelsEnabled(true);
-        plot.labelFormatter((n: number) => n.toString() + "%");
-        plot.renderTo(svg);
-        let texts = svg.selectAll("text")[0].map((n: any) => d3.select(n).text());
-        assert.lengthOf(texts, 2, "both texts drawn");
-        assert.strictEqual(texts[0], "640%", "first label is 640%");
-        assert.strictEqual(texts[1], "12345%", "first label is 12345%");
+        barPlot.labelsEnabled(true);
+        let formatter = (n: number) => n.toString() + "%";
```
bluong: Let's use string interpolation \`${n}%\`

test/plots/barPlotTests.ts
```diff
@@ -839,138 +892,192 @@ describe("Plots", () => {
...
+        barPlot.renderTo(svg);
+
+        let texts = barPlot.content().selectAll("text");
+        assert.strictEqual(texts.size(), dataset.data().length, "one label drawn per datum");
+
+        let offBarLabels = barPlot.content().selectAll(".off-bar-label");
+        assert.strictEqual(offBarLabels.size(), 3, "there are 3 off-bar labels");
+
+        let onBarLabels = barPlot.content().selectAll(".on-bar-label");
+        assert.strictEqual(onBarLabels.size(), 2, "there are 2 on-bar labels");
```
bluong: Can we get rid of the magic numbers?

src/plots/barPlot.ts
bluong: Example code would look like
```typescript
let barAboveBaseline = scaledValue < scaledBaseline;
labelContainerOrigin.y += barAboveBaseline ? offset : -offset;
yAlignment = barAboveBaseline ? "top" : "bottom";
labelOrigin.y += barAboveBaseline ? offset : containerHeight - offset - measurement.height;
```

bluong:
There was a bug desribed about setting the domain in the wrong order.  Can you elaborate?

test/plots/barPlotTests.ts
jtlan: Unfortunately, I don't think so -- with the exception of `0`, there's no value that's guaranteed to show as on-bar / off-bar. The only way to make the values computed is to essentially reimplement the drawing logic :(

test/plots/barPlotTests.ts
```diff
@@ -839,138 +892,192 @@ describe("Plots", () => {
...
+        let texts = barPlot.content().selectAll("text");
+        assert.strictEqual(texts.size(), dataset.data().length, "one label drawn per datum");
+        let expectedTexts = dataset.data().map((d) => formatter(d.y));
+        texts.each(function(d, i) {
+          assert.strictEqual(d3.select(this).text(), expectedTexts[i], `formatter is applied to the displayed value (index ${i})`);
+        });
+        svg.remove();
+      });
+
+      it("bar labels are shown inside or outside the bar as appropriate", () => {
```
bluong: Can we ake this into a smooth sentence if possible?

`it("shows labels inside or outside the bar as appropriate, ...")`

src/plots/barPlot.ts
jtlan: The `if`-block implementation here is intended to make the distinct cases for labels clearer:
* `Plot` is vertical or horizontal?
* label is on-bar or off-bar?
* value is positive or negative?

The original code used ternaries, but it was sometimes difficult to understand which case a particular line of logic was being written for.

test/plots/barPlotTests.ts
jtlan: I think `size()` is better, since it would provide more information about how many extra were drawn.

test/plots/barPlotTests.ts
bluong: How messy would it be to loop through each label, and do some DOM computation in order to figure out if they should be an `off-bar-label` or an `on-bar-label`?

jtlan: You'd have to correlate the labels with the bars somehow. Index would work I guess?

jtlan:
>There was a bug desribed about setting the domain in the wrong order.  Can you elaborate?

If you flip the domain, by using something like `xScale.domain([5, -5])` (note that the domain values are not in ascending order), the `0` label can appear on the wrong side of the baseline.

src/plots/barPlot.ts
bluong: Hmm, the code as it is is flexible, but it also is quite long.  I would say keep it for now?

test/plots/stackedBarPlotTests.ts
```diff
@@ -377,6 +391,20 @@ describe("Plots", () => {
...
+    it("doesn't show any off-bar labels", () => {
+      renderer.labelsEnabled(true);
+      xScale.domain([0, 50]);
+      let offBarLabels = renderer.content().selectAll(".off-bar-label");
+      assert.operator(offBarLabels.size(), ">", 0, "some off-bar labels are drawn");
+      let offBarLabelVisibilities: string[] = [];
+      offBarLabels.each(function() {
+        offBarLabelVisibilities.push(d3.select(this).style("visibility"));
+      });
+      assert.isTrue(offBarLabelVisibilities.every((visibility) => visibility === "hidden"), "all off-bar labels are hidden");
```
bluong: Consider using a `reduce` call here

test/plots/stackedBarPlotTests.ts
```diff
@@ -90,6 +90,20 @@ describe("Plots", () => {
...
+    it("doesn't show any off-bar labels", () => {
+      renderer.labelsEnabled(true);
+      yScale.domain([0, 30]);
+      let offBarLabels = renderer.content().selectAll(".off-bar-label");
+      assert.operator(offBarLabels.size(), ">", 0, "some off-bar labels are drawn");
+      let offBarLabelVisibilities: string[] = [];
+      offBarLabels.each(function() {
+        offBarLabelVisibilities.push(d3.select(this).style("visibility"));
+      });
+      assert.isTrue(offBarLabelVisibilities.every((visibility) => visibility === "hidden"), "all off-bar labels are hidden");
```
bluong: Consider `reduce`

crmorford:
tests from okal's PR.
I'll test more in the morning
Not enough space in y-direction http://jsfiddle.net/ts720tdr/20/
panzoom y (really cool seeing the numbers hop on and off) http://jsfiddle.net/ts720tdr/21/
panzoom x http://jsfiddle.net/ts720tdr/22/
negative zero http://jsfiddle.net/ts720tdr/23/
negative bars http://jsfiddle.net/ts720tdr/24/
time (formatted) http://jsfiddle.net/ts720tdr/25/
room to left of bar http://jsfiddle.net/ts720tdr/26/
css labels http://jsfiddle.net/ts720tdr/27/
css labels 2 http://jsfiddle.net/ts720tdr/28/
doesn't break waterfall  http://jsfiddle.net/ts720tdr/29/
clustered bar http://jsfiddle.net/ts720tdr/30/ http://jsfiddle.net/ts720tdr/31/
ModLog scale http://jsfiddle.net/ts720tdr/32/

crmorford:
I think this should have no impact on stackedBar at all.
http://jsfiddle.net/ts720tdr/33/
I know it isn't supposed to work, but I think it should fail more gracefully.

crmorford:
screen shot 2015-09-23 at 11 21 52 am
http://jsfiddle.net/ts720tdr/34/

You can get to a place where 
1. the smaller bars are too small for their labels, so the labels are off-bar labels, 
2. the bigger bars have room for labels.
3. because the domain starts at a number > 0, the bar thinks it has room for the label, so it doesn't make it an off-bar-label, but then it gets cut off by the side of the svg.

Is there a way to reorder checks so that if a label gets cut off by the edge of the svg, it becomes an off-bar-label?

jtlan:
>I think this should have no impact on stackedBar at all.
http://jsfiddle.net/ts720tdr/33/
I know it isn't supposed to work, but I think it should fail more gracefully.

Needed to load the new CSS sheet: http://jsfiddle.net/1e4jjkus/

crmorford:
stacked bars look good then :) just the last comment still stands

crmorford:
the 0 label isn't showing up here after being formatted http://jsfiddle.net/ts720tdr/37/

jtlan:
>the 0 label isn't showing up here after being formatted http://jsfiddle.net/ts720tdr/37/

Doesn't seem to happen even without the formatter: http://jsfiddle.net/8t2hz81q/1/
... but for some reason I can't get it to happen locally.

src/plots/barPlot.ts
```diff
@@ -468,78 +468,133 @@ export module Plots {
...
+        };
+        let containerWidth = barWidth;
+        let containerHeight = barHeight;
+
+        let labelOrigin = {
+          x: labelContainerOrigin.x,
+          y: labelContainerOrigin.y
+        };
+
+        let showLabelOnBar = true;
```
bluong: Based on the usage below, it seems like this variable is not well defined until entering the if blocks.

Change to
`let showLabelOnBar: boolean;`

src/plots/barPlot.ts
```diff
@@ -468,78 +468,133 @@ export module Plots {
...
-              labelPosition.x = x + w - measurement.width;
+        } else { // horizontal
+          labelOrigin.y += containerHeight / 2 - measurement.height / 2;
+
+          let barX = attrToProjector["x"](d, i, dataset);
+          let effectiveBarWidth = Utils.Math.min([
+            this.width() - barX,
+            barX + barWidth,
+            barWidth
+          ], 0);
```
bluong: Consider `Math.min` as we know that we have 3 entries here.  Having the default case does not seem useful.

src/plots/barPlot.ts
```diff
@@ -468,78 +468,133 @@ export module Plots {
...
+        if (this._isVertical) {
+          labelOrigin.x += containerWidth / 2 - measurement.width / 2;
+
+          let barY = attrToProjector["y"](d, i, dataset);
+          let effectiveBarHeight = Utils.Math.min([
+            this.height() - barY,
+            barY + barHeight,
+            barHeight
+          ], 0);
+          let offset = Bar._LABEL_VERTICAL_PADDING;
```
bluong: Consider `Math.min` as we know that we have 3 entries here.  Having the default case does not seem useful.

src/plots/barPlot.ts
jtlan: I think it's better to not have potentially `undefined` values floating around, which is why I assigned it to `true`.

src/plots/barPlot.ts
bluong: It took me a while to understand the reasoning here.  Why are we using a `min` here instead of looping through the 3 cases?

src/plots/barPlot.ts
bluong: While this is true, can we confidently say for certain that this variable is usable before it is set below?  If a developer were to put code between this line and the immediate lines below involving `showLabelOnBar`, would that code work?

jtlan: Not necessarily, but neither would it work if the value was `undefined` -- worse, Javascript would allow `if (undefined)`, so it might be less clear where the error is coming from.

src/plots/barPlot.ts
jtlan: I'll switch this to an `if`-tree which is more readable.

crmorford:
(css) http://jsfiddle.net/ts720tdr/39/

PR #2783, opened 2015-09-18
Write a test to guarantee DOM elements are removed from previous svg element when switching svg element 

### LINKED ISSUE #2739: [Component] - Switching anchored svg elements does not detach
When a Component is already `anchor()`ed to an svg element, calling `anchor()` on a different svg element does not remove the DOM elements related to the component from the previous svg element.  Likely, we just need to detach from the previous svg.

yunhaolucky:
Suppose a component 'c' is `ancher()`ed to `svg1`. And we change its svg node from `svg1` to `svg2` by running `c.archer(svg2)`. We expect all DOM Elements move from `svg1` to `svg2`. That is to say `svg1` has no child node.
This pull request includes a test to guarantee this.

close #2739

### DIFF
--- a/test/components/componentTests.ts
+++ b/test/components/componentTests.ts
@@ -78,12 +78,12 @@ describe("Component", () => {
       svg.remove();
     });
 
-    it("remove dom elements in previous svg element when switching element", () => {
+    it("remove DOM elements in previous svg when anchoring to a different svg", () => {
       c.anchor(svg);
       let svg2 = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
       c.anchor(svg2);
-      assert.isTrue(svg.select("g").empty(), "previous svg element should not have any group childnodes");
-      assert.isFalse(svg2.select("g").empty(), "new svg element should have group childnodes.");
+      assert.isTrue(svg.select("g").empty(), "previous svg element should not have any group child nodes");
+      assert.isFalse(svg2.select("g").empty(), "new svg element should have group child nodes");
       c.destroy();
       svg.remove();
       svg2.remove();
--- a/test/components/componentTests.ts
+++ b/test/components/componentTests.ts
@@ -171,6 +171,7 @@ describe("Component", () => {
       assert.isTrue((<Node> svg.node()).hasChildNodes(), "the svg has children");
       c.detach();
       assert.isFalse((<Node> svg.node()).hasChildNodes(), "the svg has no children");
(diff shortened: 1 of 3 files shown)


test/components/componentTests.ts
```diff
@@ -78,6 +78,17 @@ describe("Component", () => {
       svg.remove();
     });
 
+    it("remove dom elements in previous svg element when switching element", () => {
+      c.anchor(svg);
+      let svg2 = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      c.anchor(svg2);
+      assert.isFalse((<Element>svg.node()).hasChildNodes(), "previous svg element should not have any childnodes");
+      assert.isTrue((<Element>svg2.node()).hasChildNodes(), "new svg element should have childnodes.");
```
bluong: I think I have tried `hasChildNodes()` before with sad times from PhantomJS.  Alternatively, you can try `selection.empty()`

bluong: Can you try checking if the `svg` is empty instead to avoid the annoying cast?  If that doesn't work, this is fine it looks like

test/components/componentTests.ts
```diff
@@ -78,6 +78,17 @@ describe("Component", () => {
       svg.remove();
     });
 
+    it("remove dom elements in previous svg element when switching element", () => {
```
bluong: change sentence to "removes DOM elements in previous svg when anchoring to a different svg"

test/components/componentTests.ts
```diff
@@ -78,6 +78,17 @@ describe("Component", () => {
...
+      c.anchor(svg);
+      let svg2 = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+      c.anchor(svg2);
+      assert.isTrue(svg.select("g").empty(), "previous svg element should not have any group childnodes");
+      assert.isFalse(svg2.select("g").empty(), "new svg element should have group childnodes.");
```
bluong: Remove period I think.

bluong: "child notes"*

test/components/componentTests.ts
```diff
@@ -78,6 +78,17 @@ describe("Component", () => {
       svg.remove();
     });
 
+    it("remove DOM elements in previous svg when anchoring to a different svg", () => {
```
bluong: removes*

PR #2787, opened 2015-09-18
Support Formatter on Legend

### LINKED ISSUE #2125: Enhancement Request: Formatter on legend
I want to make it obvious these are percents
screen shot 2015-05-20 at 4 11 58 pm

yunhaolucky:
**Feature**
Display a different name on an ordinary Legend than the values in the colorScale domain using customized `Formatter`.`Formatter` is defined as `(d:any) => string`
```typescript
class Legend{
/**
* Gets the Formatter for the text.
*/
formatter(): Formatter;
/**
* Sets the Formatter for the text.
*
* @param {Formatter} formatter
* @returns {Legend} The calling Legend.
*/
formatter(formatter: Formatter): Legend;
}
```

Using  `formatter` will affect the behavior of `comparator`. We decide to make `comparator` work on formatted texts.

**JSFiddle**
Before:http://jsfiddle.net/yunhaolucky/cmf3ut7g/4/
After:http://jsfiddle.net/yunhaolucky/cmf3ut7g/3/

close #2125

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2160,6 +2160,17 @@ declare module Plottable {
             constructor(colorScale: Scales.Color);
             protected _setup(): void;
             /**
+             * Gets the Formatter for the text.
+             */
+            formatter(): Formatter;
+            /**
+             * Sets the Formatter for the text.
+             *
+             * @param {Formatter} formatter
+             * @returns {Legend} The calling Legend.
+             */
+            formatter(formatter: Formatter): Legend;
+            /**
              * Gets the maximum number of entries per row.
              *
              * @returns {number}
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2160,11 +2160,11 @@ declare module Plottable {
             constructor(colorScale: Scales.Color);
             protected _setup(): void;
             /**
-             * Gets the Formatter for the text.
+             * Gets the Formatter for the entry texts.
              */
             formatter(): Formatter;
             /**
-             * Sets the Formatter for the text.
+             * Sets the Formatter for the entry texts.
              *
              * @param {Formatter} formatter
              * @returns {Legend} The calling Legend.
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -68,7 +68,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): InterpolatedColorLegend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === null) {
+      if (formatter === undefined) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -68,7 +68,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): InterpolatedColorLegend;
     public formatter(formatter?: Formatter): any {
-      if (formatter == null) {
+      if (formatter === undefined) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -68,7 +68,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): InterpolatedColorLegend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === undefined) {
+      if (formatter === null) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/interpolatedColorLegend.ts
+++ b/src/components/interpolatedColorLegend.ts
@@ -68,7 +68,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): InterpolatedColorLegend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === undefined) {
+      if (formatter == null) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -168,9 +168,8 @@ export module Components {
       this._colorScale.domain().slice().forEach((d) => {
         entryLabels.set(this._formatter(d), d);
       });
-      let entryNames = entryLabels.keys();
-      entryNames = entryNames.sort(this.comparator()).map((d) => entryLabels.get(d));
-
+      let formattedEntryNames = entryLabels.keys();
+      let entryNames = formattedEntryNames.sort(this.comparator()).map((d) => entryLabels.get(d));
       let entryLengths: d3.Map<number> = d3.map<number>();
       let untruncatedEntryLengths: d3.Map<number> = d3.map<number>();
       entryNames.forEach((entryName) => {
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -75,7 +75,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): Legend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === undefined) {
+      if (formatter == null) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -18,6 +18,7 @@ export module Components {
 
     private _padding = 5;
     private _colorScale: Scales.Color;
+    private _formatter: Formatter;
     private _maxEntriesPerRow: number;
     private _comparator: (a: string, b: string) => number;
     private _measurer: SVGTypewriter.Measurers.Measurer;
@@ -45,7 +46,7 @@ export module Components {
       this._colorScale = colorScale;
       this._redrawCallback = (scale) => this.redraw();
       this._colorScale.onUpdate(this._redrawCallback);
-
+      this._formatter = Formatters.identity();
       this.xAlignment("right").yAlignment("top");
       this.comparator((a: string, b: string) => this._colorScale.domain().indexOf(a) - this._colorScale.domain().indexOf(b));
       this._symbolFactoryAccessor = () => SymbolFactories.circle();
@@ -62,6 +63,25 @@ export module Components {
       this._writer = new SVGTypewriter.Writers.Writer(this._measurer, this._wrapper).addTitleElement(Configs.ADD_TITLE_ELEMENTS);
     }
 
+    /**
+     * Gets the Formatter for the text.
+     */
+    public formatter(): Formatter;
+    /**
+     * Sets the Formatter for the text.
+     *
+     * @param {Formatter} formatter
+     * @returns {Legend} The calling Legend.
+     */
+    public formatter(formatter: Formatter): Legend;
+    public formatter(formatter?: Formatter): any {
+      if (formatter === null) {
+        return this._formatter;
+      }
+      this._formatter = formatter;
+      this.redraw();
+      return this;
+    }
     /**
      * Gets the maximum number of entries per row.
      *
@@ -150,7 +170,7 @@ export module Components {
       let entryLengths: d3.Map<number> = d3.map<number>();
       let untruncatedEntryLengths: d3.Map<number> = d3.map<number>();
       entryNames.forEach((entryName) => {
-        let untruncatedEntryLength = textHeight + this._measurer.measure(entryName).width + this._padding;
+        let untruncatedEntryLength = textHeight + this._measurer.measure(this._formatter(entryName)).width + this._padding;
         let entryLength = Math.min(untruncatedEntryLength, availableWidthForEntries);
         entryLengths.set(entryName, entryLength);
         untruncatedEntryLengths.set(entryName, untruncatedEntryLength);
@@ -299,8 +319,7 @@ export module Components {
                         yAlign: "top",
                         textRotation: 0
                       };
-
-                      self._writer.write(value, maxTextLength, self.height(), writeOptions);
+                      self._writer.write(self._formatter(value), maxTextLength, self.height(), writeOptions);
                     });
       return this;
     }
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -64,11 +64,11 @@ export module Components {
     }
 
     /**
-     * Gets the Formatter for the text.
+     * Gets the Formatter for the entry texts.
      */
     public formatter(): Formatter;
     /**
-     * Sets the Formatter for the text.
+     * Sets the Formatter for the entry texts.
      *
      * @param {Formatter} formatter
      * @returns {Legend} The calling Legend.
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -75,7 +75,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): Legend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === null) {
+      if (formatter === undefined) {
         return this._formatter;
       }
       this._formatter = formatter;
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -164,12 +164,7 @@ export module Components {
 
       let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
 
-      let entryLabels: d3.Map<string> = d3.map<string>();
-      this._colorScale.domain().slice().forEach((d) => {
-        entryLabels.set(this._formatter(d), d);
-      });
-      let formattedEntryNames = entryLabels.keys();
-      let entryNames = formattedEntryNames.sort(this.comparator()).map((d) => entryLabels.get(d));
+      let entryNames = this._colorScale.domain().slice().sort((a,b) => this._comparator(this._formatter(a), this._formatter(b)));
       let entryLengths: d3.Map<number> = d3.map<number>();
       let untruncatedEntryLengths: d3.Map<number> = d3.map<number>();
       entryNames.forEach((entryName) => {
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -164,7 +164,7 @@ export module Components {
 
       let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
 
-      let entryNames = this._colorScale.domain().slice().sort((a,b) => this._comparator(this._formatter(a), this._formatter(b)));
+      let entryNames = this._colorScale.domain().slice().sort((a, b) => this._comparator(this._formatter(a), this._formatter(b)));
       let entryLengths: d3.Map<number> = d3.map<number>();
       let untruncatedEntryLengths: d3.Map<number> = d3.map<number>();
       entryNames.forEach((entryName) => {
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -164,8 +164,12 @@ export module Components {
 
       let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
 
-      let entryNames = this._colorScale.domain().slice();
-      entryNames.sort(this.comparator());
+      let entryLabels: d3.Map<string> = d3.map<string>();
+      this._colorScale.domain().slice().forEach((d) => {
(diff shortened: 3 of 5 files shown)


src/components/interpolatedColorLegend.ts
```diff
@@ -68,7 +68,7 @@ export module Components {
      */
     public formatter(formatter: Formatter): InterpolatedColorLegend;
     public formatter(formatter?: Formatter): any {
-      if (formatter === undefined) {
+      if (formatter === null) {
```
yunhaolucky: This is bug in `interpolatedColorLegend`, not part of this feature.

bluong: Can you explain the bug?  The way that you have currently implemented this change completely changes the functionality of this function.  Essentially, I believe the getter is now called with `icl.formater(null);` which seems very undesirable.

bluong: If this is a bug fix for `InterpolatedColorLegend` that seems to be separate from this pull request, please move it out into a separate pull request.

src/components/legend.ts
```diff
@@ -63,6 +64,25 @@ export module Components {
...
+    public formatter(): Formatter;
+    /**
+     * Sets the Formatter for the text.
+     *
+     * @param {Formatter} formatter
+     * @returns {Legend} The calling Legend.
+     */
+    public formatter(formatter: Formatter): Legend;
+    public formatter(formatter?: Formatter): any {
+      if (formatter === null) {
```
bluong: This seems incorrect.  Now the getter can only be called with `legend.formatter(null)` it seems.

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("accept formatter and show entity labels correcly", () => {
+    color.domain(["jtao", "mschafer", "kfalter"]);
+    let data = [
+      {id: "jtao", name: "Joy Tao", score: 13},
+      {id: "mschafer", name: "Mark Schafer", score: 15},
+      {id: "kfalter", name: "Kelsey Falter", score: 11}
+    ];
```
bluong: I personally am find with using first names of people but unsure about using full names of people.  Would it make sense to just use first names?

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+    let f = (id: string) => {
```
bluong: `let formatter = ...`

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("accept formatter and show entity labels correcly", () => {
```
bluong: how about
`it("can set which formatter to use to change how entry labels are displayed"...`?

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+      {id: "mschafer", name: "Mark Schafer", score: 15},
+      {id: "kfalter", name: "Kelsey Falter", score: 11}
+    ];
+    let f = (id: string) => {
+      let pickedEntity = data.filter(function(d) {return (d.id === id); });
+      if (pickedEntity.length > 0) {
+          return pickedEntity[0].name;
+        }else {
+          return "NA";
+        }
```
bluong: Indentation issues above.

bluong: Instead of the logic above, why can't we create a map instead of a data array?  It'd be immensely easier to work with here.

Just noting that it seems you're adding logic to this formatter that seems unnecessary (this "NA" case, the `pickedEntity.length > 0` case), which can make deciphering this test harder to understand when we look at this later.

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+    };
+    legend.formatter(f);
+    legend.renderTo(svg);
+
+    let texts = svg.selectAll("title")[0].map(function(text: Element) {return text.textContent; });
```
bluong: Why are we looking for all of the `title`'s on the svg?  While I think this is one way we can check the formatter, a lot more useful check I think is to check for the text inside each entry.

bluong: I believe D3 has a `text()` API that can be incredibly useful.

https://github.com/mbostock/d3/wiki/Selections#text

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+    data.forEach(function(d, i) {
+      let idExist = (texts.indexOf(d.id) !== -1);
+      let nameExist = (texts.indexOf(d.name) !== -1);
+      assert.isTrue(nameExist, `formatter output ${d.name} should be drawn`);
+      assert.isFalse(idExist, ` formatter input ${d.id} should not be drawn`);
+    });
+     svg.remove();
```
bluong: Weird indentation issues.

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+    };
+    legend.formatter(f);
+    legend.renderTo(svg);
+
+    let texts = svg.selectAll("title")[0].map(function(text: Element) {return text.textContent; });
+    data.forEach(function(d, i) {
+      let idExist = (texts.indexOf(d.id) !== -1);
+      let nameExist = (texts.indexOf(d.name) !== -1);
+      assert.isTrue(nameExist, `formatter output ${d.name} should be drawn`);
+      assert.isFalse(idExist, ` formatter input ${d.id} should not be drawn`);
```
bluong: I think checking the latter is not important.  I think we only care that the formatter works

test/components/legendTests.ts
bluong: The `score` parameter is not used anywhere.  I think we can just delete this parameter

test/components/legendTests.ts
```diff
@@ -254,6 +254,34 @@ describe("Legend", () => {
...
+        }
+    };
+    legend.formatter(f);
+    legend.renderTo(svg);
+
+    let texts = svg.selectAll("title")[0].map(function(text: Element) {return text.textContent; });
+    data.forEach(function(d, i) {
+      let idExist = (texts.indexOf(d.id) !== -1);
+      let nameExist = (texts.indexOf(d.name) !== -1);
+      assert.isTrue(nameExist, `formatter output ${d.name} should be drawn`);
```
bluong: Instead of using `assert.isTrue` can we use something that's more useful?  Maybe loop over the appropriate domain value and apply the locally defined formatter to it?

test/components/legendTests.ts
bluong: There's no test above for the `getter` for the formatter.  This is incredibly important since I have a suspicion that the getter does not work.

plottable.d.ts
```diff
@@ -2160,6 +2160,17 @@ declare module Plottable {
             constructor(colorScale: Scales.Color);
             protected _setup(): void;
             /**
+             * Gets the Formatter for the text.
+             */
+            formatter(): Formatter;
+            /**
+             * Sets the Formatter for the text.
```
bluong: Hmm maybe be more specific to which text you're referring?  Maybe
"Sets the Formatter for the entries (or entry texts)."?  I would ping to get a second opinion

src/components/legend.ts
```diff
@@ -144,13 +164,17 @@ export module Components {
...
       let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
 
-      let entryNames = this._colorScale.domain().slice();
-      entryNames.sort(this.comparator());
+      let entryLabels: d3.Map<string> = d3.map<string>();
+      this._colorScale.domain().slice().forEach((d) => {
+        entryLabels.set(this._formatter(d), d);
+      });
+      let entryNames = entryLabels.keys();
+      entryNames = entryNames.sort(this.comparator()).map((d) => entryLabels.get(d));
```
bluong: using a map here seems very roundabout.  Why not `entryNames = entryNames.sort(this.comparator()).map(this.formatter())`?

test/components/legendTests.ts
```diff
@@ -254,6 +254,35 @@ describe("Legend", () => {
...
+    color.domain(["jtao", "mschafer", "kfalter"]);
+    let data: {[id: string]: string; } = {"jtao": "Joy", "mschafer": "Mark", "kfalter": "Kelsey"};
+    let formatter = (id: string) => {
+        return data[id];
+    };
+    legend.formatter(formatter);
+    legend.renderTo(svg);
+    assert.strictEqual(legend.formatter(), formatter, "formatter() return the formatter of legend correctly");
+    svg.remove();
+  });
```
bluong: the `renderTo` line above does not really do anything.  Remove it?

src/components/legend.ts
yunhaolucky: I want to sort the labels by `this.formatter(entryID)` instead of `entryID`.  Is there any good way to do something like `entryNames.sort(this.comparator(this.formatter()))`?

bluong: Please check if this is what people want first. If this is indeed the case, you can map the formatter over all of the labels and then apply the sort then

`entryNames.map(this.formatter()).sort(this.comparator())`

bluong:
Please check what people want out of this

src/components/legend.ts
yunhaolucky: The reason I did this is I want to have a array of scale domain names sorted by displayed texts. If I do `entryNames.map(this.formatter()).sort(this.comparator())`, I will not get the domain names back. I need domain names to pick the color in ColorScale.

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+    let data: {[id: string]: string; } = {"jtao": "Joy", "mschafer":  "Mark", "kfalter": "Kelsey"};
+    let formatter = (id: string) => data[id];
+    legend.formatter(formatter);
+    legend.renderTo(svg);
+    let legendRows = svg.selectAll(entrySelector);
+    assert.operator(legendRows.size(), ">=", 1, "There is at least one entry in the test");
+    legendRows.each(function(d: Element, i: number){
+      let expectText = formatter(legend.colorScale().domain()[i]);
+      assert.strictEqual(d3.select(this).select("text").text(), expectText,
+        `formatter output ${d3.select(this).select("text").text()} should be displayed`);
```
bluong: Shouldn't the test message include `expectText` instead of the actual response?

src/components/legend.ts
bluong: > I need domain names to pick the color in ColorScale.

What do you mean by this?

bluong: The variable names here can be quite a bit confusing.

It seems that `entryNames` is trying to be a variable for 2 different concepts.  This can be bad, because if one of the 2 concepts needs to change, then the variable here has to be drastically modified.  One is that it holds the formatted transform of the colorScale's domain.  Two is that is also eventually holds the original colorScale's domain in sorted form.

Can we separate these two variables out and see how it reads?

src/components/legend.ts
```diff
@@ -144,13 +164,17 @@ export module Components {
 
       let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
 
-      let entryNames = this._colorScale.domain().slice();
-      entryNames.sort(this.comparator());
+      let entryLabels: d3.Map<string> = d3.map<string>();
+      this._colorScale.domain().slice().forEach((d) => {
+        entryLabels.set(this._formatter(d), d);
```
bluong: Curious why this isn't `entryLabels.set(d, this._formatter(d))`.  One usually has a mapping from a raw value to a computed value, not usually the other way around.

bluong:
Semantics semantics

src/components/legend.ts
yunhaolucky: To get computed value from raw value, I can use function formatter(value). In this case, I want to get raw value from computed value. That's why I want to use a map to store those values.

src/components/legend.ts
yunhaolucky: >What do you mean by this?

 We use `this._colorScale.scale(domainNames)` to get the color of a entry symbol. And formatter can only translate from domain names to displayed texts. Therefore, I need to store domain names somewhere. That's the reason I want an array of domain names sorted by its corresponding displayed texts instead of an array of sorted displayed texts.

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+      let expectText = formatter(legend.colorScale().domain()[i]);
+      assert.strictEqual(d3.select(this).select("text").text(), expectText,
+        `formatter output ${expectText} should be displayed`);
+    });
+    svg.remove();
+  });
+
+  it("can get formatter of the legend using formatter()", () => {
+    color.domain(["jtao", "mschafer", "kfalter"]);
+    let data: {[id: string]: string; } = {"jtao": "Joy", "mschafer": "Mark", "kfalter": "Kelsey"};
```
bluong: The above two lines are not too relevant

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+    let formatter = (id: string) => data[id];
```
bluong: We can just use a very simple formatter here

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+    color.domain(colorDomain);
+    let expectedTexts = ["Z", "Y", "X"];
+    let formatter = (d: string) => expectedTexts[colorDomain.indexOf(d)];
+    legend.formatter(formatter);
+    let comparator = (a: string, b: string) => a.localeCompare(b);
+    legend.comparator(comparator);
+    legend.renderTo(svg);
+    let entryTexts = svg.selectAll(entrySelector)[0].map((node: Element) => d3.select(node).select("text").text());
+    expectedTexts.sort(comparator);
+    assert.deepEqual(expectedTexts, entryTexts, "formatted entries have been sorted as expected");
```
bluong: what does "sorted as expected" mean?

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+  it("can get formatter of the legend using formatter()", () => {
+    color.domain(["jtao", "mschafer", "kfalter"]);
+    let data: {[id: string]: string; } = {"jtao": "Joy", "mschafer": "Mark", "kfalter": "Kelsey"};
+    let formatter = (id: string) => data[id];
+    legend.formatter(formatter);
+    assert.strictEqual(legend.formatter(), formatter, "formatter() returns the formatter of legend correctly");
+    svg.remove();
+  });
+
+  it("comparator() works as expected when customized formatter is applied", () => {
```
bluong: Should comment that it sorts the formatted text

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+    let colorDomain = ["A", "B", "C"];
+    color.domain(colorDomain);
+    let expectedTexts = ["Z", "Y", "X"];
+    let formatter = (d: string) => expectedTexts[colorDomain.indexOf(d)];
+    legend.formatter(formatter);
+    let comparator = (a: string, b: string) => a.localeCompare(b);
```
bluong: Why don't we use a more simple comparator then `a.localeCompare(b)`

test/components/legendTests.ts
```diff
@@ -254,6 +254,46 @@ describe("Legend", () => {
...
+    legend.comparator(comparator);
+    legend.renderTo(svg);
+    let entryTexts = svg.selectAll(entrySelector)[0].map((node: Element) => d3.select(node).select("text").text());
+    expectedTexts.sort(comparator);
```
bluong: Why do we have to sort the `expectedTexts` as well?

src/components/legend.ts
```diff
@@ -144,13 +164,16 @@ export module Components {
...
 
-      let entryNames = this._colorScale.domain().slice();
-      entryNames.sort(this.comparator());
-
+      let entryLabels: d3.Map<string> = d3.map<string>();
+      this._colorScale.domain().slice().forEach((d) => {
+        entryLabels.set(this._formatter(d), d);
+      });
+      let formattedEntryNames = entryLabels.keys();
+      let entryNames = formattedEntryNames.sort(this.comparator()).map((d) => entryLabels.get(d));
```
bluong: Would this work?
`let entryNames = this._colorScale.domain().sort((a, b) => this.comparator()(this.formatter()(a), this.formatter()(b));`

This would get rid of the the `entryLabels` and the `formattedEntryNames` variable and I believe this is still readable

bluong:
The unit tests can be made quite a bit more simpler / easier to understand.

Also we can tighten the logic of the source code by quite a bit.

test/components/legendTests.ts
yunhaolucky: My logic is:
Given a sorted array of domain names and an unsorted array of expected texts, the actual displayed texts should be sorted. So I need to compare the actual displayed texts and sorted expected texts to see whether comparator sorts in the order of displayed texts.

src/components/legend.ts
```diff
@@ -45,9 +46,9 @@ export module Components {
       this._colorScale = colorScale;
       this._redrawCallback = (scale) => this.redraw();
       this._colorScale.onUpdate(this._redrawCallback);
-
+      this._formatter = Formatters.identity();
       this.xAlignment("right").yAlignment("top");
-      this.comparator((a: string, b: string) => this._colorScale.domain().indexOf(a) - this._colorScale.domain().indexOf(b));
+      this.comparator((a: string, b: string) => 0);
```
bluong: Why this change?

test/components/legendTests.ts
```diff
@@ -254,6 +254,43 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("can set formatter and change how entry labels are displayed", () => {
```
bluong: "can set formatter to change..."

test/components/legendTests.ts
```diff
@@ -254,6 +254,43 @@ describe("Legend", () => {
...
+    legendRows.each(function(d, i){
+      let expectText = formatter(legend.colorScale().domain()[i]);
+      assert.strictEqual(d3.select(this).select("text").text(), expectText,
+        `formatter output ${expectText} should be displayed`);
+    });
+    svg.remove();
+  });
+
+  it("can get formatter of the legend using formatter()", () => {
+    let formatter = (id: string) => String.fromCharCode(id.charCodeAt(0) + 1);
```
bluong: I personally like something even simpler like `(id: string) => \`${id}foo\``

The problem that I have with this formatter is that there are 3 things happening in order to achieve the formatted value, which is just needless overhead if people needed to decipher what happens if a test were to fail with this formatter

src/components/legend.ts
yunhaolucky: The parameters of the old `comparator` are domain names. It fails when `a` and `b` are displayed texts. The new comparator will not sort displayed texts, whose order is the same as the order of the index of domain names.

test/components/legendTests.ts
```diff
@@ -254,6 +254,43 @@ describe("Legend", () => {
...
 
+  it("can set formatter to change how entry labels are displayed", () => {
+    color.domain(["A", "B", "C"]);
+    let formatter = (id: string) => `${id}foo`;
+    legend.formatter(formatter);
+    legend.renderTo(svg);
+    let legendRows = svg.selectAll(entrySelector);
+    assert.operator(legendRows.size(), ">=", 1, "There is at least one entry in the test");
+    legendRows.each(function(d, i){
+      let expectText = formatter(legend.colorScale().domain()[i]);
```
bluong: Sorry for the late catch here.

`expectedText` please

crmorford:
http://jsfiddle.net/cmf3ut7g/5/
http://jsfiddle.net/cmf3ut7g/6/

PR #2790, opened 2015-09-21
[Typescript] - Upgrading to 1.6

bluong:
Typescript has recently released its 1.6 version: http://blogs.msdn.com/b/typescript/archive/2015/09/16/announcing-typescript-1-6.aspx.  We should upgrade our compiler so we can freely take advantage of the language's new features.

A few new features that we could be excited about:
* Class expressions (Very adhoc class creation that "may" be useful for testing)
* User defined type guards (A customizable way to type-guard variables, which can help us a ton on input validation)
* Abstract classes (Something that I think we wanted for about a year and can help us a ton in areas where we throw errors when a function is not defined (Note that this could lead us into an API break))
* Generic type aliases (There is a HACKHACK surrounding this that we can now resolve)

QE: This is a version bump on a dependency that doesn't give us any immediate features in upgrading.  A regression pass should be the only thing required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "grunt-contrib-connect": "~0.6.0",
     "grunt-blanket-mocha": "~0.4.0",
     "grunt-saucelabs": "8.6.0",
-    "typescript": "1.5.3",
+    "typescript": "1.6.2",
     "grunt-sed": "~0.1.1",
     "grunt-git": "~0.2.6",
     "grunt-contrib-clean": "~0.4.0",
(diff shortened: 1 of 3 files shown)


jtlan:
I'd argue that classes like `Component` aren't strictly abstract.

jtlan:
Probably just a quick regression pass here.

bluong:
Yes, regarding abstract classes, I think we will need to rethink how our classes function.  I believe that now we can have a clean separation between: classes that cannot operate without needing to override (abstract) and classes that can operate with its current implementation (concrete).

In this regard, I think `Component` lies in the nice spot that yes it can act as a concrete class (likely needs rethinking and making sure), even though it is a parent class.

However, `Scale` is a parent class that has an implementation that is not complete.  It will error if people try to interact with it.  In this regard, `Scale` might need to be made `abstract`, or we need to look at how we can make `Scale` concrete.

PR #2796, opened 2015-09-22
Add DragZoomLayer

teamdandelion:
DragZoomLayer creates a SelectionBox that powers a drag-to-zoom interaction. When you drag a box, the scales zoom into the bounds drawn out in that box and the box disappears. The zoom can be animated over a variable amount of time (default 750ms) with a custom easing function (default d3.cubic-in-out). Tests included.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -595,6 +595,9 @@ declare module Plottable {
         selection: d3.Selection<any>;
         component: C;
     }
+    type Numeric = number | {
+        valueOf(): number;
+    };
 }
 
 
@@ -2625,9 +2628,6 @@ declare module Plottable {
 
 declare module Plottable {
     module Components {
-        type Numeric = number | {
-            valueOf(): number;
-        };
         class DragZoomLayer extends Components.SelectionBoxLayer {
             constructor(xScale: QuantitativeScale<Numeric>, yScale: QuantitativeScale<Numeric>);
             animationTime(): number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2449,6 +2449,12 @@ declare module Plottable {
         }
         class SelectionBoxLayer extends Component {
             protected _box: d3.Selection<void>;
+            protected _xScale: QuantitativeScale<number | {
+                valueOf(): number;
+            }>;
+            protected _yScale: QuantitativeScale<number | {
+                valueOf(): number;
+            }>;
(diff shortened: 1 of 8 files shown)


jtlan:
@danmane,

I like the idea here, but I worry it might be a bit too narrowly scoped:
* What if the user wants this effect, but only in X or Y?
* What if the user doesn't want double-click to unzoom?

A more modular implementation might be to have a domain-setter on `QuantitativeScale` that allows passing in an easing function and duration:

```Javascript
scale.domain(newDomain, easingFunction, duration);
```

Calling this function in the `onDragEnd()` callback of a `DragBoxLayer` could provide the smooth zoom-in behavior. As an added bonus, it could be used with `XDragBoxLayer`, `YDragBoxLayer`, and anywhere else a user might want to smoothly transition the domain.

Let us know if you want some help implementing this.

danmane:
Yes, that does seem like a good API point to have (probably can be useful in other contexts as well).

Re: whether this is too narrowly scoped: I agree it is pretty specific to a particular usecase. I feel that having this code exist in public is still useful though, so that Plottable is a bit more "batteries included" with the ability to just do basic charting-type tasks (eg crosshairs, zooming) in a sane configuration without needing to think about XYDragBoxLayers vs XDragBoxLayers and DragEnd callbacks and so forth. Obviously I *love* the customizability and that this flexibility exists, but it might be nice to provide a path to simplicity for those who want it.

Maybe one way to tackle this would be to break out repositories, plottable-core and plottable-koans or something, where plottable-koans contains pieces like DragZoomLayer that are a bit narrower in scope but are both useful pieces of code in their own right, and provide examples of how you can use the pieces provided by plottable-core.

Or you could imagine going more modular still and encouraging people to publish this kind of code to bower and keep a registry of "plottable modules" (although now you start to have severe versioning issues if you're ever breaking backcompat, not sure if you are...)

crmorford:
Hey @danmane,
Just checking in on status -- are you working on this PR, or waiting on us to approve something? I think we only want changes to scales.domain() at this point; the discussion about -core and -koans belongs somewhere else.

Would you like to implement the changes to .domain()? Or should we try to schedule it?

danmane:
Sorry, I probably won't have time to pay attention to this for a few weeks. If you can take over putting in the domain change that would be great.

CalvinFernandez:
@danmane landing further changes here #3136 . Thanks!

Issue #2707, opened 2015-09-01
Animator behavior can be altered by calling datasets.data() ???


crmorford:
http://jsfiddle.net/k0os8wyx/
repro:
1.watch plot animate for a moment (it goes slow)
2. click on the plot
3. watch the plot re-animate, but from 2 spots

Plot will also reanimate from x spots if you click x times.
Plot will also reanimate from multiple spots if you let it finish but then click again.

jtlan:
I suspect this is because animations are being done via timeouts. We should add code that detects when the `Plot` is being re-rendered and cancel any running animations.

softwords:
The timeouts are trying to sequence the animation steps, which I think would be handled more robustly by chaining transitions. That is , instead of trying to wait for a transition to finish, before launching a new transition on the same selection, queue the second transition by chaining it from the first. The chain will all be killed if a new transition is applied directly to the underlying selection.

softwords:
@jtlan I have finally (I hope) got on top of all the architecture so that I feel confident to launch a pull request begin to address object constancy. I mention it here because I think at essence this "bug" relates to the svg nodes not "really" being destroyed - rather the animation logic tries to make it appear that they are. With a NoConstancy key, this problem goes away - see http://jsfiddle.net/46dtLjxy/1/

Issue #2512, opened 2015-07-28
Enable 'object constancy' on data refresh by supporting a key function


softwords:
'Object constancy' , as defined in http://bost.ocks.org/mike/constancy/ ,
requires a 'key' function passed to selection.data()  to provide a unique identifier for each datum. This determines what goes in enter exit and update selections.

Could dataset get a key method, and/or enhanced constructor, to specify this key function? 

Then, we'd need some way to get to the enter exit and update d3 transitions.

jtlan:
@softwords, can you explain how you're planning to use this feature? Is there a custom animation you're trying to implement?

softwords:
@jtlan: here's a quick jsfiddle http://jsfiddle.net/d4gngj2s/2/ that I hope will illustrate. 

The data comes from World Cup finals 2006,2010, 2014. For the last 32 teams, we have name, rank, goals for, goals against. The scatter chart plots a point for each team, x= GF, y = GA. The Color of the point encodes the rank.
Now - when I change from one year to another , I want a point that represents a team to move to its new position, and transition from its current color to its new color. Also, some teams drop out - these points could (e.g.) transition size=>0 or opacity=>0 before being removed. New teams will come in - these could transition size , opacity in the opposite direction. 
But to make this work, you need the key function to determine who stays in, who continues, who drops out. There are always 32 points - so if the default key (ie index) is used, there will never be enters or exits after the initial load. In my data, the key function should be: function(d) {return d.name;}

softwords:
Hello @jtlan 

I've investigated this a bit further, may I share some observations:

```Drawer._bindSelectionData()``` is the point where a key function would need to be accessed, in order that enter() and exit() are correctly identified with respect to this key. Since Drawer holds a reference to the Dataset, the key could be accessed from the Dataset metadata; e.g. by the convention that a member named _key_ in the _metadata_ object is a function returning the key.

so you'd have something like

```javascript
var ds = new Plottable.Dataset(data)
         .metadata({key: function(d,i) { return d.name;}});
```
and  ```Drawer._bindSelectionData``` could look for this:

```javascript
var dataElements;
	if (this._dataset && this._dataset.metadata && this.metadata.key) {
		dataElements = this.selection().data(data, this._dataset.metadata.key);
	}
	else {
		dataElements = this.selection().data(data);
	};
```

[...]

PR #2801, opened 2015-09-23
Enable 'object constancy' on data refresh by supporting a key function in Dataset

softwords:
see issue #2512 
This pull request 
- adds the key function to Dataset
- uses this key in Drawer when binding data to the selection
- provides 2 predefined keys:
-- NoConstancy - which will return a new value each time it is called, therefire ensuring that there is _never_ object constancy; that is, all elements in the selection are removed and recreated
-- ByIndex - returns the index of the datum n the array, which is the default behaviour of d3 anyway

- Also removes the logic on AnimateOnNextRender in ScatterPlot. This logic simulates the destruction and recreation of each svg element to create an animation effect. But, if the NoConstancy key is used, d3 destroys and recreates the svg elements itself, so there is no need for this workaround.
This also resolves issue #2707.


bluong:
There seems to be a lot of drastic whitespace changes (due to indentation, seemingly).  Can you change the pull request such that the tabs stay consistent with the rest of the code base?

softwords:
@bluong; >>drastic whitespace changes
yes I can see this now - it's a bit of a mess. I think the simplest path may be to start over

bluong:
@softwords Feel free to reach out if you need any help in development!

softwords:
@bluong thanks for the encouragement! I'll have another attempt to get this one ready, and get comfortable with the workflow.

PR #2650, opened 2015-08-18
[Do not merge] Line performance games

aicioara:
Just for the jsFiddle POC

Issue #2523, opened 2015-07-30
Improve performance of line graphs


christinaforney:
Contour has a use case with >5000 lines on a plot that is very slow. Talk to them for an example.

aicioara:
Fiddling

This is a repro of the issue http://jsfiddle.net/cvxtwuh4/
However, they are running it on `1.0.0-rc6`, so it behaves like this: //TODO

aicioara:
Possible this will be it http://jsfiddle.net/cvxtwuh4/4/

Otherwise the new trick would be to only draw lines for the viewport, start doing some smart opts. 

**Edit** The above is not performant enough, will do the viewport trick

aicioara:
POC: 

Plot with 50 000 points

Before: http://jsfiddle.net/jncLsmvd/2/
After: http://jsfiddle.net/jncLsmvd/6/

crmorford:
@aicioara what is POC an acronym for? My brain keeps autofilling "point of contact" but I know that's not right :)

aicioara:
Proof of Concept

bluong:
Welp, I personally thought it was "Point of Context"...

aicioara:
I thought POC is widely used as an acronym. Interesting.

christinaforney:
Cropped rendering and downsampling complete this.

PR #2810, opened 2015-09-24
Cropped Rendering Performance Option

aicioara:
Part of #2523 

Performance request from Contour.

Affects:
- Line Plot
- Area Plot
- Stacked Area Plot

New API points

- `Plots.Line`

```typescript
/**
 * Gets the croppedRendering preformance option state.
 *
 * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
 * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
 * then this option will boost of rendering. However, if all the data is rendered anyway, having this
 * option enabled will cause a small overhead that can be noticed in the total render time.
 */
public croppedRenderingEnabled(): boolean;

/**
 * Sets the croppedRendering performance option.
 */
public croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
```

Demo:
- http://jsfiddle.net/yxpq4uuv/4/ - with feature enabled
- http://jsfiddle.net/yxpq4uuv/3/ - without feature enabled

QE Notes:
- Visuals should be exactly the same as before
- Does not behave well when combined with the `Plot.deferredRendering()` (the two perf improvements should be incompatible)
- It should work for both horizontal line plot, vertical line plot and random tangled line plots
- Perf is only for scenarios where we have a small subset of the data in the visual area.
- Should behave very well together with downsample (PR to come)

Extra notes:
- My progress and a "merged" branch can be seen here https://github.com/palantir/plottable/pull/2650

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3410,6 +3410,19 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-open"): Line<X>;
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
+            /**
+             * Gets the croppedRendering preformance option state.
+             *
+             * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
+             * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
+             * then this option will boost of rendering. However, if all the data is rendered anyway, having this
+             * option enabled will cause a small overhead that can be noticed in the total render time.
+             */
+            croppedRenderingEnabled(): boolean;
+            /**
+             * Sets the croppedRendering performance option.
+             */
+            croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
             protected _createDrawer(dataset: Dataset): Drawer;
             protected _extentsForProperty(property: string): any[];
             protected _getResetYFunction(): (d: any, i: number, dataset: Dataset) => number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3414,8 +3414,6 @@ declare module Plottable {
              * Gets if croppedRendering is enabled
              *
              * When croppedRendering is enabled, lines that will not be visible in the viewport will not be drawn.
-             * Thus if only part of the data is in the viewport, then this option will boost of rendering.
-             * However if all the data will render, enabling will cause an insignificant overhead.
              */
             croppedRenderingEnabled(): boolean;
             /**
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3417,13 +3417,13 @@ declare module Plottable {
              * Thus if only part of the data is in the viewport, then this option will boost of rendering.
              * However if all the data will render, enabling will cause an insignificant overhead.
              */
-            croppedRendering(): boolean;
+            croppedRenderingEnabled(): boolean;
             /**
              * Sets if croppedRendering is enabled
              *
              * @returns {Plots.Line} The calling Plots.Line
              */
-            croppedRendering(croppedRendering: boolean): Plots.Line<X>;
+            croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
             protected _createDrawer(dataset: Dataset): Drawer;
             protected _extentsForProperty(property: string): any[];
             protected _getResetYFunction(): (d: any, i: number, dataset: Dataset) => number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3417,13 +3417,13 @@ declare module Plottable {
              * Thus if only part of the data is in the viewport, then this option will boost of rendering.
              * However if all the data will render, enabling will cause an insignificant overhead.
              */
-            croppedRenderingEnabled(): boolean;
+            croppedRendering(): boolean;
             /**
              * Sets if croppedRendering is enabled
              *
              * @returns {Plots.Line} The calling Plots.Line
              */
-            croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
+            croppedRendering(croppedRendering: boolean): Plots.Line<X>;
             protected _createDrawer(dataset: Dataset): Drawer;
             protected _extentsForProperty(property: string): any[];
             protected _getResetYFunction(): (d: any, i: number, dataset: Dataset) => number;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3411,16 +3411,17 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
             /**
-             * Gets the croppedRendering preformance option state.
+             * Gets if croppedRendering is enabled
              *
-             * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
-             * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
-             * then this option will boost of rendering. However, if all the data is rendered anyway, having this
-             * option enabled will cause a small overhead that can be noticed in the total render time.
+             * When croppedRendering is enabled, lines that will not be visible in the viewport will not be drawn.
+             * Thus if only part of the data is in the viewport, then this option will boost of rendering.
+             * However if all the data will render, enabling will cause an insignificant overhead.
              */
             croppedRenderingEnabled(): boolean;
             /**
-             * Sets the croppedRendering performance option.
+             * Sets if croppedRendering is enabled
+             *
+             * @returns {Plots.Line} The calling Plots.Line
              */
             croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
             protected _createDrawer(dataset: Dataset): Drawer;
--- a/plottable.js
+++ b/plottable.js
@@ -630,8 +630,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -1602,8 +1601,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -1840,8 +1838,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -1899,8 +1896,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -2091,8 +2087,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -2206,8 +2201,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -2325,8 +2319,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
@@ -2441,8 +2434,7 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    __.prototype = b.prototype;
-    d.prototype = new __();
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
 var Plottable;
 (function (Plottable) {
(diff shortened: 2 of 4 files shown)


bluong:
What are the opinions regarding `deferredRendering` and `croppedRendering` being incompatible?

bluong:
Possible dumb question: how about a `renderMode()` endpoint that can differentiate between "deferred" and "cropped"?

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-open"): Line<X>;
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
+            /**
+             * Gets the croppedRendering preformance option state.
```
bluong: "preformance" -> "performance"

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
...
+             * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
+             * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
+             * then this option will boost of rendering. However, if all the data is rendered anyway, having this
+             * option enabled will cause a small overhead that can be noticed in the total render time.
+             */
+            croppedRenderingEnabled(): boolean;
+            /**
+             * Sets the croppedRendering performance option.
+             */
+            croppedRenderingEnabled(croppedRendering: boolean): Plots.Line<X>;
```
bluong: Why is the endpoint naming different from `deferredRendering` (adding "Enabled" at the end)

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-open"): Line<X>;
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
+            /**
+             * Gets the croppedRendering preformance option state.
+             *
+             * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
+             * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
+             * then this option will boost of rendering. However, if all the data is rendered anyway, having this
+             * option enabled will cause a small overhead that can be noticed in the total render time.
```
bluong: Is it "that can be noticed" or "that cannot be noticied"?

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-open"): Line<X>;
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
+            /**
+             * Gets the croppedRendering preformance option state.
+             *
+             * When croppedRendering option is enabled, lines that will not be visible in the viewport will not be
+             * drawn anymore (will not have corresponding SVG nodes). If only part of the data is in the viewport,
```
bluong: "lines that will not be visible in the viewport will not be drawn."?  Shorter

bluong: "Thus if only part of the data is visible, then this option will boost rendering."?

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
...
+             * then this option will boost of rendering. However, if all the data is rendered anyway, having this
```
bluong: "However if all of the data will render, enabling this option will..."?

plottable.d.ts
bluong: You can end the sentence after "noticed"

plottable.d.ts
```diff
@@ -3410,6 +3410,19 @@ declare module Plottable {
...
+             * option enabled will cause a small overhead that can be noticed in the total render time.
+             */
+            croppedRenderingEnabled(): boolean;
+            /**
+             * Sets the croppedRendering performance option.
```
bluong: "Sets if croppedRendering is enabled."?

bluong: Add the "@returns" doc

plottable.d.ts
bluong: "Gets if croppedRendering is enabled"

plottable.js
```diff
@@ -630,7 +630,8 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+    __.prototype = b.prototype;
+    d.prototype = new __();
```
bluong: Looks like the compiler has not been updated yet?

src/plots/linePlot.ts
```diff
@@ -14,6 +14,9 @@ export module Plots {
 
     private _autorangeSmooth = false;
 
+    // Performance options
```
bluong: Do we have more than one?  I don't see the need for the comment here until we have more than 1?

src/plots/linePlot.ts
```diff
@@ -349,10 +374,65 @@ export module Plots {
...
+        if (!this._croppedRendering) {
+          dataToDraw.set(dataset, [data]);
+          return;
+        }
+
+        let filteredDataIndices = data.map((d, i) => i);
+
+        if (this._croppedRendering) {
+          filteredDataIndices = this._filterCroppedRendering(dataset, filteredDataIndices);
+        }
```
bluong: Due to the above if statement, the if statement here is redundant

src/plots/linePlot.ts
```diff
@@ -349,10 +374,65 @@ export module Plots {
...
+
+        dataToDraw.set(dataset, [filteredDataIndices.map((d, i) => data[d])]);
```
bluong: Curious why you didn't use a `filter` code flow instead?

Have the indices exist in a set and then filter if the index exists in the set?

src/plots/linePlot.ts
```diff
@@ -349,10 +374,65 @@ export module Plots {
...
       return dataToDraw;
     }
 
+    private _filterCroppedRendering(dataset: Dataset, indices: number[]) {
+      let xScale = this.x().scale;
+      let yScale = this.y().scale;
+      let xAccessor = this.x().accessor;
+      let yAccessor = this.y().accessor;
+      let xDomain = xScale.domain();
+      let yDomain = yScale.domain();
```
bluong: What happens if there is no x/yScale?

src/plots/linePlot.ts
```diff
@@ -349,10 +374,65 @@ export module Plots {
...
+          let nextYPoint = yAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+          shouldShow = shouldShow || xDomain[0] <= nextXPoint && nextXPoint <= xDomain[1] &&
+            yDomain[0] <= nextYPoint && nextYPoint <= yDomain[1];
+        }
+
+        if (shouldShow) {
+          filteredDataIndices.push(indices[i]);
+        }
+      }
+      return filteredDataIndices;
```
bluong: Instead of this code flow, why not check through the `width()` / `height()` pieces from the component?

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
...
+        xScale = new Plottable.Scales.Linear();
+        yScale = new Plottable.Scales.Linear();
+        plot = new Plottable.Plots.Line<number>();
+        plot.x((d) => d.x, xScale).y((d) => d.y, yScale);
+      });
+
+      it("can set the croppedRendering option", () => {
+        plot.renderTo(svg);
+
+        assert.isFalse(plot.croppedRenderingEnabled(), "croppedRendering is not enabled by default");
```
bluong: I personally like this as a separate test.  Not a big deal though.

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
...
+        plot.addDataset(new Plottable.Dataset(data));
+
+        // Only middle point is in viewport
+        xScale.domain([2.5, 3.5]);
+
+        plot.croppedRenderingEnabled(true);
+        plot.renderTo(svg);
+
+        let path = plot.content().select("path.line").attr("d");
+        let expectedRenderedData = [1, 2, 3].map((d) => data[d]);
```
bluong: Instead of using these magic numbers, can we tie these values together more cleanly?  For example, we know that only 2 lines will render since x: 3 only exists in the domain

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
...
+        path = plot.content().select("path.line").attr("d");
+        checkPathForDataPoints(path, [0, 1].map((d) => data[d]));
+
+        svg.remove();
+      });
+
+      function checkPathForDataPoints(path: string, data: any[]) {
+        let EPSILON = 0.0001;
+
+        let lineEdges = path.match(/(\-?\d+\.?\d*)(,|\s)(-?\d+\.?\d*)/g);
```
bluong: Does this work for IE?  I remember having issues on IE with this

bluong: There is a `normalizePath` function for this

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
...
+        assert.strictEqual(lineEdges.length, data.length, "correct number of edges drawn");
+
+        lineEdges.forEach((edge, i) => {
+          let coordinates = edge.split(/,|\s/);
+
+          assert.strictEqual(coordinates.length, 2, "There is an x coordinate and a y coordinate");
+          assert.closeTo(xScale.invert(+coordinates[0]), data[i].x, EPSILON,
+            `Point ${i} drawn, has correct x coordinate`);
+          assert.closeTo(yScale.invert(+coordinates[1]), data[i].y, EPSILON,
+            `Point ${i} drawn, has correct y coordinate`);
```
bluong: How do we know that data has x/y params.  Let's type data or similar?  Or we have this function be more immune to the type of data being passed in

bluong:
Mostly documentation fixes and cleanup to understand better.

src/plots/linePlot.ts
aicioara: There will be (with downsampling). It's easier to not have a merge conflict / forget to add it if I do this.

plottable.js
aicioara: Sorry

src/plots/linePlot.ts
aicioara: It will be relevant when downsampling is added. It's easier to not have a merge conflict / forget to add this if I leave it in.

src/plots/linePlot.ts
aicioara: `O(n)` vs `O(n^2)`

bluong:
> Does not behave well when combined with the Plot.deferredRendering() (the two perf improvements should be incompatible)

I personally argue that they are compatible but they can easily give an offputting visual. (The point of `deferredRendering()` is to render later and `croppedRendering()` is to render a cropped area.  Combining them means that we are rendering cropped areas later, which can be offputting but "sensical")

bluong:
@crmorford The pull request should be at the point where you can decide if we want this as a flag or how we want to model this idea as an endpoint.

src/plots/linePlot.ts
```diff
@@ -138,6 +139,29 @@ export module Plots {
...
     }
 
+    /**
+     * Gets if croppedRendering is enabled
+     *
+     * When croppedRendering is enabled, lines that will not be visible in the viewport will not be drawn.
+     * Thus if only part of the data is in the viewport, then this option will boost of rendering.
+     * However if all the data will render, enabling will cause an insignificant overhead.
+     */
+    public croppedRendering(): boolean;
```
ztsai: Is this the final decision for naming?
I thought it was either `croppedViewportEnabled` or `croppedRenderingEnabled`

bluong: I chose this name because it is consistent with the naming decision for `deferredRendering()`  I am flexible on which name we decide to choose, but for now I went with the consistent decision.

src/plots/linePlot.ts
```diff
@@ -138,6 +139,29 @@ export module Plots {
       return this;
     }
 
+    /**
+     * Gets if croppedRendering is enabled
+     *
+     * When croppedRendering is enabled, lines that will not be visible in the viewport will not be drawn.
+     * Thus if only part of the data is in the viewport, then this option will boost of rendering.
+     * However if all the data will render, enabling will cause an insignificant overhead.
```
ztsai: Did a Doc sign off on this? It seems more wordy than our other docs

bluong: I'll give this idea to Docs to see if we want to tighten the wording.  I just trimmed what was there from the previous pull request

src/plots/linePlot.ts
ztsai: I don't have a strong preference either, I just want to double check cuz I feel like that's not what people preferred in the RFC

src/plots/linePlot.ts
```diff
@@ -349,10 +373,59 @@ export module Plots {
...
+
+      this.datasets().forEach((dataset) => {
+        let data = dataset.data();
+
+        if (!this._croppedRendering) {
+          dataToDraw.set(dataset, [data]);
+          return;
+        }
+
+        let filteredDataIndices = data.map((d, i) => i);
```
ztsai: I don't understand why this is needed? isn't this just [0... data.length -1]

src/plots/linePlot.ts
```diff
@@ -349,10 +373,59 @@ export module Plots {
...
+      let xProjector = Plot._scaledAccessor(this.x());
+      let yProjector = Plot._scaledAccessor(this.y());
+
+      let data = dataset.data();
+      let filteredDataIndices: number[] = [];
+
+      for (let i = 0; i < indices.length; i++) {
+        let currXPoint = xProjector(data[indices[i]], indices[i], dataset);
+        let currYPoint = yProjector(data[indices[i]], indices[i], dataset);
+        let shouldShow = 0 <= currXPoint && currXPoint <= this.width() &&
```
ztsai: can you factor out this check into a function as it's repeated 3 times

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
 
     });
   });
+
+  describe("Line Plot", () => {
```
ztsai: why can't this be in the previous "Line Plot" describe block?

src/plots/linePlot.ts
bluong: Yeah... do we have a function that creates such an array for us?

test/plots/linePlotTests.ts
```diff
@@ -683,4 +683,157 @@ describe("Plots", () => {
...
+    describe("Cropped Rendering Performance", () => {
+
+      let svg: d3.Selection<void>;
+      let plot: Plottable.Plots.Line<number>;
+
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+
+      let SVG_WIDTH = 400;
```
ztsai: this is just default, and you didn't use this anywhere else, is this needed?

src/plots/linePlot.ts
ztsai: I am not sure, but you didn't use `indices` below to do any complex stuff
isn't `indices[i]` just `i`, `indices[i - 1]`  just `i - 1`, etc..? 
and `indices.length` just `data.length` ?

test/plots/linePlotTests.ts
bluong: I took this from the previous iteration of this pull request.  I can change it back.

src/plots/linePlot.ts
ztsai: sorry, now that I think more about it, this is needed in order to merge croppedRendering and downSampling

plottable.js
```diff
@@ -8993,6 +8993,7 @@ var Plottable;
                 _super.call(this);
                 this._interpolator = "linear";
                 this._autorangeSmooth = false;
+                this._croppedRendering = false;
```
ztsai: really minor, but we usually have the flag name to be consistent with the function name

src/plots/linePlot.ts
```diff
@@ -349,10 +371,60 @@ export module Plots {
...
+        return Utils.Math.inRange(x, 0, this.width()) &&
+          Utils.Math.inRange(y, 0, this.height());
+      };
+
+      for (let i = 0; i < indices.length; i++) {
+        let currXPoint = xProjector(data[indices[i]], indices[i], dataset);
+        let currYPoint = yProjector(data[indices[i]], indices[i], dataset);
+        let shouldShow = pointInViewport(currXPoint, currYPoint);
+
+        if (!shouldShow && indices[i - 1] != null && data[indices[i - 1]] != null) {
```
ztsai: could indices[i - 1] ever be null?

ztsai:
everything else looks good

crmorford:
chop off top and bottom http://jsfiddle.net/yxpq4uuv/5/
stacked area is still miserable :P http://jsfiddle.net/yxpq4uuv/6/

tested on everything except a windows box, i'll find one in the morning

i think croppedRendering should be on by default

crmorford:
ok looks good on ie

PR #2811, opened 2015-09-25
[Scales.Time] Change default extent to be wider and consistent.

### LINKED ISSUE #2740: Time Scale needs better default ticks
http://jsfiddle.net/korw10sr/
There are 2 ticks listed at the same location, so the time axis looks blank/broken when rendered without data.

jtlan:
Previously, the default extent was supposed to be based on the current day, but wound up defaulting instead to `[Unix epoch, + 1 second]`. The default extent has now been changed to be `[Unix epoch, + 1 day]`.

Close #2740.

### DIFF
--- a/src/scales/timeScale.ts
+++ b/src/scales/timeScale.ts
@@ -12,6 +12,7 @@ export module Scales {
     constructor() {
       super();
       this._d3Scale = d3.time.scale();
+      this.autoDomain();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/scales/timeScale.ts
```diff
@@ -12,6 +12,7 @@ export module Scales {
     constructor() {
       super();
       this._d3Scale = d3.time.scale();
+      this.autoDomain();
```
bluong: Why autodomain on instantiation?

src/scales/timeScale.ts
```diff
@@ -38,11 +39,7 @@ export module Scales {
     }
 
     protected _defaultExtent(): Date[] {
-      let now = new Date();
-      let endTimeValue = now.valueOf();
-      now.setDate(now.getDate() - 1);
-      let startTimeValue = now.valueOf();
-      return [new Date(startTimeValue), new Date(endTimeValue)];
+      return [new Date("1970-01-01"), new Date("1970-01-02")];
```
bluong: Not a big deal, but I personally like the `(year, month, day)` constructor since I find it to be less error prone.

bluong:
Just a question

src/scales/timeScale.ts
jtlan: To set the domain to padded `_defaultExtent()`. I guess we could test for it? It'd be kind of tautological though.

Issue #2815, opened 2015-09-27
Gruntfile cannot be parsed under Windows


softwords:
Visual Studio 2015 provides a range of new tools for web development, including support for node, npm, bower, grunt.
Experiments suggest the best way to work with plottable in VS2015 is to open its folder as a "Web Site" - this gives you access to edit the files in the folder without creating an unnecessary project file. You use the tasks in the gruntfile (accessed from the new Task Runner Explorer) in place of project-level build processes.
However, currently, gruntfile does not parse under Windows. The Task Runner Explorer in VS2015 shows "No tasks found".
The issue is here:
```javascript
 var updateTsFiles = function() {
    tsFiles = grunt.file.read("src/reference.ts")
      .split("\n")
```

This fails under Windows because line ends are now ```\r\n```

A solution ( taken from [here](http://stackoverflow.com/questions/6831918/node-js-read-a-text-file-into-an-array-each-line-an-item-in-the-array)) is simply to add this:

```javascript
  var updateTsFiles = function() {
    tsFiles = grunt.file.read("src/reference.ts")
      .replace(/\r\n/g,"\n")
      .split("\n")
```

jtlan:
@softwords Was this fixed by #2816?

softwords:
Yes, but, this code is now gone due to the restructure of the compile process to exploit the options of the compiler. So the only issue  remaining under windows (also relating to line-ends) is #2914.

PR #2816, opened 2015-09-27
Changes to gruntfile to support Visual Studio 2015

softwords:
see issue #2815

These minor changes to the gruntfile allow the Task Runner Explorer in VS2015 to parse the gruntfile and display the tasks.


Gruntfile.js
```diff
@@ -92,6 +92,7 @@ module.exports = function(grunt) {
   // on each recompile
   var updateTsFiles = function() {
     tsFiles = grunt.file.read("src/reference.ts")
+      .replace(/\r\n/g, "\n")
```
aicioara: This sounds a bit odd. The reference.ts file should not contain any Windows endings (`\r\n`). Would this work if you set VS to use UNIX line endings. [how to](http://stackoverflow.com/questions/3802406/configure-visual-studio-with-unix-end-of-lines)

bluong: I would prefer not having Windows endings as well in our files if we can help it.  @softwords how feasible is this?

softwords: _Would this work if you set VS to use UNIX line endings?_
No, that doesn't seem to work, because every time you fetch the project the setting for the current file is lost. It's git that decides what line endings to use when it writes out the files. In general, it's fine to use the default line ends for the operating system, but plottable's gruntfile assumes the line ends are \n, regardless of the operating system.
_I would prefer not having Windows endings as well in our files if we can help it._
So -  a better solution than what I have proposed here may to to introduce a .gitattributes file. This gives git some project specific information. With .gitattributes, you can force the line ends to use for various file types  - see [here](https://help.github.com/articles/dealing-with-line-endings/).
Is this a better approach, and if so, how would you prefer to move forward with this?

softwords: Doing some more testing, sed:privateDefinitions also gives unwanted results with windows line ends, and this would be a lot harder to resolve... so fixing up the line ends with .gitattributes seems the most robust solution.

bluong: sed:privateDefinitions works on the current files in the directory, though.  How does fixing up .gitattributes affect the functionality of sed:privateDefinitions?

softwords: You´re right, it won´t help and the problem then runs deeper, becuase all the typescript compiles and the concats are going to use the windows linefeed, and neither git nor visual studio can affect that.

But another possibility could be to use
grunt.util.linefeed
instead of the literal \n

```javascript
   var updateTsFiles = function() {
     tsFiles = grunt.file.read("src/reference.ts")
       .split(grunt.util.linefeed)
```
If this were done everywhere, .gitattributes would not be needed?

bluong: I think I like this approach a lot better.  What are your results if you run with this?

softwords: Yes this (grunt.util.linefeed) is a better solution than my first attempt. The key thing I want to achieve is to get the same behaviour from the grunt tasks under Windows as elsewhere, including
- gruntfile parses and can execute;
- outputs of compiles give identical plottable.js and plottable.d.ts
- tests execute and give the same results

At the moment, I get blank lines through plottable.d.ts. This is caused by the sed task removing the private declarations - it doesn't remove the line-end properly. This can probably be fixed by substituting grunt.util.linefeed into these statements where needed:
```javascript
   var prefixMatch = "\\n *(function |var |static )?";
    var varNameMatch = "[^(:;]*(\\([^)]*\\))?"; // catch function args too
    var nestedBraceMatch = ": \\{[^{}]*\\}";
    var typeNameMatch = ": [^;]*";
    var finalMatch = "((" + nestedBraceMatch + ")|(" + typeNameMatch + "))?\\n?;";
    var jsdocInit = "\\n *\\/\\*\\* *\\n";
    var jsdocMid = "( *\\*[^\\n]*\\n)+";
    var jsdocEnd = " *\\*\\/ *";
    var jsdoc = "(" + jsdocInit + jsdocMid + jsdocEnd + ")?";
```

(i will try then report back)

Also, currently I get a few test fails just fetching the current develop branch and executing test - am I right to think this shouldn't happen? e.g.
```

  1) NumericAxis draws ticks correctly (vertical):
     AssertionError: tick label is centered on mark: expected 3501.5 to be close to 3503 +/- 1
```

bluong: This is likely a precision error due to being on a different operating system (This looks like a text measurement error and text is known to be very inconsistently measured across browsers).  If you run into any more errors moving forward.  Feel free to skip them in development if it causes trouble so we can inspect the test to hopefully make it less fragile.

softwords:
This seems to fix the gruntfile issues under Windows / Visual Studio, including generation of plottable.d.ts

Gruntfile.js
```diff
@@ -51,13 +51,14 @@ module.exports = function(grunt) {
...
+  var prefixMatch = "\\n *(function |var |static )?".replace("\\n", regexLF);
   var varNameMatch = "[^(:;]*(\\([^)]*\\))?"; // catch function args too
   var nestedBraceMatch = ": \\{[^{}]*\\}";
   var typeNameMatch = ": [^;]*";
-  var finalMatch = "((" + nestedBraceMatch + ")|(" + typeNameMatch + "))?\\n?;";
-  var jsdocInit = "\\n *\\/\\*\\* *\\n";
-  var jsdocMid = "( *\\*[^\\n]*\\n)+";
+  var finalMatch = "((" + nestedBraceMatch + ")|(" + typeNameMatch + "))?\\n?;".replace("\\n", regexLF);
+  var jsdocInit = "\\n *\\/\\*\\* *\\n".replace("\\n", regexLF);
+  var jsdocMid = "( *\\*[^\\n]*\\n)+".replace("\\n", regexLF);
```
bluong: Instead of all these replacements, why not just interpolate regexLF where it's needed?

Gruntfile.js
```diff
@@ -51,13 +51,14 @@ module.exports = function(grunt) {
     }
   };
 
-  var prefixMatch = "\\n *(function |var |static )?";
+  var regexLF = (grunt.util.linefeed === "\r\n" ? "(\\u000D\\n)" : "\\n"); //get the correct Lf for the operating system
```
bluong: Can you provide a comment or a link that references why the LF has to be as you suggest?  Why does `grunt.util.linefeed` not work?

bluong:
Thanks a lot for cleaning up the pull request.  It looks like a much more straightforward fix.  Can you reply / make the changes suggested and also send us the signed CLA when you have the chance?

Gruntfile.js
softwords: Well, I thought that then each expression would be a concatenation: e.g.
```var jsdocMid = "( *\\*[^" + regexLF + "]*" + regexLF + ")+"```
which could be harder to maintain, and harder to understand.

File: Gruntfile.js (hunk 3 above)
Comment (softwords) [inline Gruntfile.js]: What we are building here is a regular expression, and we need to have the string "\n" not the character \n. That's why all these lines escape the \ character ```\\```.

**Comment by softwords** [comment]
I have submitted the CLA a few weeks back. :)

File: Gruntfile.js (hunk 3 above)
Comment (softwords) [inline Gruntfile.js]: Also, ```\\u000D``` represents a unicode Carriage Return in the regular expression. I played with various alternatives, this was the one that gave the result.

Comment (bluong) [inline Gruntfile.js]: Just curious.  if we take the `escapeRegExp` function from here https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex , does that work here (Please reference the stackoverflow post if you do use it)?

Otherwise if this does not work, please put a comment above this line that details what you are doing here.

File: Gruntfile.js (hunk 2 above)
Comment (bluong) [inline Gruntfile.js]: Curious.  Does using template strings help here?  As in, does
```javascript
var jsdocMid = `( *\\*[^${regexLF}]*${regexLF})+...`
```
work?

While a replace idea could work, I personally think some form of interpolation is cleaner.  If you still think this idea is better, feel free to respond.

bluong:
After discussion, I submitted a pull request here in order to remove our actively trying to remove the private members, which hopefully should reduce a bunch of your pain.

https://github.com/palantir/plottable/pull/2846

You can try rebasing to see if keeping the private members does not affect the "plottable.d.ts" so much when running grunt.

softwords:
I imagine you do that with a sense of regret, it seems untidy to have private defs in the type definition, but  the typescript issue seems to explain that it is a necessary evil.
Perhaps your PR could add the two references I have in here to replace ```split('\n')``` with ```split(grunt.util.linefeed)```
Then the current PR can be closed?

Now - the bad news - is that it doesn't solve my problem that git thinks every line I write into plottable.d.ts is a new line, compared to all versions of the file that don't originate from me. I'm getting a feeling this is something to do with file encoding... there is a single non-ascii character is arrayUtils.ts that finds its way into plottable.d.ts
```     * Will work iff ∀ a, b, a.toString() == b.toString() => a == b; will break on Object inputs```

and that makes VS want to save this file as unicode whereas everything else seems to be codepage 1252. This is just a hypothesis, but I'll keep looking... clearly if I can't make a plottable.d.ts I can't contribute without completely retooling ( e.g. buying a mac!)

bluong:
If that non-ascii character is removed, do you finally get to a good state?

If all else fails, since the pull request has gotten a whole lot more readable, I can review it and then I can personally play around with the "plottable.d.ts" situation.  Let's try to avoid this if we can though.

bluong:
Also, I did not include the `grunt.util.linefeed` in that pull request so I can just incorporate yours into it.

Hopefully we can get your dev environment to a good state.

bluong:
This looks good to merge I think.  Can you check if this does help your working environment?

softwords:
Yes, certainly it does. With this change, the Task Runner Explorer shows the grunt tasks as expected. Currently, I have to unstash my modified gruntfile every time I fetch or checkout the develop branch. this will streamline things for me - and anybody using VS2015 to follow!

softwords:
On that note, my latest PR ensures that, when checking out develop , and running test, all the current tests execute correctly under Windows - there's currently only one that doesn't.
So these two things taken together are a big step forward.
I am still left with the plottabe.d.ts issue; will keep playing with that today and let you now if I have any joy.

bluong:
I hope we can hopefully resolve your issues.  In your undertaking to get a working environment on your side, I'm also trying to simplify the Gruntfile to have our build process be massively more simplified https://github.com/palantir/plottable/pull/2851.

Unfortunately if that one goes in, it essentially removes your changes, but your environment should hopefully have less problems.

Again, thanks a lot for trying your environment set up.  I'll look into your pull request in the meantime, since it seems that worst case scenario, I can build the plottable.d.ts file myself (unpreferred but feasible for now)

PR #2821, opened 2015-09-29
consider overlay for pointer interactions

### LINKED ISSUE #2416: Pointer (and Key) Interactions work through overlays
Repro: http://jsfiddle.net/79tL9n7k/

Forked off of #2142.

ztsai:
pointerMove should not be called under overlay
overlay is considered as outside of component for pointerExit and pointerEnter

closes #2416

before: http://jsfiddle.net/ztsai/crjz5531/1/
after: http://jsfiddle.net/ztsai/crjz5531/


plottable.js
```diff
@@ -630,7 +630,8 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+    __.prototype = b.prototype;
+    d.prototype = new __();
```
bluong: Looks like typescript did not update.  Please update.

bluong:
Can you give more context from what the code did before and what the code does now?  I'm having trouble parsing your changes

src/interactions/pointerInteraction.ts
```diff
@@ -34,19 +35,32 @@ export module Interactions {
...
         }
         this._pointerMoveCallbacks.callCallbacks(translatedP);
-      } else if (this._overComponent) {
+      } else if (this._overComponent && this._insideSVG) {
         this._overComponent = false;
         this._pointerExitCallbacks.callCallbacks(translatedP);
       }
+
+      this._overComponent = overComponent;
+      this._insideSVG = insideSVG;
```
bluong: Does it make sense to consolidate these 2 variables?

src/interactions/pointerInteraction.ts
```diff
@@ -34,19 +35,32 @@ export module Interactions {
...
-    private _handlePointerEvent(p: Point) {
+    private _handleMouseEvent(p: Point, e: MouseEvent) {
+      let insideSVG = this._mouseDispatcher.eventInsideSVG(e);
+      this._handlePointerEvent(p, insideSVG);
+    }
+
+    private _handleTouchEvent(p: Point, e: TouchEvent) {
+      let insideSVG = this._touchDispatcher.eventInsideSVG(e);
+      this._handlePointerEvent(p, insideSVG);
+    }
```
bluong: I'm trying to interpret what happened before.

Is this behavior specific to mousemove / touch?

ztsai: the only difference is that it checks against `mouseDispatcher` for `MouseEvent` and `touchDispatcher` for `TouchEvent`.

I could move this check inside `_handlePointerEvent` but I feel it's hacky to check the type of `e` and decide which `Dispatcher` to use

ztsai:
basically, before we consider pointer to be "inside" of a component by its position, so mousemove would be triggered under overlay.
the correct behavior should be that a pointer is considered inside if and only its position is contained by the component AND the event target is the SVG

crmorford:
If i drag really fast, the point can show up under the overlay. Why does this happen?
ezgif com-video-to-gif

ztsai:
the green point is where the pointerExit event got triggered
if you move really fast, the end point of that move will be inside the overlay

crmorford:
My initial thought is that the pointerExit event should never be triggered under the overlay.
This is definitely an improvement though, and I'm not sure how important that is, so I'm gonna move ahead with QE.

ztsai:
if you move really fast toward outside of the component, the exit point would be outside of the component too ( x > 300 if you move to the right in the fiddle example)

crmorford:
ooh okay that makes more sense

crmorford:
looks good on chrome and firefox,
i ran into trouble testing on safari, since i don't know how to get pointer events to trigger. i tried layering the scatter plot with a colored component, but no luck http://jsfiddle.net/crjz5531/4/

ztsai:
this is a regression caused by this [PR](https://github.com/palantir/plottable/pull/2749)
the original fix was [this](https://github.com/palantir/plottable/pull/2575)

in the meanwhile, you can test in safari with this hack: http://jsfiddle.net/crjz5531/6/

crmorford:
ok, looks good in safari and mobile!

Issue #2794, opened 2015-09-21
[Plots.Bar] entities() returns wrong positions when domain is reversed


jtlan:
<img width="316" alt="screen shot 2015-09-21 at 4 23 30 pm" src="https://cloud.githubusercontent.com/assets/1440449/10007807/5435fa16-607d-11e5-9736-ad9afb409962.png">

**Repro**: http://jsfiddle.net/ogL16g2f/
**Actual**: `entity.position` is at the top of some bars, and at the base of others
**Expected**: `entity.position` is at the top of each bar

jtlan:
Even goofier things happen if more data points are used, and some lie outside the visible domain:
<img width="311" alt="screen shot 2015-09-21 at 4 26 30 pm" src="https://cloud.githubusercontent.com/assets/1440449/10007831/89c5599c-607d-11e5-867a-c71d98ee6ef1.png">
http://jsfiddle.net/ogL16g2f/2/

jtlan:
Probably related to https://github.com/palantir/plottable/issues/2793

crmorford:
GROSS

crmorford:
@yunhaolucky  fixed this :)

PR #2825, opened 2015-09-30
Barplot's pixelpoint return data point

### LINKED ISSUE #2793: [Plots.Bar] _pixelPoint() returns wrong value
The `_pixelPoint()` of `Plots.Bar` returns the wrong value:

screen shot 2015-09-21 at 3 59 52 pm
(green shows expected positions, red shows actual ones)

**Repro**: http://jsfiddle.net/8gxszz87/
**Actual**: points for negative values appear at the top of the bar, rather than the end.
**Expected**: return value for `_pixelPoint()` should match `entities().position`

It looks like `entities()` accounts for the wrong position somehow.

yunhaolucky:
**Issue** 
`entity.position` are used to calculate the nearest entity. It should return the data point calculated by scales (for stackedBarPlot, it should return data point plus offset). Previously, it returns wrong points when scale domain are reversed or bar datum values are negative. Also, `barPlot._pixelPoint()` should return the same result as `entity.position`.

**Fix** 
I change `barPlot._pixelPoint()` function and calculated `pixelPoint()` separately when the datum value is negative. Also, I also call`barPlot._pixelPoint()` inside `entity.position` to make them return the same point.
This fix should work in barPlot, stackedBarPlot and clusteredBarPlot.

**JSFiddle**:
`barPlot._pixelPoint()` vs `entity.position`
Before: http://jsfiddle.net/8gxszz87/1/
After: http://jsfiddle.net/yunhaolucky/8gxszz87/2/
(Red points are `_pixelPoint`, green points are `entity.position`).

`entity.position` returns wrong points
Before: http://jsfiddle.net/yunhaolucky/ogL16g2f/6/
After: http://jsfiddle.net/yunhaolucky/ogL16g2f/4/

close #2793 #2794

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3357,8 +3357,8 @@ declare module Plottable {
             protected _getBarPixelWidth(): number;
             entities(datasets?: Dataset[]): PlotEntity[];
             protected _pixelPoint(datum: any, index: number, dataset: Dataset): {
-                x: any;
-                y: any;
+                x: number;
+                y: number;
             };
             protected _uninstallScaleForKey(scale: Scale<any, number>, key: string): void;
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3357,8 +3357,8 @@ declare module Plottable {
             protected _getBarPixelWidth(): number;
             entities(datasets?: Dataset[]): PlotEntity[];
             protected _pixelPoint(datum: any, index: number, dataset: Dataset): {
-                x: number;
-                y: number;
+                x: any;
+                y: any;
             };
             protected _uninstallScaleForKey(scale: Scale<any, number>, key: string): void;
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3356,10 +3356,7 @@ declare module Plottable {
              */
             protected _getBarPixelWidth(): number;
             entities(datasets?: Dataset[]): PlotEntity[];
-            protected _pixelPoint(datum: any, index: number, dataset: Dataset): {
-                x: number;
-                y: number;
-            };
+            protected _pixelPoint(datum: any, index: number, dataset: Dataset): Point;
             protected _uninstallScaleForKey(scale: Scale<any, number>, key: string): void;
             protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
         }
--- a/plottable.js
+++ b/plottable.js
@@ -8559,16 +8559,8 @@ var Plottable;
                         var plotPtPrimary = _this._isVertical ? plotPt.x : plotPt.y;
                         primaryDist = Math.abs(queryPtPrimary - plotPtPrimary);
                         // compute this bar's min and max along the secondary axis
-                        var barMinSecondary;
-                        var barMaxSecondary;
-                        if (_this._isVertical) {
-                            barMinSecondary = barBBox.y - (entity.datum.y > _this.baselineValue() ? 0 : barBBox.height);
-                            barMaxSecondary = barBBox.y + (entity.datum.y > _this.baselineValue() ? barBBox.height : 0);
-                        }
-                        else {
-                            barMinSecondary = barBBox.x - (entity.datum.x < _this.baselineValue() ? 0 : barBBox.width);
-                            barMaxSecondary = barBBox.x + (entity.datum.x < _this.baselineValue() ? barBBox.width : 0);
-                        }
+                        var barMinSecondary = _this._isVertical ? barBBox.y : barBBox.x;
+                        var barMaxSecondary = barMinSecondary + (_this._isVertical ? barBBox.height : barBBox.width);
                         if (queryPtSecondary >= barMinSecondary - tolerance && queryPtSecondary <= barMaxSecondary + tolerance) {
                             // if we're within a bar's secondary axis span, it is closest in that direction
                             secondaryDist = 0;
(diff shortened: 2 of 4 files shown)


plottable.js
```diff
@@ -630,7 +630,8 @@ var Plottable;
 var __extends = (this && this.__extends) || function (d, b) {
     for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
     function __() { this.constructor = d; }
-    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
```
bluong: Please update your typescript

src/plots/barPlot.ts
```diff
@@ -720,8 +711,17 @@ export module Plots {
       let rectY = attrToProjector["y"](datum, index, dataset);
       let rectWidth = attrToProjector["width"](datum, index, dataset);
       let rectHeight = attrToProjector["height"](datum, index, dataset);
-      let x = this._isVertical ? rectX + rectWidth / 2 : rectX + rectWidth;
-      let y = this._isVertical ? rectY : rectY + rectHeight / 2;
+      let x: any;
+      let y: any;
```
bluong: Are these `number`s?

src/plots/barPlot.ts
```diff
@@ -290,8 +290,15 @@ export module Plots {
...
 
           // compute this bar's min and max along the secondary axis
-          let barMinSecondary = this._isVertical ? barBBox.y : barBBox.x;
-          let barMaxSecondary = barMinSecondary + (this._isVertical ? barBBox.height : barBBox.width);
+          let barMinSecondary: number;
+          let barMaxSecondary: number;
+          if (this._isVertical) {
+            barMinSecondary = barBBox.y - (entity.datum.y > this.baselineValue() ? 0 : barBBox.height);
+            barMaxSecondary = barBBox.y + (entity.datum.y > this.baselineValue() ? barBBox.height : 0);
+          }else {
```
bluong: There should be a space before `else`

src/plots/barPlot.ts
```diff
@@ -290,8 +290,15 @@ export module Plots {
...
+            barMinSecondary = barBBox.x - (entity.datum.x < this.baselineValue() ? 0 : barBBox.width);
+            barMaxSecondary = barBBox.x + (entity.datum.x < this.baselineValue() ? barBBox.width : 0);
+          }
```
bluong: Can you explain the changes to `entitiesNearest()`?

test/plots/barPlotTests.ts
```diff
@@ -252,8 +252,27 @@ describe("Plots", () => {
...
-        });
+          it("entities().position and pixelPoint() return the same point", () => {
+            let entities = barPlot.entities();
+            entities.forEach((entity) => {
+              let pixelPoint = (<any> barPlot)._pixelPoint(entity.datum, entity.index, entity.dataset);
+              assert.strictEqual(pixelPoint.x, entity.position.x, "pixelPoint and entities().position should have same x value");
+              assert.strictEqual(pixelPoint.y, entity.position.y, "pixelPoint and entities().position should have same y value");
+            });
+            svg.remove();
+          });
```
bluong: I personally don't think this test is that useful.

`_pixelPoint` is a non-public function, which means that we are testing against pieces of the code that users will never touch.  While this may be useful if we really want to assert this, if the behavior derived from `_pixelPoint` were to work too, then we don't need to test this directly.

Also, there is some fragility here in that if `_pixelPoint` were to have its name changed, then we would get a runtime error since our compiler will not be able to catch the name change.

test/plots/barPlotTests.ts
```diff
@@ -252,8 +252,27 @@ describe("Plots", () => {
...
 
+          it("entities().position returns the position of data point", () => {
+            let entities = barPlot.entities();
+            entities.forEach((entity) => {
+              let dataX = (<any> Plottable.Plot)._scaledAccessor(barPlot.x())(entity.datum, entity.index, entity.dataset);
+              let dataY = (<any> Plottable.Plot)._scaledAccessor(barPlot.y())(entity.datum, entity.index, entity.dataset);
```
bluong: Instead of using `_scaledAccessor`, we could use the xScale/yScale with the xAccessor/yAccessor to check.

bluong:
Can we put a description in the pull request such that readers of the pull request have better context?

bluong:
One of the tests is testing a non-public function which has very limited use.

Also a question about `entitiesNearest()`

plottable.d.ts
```diff
@@ -3357,8 +3357,8 @@ declare module Plottable {
             protected _getBarPixelWidth(): number;
             entities(datasets?: Dataset[]): PlotEntity[];
             protected _pixelPoint(datum: any, index: number, dataset: Dataset): {
```
bluong: Let's explicitly type this as `Point`

crmorford:
Testing Plan:
Regular/reversed xDomain for horizontal bar plots, stackedbar, clusteredbar
Regular/reversed yDomain for vertical bar plots, stackedbar, clusteredbar
category axes, linear, log
full regression pass

crmorford:
vertical, clustered, flipped Y http://jsfiddle.net/ogL16g2f/8/

crmorford:
develop: http://jsfiddle.net/ogL16g2f/9/
normal: http://jsfiddle.net/ogL16g2f/10/

stacked bar doesn't do well with a flipped domain :fearful: 
this could be the right place to fix that

yunhaolucky:
The issue here is stackedBarPlot does draw bars correctly with a reversed scale domain. I open an issue about this #2830.

bluong:
does or does not?

PR #2829, opened 2015-09-30
correct end tick marks length when tier label positions are set to "center"

### LINKED ISSUE #2800: Wrong length of end ticks in TimeAxis when `tierlabelposition` is set to `center`.
In a TimeAxis, the end ticks have different lengths when the `tierlabelposition` is set to `center`.
This will cause a problem as it make end tick labels overlap with end tick marks.

repro: http://jsfiddle.net/yunhaolucky/o87wsq9h/5/
screen shot 2015-09-23 at 10 56 42 am

yunhaolucky:
**Issue**
Previously in a TimeAxis, when its `tierLabelPosition` is set to "center", the end tick marks on two sides have different lengths. The first end tick mark's length is the height of the tier.

**Fix**
Now the lengths of the end tick marks are `axis.endTickLength()`, whose default value is 5.

**JSFiddle**
Before:http://jsfiddle.net/yunhaolucky/o87wsq9h/5/
After:http://jsfiddle.net/yunhaolucky/o87wsq9h/6/
close #2800

### DIFF
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -130,6 +130,21 @@ describe("TimeAxis", () => {
     svg.remove();
   });
 
+  it("end ticks' lengths equal to endTickLength() when tierLabelPosition is set to center", () => {
+    let width = 500;
+    let svg = TestMethods.generateSVG(width, 100);
+    axis.tierLabelPositions(["center", "center"]);
+    scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
+    axis.renderTo(svg);
+    let endTicks = axis.content().selectAll(`.${Plottable.Axis.END_TICK_MARK_CLASS}`);
+    assert.operator(endTicks.size(), ">=", 1, "At least one end tick mark is selected in the test");
+    endTicks.each(function(d, i){
+      let tickLength = Math.abs(+this.getAttribute("y1") - +this.getAttribute("y2"));
+      assert.strictEqual(tickLength, axis.endTickLength(), "end tick marks's length should equal to endTickLength()");
+    });
+    svg.remove();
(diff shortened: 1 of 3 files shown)


test/axes/timeAxisTests.ts
```diff
@@ -130,6 +130,20 @@ describe("TimeAxis", () => {
...
 
+  it("end ticks' lengths equal to endTickLength() when tierLabelPosition is set to center", () => {
+    let svg = TestMethods.generateSVG(400, 100);
+    axis.tierLabelPositions(["center", "center"]);
+    scale.domain([new Date("2010-01-01"), new Date("2014-01-01")]);
+    axis.renderTo(svg);
+    let endTicks = axis.content().selectAll(`.${Plottable.Axis.END_TICK_MARK_CLASS}`);
+    assert.operator(endTicks.size(), ">=", 1, "At least one end tick mark is selected in the test");
+    endTicks.each(function(d, i){
+      let tickLength = Math.abs(+this.getAttribute("y1") - +this.getAttribute("y2"));
```
bluong: Instead let's use this to be more consistent
```typescript
let endTick = d3.select(this);
let tickLength = Math.abs(TestMethods.numAttr("y1") - TestMethods.numAttr("y2"));
```

test/axes/timeAxisTests.ts
```diff
@@ -130,6 +130,20 @@ describe("TimeAxis", () => {
...
+      assert.strictEqual(tickLength, axis.endTickLength(), "end tick marks's length should equal to endTickLength()");
```
bluong: We're comparing pixels to our own values.  Let's use `assert.closeTo` to not have this test fail due to pixel diffing reasons.

PR #2833, opened 2015-10-01
Enable 'object constancy' on data refresh by supporting a key function in Dataset

softwords:
Resubmitting with cleanups to formatting - previously #2801
also see issue #2512 
This pull request

adds the key function to Dataset
uses this key in Drawer when binding data to the selection
provides 2 predefined keys:
-- NoConstancy - which will return a new value each time it is called, therefire ensuring that there is never object constancy; that is, all elements in the selection are removed and recreated
-- ByIndex - returns the index of the datum n the array, which is the default behaviour of d3 anyway

Also removes the logic on AnimateOnNextRender in ScatterPlot. This logic simulates the destruction and recreation of each svg element to create an animation effect. But, if the NoConstancy key is used, d3 destroys and recreates the svg elements itself, so there is no need for this workaround.
This also resolves issue #2707.


bluong:
Can you please compile using `grunt` so we can see the output.

softwords:
@bluong I have loaded the plottable.js and plottable.d.ts that I get running ```test``` in grunt from the VS Task runner Explorer.
You'll see the issue I mentioned yesterday - that whereever a private declaration has been removed from plottable.d.ts, I have a blank row.

bluong:
So there is some functionality in our `Gruntfile.js` that removes the private declarations since it seems that having them is unnecessary.  That functionality might be what you're running into.

softwords:
Yes I've had a look at that - it's the sed:privateDeclarations task. I've played with it a bit but haven't yet managed to find a way to get the same result in Windows - becuase those regex expressions are looking for \n line-ends I think.
But just on that, I wondered why typescript would even put those private declarations in the .d.ts file - you would think that only the publicly exposed interface matters. There is an explanation [here](https://github.com/Microsoft/TypeScript/issues/1532) in the typescript github:

_Private members are emitted in a declaration file to indicate that a private property with that name exists so that a derived class doesn't accidentally use the name (which would step on the private members)._

It certainly looks nicer for the .d.ts file not to have them, and chances of accidently reusing the private variable would seem slim; but by including those private names in the .d.ts you'll get a compile error if you try to override. 
So - it may be worth considering leaving them there?

softwords:
A lot of things I don't fully understand are happening with the testing stage: as you can see in the travis output phamtomjs is throwing an error. Sometimes I get this locally, but when it runs to completion, I get a number of tasks _pending_. These tasks are also pending if I run coverage.html in chrome - they all relate to testings for type, or non-numeric values. e.g. _currency() - throws exception for non-numeric values._
Other errors are rounding type problems: e.g. pointerinteraction is typical:
```
calls the onPointerEnter callback (mouse) â€£
AssertionError: was passed correct point (mouse): expected { x: 200, y: 199.33203125 } to deeply equal { x: 200, y: 200 }
    at Context.<anonymous> (tests.js:13773:24)
```
Again, see the same thing in chrome and phantomjs, although sometimes the fail in phantomjs also reports _out of memory_:
```

  10) Plots Bar Plot Horizontal Bar Plot entityNearest() considers only in-view bars:
     Error: Out of memory
```
 chrome reports the error a bit differently:
```
considers only in-view bars â€£
AssertionError: closest plot data is on-plot data (selection contents): expected <rect fill="#5279c7" width="150" height="103.7037037037037" y="222.22222222222223" x="150"></rect> to equal <rect fill="#5279c7" width="473.6842105263158" height="103.7037037037037" y="222.22222222222223" x="157.89473684210526"></rect>
    at tests.js:109:20
    at Array.forEach (native)
    at assertEntitiesEqual (tests.js:108:29)
    at Object.assertPlotEntitiesEqual (tests.js:115:9)
    at Context.<anonymous> (tests.js:7577:33)
```

So any guidance you can offer as to what could be going on here would be much appreciated.

bluong:
One thing that could be happening is inaccuracies due to being on a different OS / different version of a certain browser.  In terms of the `out of memory error`, I am unsure why that would be the case.  In terms of the `closest plot data` errorin and the `onPointerEnter` callback erroring, it seems to be precision errors?

Can you `skip` the tests that are failing and then continue with the pull request?

softwords:
Looking at these precision errors, running in both Chrome and IE, I have found something interesting:

in this code in TestMethods:

```
 export function triggerFakeMouseEvent(type: string, target: d3.Selection<void>, relativeX: number, relativeY: number, button = 0) {
    let clientRect = (<Element> target.node()).getBoundingClientRect();
    let xPos = clientRect.left + relativeX;
    let yPos = clientRect.top + relativeY;
    let e = <MouseEvent> document.createEvent("MouseEvents");
    e.initMouseEvent(type, true, true, window, 1,
      xPos, yPos,
      xPos, yPos,
      false, false, false, false,
      button, null);
    target.node().dispatchEvent(e);
  }
```

The members of the object ```ClientRect``` returned by ```getBoundingClientRect()``` are defined in the standard as ```float``` .  But, if these values are passed to ```MouseEvent.InitMouseEvent``` they are converted to integers. This is where all these precision errors are originating. 

On circumstance where the members of ```ClientRect``` may not be integers is when the zoom factor of the browser is < 1. 
Try running the coverage.html in chrome, with zoom factor set to e.g. 75% and you may see some errors you don't otherwise see.

IE10 apparently returns fractional values like this:
```
can register callback using onClick() 
â€£
AssertionError: was passed correct point: expected { x: 200.5, y: 200.5 } to deeply equal { x: 200, y: 200 }
   at Anonymous function (tests.js:14028:17)
```
 by design: see [here](http://blogs.msdn.com/b/ie/archive/2012/02/17/sub-pixel-rendering-and-the-css-object-model.aspx)

 Not sure why phantomjs is returning these non-integer values though, or why the test run under travis fails.

softwords:
Clean test now, but I have made the byIndex function the default key instead of noConstancy. noConstancy solves the issues around AnimateOnNextRender, but in other tests 9 e.g. _metaDataTests=>Dataset is passed in_ are expecting constancy by index to work... so there is a bit of contradiction.

Needs more discussion before pulling I think.

bluong:
A huge step forward as the pull request is now a lot more reviewable :smiley: 

I can review it on my end now.  In the meantime, can you look to resolve the "plottable.d.ts" issues.  Also, if you haven't already, can you submit a signed version of the CLA in the repo?  Thanks!

bluong:
Hmm, if you merged in the other pull request into this one, it still seems that there is a huge diff in "plottable.d.ts"

softwords:
Yes, I know. But, when I look at the raw files of plottable.d.ts in this PR and in the develop branch, I can't see any difference at all to the naked eye, so it's very puzzling. No other diff tool (ie in Visual Studio or notepad++) sees them as different either. Could it be that, because the earlier commit 'Compiled Outputs', effectively threw the whole file away and wrote it again, that all diffs on subsequent commits accumulate that change? In other words, does git diff compare the 2 files absolutely ( regardless of their history) or does it somehow compute the diff by accumulating the changes in all the commits between them? I'll continue to research this.
Either way, we can't release this file into the develop branch. 
Question: when you merge a PR - do you take the plottable.js and .d.ts included in the PR into develop, or do you recalculate them?

softwords:
fyi for anybody interested, I'm finding that the git support in VS2015 is too simplistic really for a project of this complexity, I'm running Atlassian sourceTree in parallel to VS, changes made in SourceTree do get immediately reflected in VS.

bluong:
I believe that `git diff` only looks at the absolute changes without regarding the history

bluong:
We do indeed take the `.js` / `.d.ts` file included in the PR since we don't currently have a way to compile these built files to the develop branch after-merge.

src/core/dataset.ts
```diff
@@ -4,9 +4,20 @@ module Plottable {
...
 export type DatasetCallback = (dataset: Dataset) => void;
 
+export class KeyFunctions {
+  protected static counter: number = 0;
+  public static noConstancy: (d: any, i: number) => any = (d: any, i: number) => {
+    return KeyFunctions.counter++;
+  };
+  public static byIndex: (d: any, i: number) => any = (d: any, i: number) => {
+    return i;
+  };
```
bluong: We don't need to explicitly type the above

bluong: Rename `byIndex` to `index` maybe?

src/core/dataset.ts
```diff
@@ -90,5 +101,23 @@ export class Dataset {
...
+
+  public key(): (datum: any, index: number) => any;
+  /**
+   * Sets the key.
+   *
+   * @param { (d: any, i: number) => any} key
+   * @returns {Dataset} The calling Dataset.
+   */
+  public key(key: (datum: any, index: number) => any): Dataset;
+  public key(key?: (datum: any, index: number) => any): any {
```
bluong: More like `keyFunction` right?

src/drawers/drawer.ts
```diff
@@ -76,7 +76,14 @@ export class Drawer {
    * @param{any[]} data The data to be drawn
    */
   private _bindSelectionData(data: any[]) {
-    let dataElements = this.selection().data(data);
+    // if the dataset has a key, use it when binding the data   
+    let dataElements: d3.selection.Update<any>;
+    if (this._dataset && this._dataset.key()) {
```
bluong: I believe it can be assumed that `this._dataset` always exists.  Verify if this is the case?

src/plots/scatterPlot.ts
```diff
@@ -83,12 +83,6 @@ export module Plots {
 
     protected _generateDrawSteps(): Drawers.DrawStep[] {
       let drawSteps: Drawers.DrawStep[] = [];
-      if (this._animateOnNextRender()) {
-        let resetAttrToProjector = this._generateAttrToProjector();
-        resetAttrToProjector["d"] = () => "";
-        drawSteps.push({attrToProjector: resetAttrToProjector, animator: this._getAnimator(Plots.Animator.RESET)});
-      }
-
```
bluong: Does this mean we can take out other similar calls from the other plots?

src/drawers/drawer.ts
```diff
@@ -76,7 +76,14 @@ export class Drawer {
...
+      dataElements = this.selection().data(data, this._dataset.key());
+    } else {
+      dataElements = this.selection().data(data);
+    }
```
bluong: Mainly curious: usually we have users re-animate their plot using `dataset.data(data)` even though that is hacky in my opinion.  After this change, would that idea still work?

src/core/dataset.ts
softwords: how about ```useIndex``` ?

src/core/dataset.ts
softwords: Agreed, that's probably better, and in keeping with the terminology here: http://bost.ocks.org/mike/constancy/

src/drawers/drawer.ts
softwords: I agree with you that it feels hacky, but I don't want to break that.
Stepping back a bit, the goal here ultimately is to support animations that can enrich the data visualisation by allowing the user to see how data points change from dataset to dataset, and to be made aware of points that enter the dataset, and points that leave. 
This [quote] (http://bost.ocks.org/mike/constancy/) from Mike Bostok sums this up I think:
_Above all, animation should be meaningful. While it may be visually impressive for bars to fly around the screen during transitions, animation should only be used when it enhances understanding._

I have banged on about this at some length in this issue:  #2512 
To get to that point in plottable, there are a couple of things that need to happen, and this PR is the first of them:
1) we need a way to specify a key function that allows a meaningful connection to be made between data points in different datasets (This PR does that much)
2) we need to support distinct animations on entering and exiting points - the [enter/update/exit] (https://gist.github.com/mbostock/3808218) is hidden inside Drawer - I believe that it needs to be exposed to the Animator. 
3) Once there is a key to define the relationship between elements, and the Animator can manipulate each of the update, enter and exit selections, we need new versatile concrete animators to take advantage of that. 

If this paradigm is in place, then resetting the data to identical data should not cause any animation. If you want the animation as it is currently implemented, then you could explicitly use the ```noConstancy``` key function to ensure there is never any join between the incoming and outgoing datasets - this means that every datum will be included in the enter() selection, and handled by the Animator accordingly.

But the current ```animateOnNextRender``` implementations I would argue are incompatible with this "object constancy" approach to animation, so they probably should go.

Now step 2) above is probably more radical and I'm fleshing this out now - but I'd need guidance from you about changing (breaking?) the IAnimator interface,and I can appreciate you may want to take a careful look at that before heading down this path.
But if we put back the piece in ScatterPlot, the other changes would not affect any existing code - data binding would use the default key ```useIndex``` which is the d3 default anyway.

bluong: I think I see what you are getting at.  I think the idea that you are proposing definitely makes sense.  I am mainly curious in what you thought about us removing that previous workflow inherently (even though it seems that if users were to use `noConstancy` then they would be in the clear in the first place).

src/core/dataset.ts
```diff
@@ -4,9 +4,20 @@ module Plottable {
 
 export type DatasetCallback = (dataset: Dataset) => void;
 
+export class KeyFunctions {
+  protected static counter: number = 0;
+  public static noConstancy: (d: any, i: number) => any = (d: any, i: number) => {
+    return KeyFunctions.counter++;
+  };
```
bluong: Just curious, what's the difference between this and not using a key function at all?

bluong:
Is it possible for you to put up a JSFiddle so that we can see what the results of this would be?  To my understanding, this essentially enables that data knows if they are being newly entered or newly removed, meaning that we can have only new data points on a `Plots.Scatter` to render instead of rerendering the entire plot.  Is this correct?

bluong:
Here is a very generic JSFiddle that I use as a baseline: http://jsfiddle.net/bluong63/zsfkg32b/

Feel free to fork from it (You'll need to add a script tag to the html section to retrieve the plottable.js from this branch.  You can check the External Resources tab on the left to see how develop's plottable is being loaded in by default using rawgithub).

src/core/dataset.ts
softwords: If you call ```selection.bind(data)``` without supplying a key function, d3 defaults to using the index of the datum in the data array as the key - ie same as the ```useIndex``` function here.  This is why, currently, if you simply replace the data with itself - ```dataset.data(data)``` - no animation would occur without the ```animateOnNextRender``` logic. That forces everything back to a base position so it can then move back to where it was before.

softwords:
Here's some of the fiddles I prepared some time back to demonstrate ome of these ideas (these were created as quick-and-dirty proof-of-concept by directly editing the plottable javascript) 
Data is the final 32 countries in the football world cup for 2006/10/14: name, rank (1-32), goals for, goals against.  Some countries come in and go out each year,

http://jsfiddle.net/oaskqta0/5/ bar chart - the bar representing a country moves to the rank for that country in each dataset. Watch Uruguay come in (2010) and go out. this one lets you see the effect of changing the key function.

http://jsfiddle.net/d4gngj2s/3/ - http://jsfiddle.net/d4gngj2s/4/ these are scatter plot - a dot that represents a country stays with that country when you change datasets. they use different animations for exit and enter.

src/core/dataset.ts
bluong: I think `byIndex` is fine

bluong:
These fiddles look really nice.  Thanks for the proof-of-concept!

Mainly so that we can get a better gathering of this pull request, can you generate a JSFiddle using only the additions made in this pull request?  It seems that there are prototyped `plot.enter()` / `plot.exit()` endpoints.

I'm curious on what additional functionality only this JSFiddle provides us, so that users / we know what features we get with this pull request.

On an overall note, can you respond to the above comments / make the appropriate changes.  Also can you make the similar changes to the other `Plots`?

softwords:
The ```plot.enter``` and ```plot.exit``` endpoints were an earlier idea, @jtlan  and @aicioara talked me out of that. What I am working towards now is that the Animator will manage that. So those fiddles are useful for the "ends", not the "means".

plottable.js
```diff
@@ -3,7 +3,7 @@ Plottable 1.14.0 (https://github.com/palantir/plottable)
 Copyright 2014-2015 Palantir Technologies
 Licensed under MIT (https://github.com/palantir/plottable/blob/master/LICENSE)
 */
-
+//# sourceURL=plottable.js
```
bluong: Why is this here?

src/core/dataset.ts
```diff
@@ -4,9 +4,23 @@ module Plottable {
...
+  public static useIndex: (d: any, i: number) => any = (d: any, i: number) => {
```
bluong: Remove explicit typing.

src/core/dataset.ts
```diff
@@ -4,9 +4,23 @@ module Plottable {
 
 export type DatasetCallback = (dataset: Dataset) => void;
 
+export class KeyFunctions {
+  protected static counter: number = 0;
+  public static noConstancy: (d: any, i: number) => any = (d: any, i: number) => {
```
bluong: Remove explicit typing.

src/core/dataset.ts
```diff
@@ -4,9 +4,23 @@ module Plottable {
 
 export type DatasetCallback = (dataset: Dataset) => void;
 
+export class KeyFunctions {
+  protected static counter: number = 0;
```
bluong: Why `protected`?

src/core/dataset.ts
```diff
@@ -4,9 +4,23 @@ module Plottable {
 
 export type DatasetCallback = (dataset: Dataset) => void;
 
+export class KeyFunctions {
```
bluong: does it make more sense to store this in a module
`export module KeyFunctions {`

src/core/dataset.ts
```diff
@@ -90,5 +104,23 @@ export class Dataset {
       return this;
     }
   }
+
+  public keyFunction(): (datum: any, index: number) => any;
+  /**
+   * Sets the key.
```
bluong: "Sets the keyFunction."

Also can you explain what a keyFunction is here in a very brief summary (1-3 lines)

src/core/dataset.ts
```diff
@@ -90,5 +104,23 @@ export class Dataset {
...
+   *
+   * @param { (d: any, i: number) => any} key
```
bluong: keyFunction

src/plots/scatterPlot.ts
bluong: Curious about this

src/drawers/drawer.ts
```diff
@@ -76,7 +76,14 @@ export class Drawer {
    * @param{any[]} data The data to be drawn
    */
   private _bindSelectionData(data: any[]) {
-    let dataElements = this.selection().data(data);
+    // if the dataset has a key, use it when binding the data   
+    let dataElements: d3.selection.Update<any>;
+    if (this._dataset && this._dataset.keyFunction()) {
```
bluong: `Drawer`s are instantiated with a `dataset`.  It can be assumed that `this._dataset` always exists

src/core/dataset.ts
```diff
@@ -4,9 +4,23 @@ module Plottable {
...
+  };
+  public static useProperty(propertyname: string) {
+    return (d: any, i: number) => { return d[propertyname]; };
+  };
+}
+
 export class Dataset {
   private _data: any[];
   private _metadata: any;
+  private _keyFunction: (datum: any, index: number) => any = KeyFunctions.useIndex;
```
bluong: Remove explicit typing

src/drawers/drawer.ts
```diff
@@ -76,7 +76,14 @@ export class Drawer {
...
+      dataElements = this.selection().data(data, this._dataset.keyFunction());
+    } else {
+      dataElements = this.selection().data(data);
+    }
```
bluong: It seems that based in `Dataset` that we always have a `keyFunction()` (since we start with a default).  Thus, can we remove the if block?

test/core/keyTests.ts
```diff
@@ -0,0 +1,41 @@
...
+    let d: any = { foo: "bar" };
+    let a: number = ds.keyFunction()(d, 1);
+    let b: number = ds.keyFunction()(d, 1);
+    assert.isTrue(b - a === 1, "invocations give numerically increasing results");
+  });
+  it(" defaults to useIndex predefined key function", () => {
+    let ds: Plottable.Dataset = new Plottable.Dataset();
+    assert.isTrue(ds.keyFunction() === Plottable.KeyFunctions.useIndex, "useIndex is default");
+  });
+  it(" may accept useIndex predefined key function", () => {
```
bluong: Remove above leading whitespace please in the tests

test/core/keyTests.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+  describe("DatasetKey noConstancy", () => {
+    it("generates a different value each time invoked", () => {
+      let keyFunction = Plottable.KeyFunctions.noConstancy;
+      let d: any = { foo: "bar" };
+      let a: number = keyFunction(d, 1);
+      let b: number = keyFunction(d, 1);
+
+      assert.isTrue(b - a === 1, "invocations give numerically increasing results");
+    });
```
bluong: Remove the explicit typings in the above variable declarations

plottable.js
softwords: Sorry my mistake leaving that in there - I've been working with your (elegant and ingenious) quicktests... this line enables the chrome developer tools to display plottable.js in the sources page, even when it is dynamically loaded. this makes it easier to set breakpoints. I think it would be a useful thing to have in there for that reason (and similarly, in each of the quicktests) but - that's a separate issue for another day.

bluong:
@softwords Yep, I think I see what you are saying.  I am mainly suggesting that you make a JSFiddle so that I, other viewers, and our QE can have a proof-of-concept of this specific pull request change.

In interacting with your Scatter Plot example, it seems that this can work with some tweaking (Given the src attribute, it is hard to tell that the input "plottable.js" actually comes from pull request.  We generally use rawgit.com for these situations.)

src/core/dataset.ts
softwords: You're right, I'll change to private.

bluong:
In terms of the overall pull request, most of my comments are essentially very small changes / cleaning up.  My only concern revolves around the Scatter Plot change.  Do we want to now make this change across the rest of the plots?  Or do we just have it on Scatter Plot?

src/core/dataset.ts
softwords: noted

softwords:
Here's two variations on the previous fiddle:
http://jsfiddle.net/oaskqta0/6/
This one is built using the current PR. This uses the existing animateOnRender logic in BarPlot. There are no longer transitions on enter or exit. The main difference you'll see between constancy and noconstacy is that, with constancy, the bars remember their height. With noConstancy each bar grows from the origin in its new position every time. In both cases they do not transition to their new positions - they just jump - this is because of the extra drawstep inserted (with Null Animator) by the animateOnRender logic.
http://jsfiddle.net/oaskqta0/7/
this one is built from a new branch ( no PR) . It removes the animateOnRender logic from BarPlot. Now you can see the updated selection (ie those countries appearing in both datasets) transition to both their new height and new position. But, there is still no transition on the enter or exit selections, because the Animator can't get at these yet.

softwords:
http://jsfiddle.net/oaskqta0/8/
Same data, current PR, scatter chart. The enter() selection squirts in from top left. This is quite dramatic with noConstancy. The reason for this is that they are not initialised with any properties when they are created, so the transition to their ultimate values begins at 0,0. (This is a bit different to what happens now  - the animateOnRender kills only the ```d``` attribute and retains the transform which fixes the x,y) . In the d3 enter/update/exit pattern,  there is usually some special setup done on the entering elements to initialise various attributes even before they are transitioned to their final size and position. With a richer IAnimator interface, we can make a hook to do this.

test/core/keyTests.ts
softwords: not sure what you're after... the indenting here looks the same as in other tests?

bluong: Oh I mean the whitespace in front of the strings
the
it(" may... change it to
it("may...

bluong: Also, can you place a newline in between the it blocks?  Mainly to get a consistent look across the tests

bluong:
Do you plan on having the `Plots.Bar` scenario to be part of this pull request?  Also, for that scenario, we usually have the first animation start from the baseline.  Is that incorporated?

softwords:
>> Do you plan on having the Plots.Bar scenario to be part of this pull request? 

I'll be guided by you as to how much / how little of this you want to introduce in this PR or delay to subsequent PR.  This change is fundamental ( see the 3 steps I described earlier) , but of itself, doesn't add a lot of functionality. From your perspective, I can appreciate that you may want a better idea of where all this is heading before committing anything. 
If you commit now, the thing people will see is a change in the animation behaviour of scatter plots. They will have some control over this, but the "squirting in from top-left" behaviour is what they'll see at first.
On the other hand, we could reduce this PR ( by taking out the change to ScatterPlot) so that it makes no impact on any existing code at all. But by doing that, setting KeyFunction will never have any effect.
Or, we can hold off while I work on the second step - which involves IAnimator changes - and roll it all in in one go. Then there will be more fine-grained control over animations, but it becomes a big PR.
My suggestion would be, let's put ScatterPlot back the way it was, QE won't see any difference, but we can then move forward on a clean page to build on this foundation.

src/drawers/drawer.ts
softwords: I gets errors in the test suite without the test  for Dataset
Abstract Drawer drawer timing works as expected for null animators:
Abstract Drawer drawer timing works for non-null animators:

I suspect that the test just sets up an unrealistic situation, but it does demonstrate that its possible to pass the Dataset as null to the constructor of Drawer.

bluong: What probably happened is that the tests explicitly passed in `null`.  Let's leave this in for now then.  I can go look for a fix on my end.

softwords:
by the way, thanks for all your patience and assistance, it's much appreciated

bluong:
In talking with QE, I think a solution that came up with was essentially to create a side branch that branches off of develop, and then you can make pull requests to that side branch.  We'll then iterate on each of those pull requests and then once we feel that the side branch is ready to go, we hopefully simply merge that in.  How does that sound to you?

In this regard, yes let's remove the Scatter Plot functionality and stick with only the KeyFunction idea.  Place that functionality in a separate pull request only dealing with the Scatter Plot change.

To my understanding I don't think non-contributors can create branches on our end, I will create a branch at dataConstancy .  If this makes sense, can you make a new pull request that targets that branch instead?  If so, make have this pull request reference that pull request and vice versa.

bluong:
I definitely want to shout my thanks back to you for sticking with our development cycle.  Your contribution definitely looks really promising and I would like to find a way to include it into Plottable in the most effective manner.

softwords:
>>  create a side branch that branches off of develop

that's sound like a good plan - can I suggest you call it _objectConstancy_, in keeping with Mike Bostock's terminology?

I'll wait to hear back on that.
Meantime, I'm hoping to see your changes to the compile sequence in the gruntfile soon, which (mysteriously) fixes my ```plottable.d.ts``` troubles.

bluong:
I'll change the branch to objectConstancy.

In regards to the compile sequence, you can try testing your pull request by rebasing on my pull request, possibly.  We are expecting that to merge by hopefully early next week.

bluong:
Branch created, feel free to make a pull request to that branch

bluong:
Just in case, Github does not allow the changing of the target branch on a pull request.  Instead, you will need to create a new pull request with the same source branch and change the target branch to "objectConstancy".

bluong:
Closing this pull request since it seems we have a reference to the new pull request.

PR #2834, opened 2015-10-02
[Plots.Line] - croppedRendering by default

bluong:
Having this be on by default.

Tests change because having an array go through a map does not return the same reference, but `assert.include` uses reference checking I believe -___-


crmorford:
did a full regression pass

PR #2837, opened 2015-10-02
Add downsampling performance option

yunhaolucky:
New Performance option for `Plots.Line`.

When downsampling is enabled in a Line Plot, the number of lines will decrease significantly if there are lots of lines whose ends have the same x/y values and overlap with each other. This change will also affect Area Plot and Stacked Area Plot.

API change:
```typescript
/**
* Gets the downsampling performance option state
*
* When downsampling is enabled, lines whose two ends have the same scaled x/y values will not be drawn on top of each other.
*/
downsamplingEnabled(): boolean;
/**
* Sets if the downsampling performance option is enabled
*
* @returns {Plots.Line} The calling Plots.Line
*/
downsamplingEnabled(downsample: boolean): Plots.Line<X>;
```

Demo:
Before: http://jsfiddle.net/yunhaolucky/yxpq4uuv/11/
After: http://jsfiddle.net/yunhaolucky/yxpq4uuv/15/

Part of #2523


plottable.d.ts
```diff
@@ -3419,6 +3419,18 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
             /**
+             * Gets the downsampling performance option state
+             *
+             * When downsampling is enabled, lines whose two ends have the same scaled x/y values will not be drawn on top of each other.
```
ztsai: update the docs

src/plots/linePlot.ts
```diff
@@ -138,6 +140,25 @@ export module Plots {
       this.render();
       return this;
     }
+    /**
+     * Gets the downsampling performance option state
+     * 
+     * When downsampling is enabled, lines whose two ends have the same scaled x/y values will not be drawn on top of each other.
```
ztsai: update the docs

src/plots/linePlot.ts
```diff
@@ -375,14 +396,18 @@ export module Plots {
...
-        if (!this._croppedRenderingEnabled) {
+        if (!this._croppedRenderingEnabled && !this._downsamplingEnabled) {
           dataToDraw.set(dataset, [data]);
           return;
         }
 
         let filteredDataIndices = data.map((d, i) => i);
-
+        if (this._croppedRenderingEnabled) {
         filteredDataIndices = this._filterCroppedRendering(dataset, filteredDataIndices);
```
ztsai: indent

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
       return filteredDataIndices;
     }
 
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
```
ztsai: why do you need 2 functions here? it seems like _filterDownsampling doesn't do much other than getting accessors

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+      let xAccessor = this.x().accessor;
+      let yAccessor = this.y().accessor;
+      let filteredIndices = this._filterDownsamplingSlope(dataset, indices, this.x().scale, this.y().scale, xAccessor, yAccessor);
+      return filteredIndices;
+    }
+
+    private _filterDownsamplingSlope(dataset: Dataset, indices: number[],
+      xScale: Scale<any, number>, yScale: Scale<any, number>, xAcessor: Accessor<any>, yAcessor: Accessor<any>) {
```
ztsai: is there any reason why you need to pass in scales and accessors here? they should be accessible via `this`

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+      let filteredIndices: number[] = [];
+      let min: number;
+      let max: number;
+      let currentSlope: number;
+      if (indices.length === 0) {
+        return filteredIndices;
+      }
+      filteredIndices.push(indices[0]);
+      function belongToCurBucket(i: number) {
+        let p1x = xScale.scale(xAcessor(data[indices[i]], indices[i], dataset));
```
ztsai: you made an assumption that xScale.scale exists which might not be the case
you can use `Plot._scaledAccessor` function instead

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+        let p1y = yScale.scale(yAcessor(data[indices[i]], indices[i], dataset));
+        let p2x = xScale.scale(xAcessor(data[indices[i + 1]], indices[i + 1], dataset));
+        let p2y = yScale.scale(yAcessor(data[indices[i + 1]], indices[i + 1], dataset));
+        if (currentSlope == null) {
+          currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
+          min = (currentSlope === Infinity) ? yAcessor(data[indices[i]], indices[i], dataset) :
```
ztsai: `yAcessor(data[indices[i]], indices[i], dataset)` is just `p1y`

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+            xAcessor(data[indices[i]], indices[i], dataset);
+          max = min;
+          return true;
+        }
+        if (currentSlope === Infinity) {
+          return Math.floor(p1x) === Math.floor(p2x);
+        }else {
```
ztsai: space

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+          let expectedP2y = p1y + (p2x - p1x) * currentSlope;
+          return Math.floor(p2y) === Math.floor(expectedP2y);
+        }
+      }
+      for (let i = 0; i < indices.length - 1; ) {
```
ztsai: personally prefer while loop but fine either way

plottable.js
```diff
@@ -9296,6 +9309,73 @@ var Plottable;
...
+            Line.prototype._filterDownsampling = function (dataset, indices) {
+                var xAccessor = this.x().accessor;
+                var yAccessor = this.y().accessor;
+                var filteredIndices = this._filterDownsamplingSlope(dataset, indices, this.x().scale, this.y().scale, xAccessor, yAccessor);
+                return filteredIndices;
+            };
+            Line.prototype._filterDownsamplingSlope = function (dataset, indices, xScale, yScale, xAcessor, yAcessor) {
+                var data = dataset.data();
+                var filteredIndices = [];
+                var min;
```
ztsai: initialize min to `Infinity` and max to `-Infinity`

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+        return filteredIndices;
+      }
+      filteredIndices.push(indices[0]);
+      function belongToCurBucket(i: number) {
+        let p1x = xScale.scale(xAcessor(data[indices[i]], indices[i], dataset));
+        let p1y = yScale.scale(yAcessor(data[indices[i]], indices[i], dataset));
+        let p2x = xScale.scale(xAcessor(data[indices[i + 1]], indices[i + 1], dataset));
+        let p2y = yScale.scale(yAcessor(data[indices[i + 1]], indices[i + 1], dataset));
+        if (currentSlope == null) {
+          currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
```
ztsai: I feel like checking `Math.abs(p2x - p1x)` is less a certain threshold is more accurate than comparing the floored values
but the difference is probably very minor

src/plots/linePlot.ts
```diff
@@ -425,6 +450,74 @@ export module Plots {
...
+          i++;
+        }
+        let pLast = indices[i];
+        if (pMin !== pFirst) {
+          filteredIndices.push(pMin);
+        }
+        if (pMax !== pMin && pMax !== pFirst) {
+          filteredIndices.push(pMax);
+        }
+        if (pLast && pLast !== pFirst && pLast !== pMin && pLast !== pMax) {
```
ztsai: why pLast?

test/plots/linePlotTests.ts
```diff
@@ -830,5 +829,119 @@ describe("Plots", () => {
...
+        svg.remove();
+      });
+
+      it("does not render points that should be removed in downsampling", () => {
+        let data = [
+          {x: -10, y: -1}, // last element in previous bucket
+          {x: 0, y: 2}, // first element in current bucket
+          {x: 0.5, y: 1.5}, // the point to be removed 
+          {x: 1, y: 1}, // minimum y in current bucket
+          {x: 2, y: 4}, // maximum y in current bucket
```
ztsai: why min/max Y matters? you are comparing `x` values for all finite slope lines

test/plots/linePlotTests.ts
```diff
@@ -830,5 +829,119 @@ describe("Plots", () => {
...
+        assert.strictEqual(plot.downsamplingEnabled(true), plot, "enabling the downsampling option returns the plot");
+        assert.isTrue(plot.downsamplingEnabled(), "can enable the downsampling option");
+
+        plot.downsamplingEnabled(false);
+        assert.isFalse(plot.downsamplingEnabled(), "can disable the downsampling option");
+
+        svg.remove();
+      });
+
+      it("does not render points that should be removed in downsampling", () => {
```
ztsai: if you are testing for horizontal lines can you include that in the test case?

test/plots/linePlotTests.ts
```diff
@@ -830,5 +829,119 @@ describe("Plots", () => {
...
+        let data = [
+          {x: 2, y: 2}, // first element in the bucket
+          {x: 1, y: 1}, // minimum element in the bucket
+          {x: 10, y: 10}, // maximum element in the bucket
+          {x: 2.5, y: 2.5}, // the point to be removed
+          {x: 3, y: 3}, // last element in the bucket
+        ];
+        plot.addDataset(new Plottable.Dataset(data));
+
+        yScale.domain([0, 10]);
```
ztsai: is this needed?

src/plots/linePlot.ts
yunhaolucky: My concert about using `Math.abs(p2x - p1x)` is suppose we have 3 points with x value(1.1, 2.0, 2.9), they will be put in the same bucket if we use `Math.floor(Math.abs(p2x - p1x)) === 0`. But they are supposed to be in different buckets.

src/plots/linePlot.ts
yunhaolucky: Merged into one function

test/plots/linePlotTests.ts
yunhaolucky: I think this is needed because we need a test to make sure slope method works.

plottable.js
yunhaolucky: I initialize min and max to the  x/y value of the first point in each bucket.

src/plots/linePlot.ts
yunhaolucky: `p1y` is the scaled y value, which means the maximum of `p1y` actually has the smallest y value in dataset. But as scale is a linear function, this shouldn't be a problem. I change  `yAcessor(data[indices[i]], indices[i], dataset)` to `p1y`.

test/plots/linePlotTests.ts
```diff
@@ -830,5 +829,121 @@ describe("Plots", () => {
...
+          {x: 2, y: 2}, // first element in the bucket
+          {x: 1, y: 1}, // minimum element in the bucket
+          {x: 10, y: 10}, // maximum element in the bucket
+          {x: 2.5, y: 2.5}, // the point to be removed
+          {x: 3, y: 3}, // last element in the bucket
+          {x: 3, y: 1}, // first element in next bucket
+        ];
+        plot.addDataset(new Plottable.Dataset(data));
+
+        yScale.domain([0, 10]);
```
ztsai: I still don't understand why this is needed.  all your data points lie in the same line already without scaling

test/plots/linePlotTests.ts
```diff
@@ -830,5 +829,121 @@ describe("Plots", () => {
...
+        plot.renderTo(svg);
+
+        let path = plot.content().select("path.line").attr("d");
+        let expectedRenderedData = [0, 1, 2, 3, 5, 6].map((d) => data[d]);
+        checkPathForDataPoints(path, expectedRenderedData);
+
+        svg.remove();
+      });
+
+      function checkPathForDataPoints(path: string, data: {x: number, y: number}[]) {
```
ztsai: this method seems identical to the one in the previous describe block
can you factor this out?
maybe put it in TestMethods?

ztsai:
@bluong, would you mind checking the logic?  I just want to double check as I feel like I might have blind spot

test/plots/linePlotTests.ts
yunhaolucky: There are 5 points on that line but only 4 points are drawn. Which means we remove the point that is not the first, the last or the min or the max.

plottable.d.ts
```diff
@@ -3419,6 +3419,18 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
             /**
+             * Gets the downsampling performance option state
+             *
+             * When downsampling is enabled, two consecutive lines with the same loop will be merged to one line.
```
bluong: What does this mean
"two consecutive lines with the same loop"?

plottable.d.ts
```diff
@@ -3419,6 +3419,18 @@ declare module Plottable {
             interpolator(interpolator: "cardinal-closed"): Line<X>;
             interpolator(interpolator: "monotone"): Line<X>;
             /**
+             * Gets the downsampling performance option state
```
bluong: "Gets if downsampling is enabled."

plottable.d.ts
```diff
@@ -3419,6 +3419,18 @@ declare module Plottable {
...
+             *
+             * When downsampling is enabled, two consecutive lines with the same loop will be merged to one line.
+             */
+            downsamplingEnabled(): boolean;
+            /**
+             * Sets if the downsampling performance option is enabled
```
bluong: "Sets if downsampling is enabled."

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+      let currentSlope: number;
+      let scaledXAcessor = Plot._scaledAccessor(this.x());
+      let scaledYAcessor = Plot._scaledAccessor(this.y());
+
+      if (indices.length === 0) {
+        return filteredIndices;
+      }
+      filteredIndices.push(indices[0]);
+
+      function belongToCurBucket(i: number) {
```
bluong: Let's not abbreviate the names here

I assume that `Cur` is `Current` but I am unsure.

Also, what are you checking is "belonging to the "Cur" Bucket"?  Is it the index?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
     }
 
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
+      let data = dataset.data();
+      let filteredIndices: number[] = [];
+      let min: number;
+      let max: number;
+      let currentSlope: number;
+      let scaledXAcessor = Plot._scaledAccessor(this.x());
+      let scaledYAcessor = Plot._scaledAccessor(this.y());
```
bluong: Spelling is incorrect

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+      }
+      filteredIndices.push(indices[0]);
+
+      function belongToCurBucket(i: number) {
+        let p1x = scaledXAcessor(data[indices[i]], indices[i], dataset);
+        let p1y = scaledYAcessor(data[indices[i]], indices[i], dataset);
+        let p2x = scaledXAcessor(data[indices[i + 1]], indices[i + 1], dataset);
+        let p2y = scaledYAcessor(data[indices[i + 1]], indices[i + 1], dataset);
+        if (currentSlope == null) {
+          currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
```
bluong: Why usage of `Math.floor` here?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
       return filteredDataIndices;
     }
 
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
+      let data = dataset.data();
+      let filteredIndices: number[] = [];
+      let min: number;
+      let max: number;
```
bluong: What are these the `max` and `min` of?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+
+      function belongToCurBucket(i: number) {
+        let p1x = scaledXAcessor(data[indices[i]], indices[i], dataset);
+        let p1y = scaledYAcessor(data[indices[i]], indices[i], dataset);
+        let p2x = scaledXAcessor(data[indices[i + 1]], indices[i + 1], dataset);
+        let p2y = scaledYAcessor(data[indices[i + 1]], indices[i + 1], dataset);
+        if (currentSlope == null) {
+          currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
+          min = (currentSlope === Infinity) ? p1y : p1x;
+          max = min;
```
bluong: Since there is very little context on `min`/`max`, the above two lines are not too understandable

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+          return true;
+        }
+        if (currentSlope === Infinity) {
+          return Math.floor(p1x) === Math.floor(p2x);
+        } else {
+          let expectedP2y = p1y + (p2x - p1x) * currentSlope;
+          return Math.floor(p2y) === Math.floor(expectedP2y);
+        }
+      }
```
bluong: Can you add a newline after this function block.  This is a lot to take in without using a newline to break up the flow.

src/plots/linePlot.ts
bluong: I so far do not see a concept of a bucket.  Where is this "bucket" referenced?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+      for (let i = 0; i < indices.length - 1; ) {
+        currentSlope = null;
```
bluong: If we're going to be resetting the currentSlope every time, why not scope the variable to inside this for loop?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+        let pFirst = indices[i];
+        let pMin = indices[i];
+        let pMax = indices[i];
+        while (i < indices.length - 1 && belongToCurBucket(i)) {
+          let currPoint = (currentSlope === Infinity ? scaledYAcessor(data[indices[i + 1]], indices[i + 1], dataset) :
```
bluong: Why need the outer parens?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+        } else {
+          let expectedP2y = p1y + (p2x - p1x) * currentSlope;
+          return Math.floor(p2y) === Math.floor(expectedP2y);
+        }
+      }
+      for (let i = 0; i < indices.length - 1; ) {
+        currentSlope = null;
+        let pFirst = indices[i];
+        let pMin = indices[i];
+        let pMax = indices[i];
```
bluong: Please expand the variable names to explain what they mean

src/plots/linePlot.ts
```diff
@@ -425,6 +449,69 @@ export module Plots {
...
+        let pLast = indices[i];
+        if (pMin !== pFirst) {
+          filteredIndices.push(pMin);
+        }
+        if (pMax !== pMin && pMax !== pFirst) {
+          filteredIndices.push(pMax);
+        }
+        if (pLast !== pFirst && pLast !== pMin && pLast !== pMax) {
+          filteredIndices.push(pLast);
+        }
```
bluong: Since little context on `pMin` `pMax` `pLast` and `pFirst`, it is very hard to understand what's happening here

bluong:
It's pretty hard to understand how the downsampling algorithm works.  Can we see if there's a way to break up the logic / expand on the small-contexted items to have it be more readable?

src/plots/linePlot.ts
yunhaolucky: If the p1.x and p2.x are close enough, their slope should be infinite and they should belong to the same bucket.

src/plots/linePlot.ts
yunhaolucky: Because `belongToCurrentBucket` needs to set `currentSlope`, therefore I need to make it outside the for loop.

src/plots/linePlot.ts
```diff
@@ -425,6 +449,72 @@ export module Plots {
...
+      let currentSlope: number;
+      let scaledXAccessor = Plot._scaledAccessor(this.x());
+      let scaledYAccessor = Plot._scaledAccessor(this.y());
+
+      if (indices.length === 0) {
+        return filteredIndices;
+      }
+      filteredIndices.push(indices[0]);
+
+      let belongingToCurrentBucket = (i: number) => {
```
bluong: Would the function be better named to `indexBelongsToCurrentBucket`?  I guess the thing that is also not obvious is what is the "CurrentBucket"?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,72 @@ export module Plots {
...
+          maxScaledValue = minScaledValue;
+          return true;
+        }
+        if (currentSlope === Infinity) {
+          return Math.floor(p1x) === Math.floor(p2x);
+        } else {
+          let expectedP2y = p1y + (p2x - p1x) * currentSlope;
+          return Math.floor(p2y) === Math.floor(expectedP2y);
+        }
+      };
```
bluong: After This function seems to be doing a lot (It sets the `minScaledValue`, `maxScaledValue`, and `currentSlope` variables as well as returning a `boolean`.

Can we split up these two pieces of functionality?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,72 @@ export module Plots {
...
+
+      for (let i = 0; i < indices.length - 1; ) {
+        currentSlope = null;
```
bluong: The fact that we have to reset `currentSlope` every time signals the fact that this variable should only be scoped inside the for loop.

If the item prventing this is that `currentSlope` is defined inside the function, then let's use `currentSlope` as an argument to a function rather than having it be a more globalized variable.

src/plots/linePlot.ts
```diff
@@ -425,6 +449,72 @@ export module Plots {
...
+        let indexFirst = indices[i];
+        let indexMin = indices[i];
+        let indexMax = indices[i];
```
bluong: Based on reading this, I also think `indices.forEach` is cleaner

bluong:
Functionality got slightly cleaner to read, but we can still work on cleaning it up.

src/plots/linePlot.ts
yunhaolucky: I feel the function itself defines bucket. In the function, we put the end points of adjacent lines that have same slopes into one bucket.

src/plots/linePlot.ts
yunhaolucky: Because there is a while loop inside, indices.forEach might not work.

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
+      let data = dataset.data();
+      let filteredIndices: number[] = [];
+
+      let scaledXAccessor = Plot._scaledAccessor(this.x());
+      let scaledYAccessor = Plot._scaledAccessor(this.y());
+
+      if (indices.length === 0) {
+        return filteredIndices;
+      }
```
bluong: This looks like a piece of code that would be more clear + miniscule more performant if it was placed first
```typescript
if (indices.length === 0) {
  return [];
}
```

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+        }
+      };
+
+      for (let i = 0; i < indices.length - 1; ) {
+        let indexFirst = indices[i];
+        let p1x = scaledXAccessor(data[indices[i]], indices[i], dataset);
+        let p1y = scaledYAccessor(data[indices[i]], indices[i], dataset);
+        let p2x = scaledXAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+        let p2y = scaledYAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+        let currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
```
bluong: Why do we need to edge case like this?  If (p2x were to equal p1x), then `currentSlope` would be `Infinity`

src/plots/linePlot.ts
bluong: Are we using `Math.floor` as a means so non-integer values can be mapped to the same value?

src/plots/linePlot.ts
ztsai: if p1x = p2x and p1y = p2y, then 0 / 0 = NaN

test/plots/linePlotTests.ts
ztsai: my question is that why it doesn't work with auto domain? why does it have to be [0, 10]?

src/plots/linePlot.ts
bluong: While the meta / concept is bucketing, this is not obvious unless you use a more direct name.  Are you suggesting that this index lies on the current slope?  Something like `indexOnCurrentSlope`?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+        let indexMin = indices[i];
+        let minScaledValue = (currentSlope === Infinity) ? p1y : p1x;
```
bluong: Why this edge case?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+            scaledXAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+          if (currPoint > maxScaledValue) {
+            maxScaledValue = currPoint;
+            indexMax = indices[i + 1];
+          }
+          if (currPoint < minScaledValue) {
+            minScaledValue = currPoint;
+            indexMin = indices[i + 1];
+          }
+          i++;
```
bluong: Would it be more clear to place `i++` above the rest of the while loop code?  All of the code above uses `i + 1`, and this is more evidenced by the fact that the variable `currPoint` uses `i + 1` instead of a more expected `i`.

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+        let p2y = scaledYAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+        let currentSlope = (Math.floor(p1x) === Math.floor(p2x)) ? Infinity : (p2y - p1y) / (p2x - p1x);
+        let indexMin = indices[i];
+        let minScaledValue = (currentSlope === Infinity) ? p1y : p1x;
+        let indexMax = indexMin;
+        let maxScaledValue = minScaledValue;
+
+        while (i < indices.length - 1 && indexBelongsToCurrentBucket(i, currentSlope)) {
+          let currPoint = currentSlope === Infinity ? scaledYAccessor(data[indices[i + 1]], indices[i + 1], dataset) :
+            scaledXAccessor(data[indices[i + 1]], indices[i + 1], dataset);
```
bluong: The value of this variable is not a `Point`.  We should rename `currPoint` to maybe like `currentPixelValue` possibly? or `currentPixel`?

src/plots/linePlot.ts
```diff
@@ -425,6 +449,71 @@ export module Plots {
...
+        }
+
+        let indexLast = indices[i];
+
+        if (indexMin !== indexFirst) {
+          filteredIndices.push(indexMin);
+        }
+        if (indexMax !== indexMin && indexMax !== indexFirst) {
+          filteredIndices.push(indexMax);
+        }
```
bluong: In reading the above code, what is the purpose of `indexMin` and `indexMax`?  Why aren't `indexLast` and `indexFirst` enough?

bluong:
It's gotten significantly more understandable, but it still needs some work

src/plots/linePlot.ts
yunhaolucky: Suppose we have several consecutive lines with same slopes. We need the max and min points to draw the entire line. We need the first point and last point to draw the lines to connect the current line with the lines before and after it.

src/plots/linePlot.ts
yunhaolucky: If the line is vertical, we need to compare y values because all of them have the same x values. If the line is not vertical, we can use x to pick the end point of the line.

src/plots/linePlot.ts
bluong: If the lines are consecutive with the same slope, why isn't this covered by `indexFirst` / `indexLast`?

yunhaolucky: Because the lines can go upward and downward. 
For example, we have a line with slope 1.
(1,1)(0,0),(2,2) are on the same line. But the min point is (0,0), not the first point.

bluong: Ok this makes sense now

src/plots/linePlot.ts
```diff
@@ -425,6 +449,70 @@ export module Plots {
...
     }
 
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
+      let data = dataset.data();
+      let scaledXAccessor = Plot._scaledAccessor(this.x());
+      let scaledYAccessor = Plot._scaledAccessor(this.y());
+
+      if (indices.length === 0) {
+        return [];
+      }
```
bluong: We can put this at the very start of the function

src/plots/linePlot.ts
```diff
@@ -425,6 +449,70 @@ export module Plots {
...
+      let filteredIndices: number[] = [];
+      filteredIndices.push(indices[0]);
```
bluong: `let filteredIndices = [indices[0]];`

src/plots/linePlot.ts
bluong: Still unsure about the name of the function.  There is no mention of a `Bucket` anywhere in the code

src/plots/linePlot.ts
```diff
@@ -425,6 +449,68 @@ export module Plots {
...
+        let p2y = scaledYAccessor(data[indices[i + 1]], indices[i + 1], dataset);
+        if (currentSlope === Infinity) {
+          return Math.floor(p1x) === Math.floor(p2x);
+        } else {
+          let expectedP2y = p1y + (p2x - p1x) * currentSlope;
+          return Math.floor(p2y) === Math.floor(expectedP2y);
+        }
+      };
+
+      for (let i = 0; i < indices.length - 1; ) {
```
bluong: Remove the whitespace at the end of the for

src/plots/linePlot.ts
```diff
@@ -425,6 +449,68 @@ export module Plots {
...
     }
 
+    private _filterDownsampling(dataset: Dataset, indices: number[]) {
+      if (indices.length === 0) {
+        return [];
+      }
+      let data = dataset.data();
+      let scaledXAccessor = Plot._scaledAccessor(this.x());
+      let scaledYAccessor = Plot._scaledAccessor(this.y());
+      let filteredIndices: number[] = [indices[0]];
```
bluong: Remove the explicit `number[]` typing

crmorford:
This case looks a lot more different than I'd like :/
http://jsfiddle.net/yxpq4uuv/19/

crmorford:
Area is freezing up when downsampling is enabled, I can barely pan zoom before the page locks up :(
enabled: http://jsfiddle.net/yxpq4uuv/22/
not enabled: http://jsfiddle.net/yxpq4uuv/23/

crmorford:
stacked area :( http://jsfiddle.net/yxpq4uuv/25/

crmorford:
Stacked Area has a ton of white spaces showing up where they don't on the original plot (in between the blue and red areas)
Why is this happening?
screen shot 2015-10-08 at 8 46 25 pm

crmorford:
Combined with other perf options: http://jsfiddle.net/yxpq4uuv/32/
Combined with other plots: http://jsfiddle.net/yxpq4uuv/33/
Area http://jsfiddle.net/yxpq4uuv/35/
Line http://jsfiddle.net/yxpq4uuv/34/
Stacked Area http://jsfiddle.net/yxpq4uuv/36/
Time Axis http://jsfiddle.net/yxpq4uuv/35/
Linear Axis http://jsfiddle.net/yxpq4uuv/36/
Log Axis http://jsfiddle.net/yxpq4uuv/37/
yAxis = category http://jsfiddle.net/yxpq4uuv/38/

yunhaolucky:
For now, we disable downsampling option for Stacked Area Plot.

crmorford:
Okay, that seems like a fair tradeoff since Contour wants this and it works fine on Line / Area

crmorford:
If this passes dev CR, looks good to me

src/plots/stackedAreaPlot.ts
```diff
@@ -67,6 +67,17 @@ export module Plots {
       return this;
     }
 
+    public downsamplingEnabled(downsampling?: boolean): any {
+      if (downsampling == null) {
+        return super.downsamplingEnabled();
+      }
+      if (downsampling === true) {
+        Utils.Window.warn("Warning: Stacked Area Plot does not support downsampling");
```
ztsai: remove everything else except this line
also add a test for it

plottable.d.ts
```diff
@@ -3530,6 +3542,7 @@ declare module Plottable {
             y(): Plots.AccessorScaleBinding<number, number>;
             y(y: number | Accessor<number>): StackedArea<X>;
             y(y: number | Accessor<number>, yScale: QuantitativeScale<number>): StackedArea<X>;
+            downsamplingEnabled(downsampling?: boolean): any;
```
bluong: Copy paste the declarations into `StackedArea` from `Line`

Issue #2840, opened 2015-10-05
[Testing Audit]


ztsai:
Dispatchers
- [x] `keyDispatcherTests.ts`
- [x] `mouseDispatcherTests.ts`
- [x] `touchDispatcherTests.ts`

make sure subclass of `Plottable.Dispatcher` test for the functionality of its protested and private methods:

- [ ] can set and unset callbacks
- [ ] connects dispatcher to the DOM when callback is set
- [ ] unsets one callback doesn't disconnect other callbacks

jtlan:
> connects dispatcher to the DOM when callback is set

I don't think there's a way to test directly for the existence of callbacks attached to DOM elements.

PR #2841, opened 2015-10-05
removed dispatcherTests as Dispatcher consists only of private and protected methods

ztsai:
part of #2840


bluong:
Remove the entry from `testReference.ts`

ztsai:
close it for now; will reopen and merge once I verified the other dispatchers are checking the right things

PR #2843, opened 2015-10-05
change the default comparator of legend

### LINKED ISSUE #2839: Comparator issues on legend
screen shot 2015-10-05 at 12 38 12 pm

https://jsfiddle.net/5pzsk1nv/

the comparator is scrambling the alphabet :O

yunhaolucky:
**Issue**: the order of entries in a Legend are different from its order in domains.
**Fix**: The previous comparator assumes all elements are equal, which cause problems when using different sorting methods. The new default comparator sorts the  formatted texts based on their corresponding indexes of domain names.

Before:https://jsfiddle.net/5pzsk1nv/
After:https://jsfiddle.net/yunhaolucky/5pzsk1nv/2/

This fix might not work if several entries have the same formatted text.

close #2839

### DIFF
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -48,7 +48,10 @@ export module Components {
       this._colorScale.onUpdate(this._redrawCallback);
       this._formatter = Formatters.identity();
       this.xAlignment("right").yAlignment("top");
-      this.comparator((a: string, b: string) => 0);
+      this.comparator(function(a: string, b: string) {
+        let formattedText = this._colorScale.domain().slice().map((d: string) => this._formatter(d));
+        return formattedText.indexOf(a) - formattedText.indexOf(b);
+      });
       this._symbolFactoryAccessor = () => SymbolFactories.circle();
       this._symbolOpacityAccessor = () => 1;
     }
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -48,7 +48,7 @@ export module Components {
       this._colorScale.onUpdate(this._redrawCallback);
       this._formatter = Formatters.identity();
       this.xAlignment("right").yAlignment("top");
-      this.comparator(function(a: string, b: string) {
+      this.comparator((a: string, b: string) => {
         let formattedText = this._colorScale.domain().slice().map((d: string) => this._formatter(d));
         return formattedText.indexOf(a) - formattedText.indexOf(b);
       });
--- a/test/components/legendTests.ts
+++ b/test/components/legendTests.ts
@@ -291,6 +291,15 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("do not change the order of legend entries when using default comparator", () => {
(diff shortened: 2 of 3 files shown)


src/components/legend.ts
```diff
@@ -48,7 +48,10 @@ export module Components {
       this._colorScale.onUpdate(this._redrawCallback);
       this._formatter = Formatters.identity();
       this.xAlignment("right").yAlignment("top");
-      this.comparator((a: string, b: string) => 0);
+      this.comparator(function(a: string, b: string) {
```
bluong: Use `() => `notation please

test/components/legendTests.ts
```diff
@@ -291,6 +291,15 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("do not change the order of legend entries when using default comparator", () => {
+    let colorDomain = ["a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k"];
+    color.domain(colorDomain);
+    legend.renderTo(svg);
+    let entryTexts = svg.selectAll(entrySelector)[0].map((node: Element) => d3.select(node).select("text").text());
```
bluong: I would prefer to use `legend.content()` instead of `svg` as that's more precise

bluong: You can also use `.data()` and then check if the data is in order

yunhaolucky: If I use data(), I will get the data that is assigned to that entry, which are not the entity text.

test/components/legendTests.ts
```diff
@@ -291,6 +291,15 @@ describe("Legend", () => {
     svg.remove();
   });
 
+  it("do not change the order of legend entries when using default comparator", () => {
```
bluong: it("does not change...

crmorford:
add comparator that reverses entries https://jsfiddle.net/5pzsk1nv/4/
add formatter https://jsfiddle.net/5pzsk1nv/5/
comparator + formatter...why does N always want to come first?? https://jsfiddle.net/5pzsk1nv/6/
** we need to figure out what's going wrong in /6/**
i printed out some of the comparator stuff and it all looks good as far as i can tell :/ https://jsfiddle.net/5pzsk1nv/7/

yunhaolucky:
Maybe we should use `return a.charCodeAt() - b.charCodeAt();` instead of   `return a.charCodeAt() > b.charCodeAt();` ?
It seems work in comparator + formatter. 
https://jsfiddle.net/yunhaolucky/5pzsk1nv/9/

crmorford:
ooops don't know why i went to ">" by default. that looks fine.

but why does using ">" break it?? i might take a closer look at comparators soon.

yunhaolucky:
It seems comparator should return numbers. (positive for a > b, negative for a < b). And it seems true > 0 and false == 0.  That's why `return a.charCodeAt() > b.charCodeAt();` will not return the correct result.

PR #2850, opened 2015-10-06
tweak test for Windows

softwords:
This test is the only test in the current test suite that fails when executed under grunt in Visual Studio 2015. Resolved by changing "closeTo" value from 1 to 1.5

### DIFF
--- a/test/axes/numericAxisTests.ts
+++ b/test/axes/numericAxisTests.ts
@@ -97,7 +97,7 @@ describe("NumericAxis", () => {
       let markCenter = (markBB.top + markBB.bottom) / 2;
       labelBB = tickLabels[0][i].getBoundingClientRect();
       let labelCenter = (labelBB.top + labelBB.bottom) / 2;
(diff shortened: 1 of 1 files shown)


test/axes/numericAxisTests.ts
```diff
@@ -97,7 +97,7 @@ describe("NumericAxis", () => {
       let markCenter = (markBB.top + markBB.bottom) / 2;
       labelBB = tickLabels[0][i].getBoundingClientRect();
       let labelCenter = (labelBB.top + labelBB.bottom) / 2;
-      assert.closeTo(labelCenter, markCenter, 1, "tick label is centered on mark");
+      assert.closeTo(labelCenter, markCenter, 1.5, "tick label is centered on mark");
```
bluong: We have a `window.Pixel_closeToRequirement` or so variable.  Can we use that one instead?

bluong: Actually we are in process of refactoring our tests, and refactoring this class is assigned to me.  I got this part.

bluong:
Thanks for the help!

PR #2851, opened 2015-10-06
[Gruntfile] - Massively simplify compilation

bluong:
Currently the way that "plottable.js" and "plottable.d.ts" is built is through quite a bit of processing on our side regarding concatenating individual javascript files, renaming, replacing, and also some updating.  Essentially, our build pattern is quite a bit complex.

In looking at Typescript's [compiler options](https://github.com/TypeStrong/grunt-ts), a bunch of this processing was found to be pretty needless (Why do we need to concatenate our javascript files when there is an explicit flag to compile to a single javascript file?, for instance).

Additionally, with the release of Typescript 1.5, there is now an idea of a [tsconfig.json](https://github.com/Microsoft/TypeScript/wiki/tsconfig.json).  One decent way to think about this file is to think as a "reference.ts" file that a developer does not have to reference in their source files.  It is the compiler's way of knowing which files it needs to include, instead of needing to do a 2-way referencing using the "reference.ts" idea.  As a result, the reference comments are taken out, and we only need to add the names to tsconfig.json.

Developer note: The `grunt-ts` package has been updated to take advantage of the `tsconfig` option.  Please update with latest develop + run `npm install` after this merge.

QE: Based on the "plottable.js" file, essentially nothing has changed (whitespace removal and removal of redundant code).  If you want to do a regression pass, feel free.  Otherwise I think this is safe.


plottable.js
```diff
@@ -12698,6 +12347,90 @@ var Plottable;
...
+/// <reference path="interactions/doubleClickInteraction.ts" />
+/// <reference path="interactions/keyInteraction.ts" />
+/// <reference path="interactions/pointerInteraction.ts" />
+/// <reference path="interactions/panZoomInteraction.ts" />
+/// <reference path="interactions/dragInteraction.ts" />
+/// <reference path="components/dragBoxLayer.ts" />
+/// <reference path="components/xDragBoxLayer.ts" />
+/// <reference path="components/yDragBoxLayer.ts" />
+/// <reference path="components/dragLineLayer.ts" />
+///<reference path="../reference.ts" />
```
aicioara: These look a bit odd. Seems like the compiler no longer uses the reference paths?

Gruntfile.js
```diff
@@ -5,26 +5,24 @@ module.exports = function(grunt) {
...
-      src: ["src/**/*.ts", "typings/**/*.d.ts"],
-      outDir: "build/src/",
+      src: ["src/**/*.ts"],
+      out: "plottable.js",
       options: {
         target: "es5",
         noImplicitAny: true,
         sourceMap: false,
         declaration: true,
-        compiler: "./node_modules/typescript/bin/tsc",
```
aicioara: By doing this, you downgrade to ts `v1.5.3`

plottable.js
```diff
@@ -3542,13 +3431,7 @@ var Plottable;
     })(Plottable.Component);
     Plottable.ComponentContainer = ComponentContainer;
 })(Plottable || (Plottable = {}));
-
 ///<reference path="../reference.ts" />
-var __extends = (this && this.__extends) || function (d, b) {
-    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
-    function __() { this.constructor = d; }
-    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
-};
```
aicioara: Proof here

Gruntfile.js
```diff
@@ -5,26 +5,24 @@ module.exports = function(grunt) {
...
         removeComments: false
       }
     },
     test: {
-      src: ["test/**/*.ts", "typings/**/*.d.ts", "build/plottable.d.ts"],
```
aicioara: How do the tests now know where to get the Plottable typings from?

Gruntfile.js
```diff
@@ -325,14 +267,11 @@ module.exports = function(grunt) {
   ]);
 
   grunt.registerTask("generateJS", [
-    "concat:plottable",
     "concat:svgtypewriter",
-    "concat:definitions",
     "sed:definitions",
```
aicioara: This will now get out of sync, as you no longer have the `build/` folder. Try deleting the `build/` folder and check what blows up.

plottable.js
bluong: I'm unsure why this got printed here.  Was trying to investigate it to no avail.  Note that this ultimately doesn't affect anything, but it is sad to see in the build.

Gruntfile.js
bluong: As long as our package.json has a specified typescript package, it's fine.  https://github.com/TypeStrong/grunt-ts/issues/280 Also, `grunt-ts` is meaning to upgrade likely today, so maybe we don't even have to worry about this.

plottable.js
bluong: Proof of what?  This is a strict removal, without adding the typescript 1.5.3 items.

Gruntfile.js
bluong: `testReference.ts` references the typings.

Gruntfile.js
bluong: Nothing should blow up since there is no longer a reference to the `build/` directory.

aicioara: My bad. You're right.

Gruntfile.js
aicioara: Okay

plottable.js
aicioara: I believe this line should not disappear when TS1.6 is used for compilation

Gruntfile.js
aicioara: If we delete this line, we don't even need to specify `typescript` as a dependency in our `package.json`, as it will always use `grunt-ts` dependency all the time

plottable.js
bluong: The console output explicitly says that it uses 1.6 in its compilation

softwords:
In my Windows environment, this simplification produces a plottable.d.ts that is compatible with earlier ones - ie it eliminates the massive diff previously produced by git deciding that every line had been changed.

plottable.js
softwords: Could it be the reason why these lines are disappearing is that, previously, by compiling file-by-file, the definition of ```__extends``` was part of every file. Hence it appeared multiple times when plottable.js was concatenated. Now it is included by the compiler only once per closure.

Gruntfile.js
aicioara: You are right, my bad again. Deleting this line is good.

bluong:
I made an additional update to use the new `tsconfig.json` feature introduced in TS 1.5.  It should hopefully ease development a bit.

bluong:
The next step I would personally prefer to do is to be able to remove the reference tags from the tests, but to my knowledge, that might be a bit harder.

crmorford:
When I change the gruntfile like this
  var connectConfig = {
    server: {
      options: {
        port: 9999,
        protocol: "https",
        hostname: "*",
        base: "",
        livereload: true
      }
    }
  };

and run > grunt

[these local tests](https://jsfiddle.net/9djcfvh7/121/) no longer work, i get console errors about plottable not found :( 

They did work with the last grunt file

PR #2852, opened 2015-10-06
removed computedWidth and computedHeight

### LINKED ISSUE #2589: Timeseires y-axis label formatting for large numbers
y-axis shows labels like "7..." rather than 7,000 or 70,000,000 - so it is hard to interpret. There is a workaround for this by wrapping in a $timeout to delay. Figure out why the timeout fixes this, and if there's actually an issue with the implementation. 

Might be able to be fixed by: https://github.com/palantir/plottable/issues/1353

ztsai:
closes #2589 
before: http://jsfiddle.net/ztsai/zg13fjt2/5/
after: http://jsfiddle.net/ztsai/zg13fjt2/6/

basically calling `table.redraw` doesn't invalid cached computedWidth and computedHeight, but calling `axis.redraw` would

it seems like there's really no reason for us to cached the computed values

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1840,6 +1840,8 @@ declare module Plottable {
         protected _scale: Scale<D, number>;
         private _formatter;
         private _orientation;
+        protected _computedWidth: number;
+        protected _computedHeight: number;
         private _endTickLength;
         private _innerTickLength;
         private _tickLabelPadding;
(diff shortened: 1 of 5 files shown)


plottable.d.ts
```diff
@@ -1840,8 +1840,6 @@ declare module Plottable {
         protected _scale: Scale<D, number>;
         private _formatter;
         private _orientation;
-        protected _computedWidth: number;
-        protected _computedHeight: number;
```
bluong: This is an API break.  Instead of removing this, can we put a comment above the function stating that these variables are deprecated?

bluong:
Just the API break issue.

bluong:
@crmorford Mostly a regression pass on Axes.  Specifically, check the performance with Axes as we are removing supposedly a very minor increase in performance.  If the performance is still comparable with develop's version, then this is fine.

crmorford:
all my regression tests look good!

Issue #2632, opened 2015-08-17
[Testing Audit] Area and Line


crmorford:

aicioara:
Block on this please. https://github.com/palantir/plottable/pull/2520 It should get in soon.

PR #2860, opened 2015-10-08
areaPlotTests refactor

ztsai:
part of #2632

### DIFF
--- a/test/plots/areaPlotTests.ts
+++ b/test/plots/areaPlotTests.ts
@@ -3,15 +3,13 @@
 describe("Plots", () => {
   describe("AreaPlot", () => {
     describe("Basic Usage", () => {
-      let SVG_WIDTH = 400;
-      let SVG_HEIGHT = 500;
       let svg: d3.Selection<void>;
       let xScale: Plottable.Scales.Linear;
       let yScale: Plottable.Scales.Linear;
       let plot: Plottable.Plots.Area<{}>;
 
       beforeEach(() => {
-        svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+        svg = TestMethods.generateSVG();
         xScale = new Plottable.Scales.Linear();
         yScale = new Plottable.Scales.Linear();
         plot = new Plottable.Plots.Area();
@@ -19,10 +17,8 @@ describe("Plots", () => {
         plot.y((d) => d.y, yScale);
       });
 
-      it("renders correctly with no data", () => {
-        assert.doesNotThrow(() => plot.renderTo(svg), Error);
-        assert.strictEqual(plot.width(), SVG_WIDTH, "was allocated width");
-        assert.strictEqual(plot.height(), SVG_HEIGHT, "was allocated height");
+      it("does not throw error when rendering without data", () => {
+        assert.doesNotThrow(() => plot.renderTo(svg), Error, "rendering without data does not throw an error");
         svg.remove();
       });
 
@@ -64,48 +60,54 @@ describe("Plots", () => {
       it("draws area and line correctly", () => {
         let content = areaPlot.content();
         let areaPath = content.select(".area");
-        let data = areaPlot.datasets()[0].data();
-        let pointsInArea = getPointsInArea(areaPlot, data);
-        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
-        assert.strictEqual(areaPath.attr("fill"), fill, "area fill was set correctly");
-        assert.strictEqual(areaPath.style("stroke"), "none", "area stroke renders as \"none\"");
+        areaPlot.datasets().forEach((dataset: Plottable.Dataset) => {
+          let data = dataset.data();
+          let pointsInArea = getPointsInArea(areaPlot, data);
+          TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+          assert.strictEqual(areaPath.attr("fill"), fill, "area fill was set correctly");
+          assert.strictEqual(areaPath.style("stroke"), "none", "area stroke renders as \"none\"");
 
-        let linePath = content.select(".line");
-        TestMethods.assertLinePathEqualToDataPoints(linePath.attr("d"), data, xScale, yScale);
-        assert.strictEqual(linePath.attr("stroke"), "#000000", "line stroke was set correctly");
-        assert.strictEqual(linePath.style("fill"), "none", "line fill renders as \"none\"");
+          let linePath = content.select(".line");
+          TestMethods.assertLinePathEqualToDataPoints(linePath.attr("d"), data, xScale, yScale);
+          assert.strictEqual(linePath.attr("stroke"), "#000000", "line stroke was set correctly");
+          assert.strictEqual(linePath.style("fill"), "none", "line fill renders as \"none\"");
+        });
         svg.remove();
       });
 
       it("fills for non-zero floor values appropriately", () => {
         areaPlot.y0((d) => d.y / 2);
         areaPlot.renderTo(svg);
-        let data = areaPlot.datasets()[0].data();
-        let areaPath = areaPlot.content().select(".area");
-        let pointsInArea = getPointsInArea(areaPlot, data);
-        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+        areaPlot.datasets().forEach((dataset: Plottable.Dataset) => {
+          let data = dataset.data();
+          let areaPath = areaPlot.content().select(".area");
+          let pointsInArea = getPointsInArea(areaPlot, data);
+          TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+        });
         svg.remove();
       });
 
       it("places the area before line", () => {
         let content = areaPlot.content();
         let paths = content.selectAll("path")[0];
-        let areaSelection = content.select(".area")[0][0];
-        let lineSelection = content.select(".line")[0][0];
-        assert.operator(paths.indexOf(areaSelection), "<", paths.indexOf(lineSelection), "area appended before line");
+        let areaElement = content.select(".area")[0][0];
+        let lineElement = content.select(".line")[0][0];
+        assert.operator(paths.indexOf(areaElement), "<", paths.indexOf(lineElement), "area appended before line");
         svg.remove();
       });
 
       it("cleans up correctly when removing Datasets", () => {
         areaPlot.renderTo(svg);
+        let paths = areaPlot.content().selectAll("path");
+        let pathSize = paths.size();
         let dataset = areaPlot.datasets()[0];
         areaPlot.removeDataset(dataset);
-        let paths = areaPlot.content().selectAll("path");
-        assert.strictEqual(paths.size(), 0, "removing a Dataset cleans up all <path>s associated with it");
+        paths = areaPlot.content().selectAll("path");
+        assert.strictEqual(paths.size(), pathSize - 2, "removing a Dataset cleans up both <path>s associated with it");
         svg.remove();
       });
 
-      it("correctly handles NaN and undefined x and y values", () => {
+      it("skips data points consisting of NaN and undefined x and y values", () => {
         let areaData = [
           { x: 0.0, y: 0.0 },
           { x: 0.2, y: 0.2 },
@@ -143,11 +145,11 @@ describe("Plots", () => {
       });
 
       it("retains original classes when setting css class via attr", () => {
-        areaPlot.attr("class", () => "pink");
+        areaPlot.attr("class", "pink");
         areaPlot.renderTo(svg);
         let areaPath = areaPlot.content().select(".area");
-        assert.isTrue(areaPath.classed("pink"));
-        assert.isTrue(areaPath.classed("area"));
+        assert.isTrue(areaPath.classed("pink"), "it has css class set via attr");
+        assert.isTrue(areaPath.classed("area"), "it has default css class");
         svg.remove();
       });
     });
@@ -157,7 +159,6 @@ describe("Plots", () => {
       let xScale: Plottable.Scales.Linear;
       let yScale: Plottable.Scales.Linear;
       let areaPlot: Plottable.Plots.Area<number>;
-      let renderArea: d3.Selection<void>;
 
       beforeEach(() => {
         xScale = new Plottable.Scales.Linear();
@@ -172,7 +173,6 @@ describe("Plots", () => {
                 .attr("fill", "steelblue")
                 .attr("stroke", (d: any, i: number, m: any) => d3.rgb(d.foo, d.bar, i).toString())
                 .renderTo(svg);
-        renderArea = areaPlot.content();
       });
 
       it("retrieves all selections with no args", () => {
@@ -199,12 +199,13 @@ describe("Plots", () => {
 
     function getPointsInArea(plot: Plottable.Plots.Area<any>, points: Plottable.Point[]) {
       let y0Binding = plot.y0();
-      let firstPoint = points[0];
-      let lastPoint = points[points.length - 1];
       let ds = plot.datasets()[0];
-      let lasty0 = y0Binding.accessor(lastPoint, points.length - 1, ds);
-      let firsty0 = y0Binding.accessor(firstPoint, 0, ds);
-      return points.concat([{ x: lastPoint.x, y: lasty0 }, { x: firstPoint.x, y: firsty0 }]);
+      let bottomPoints = points.map((d, i) => {
+        let x = d.x;
+        let y0 = y0Binding.accessor(d, i, ds);
+        return { x: x, y: y0 };
+      }).reverse();
+      return points.concat(bottomPoints);
     };
   });
 });
--- a/test/plots/areaPlotTests.ts
+++ b/test/plots/areaPlotTests.ts
@@ -57,32 +57,32 @@ describe("Plots", () => {
                 .renderTo(svg);
       });
 
-      it("draws area and line correctly", () => {
+      it("draws area and line with correct data points and correct fill and stroke attributes", () => {
         let content = areaPlot.content();
         let areaPath = content.select(".area");
         areaPlot.datasets().forEach((dataset: Plottable.Dataset) => {
           let data = dataset.data();
           let pointsInArea = getPointsInArea(areaPlot, data);
-          TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+          TestMethods.assertPathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
           assert.strictEqual(areaPath.attr("fill"), fill, "area fill was set correctly");
           assert.strictEqual(areaPath.style("stroke"), "none", "area stroke renders as \"none\"");
 
           let linePath = content.select(".line");
-          TestMethods.assertLinePathEqualToDataPoints(linePath.attr("d"), data, xScale, yScale);
+          TestMethods.assertPathEqualToDataPoints(linePath.attr("d"), data, xScale, yScale);
           assert.strictEqual(linePath.attr("stroke"), "#000000", "line stroke was set correctly");
           assert.strictEqual(linePath.style("fill"), "none", "line fill renders as \"none\"");
         });
         svg.remove();
       });
 
-      it("fills for non-zero floor values appropriately", () => {
+      it("draws area appropriately with non-constant y0 values", () => {
         areaPlot.y0((d) => d.y / 2);
         areaPlot.renderTo(svg);
         areaPlot.datasets().forEach((dataset: Plottable.Dataset) => {
           let data = dataset.data();
           let areaPath = areaPlot.content().select(".area");
           let pointsInArea = getPointsInArea(areaPlot, data);
-          TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+          TestMethods.assertPathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
         });
         svg.remove();
       });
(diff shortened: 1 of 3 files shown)


test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
       plot.y((d) => d.y, yScale);
+      yScale.padProportion(0.1);
+      let constantY0 = 30;
+      yScale.addIncludedValuesProvider((scale: Plottable.Scales.Linear) => [constantY0, constantY0 + 10]);
       plot.y0(constantY0);
       plot.addDataset(new Plottable.Dataset([{ x: 0, y: constantY0 + 5 }]));
       plot.renderTo(svg);
       assert.strictEqual(yScale.domain()[0], constantY0, "y Scale doesn't pad beyond 0 when used in a Plots.Area");
       svg.remove();
     });
```
bluong: Put the above in an inner `describe`?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
       let yScale = new Plottable.Scales.Linear();
       let plot = new Plottable.Plots.Area();
-      plot.x((d: any) => d.x, xScale);
+      plot.x((d) => d.x, xScale);
       plot.y((d) => d.y, yScale);
       assert.doesNotThrow(() => plot.renderTo(svg), Error);
-      assert.strictEqual(plot.width(), 400, "was allocated width");
-      assert.strictEqual(plot.height(), 400, "was allocated height");
+      assert.strictEqual(plot.width(), SVG_WIDTH, "was allocated width");
+      assert.strictEqual(plot.height(), SVG_HEIGHT, "was allocated height");
```
bluong: Unsure if the above two tests really are useful.  I would remove them; your call.

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
-              .attr("fill", fillAccessor)
-              .attr("stroke", colorAccessor)
-              .renderTo(svg);
-      renderArea = (<any> areaPlot)._renderArea;
-    });
+      it("draws area and line correctly", () => {
+        let areaPath = renderArea.select(".area");
+        let pointsInArea = getPointsInArea(areaPlot, twoPointData);
+        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+        assert.strictEqual(areaPath.attr("fill"), "steelblue", "area fill was set correctly");
```
bluong: Let's factor our the "steelblue" part

Also, why do we want to test the "fill" attribute specifically?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
+        simpleDataset = new Plottable.Dataset(twoPointData);
+        areaPlot = new Plottable.Plots.Area<number>();
+        areaPlot.addDataset(simpleDataset);
+        areaPlot.x((d: any) => d.x, xScale)
+                .y((d: any) => d.y, yScale);
+        areaPlot.y0(() => 0)
+                .attr("fill", () => "steelblue")
+                .attr("stroke", (d, i) => d3.rgb(d.x, d.y, i).toString())
+                .renderTo(svg);
+        renderArea = areaPlot.content();
```
bluong: Rename `renderArea` to `content`?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
-      svg.remove();
-    });
+      it("fills for non-zero floor values appropriately", () => {
+        areaPlot.y0((d) => d.y / 2);
+        areaPlot.renderTo(svg);
+        let areaPath = renderArea.select(".area");
+        let pointsInArea = getPointsInArea(areaPlot, twoPointData);
+        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
+        svg.remove();
+      });
```
bluong: What this is this test testing?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
 
-    it("area fill works for non-zero floor values appropriately, e.g. half the height of the line", () => {
-      areaPlot.y0((d) => d.bar / 2);
-      areaPlot.renderTo(svg);
-      renderArea = (<any> areaPlot)._renderArea;
-      let areaPath = renderArea.select(".area");
-      assert.strictEqual(TestMethods.normalizePath(areaPath.attr("d")), "M0,500L500,0L500,250L0,500Z");
-      svg.remove();
-    });
+      it("appends before line", () => {
```
bluong: it("appends the area before line", () => {

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
 
-      areaPathString = TestMethods.normalizePath(areaPath.attr("d"));
-      TestMethods.assertAreaPathCloseTo(areaPathString, expectedPath, 0.1, "area d was set correctly (x=undefined case)");
+        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
 
-      svg.remove();
+        svg.remove();
+      });
+
+      it("retains original classes when class is projected", () => {
```
bluong: the term "projected" does not apply anymore.  Let's change the test name?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
-      yAccessor = (d: any) => d.bar;
-      y0Accessor = () => 0;
-      colorAccessor = (d: any, i: number, m: any) => d3.rgb(d.foo, d.bar, i).toString();
-      fillAccessor = () => "steelblue";
-    });
+    describe("Rendering", () => {
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let twoPointData = [{x: 0, y: 0}, {x: 1, y: 1}];
```
bluong: Don't think this variable needs to exist

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
+      let simpleDataset: Plottable.Dataset;
+      let areaPlot: Plottable.Plots.Area<number>;
+      let renderArea: d3.Selection<void>;
```
bluong: Don't think this variable needs to exist

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
-      y0Accessor = () => 0;
-      colorAccessor = (d: any, i: number, m: any) => d3.rgb(d.foo, d.bar, i).toString();
-      fillAccessor = () => "steelblue";
-    });
+    describe("Rendering", () => {
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let twoPointData = [{x: 0, y: 0}, {x: 1, y: 1}];
+      let simpleDataset: Plottable.Dataset;
```
bluong: Don't think this variable needs to exist

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,177 @@
...
 
-    describe("selections()", () => {
+    describe("Selections", () => {
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let twoPointData = [{x: 0, y: 0}, {x: 1, y: 1}];
+      let simpleDataset: Plottable.Dataset;
+      let areaPlot: Plottable.Plots.Area<number>;
+      let renderArea: d3.Selection<void>;
```
bluong: I believe `twoPointData` and `simpleDataset` and `renderArea` are unneeded variables here

test/plots/areaPlotTests.ts
```diff
@@ -176,28 +191,16 @@ describe("Plots", () => {
...
-    });
+    function getPointsInArea(plot: Plottable.Plots.Area<any>, points: Plottable.Point[]) {
+      let y0Binding = plot.y0();
+      let firstPoint = points[0];
+      let lastPoint = points[points.length - 1];
+      let ds = plot.datasets()[0];
+      let lasty0 = y0Binding.accessor(lastPoint, points.length - 1, ds);
+      let firsty0 = y0Binding.accessor(firstPoint, 0, ds);
+      return points.concat([{ x: lastPoint.x, y: lasty0 }, { x: firstPoint.x, y: firsty0 }]);
+    };
```
bluong: is this to include the y0 points?  If so, don't we need to include the y0 for each point?

test/plots/areaPlotTests.ts
ztsai: non static y0

test/plots/areaPlotTests.ts
ztsai: y0 could be different each data point

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      plot.x((d) => d.x, xScale);
-      plot.y((d) => d.y, yScale);
-      plot.y0(constantY0);
-      plot.addDataset(new Plottable.Dataset([{ x: 0, y: constantY0 + 5 }]));
-      plot.renderTo(svg);
-      assert.strictEqual(yScale.domain()[0], constantY0, "y Scale doesn't pad beyond 0 when used in a Plots.Area");
-      svg.remove();
-    });
-  });
+      it("renders correctly with no data", () => {
```
bluong: Test name seems too general for me.  Instead can we say that it does not error upon rendering if no data is inputted?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+        assert.doesNotThrow(() => plot.renderTo(svg), Error);
+        assert.strictEqual(plot.width(), SVG_WIDTH, "was allocated width");
+        assert.strictEqual(plot.height(), SVG_HEIGHT, "was allocated height");
```
bluong: Do we need the latter 2 statements?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      plot.x((d: any) => d.x, xScale);
-      plot.y((d) => d.y, yScale);
-      assert.doesNotThrow(() => plot.renderTo(svg), Error);
-      assert.strictEqual(plot.width(), 400, "was allocated width");
-      assert.strictEqual(plot.height(), 400, "was allocated height");
-      svg.remove();
-    });
+    describe("Basic Usage", () => {
+      let SVG_WIDTH = 400;
+      let SVG_HEIGHT = 500;
```
bluong: Do we need these 2 variables?  I personally don't really see the need to assert the width/height of the plot.  Do you think it's necessary?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let plot: Plottable.Plots.Area<{}>;
```
bluong: Rename to `areaPlot` maybe?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      assert.strictEqual(TestMethods.normalizePath(linePath.attr("d")), "M0,500L500,0", "line d was set correctly");
-      assert.strictEqual(linePath.attr("stroke"), "#000000", "line stroke was set correctly");
-      let lineComputedStyle = window.getComputedStyle(<Element> linePath.node());
-      assert.strictEqual(lineComputedStyle.fill, "none", "line fill renders as \"none\"");
-      svg.remove();
-    });
+      it("draws area and line correctly", () => {
+        let content = areaPlot.content();
+        let areaPath = content.select(".area");
+        let data = areaPlot.datasets()[0].data();
```
bluong: Instead of grabbing the first one, do you think it's worth it to just loop?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      let areaSelection = renderArea.select(".area")[0][0];
-      let lineSelection = renderArea.select(".line")[0][0];
-      assert.operator(paths.indexOf(areaSelection), "<", paths.indexOf(lineSelection), "area appended before line");
-      svg.remove();
-    });
+      it("places the area before line", () => {
+        let content = areaPlot.content();
+        let paths = content.selectAll("path")[0];
+        let areaSelection = content.select(".area")[0][0];
+        let lineSelection = content.select(".line")[0][0];
```
bluong: These aren't `Selection`s, they are the actual DOM elements.  We should rename the variable names to reflect this point.

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      let paths = areaPlot.content().selectAll("path");
-      assert.strictEqual(paths.size(), 0, "removing a Dataset cleans up all <path>s associated with it");
-      svg.remove();
-    });
+      it("cleans up correctly when removing Datasets", () => {
+        areaPlot.renderTo(svg);
+        let dataset = areaPlot.datasets()[0];
+        areaPlot.removeDataset(dataset);
+        let paths = areaPlot.content().selectAll("path");
+        assert.strictEqual(paths.size(), 0, "removing a Dataset cleans up all <path>s associated with it");
```
bluong: Why 0?  Why not previousSize - 1?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+        svg.remove();
+      });
+
+      it("correctly handles NaN and undefined x and y values", () => {
```
bluong: Can we slightly elaborate what correctly means?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
 
-      areaPathString = TestMethods.normalizePath(areaPath.attr("d"));
-      TestMethods.assertAreaPathCloseTo(areaPathString, expectedPath, 0.1, "area d was set correctly (x=undefined case)");
+        svg.remove();
+      });
 
-      svg.remove();
+      it("retains original classes when setting css class via attr", () => {
+        areaPlot.attr("class", () => "pink");
```
bluong: Can we simplify this?
`areaPlot.attr("class", "pink")`

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+        areaPlot.renderTo(svg);
+        let areaPath = areaPlot.content().select(".area");
+        assert.isTrue(areaPath.classed("pink"));
+        assert.isTrue(areaPath.classed("area"));
```
bluong: Let's give messages here

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+      });
     });
 
-    describe("selections()", () => {
+    describe("Selections", () => {
+      let svg: d3.Selection<void>;
+      let xScale: Plottable.Scales.Linear;
+      let yScale: Plottable.Scales.Linear;
+      let areaPlot: Plottable.Plots.Area<number>;
+      let renderArea: d3.Selection<void>;
```
bluong: Remove `renderArea` variable?

test/plots/areaPlotTests.ts
bluong: > y0 could be different at each data point

Isn't this more reason why this function is incomplete?  If we set a variable `y0` function, does this function still get all the points?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      assert.strictEqual(areaComputedStyle.stroke, "none", "area stroke renders as \"none\"");
-
-      let linePath = renderArea.select(".line");
-      assert.strictEqual(TestMethods.normalizePath(linePath.attr("d")), "M0,500L500,0", "line d was set correctly");
-      assert.strictEqual(linePath.attr("stroke"), "#000000", "line stroke was set correctly");
-      let lineComputedStyle = window.getComputedStyle(<Element> linePath.node());
-      assert.strictEqual(lineComputedStyle.fill, "none", "line fill renders as \"none\"");
-      svg.remove();
-    });
+      it("draws area and line correctly", () => {
```
bluong: can we be more exact with what draws correctly means?

it seems it involves that the line path is the data points
it also seems that the area has the specified fill
it also seems that the line is not filled
it also seems that the area has no stroke
it also seems that the line has the specified stroke

I personally think this may be different tests since the main problem is that "draws correctly" is incredibly vague

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
 
-    it("area fill works for non-zero floor values appropriately, e.g. half the height of the line", () => {
-      areaPlot.y0((d) => d.bar / 2);
-      areaPlot.renderTo(svg);
-      renderArea = (<any> areaPlot)._renderArea;
-      let areaPath = renderArea.select(".area");
-      assert.strictEqual(TestMethods.normalizePath(areaPath.attr("d")), "M0,500L500,0L500,250L0,500Z");
-      svg.remove();
-    });
+      it("fills for non-zero floor values appropriately", () => {
```
bluong: What does this mean by "fills"?

ztsai: area under the line?

bluong: Let's use a better test name here.  It seems to be something along the lines that we are checking the `y0` points?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+        areaPlot.y0((d) => d.y / 2);
+        areaPlot.renderTo(svg);
+        areaPlot.datasets().forEach((dataset: Plottable.Dataset) => {
+          let data = dataset.data();
+          let areaPath = areaPlot.content().select(".area");
+          let pointsInArea = getPointsInArea(areaPlot, data);
+          TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
```
bluong: Is it correct to assert a `LinePath` is equal when inputting an `areaPath`?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
+      });
 
-    it("cleans up correctly when removing Datasets", () => {
-      areaPlot.renderTo(svg);
-      areaPlot.removeDataset(simpleDataset);
-      let paths = areaPlot.content().selectAll("path");
-      assert.strictEqual(paths.size(), 0, "removing a Dataset cleans up all <path>s associated with it");
-      svg.remove();
-    });
+      it("cleans up correctly when removing Datasets", () => {
```
bluong: maybe a better description is
`it("removes the plot svg elements when removing datasets"`?

bluong: Note that I am growing to be more against using the word "correctly" since it is a really vague term.  Let's define what "correctly" means and then use that as the test name

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,181 @@
...
-      simpleDataset.data(dataWithUndefined);
+        TestMethods.assertLinePathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
 
-      areaPathString = TestMethods.normalizePath(areaPath.attr("d"));
-      TestMethods.assertAreaPathCloseTo(areaPathString, expectedPath, 0.1, "area d was set correctly (x=undefined case)");
+        svg.remove();
+      });
 
-      svg.remove();
+      it("retains original classes when setting css class via attr", () => {
```
bluong: "retains the area class..."?

test/plots/areaPlotTests.ts
```diff
@@ -176,28 +195,17 @@ describe("Plots", () => {
...
+    function getPointsInArea(plot: Plottable.Plots.Area<any>, points: Plottable.Point[]) {
+      let y0Binding = plot.y0();
+      let ds = plot.datasets()[0];
+      let bottomPoints = points.map((d, i) => {
+        let x = d.x;
+        let y0 = y0Binding.accessor(d, i, ds);
+        return { x: x, y: y0 };
+      }).reverse();
+      return points.concat(bottomPoints);
+    };
```
bluong: Why y0 instead of y?

ztsai: `(x, y0)` is used for the lower bound, `(x, y)` is included in `points`

bluong: Would it make more sense to rename such that you're not getting the points in the area but maybe the y0 points?  Unsure of the function name, but it should at least be named such that I should expect to receive y0 points

ztsai: well this function gets all points contained in the area, upper bounds and lower bounds
splitting into 2 functions or concatenating in the calling functions seems redundant and will make code harder to read ( in the invalid data tests for example)

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,182 @@
...
-      areaPlot.y0((d) => d.bar / 2);
-      areaPlot.renderTo(svg);
-      renderArea = (<any> areaPlot)._renderArea;
-      let areaPath = renderArea.select(".area");
-      assert.strictEqual(TestMethods.normalizePath(areaPath.attr("d")), "M0,500L500,0L500,250L0,500Z");
-      svg.remove();
-    });
+      it("draws area appropriately with non-constant y0 values", () => {
+        areaPlot.y0((d) => d.y / 2);
+        areaPlot.renderTo(svg);
```
bluong: This `renderTo` call seems unnecessary?

test/plots/areaPlotTests.ts
```diff
@@ -2,162 +2,182 @@
...
-      simpleDataset.data(dataWithUndefined);
+        TestMethods.assertPathEqualToDataPoints(areaPath.attr("d"), pointsInArea, xScale, yScale);
 
-      areaPathString = TestMethods.normalizePath(areaPath.attr("d"));
-      TestMethods.assertAreaPathCloseTo(areaPathString, expectedPath, 0.1, "area d was set correctly (x=undefined case)");
+        svg.remove();
+      });
 
-      svg.remove();
+      it("retains the area class. when setting css class via attr", () => {
```
bluong: Remove the period

bluong:
Few pieces

PR #2862, opened 2015-10-08
[StackedAreaPlotTests] - Revamp

bluong:
Trying to scope down on each of the uses of the derived functions on `StackedAreaPlot`

Test screenshot:
screen shot 2015-10-26 at 2 38 04 pm


ztsai:
can you resolve the merge conflicts?
also, is there any reason why you are not using `assertLinePathEqualToDataPoints` and write a new method instead?

bluong:
`assertLinePathEqualToDataPoints` is not enough, unless I create a data object that contains the stack offsets instead of the data itself.  Would the be preferred?

ztsai:
I see
It seems to me there's some duplicated functionality in `assertLinePathEqualToDataPoints` and `areaVertices`
does it make sense for me to remove `assertLinePathEqualToDataPoints` and use `areaVertices` in the tests for LinePlot and AreaPlot?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
-      let xAxis = new Plottable.Axes.Numeric(xScale, "bottom");
-      new Plottable.Components.Table([[renderer], [xAxis]]).renderTo(svg);
+    describe("setting the x property", () => {
+      let svg: d3.Selection<void>;
+      let stackedAreaPlot: Plottable.Plots.StackedArea<number>;
+
+      beforeEach(() => {
+        svg = TestMethods.generateSVG();
+        let data1 = [
+          {foo: 1},
```
ztsai: probably just personal preference, but `foo`, `bar` are not good variable names

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+        let stackedAreas = stackedAreaPlot.content().selectAll("path");
+        assert.strictEqual(stackedAreas.size(),
+          stackedAreaPlot.datasets().length, "same number of area selections as datasets");
+
+        stackedAreas.each(function(d, i) {
+          let stackedAreaSelection = d3.select(this);
+          let areaVertices = TestMethods.areaVertices(stackedAreaSelection);
+          let areaXs = areaVertices.map((areaVertex) => areaVertex.x).slice(0, -2);
+          d.forEach((datum: any, index: number) => {
+            assert.closeTo(areaXs[index], constantValue, window.Pixel_CloseTo_Requirement, "cools");
```
ztsai: message

test/plots/stackedAreaPlotTests.ts
bluong: These are attributes however.  What do you think we should rename them to be?

ztsai: from the usage, `foo` could be `x` and `bar` could be `offset`, but then again, it's just personal preference as there's not a clear mapping

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+        assert.strictEqual(stackedAreas.size(),
+          stackedAreaPlot.datasets().length, "same number of area selections as datasets");
+
+        stackedAreas.each(function(data, i) {
+          let stackedAreaSelection = d3.select(this);
+          let areaVertices = TestMethods.areaVertices(stackedAreaSelection);
+          let areaXs = areaVertices.map((areaVertex) => areaVertex.x).slice(0, -2);
+          data.forEach((datum: any, datumIndex: number) => {
+            let x = accessor(datum, datumIndex, stackedAreaPlot.datasets()[i]);
+            assert.closeTo(areaXs[datumIndex], x, window.Pixel_CloseTo_Requirement, "cools");
```
ztsai: message

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+        stackedAreaPlot.renderTo(svg);
+
+        let stackedAreas = stackedAreaPlot.content().selectAll("path");
+        assert.strictEqual(stackedAreas.size(),
+          stackedAreaPlot.datasets().length, "same number of area selections as datasets");
+
+        stackedAreas.each(function(d, i) {
+          let stackedAreaSelection = d3.select(this);
+          let areaVertices = TestMethods.areaVertices(stackedAreaSelection);
+          let areaXs = areaVertices.map((areaVertex) => areaVertex.x).slice(0, -2);
```
ztsai: can you check the last 2 points as well?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+      });
+
+      it("can set the y0 property accessor to a constant value", () => {
+        let constantValue = 10;
+        stackedAreaPlot.y0(constantValue);
+        assert.strictEqual(stackedAreaPlot.y0().accessor(null, 0, null), constantValue, "property set to constant value");
+        stackedAreaPlot.destroy();
+      });
+
+      it("can set the y0 property accessor to be dependent on the input data", () => {
```
ztsai: can you also checked the rendered points?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
-      svg.remove();
+    describe("rendering on edge case scenarios", () => {
+      it("renders nothing when no data", () => {
+        let svg = TestMethods.generateSVG();
+        let stackedAreaPlot = new Plottable.Plots.StackedArea();
+        stackedAreaPlot.x(() => null);
+        // HACKHACK https://github.com/palantir/plottable/issues/2712 y scale must be set.
+        stackedAreaPlot.y(() => null, new Plottable.Scales.Linear());
+        stackedAreaPlot.renderTo(svg);
+        assert.strictEqual(stackedAreaPlot.selections().size(), 0, "test");
```
ztsai: message

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+        stackedAreaPlot.destroy();
+        svg.remove();
+      });
+
+      it("0 as a string coerces correctly and is not subject to off by one errors", () => {
```
ztsai: can you rewrite the message so that it reads semantically along with `it`?

ztsai: also, I believe we usually consider number string as invalid input in other plots
"0" is an edge case, can you test against non-zero number string and see it's skipped?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -2,485 +2,346 @@
...
+        stackedAreaPlot.destroy();
+      });
     });
 
-    it("disables downsampling", () => {
-      assert.strictEqual(renderer.downsamplingEnabled(), false, "downsampling is disabled by default");
-      renderer.downsamplingEnabled(true);
-      assert.strictEqual(renderer.downsamplingEnabled(), false, "downsampling will not be enabled by user");
-      svg.remove();
+    describe("rendering on edge case scenarios", () => {
```
ztsai: can you also test if x values of datasets don't align?

test/plots/stackedAreaPlotTests.ts
bluong: I believe there is a bug here
https://github.com/palantir/plottable/issues/2861

bluong:
I am unsure about using `assert` vs the `areaVertices` function.  My understanding is that I did not want to necessarily assert but only extract the relevant pieces of information and then continue with my test flow.

test/plots/stackedAreaPlotTests.ts
bluong: Why so?

test/plots/stackedAreaPlotTests.ts
bluong: `stackedAreaPlot` does not support this scenario.  How do you want me to test this?

ztsai: is this documented somewhere?

bluong: Well it's just that `stackedAreaPlot` will operate in a way that we would not recommend users to use.  If someone can come up with a detailed spec on how we want to do this, then we can align to that idea, but right now that spec is very vague

test/plots/stackedAreaPlotTests.ts
ztsai: what about adding a test and then skip it?

test/plots/stackedAreaPlotTests.ts
ztsai: for correctness of the endpoints?

bluong: I'll see how hard / useful it is.

test/plots/stackedAreaPlotTests.ts
bluong: I personally don't see too much immediate value in adding a skipped test.  I would much rather create an issue for the bug and then make the test at that point.

test/plots/stackedAreaPlotTests.ts
```diff
@@ -1,486 +1,373 @@
...
+        let stackedAreaSelection2 = d3.select(stackedAreaSelections[0][2]);
+
+        let areaVertices0 = TestMethods.areaVertices(stackedAreaSelection0);
+        let areaYs0 = areaVertices0.map((areaVertex) => areaVertex.y).slice(0, -2);
+        let areaVertices1 = TestMethods.areaVertices(stackedAreaSelection1);
+        let areaYs1 = areaVertices1.map((areaVertex) => areaVertex.y).slice(0, -2);
+        let areaVertices2 = TestMethods.areaVertices(stackedAreaSelection2);
+        let areaYs2 = areaVertices2.map((areaVertex) => areaVertex.y).slice(0, -2);
+
+        assert.closeTo(areaYs0[0], yScale.scale(<any> data0[0].y), window.Pixel_CloseTo_Requirement,
```
ztsai: why do you need to any cast data0?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -1,486 +1,373 @@
...
+          "dataset0 should have no offset");
+        assert.closeTo(areaYs1[0], yScale.scale(<any> data0[0].y), window.Pixel_CloseTo_Requirement,
+          "dataset1 should be offset with dataset0");
+        assert.closeTo(areaYs2[0], yScale.scale(data2[0].y + parseInt(<any> data0[0].y, 10)), window.Pixel_CloseTo_Requirement,
+          "dataset2 should be offset with only dataset0");
```
ztsai: this error message is misleading.  it's not offset with dataset1 because the value happens to be "0"

test/plots/stackedAreaPlotTests.ts
bluong: `data0[0]` is a string and our current behavior is such that there is a coercion done when stacking from strings to numbers.  In accordance there must be a coercion done from `string`s to `number`s for the scale

test/plots/stackedAreaPlotTests.ts
```diff
@@ -1,486 +1,373 @@
...
+        let areaVertices1 = TestMethods.areaVertices(stackedAreaSelection1);
+        let areaYs1 = areaVertices1.map((areaVertex) => areaVertex.y).slice(0, -2);
+        let areaVertices2 = TestMethods.areaVertices(stackedAreaSelection2);
+        let areaYs2 = areaVertices2.map((areaVertex) => areaVertex.y).slice(0, -2);
+
+        assert.closeTo(areaYs0[0], yScale.scale(<any> data0[0].y), window.Pixel_CloseTo_Requirement,
+          "dataset0 should have no offset");
+        assert.closeTo(areaYs1[0], yScale.scale(<any> data0[0].y), window.Pixel_CloseTo_Requirement,
+          "dataset1 should be offset with previous datasets");
+        assert.closeTo(areaYs2[0], yScale.scale(data2[0].y + parseInt(<any> data0[0].y, 10)), window.Pixel_CloseTo_Requirement,
```
ztsai: should be data2[0].y + data1[0].y + data1[0].y ?

PR #2870, opened 2015-10-12
Implement Data.KeyFunction

softwords:
Resubmitted against objectConstancy branch - previously #2833
also see issue #2512 
This pull request

adds the key function to Dataset
uses this key in Drawer when binding data to the selection
provides 2 predefined keys:
-- NoConstancy - which will return a new value each time it is called, therefore ensuring that there is never object constancy; that is, all elements in the selection are removed and recreated
-- useIndex - returns the index of the datum n the array, which is the default behaviour of d3 anyway

This PR provides a necessary foundation for developing animations in plottable that respect "object constancy" - proving visual representation of elements moving, entering or exiting the datasets being plotted.


bluong:
Can you edit your pull request description to reference the other pull request and vice versa?  After that, we can close the other pull request, and we can continue discussion here.

bluong:
Feel free to copy paste the original pull request description here as well.  Specify what exactly this pull request brings to the table (if it's nothing at all from user-perspective, then mention this).

bluong:
I just pushed in the changes for our Gruntfile.  Can you check to see if this helps?

softwords:
Solved the plottable.d.ts issue by resorting to explicitly changing the line ends before saving.
Not sure why this is required for this file and not any other...?

bluong:
Weird, we'll run with this for now.  When making further pull requests, can you report if you're running into more of the same issue?

By the way, with the way our Gruntfile is set up, all of the creation of plottable.d.ts is on Typescript / Grunt-TS .  It might be a good idea to ping them if you continue to run into issues.

src/core/dataset.ts
```diff
@@ -2,9 +2,23 @@ module Plottable {
...
+  private static counter: number = 0;
+  public static noConstancy = (d: any, i: number) => {
+    return KeyFunctions.counter++;
+  };
+  public static useIndex = (d: any, i: number) => {
+    return i;
+  };
+  public static useProperty(propertyname: string) {
+    return (d: any, i: number) => { return d[propertyname]; };
+  };
```
bluong: Remove the unnecessary parameters from the functions as necessary.

src/core/dataset.ts
```diff
@@ -88,5 +102,26 @@ export class Dataset {
       return this;
     }
   }
+
+  public keyFunction(): (datum: any, index: number) => any;
+  /**
+   * Sets the keyFunction.
+   * in d3, when binding data using selection.data(), a keyFunction may be supplied
```
bluong: "In D3, ..."

src/core/dataset.ts
```diff
@@ -88,5 +102,26 @@ export class Dataset {
...
+   * to generate a unique identifier for each datum. When data is updated, d3 uses this identifier to
+   * determine which data points have entered or exited the visualisation.
```
bluong: visualization

test/core/keyTests.ts
```diff
@@ -0,0 +1,45 @@
...
+    assert.deepEqual(ds.keyFunction(), key, "key is passed to dataset");
+  });
+
+  it("may accept noConstancy predefined key function", () => {
+    let ds: Plottable.Dataset = new Plottable.Dataset();
+    ds.keyFunction(Plottable.KeyFunctions.noConstancy);
+    let d = { foo: "bar" };
+    let a = ds.keyFunction()(d, 1);
+    let b = ds.keyFunction()(d, 1);
+    assert.isTrue(b - a === 1, "invocations give numerically increasing results");
```
bluong: Let's have better variable names than the above

test/core/keyTests.ts
```diff
@@ -0,0 +1,45 @@
+﻿///<reference path="../testReference.ts" />
+
+describe("DatasetKey", () => {
+  it("is passed to Dataset", () => {
+    let ds: Plottable.Dataset = new Plottable.Dataset();
```
bluong: Remove the explicit typing here and in all other occurances

test/core/keyTests.ts
```diff
@@ -0,0 +1,45 @@
+﻿///<reference path="../testReference.ts" />
+
+describe("DatasetKey", () => {
```
bluong: Dataset KeyFunctions?

test/core/keyTests.ts
bluong: Can you apply this comment to the rest of the tests?

test/core/keyTests.ts
```diff
@@ -0,0 +1,45 @@
...
+  it("is passed to Dataset", () => {
+    let ds: Plottable.Dataset = new Plottable.Dataset();
+    let key = (d: any, i: number) => { return i; };
+    ds.keyFunction(key);
+    assert.deepEqual(ds.keyFunction(), key, "key is passed to dataset");
```
bluong: Please use `assert.strictEqual` here.  `deepEqual` is for arrays and objects I believe.

test/core/keyTests.ts
bluong: For this assertion, can we change it to
`assert.strictEqual(b, a + 1...)`?

`assert.strictEqual` is significantly more useful than `assert.isTrue`.  Please apply this idea to other occurances

bluong:
It seems that due to this pull request, users will not see anything unless they opt into the `keyFunction` endpoint.  However, it seems that based on prior discussion, we cannot do that truly just yet (due to awkwardnesses from `_animateOnNextRender`.

Can you up a JSFiddle that shows just the current behavior, so that our QE knows the value that is added and what to look for going forward with the other pull requests?

test/core/keyTests.ts
softwords: new to typescript - can you give some guidance on when it is appropriate or not appropriate to use explicit typing?

softwords:
>>It seems that due to this pull request, users will not see anything unless they opt into the keyFunction endpoint. However, it seems that based on prior discussion, we cannot do that truly just yet (due to awkwardnesses from _animateOnNextRender.

That is a good summary. As I mentioned in the other PR, this is the first of 3 steps to getting some enhanced object-constancy aware animations into plots.
The second step is to expose the enter-update-exit pattern to the animator, the third is to build animations on this. This first step doesn't give functionality for the reasons you've noted.

This [fiddle](http://jsfiddle.net/oaskqta0/12/) is based on code I have in another branch building forward from this to address steps 2 and 3 - it is genuinely created from the plottable typescript framework - no javascript hacking!
In it, you see a new animator ```Bar``` acting on the bar chart from earlier. But now, you have animations showing the outgoing elements, staged animation on the updating selection, altering first the height then the position, and finally, animation on the incoming elements that grow into the vacant slots. (Turn off the constancy key, to see more mundane behaviour.)

This staged visualistion model is based on the ideas in this [paper ](http://vis.berkeley.edu/papers/animated_transitions/) that Mike Bostock has referred to a number of times.

test/core/keyTests.ts
bluong: Explicit typing is mainly unnecessary if Typescript can infer the type.

In this scenario, you are setting the variable to `new Plottable.Dataset()` (which is of type `Plottable.Dataset).  Thus, the variable can be inferred to be of type `Plottable.Dataset`

bluong:
Oh it seems I misinterpreted the pull request.  In using this pull request in a JSFiddle, it seems that in fact this pull request does nothing even if KeyFunctions are set due to the nature of having a reset animator.

Can you verify that this is the case?  After you do that I can assign it to our QE to ensure that nothing broke.  In the meantime, feel free to proceed to the next step.

softwords:
>>it seems that in fact this pull request does nothing even if KeyFunctions are set due to the nature of having a reset animator

That's correct. But in the next proposed PR I am working on, you will need to set the noConstancy key on the the dataset to get this same default behaviour ( I am building all this into quicktests)

bluong:
Thanks a lot @softwords !  I'm assigning the pull request to our QE to make sure that nothing broke.  In the meanwhile, feel free to keep developing the next stage.

If it is already ready, feel free to make the pull request aiming at `objectConstancy`.  Since I have context, I can sift through the `KeyFunction` additions you made in this pull request to get at your new code added.

src/drawers/drawer.ts
```diff
@@ -74,7 +74,14 @@ export class Drawer {
    * @param{any[]} data The data to be drawn
    */
   private _bindSelectionData(data: any[]) {
-    let dataElements = this.selection().data(data);
+    // if the dataset has a key, use it when binding the data   
+    let dataElements: d3.selection.Update<any>;
+    if (this._dataset && this._dataset.keyFunction()) {
```
bluong: Since it seems that `this._dataset.keyFunction()` will never be `null` (it is by default using the index as the key, and there is a guard surrounding setting it to `null`), do we need the above guard?

bluong:
Whoops, it seems that I missed something.  @softwords can you look into my comment?  After you follow up, feel free to proceed with your next pull request.

src/drawers/drawer.ts
softwords: Noted, done. Remember that the other test (for the existence of dataset) we have already discussed.

crmorford:
Hey @softwords, I just wanna say sorry for the holdup, I was pretty busy at the end of last week. I'm going to take a look at this now :)

crmorford:
**Testing Fiddles** 
**No action required by softwords**

Plot Types:
Horizontal Bar (normal, stacked, clustered)
Vertical Bar (normal, stacked, clustered)
Scatter
Line
Area
Pie

Axis Types:
Linear 
Log
Time
Category

Data stuff:
2 datasets with no overlap
2 datasets with all overlap
dataset doesn't have key identified
change key function from category1 --> category2
key function uses combination of 2 categories?
change fill function based on index http://jsfiddle.net/oaskqta0/13/

crmorford:
Hi @softwords,


In this [fiddle](http://jsfiddle.net/oaskqta0/13/), I'm using the rankcolor function that you had written (with some tweaks), and there's somthing I'm confused about. 

rankColor uses the index to color bars. It seems to work correctly when there is no object constancy or when it's joined on the index, but I'm seeing something I don't expect when object constancy is enabled.

It starts out looking good:
screen shot 2015-10-19 at 11 42 20 am

But when I press the "2010" button, the blues and purples get mixed up:
screen shot 2015-10-19 at 11 42 37 am

I would expect the fill function, rankColor, to just be based on the index, but it seems like something else is going on. Can you explain that?

crmorford:
adding on to last comment:
on plottable develop the same fiddle (minus some code) uses the rankColor function the way i'd expect when you click 2006/2010/2014 buttons -- the colors always go in rainbow order
http://jsfiddle.net/oaskqta0/14/

on this branch:
uncommenting the ds.keyFunction() lines, and clicking the 2006/2010/2014 buttons causes different coloring than expected -- not in rainbow order
http://jsfiddle.net/oaskqta0/15/

I don't want to QE +1 this until I understand what's happening, but this is the only thing i've found so far. regression pass looks good.

bluong:
Looked into your JSFiddle @crmorford and I found why there is an inconsistency (which seems to me that we should ship with the index constancy which I believe keeps the default behavior).  It has to do with how D3 does its enter/exit selections in accordance with the data.

Unless I'm mistaken, let's say I have an array `[o1, o2, o3]`.  Now let's say I change the data to `[c1, c2, c3]`, however let's say that `key(o1) === key(c2)`, then I believe D3 will actually count `c2` to be existing where `o1` used to exist.  Thus, the input indices are modified.

bluong:
The realization that the "same" datum maintains their index across changing data is something that caught us off guard, but it makes sense after a bunch of discussion.  We're going to need some time to think about think specific realization, but we should come up with an answer soon.

crmorford:
Ok, this looks good for now. I think it'll be far easier to test intensively when the next PR comes in, and I don't see it causing any problems with existing use cases.

bluong:
@softwords We are awaiting your next pull request!

softwords:
@bluong I was just replying to @crmorford when the merge came through, so I am overtaken by the speed of events!
Ok - I think now, for your risk management point of view, I should put up _everything_ that I now have on this. You can see the strategy (and code!) and decide sooner rather than later on architecture issues, (or indeed if you want to proceed) and gauge the amount of effort involved in getting it over the line. 
Take a look at it, and we can then think about whether you want to try to push it all through in one big PR, or try to break it up into smaller bits.
stay tuned!

softwords:
hello @crmorford, 
-- I just wanna say sorry
no apology required! :)
Looking into your [fiddle](http://jsfiddle.net/oaskqta0/15/), notice that the x co-ordinate of each bar is taken from the ```R``` property in the data, rather than from the index.
If your color function becomes
``` javascript
    rankColor = function (d) {
        if (d.R == 1) return "#FF0000";
        if (d.R == 2) return "#FFA500";
        if (d.R <= 4) return "#FFFF00";
        if (d.R <= 8) return "#00FF00";
        if (d.R <= 16) return "#0000FF";
        return "#FF00FF";
    }
```
then it works as you would [expect](http://jsfiddle.net/oaskqta0/16/). 
But, this data is sorted in order of ```R``` anyway, ie, you would think that ```d.R === i```
Using this color function
``` javascript
    rankColor = function (d, i) {
        if (d.R - 1 === i) return "#FF0000";
        return "#FF00FF";
    }
```
[demonstrates this is not so](http://jsfiddle.net/oaskqta0/17/) when constancy is on.
So @bluong you are right that the issue is to do with the index.
Currently in plottable's animators the attributes are applied to the entire selection - which is in one-to-one correspondence with the incoming data. So in this circumstance you __can__ assume that the index ```i``` on the attribute function is the index into the data array.  But - when the Animator can work with the ```enter```, ```update``` and ```exit``` selections created by ```selection.bind()``` then the parameter ```i``` is passed the index into the targeted selection, which will in general not be the same as the index into the original data.

bluong:
@softwords , I am mainly curious in your approach regarding the `Animator`s.  In earlier pull requests, you seem to reference an `Animators.Bar`, etc.  Can you elaborate on this?  If you have a pull request concerning this, I'd be happy to look at this as well.

softwords:
@bluong @crmorford I have been putting together these notes explaining what is in my next PR from branch 'DrawingTarget'. (within 24 hours)

__DrawingTarget__

A key change in this PR is to expose the D3 "enter / update / exit" pattern to the animators.
This introduces the type ```DrawingTarget```, which has properties ```enter``` ```exit``` ```update``` and ```merge``` corresponding to these selections. ```merge``` is the combination of update and enter - in d3, once the ```enter().append()``` method is called on the bound data, that bound data is automatically updated to include the entering elements. ```merge``` is therefore in one-to-one correspondence with the incoming data.
The DrawingTarget is created by Drawer in _bindSelectionData. It is then passed to the animators, by extending the ```IAnimator``` interface:
```javascript
  animate(selection: d3.Selection<any>|d3.Transition<any>, attrToAppliedProjector: AttributeToAppliedProjector,
    drawingTarget?: Drawers.DrawingTarget, drawer?: Drawer): d3.Selection<any> | d3.Transition<any>;
```
Note that _bindSelectionData no longer calls ```.remove()``` on the exit selection - otherwise you still could not do anything with it! ```remove()```ing the ```exit``` selection is now a responsibility of the animator; see below.

__Initializer__

Usually, in the d3 "enter update exit" pattern, there is some initialization of attributes when the ```enter()``` elements are created; e.g.:
```
	selection.enter()
	  .append(_svgElement)
	  .attr(....)
```

```initializer``` is a new property on Drawer that takes a function ```()=> AttributeToProjector```. Drawer also gets the method ```appliedInitializer(): AttributeToAppliedProjector``` which is the result of the initializer function converted to an ```AttributeToAppliedProjector```. 
This is applied to the ```enter()``` elements to initialize their state. 
The ```initializer``` is passed by the plot in its ```_createDrawer``` method, so the ```initializer``` may have access to the internals of the calling plot. In fact, a number of plots already have the code needed for the initializer - it's precisely the code in the ```_animateOnNextRender``` block. So this refactor does away with the pesky ```_animateOnNextRender``` logic, and makes the initialization of new elements more "mainstream" d3.

__Responsibilities of Animators__

Animators now get a lot of opportunities to work in detail with the "enter update exit" pattern, and in particular, to apply transitions to the ```enter``` and ```exit``` selections. Animators have two responsibilities: 
-- remove() the exit selection. 
-- update the DrawingTarget when transitions are applied.
For example, if the Drawer derives a transition from drawingTarget.enter, it should update drawingTarget.enter with that transition:
```javascript
drawingTarget.enter = this.getTransition(drawingTarget.enter, 1000);
```
The point of this is that a subsequent transition derived from ```drawingTarget.enter``` will be a "subtransition", and (if you are careful not to interfere) d3 will handle scheduling this subtransition to chain from the first (see [here](https://github.com/mbostock/d3/wiki/Transitions#transition)).  This is one way to create more complex staged animations.

__Easing Functions__

This PR introduces some custom easing functions:
```atStart``` - everything happens immediately at the start of the transition. ie this create a "wait" after applying the attribues.
```atEnd``` - everything happens in one go at the end of the transition - ie this creates a "wait" before applying the attributes.
```squEase``` - this applies the transition squeezed into a subinterval [a,b] where 0 <= a <= b <= 1. The actual easing to apply in that interval is passed as an argument. 
These are designed to help animators schedule animations between the enter update and exit selections - each can have a transition of the same duration, but the actual timing of activity will be determined by the easing function.

__Animator Hierarchy__

There are a number of new animator classes:
```Base``` - basically equivalent to the current Easing, but with various protected helper methods available to subclasses.
```Attr``` - apply a set of attributes to enter coming in - and to exit going out. 
```Opacity``` - derived from Attr, fades incoming/outgoing in and out.
```Callback``` - gets the actual ```animate``` function as a property. This allows the javascript client full access to the "enter update exit" selections. I think this is in keeping with the philosophy that plottable will always provide a way to get directly to d3 for occasions when you need something not provided in higher level code.  
```Reset``` - applies the ```Drawer```'s ```appliedInitializer``` to the entire ```merge``` selection. (this is a key reason for ```Drawer``` passing itself to the animator).
```Bar``` - a staged animation specific to bar charts, that exits, then moves, then enters elements in turn. Demonstrates multiple transitions, custom easing functions, and the various helper functions in ```Animators.Base```.

__Legacy behaviour__

Current animation resets all elements to a "start" state, and animates them back to their final state. To get this to happen in the new architecture, the ```Drawer```'s ```appliedIinitializer``` has to be applied to all elements, not just the ```enter()```ing elements. Two ways to do this:
-- use the ```noConstancy``` key. With ```noConstancy```, every element is always in ```enter```, so it gets initialised.
-- use the ```Reset``` animator. 

__Quicktests__

See the animation.html in the quicktests/overlay folder. Click on any of these animations to toggle between two datasets. The data items in these datasets have a ```name``` property, a new utility function ```makeRandomNamedData()``` generates names with some degree of overlap - so you get ```enter``` ```update``` and ```exit``` selections to observe. The dropdown bar lets you choose between key functions - useProperty("name") (ie constancy), noConstancy, and useIndex.

__function.prototype.bind__

To maintain ```this``` as the context of the hosting plot, initializer functions use ```function.prototype.bind```, which it turns out is not supported by phantomjs versions < 2.0.0. I have added a polyfill into coverage.html to get around this.

__//# sourceURL__

The new quicktests use the annotation
```//# sourceURL=<url> ```
as the last line of the file. When chrome loads the file dynamically, the sourceURL annotation causes it to appear in the Source page list of files, which makes it much easier to examine and debug. I found it's better placed at the end rather than the start, because chrome removes the line, and thereafter seems to be one line out when you are stepping through the code.

PR #2874, opened 2015-10-12
[Table] Add method to return Component at a specified row, column index.

### LINKED ISSUE #2842: [Table] - `componentAt(row, column)` endpoint
There currently does not seem to be a function to get a Component at a specified location on a `Components.Table`

jtlan:
New API Point:
```Typescript
/**
 * Returns the Component at the specified row and column index.
 *
 * @param {number} rowIndex
 * @param {number} columnIndex
 * @returns {Component} The Component at the specified position, or null if no Component is there.
 */
componentAt(rowIndex: number, columnIndex: number): Component;
```

Fiddle: http://jsfiddle.net/pbotk0v8/

Close #2842.

### DIFF
--- a/src/components/table.ts
+++ b/src/components/table.ts
@@ -81,6 +81,20 @@ export module Components {
     }
 
     /**
+     * Returns the Component at the specified row and column index.
+     *
+     * @param {number} rowIndex
+     * @param {number} columnIndex
+     * @returns {Component} The Component at the specified position, or null if no Component is there.
+     */
+    public componentAt(rowIndex: number, columnIndex: number) {
+      if (rowIndex < 0 || rowIndex >= this._nRows || columnIndex < 0 || columnIndex >= this._nCols) {
+        return null;
+      }
+      return this._rows[rowIndex][columnIndex];
+    };
+
+    /**
      * Adds a Component in the specified row and column position.
      *
      * For example, instead of calling `new Table([[a, b], [null, c]])`, you
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -2624,6 +2624,14 @@ declare module Plottable {
              */
             has(component: Component): boolean;
             /**
+             * Returns the Component at the specified row and column index.
+             *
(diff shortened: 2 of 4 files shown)


src/components/table.ts
```diff
@@ -81,6 +81,20 @@ export module Components {
...
+     * @param {number} rowIndex
+     * @param {number} columnIndex
+     * @returns {Component} The Component at the specified position, or null if no Component is there.
+     */
+    public componentAt(rowIndex: number, columnIndex: number) {
+      if (rowIndex < 0 || rowIndex >= this._nRows || columnIndex < 0 || columnIndex >= this._nCols) {
+        return null;
+      }
+      return this._rows[rowIndex][columnIndex];
+    };
```
bluong: Just curious (no code action needed, necessarily): let's say that we want to add an endpoint to set the component at a specified index, would we add it through extending this endpoint?

src/components/table.ts
```diff
@@ -81,6 +81,20 @@ export module Components {
...
 
     /**
+     * Returns the Component at the specified row and column index.
+     *
+     * @param {number} rowIndex
+     * @param {number} columnIndex
+     * @returns {Component} The Component at the specified position, or null if no Component is there.
+     */
+    public componentAt(rowIndex: number, columnIndex: number) {
+      if (rowIndex < 0 || rowIndex >= this._nRows || columnIndex < 0 || columnIndex >= this._nCols) {
```
bluong: Use `Utils.Math.inRange` instead?

src/components/table.ts
jtlan: You mean `add()`? Or something like that? There might not be a way to name it that makes it clear this is a getter under our chosen conventions.

crmorford:
copy a table with interactions on components http://jsfiddle.net/j5b9f60c/54/
call methods on componentAt(i. j) http://jsfiddle.net/j5b9f60c/55/

PR #2875, opened 2015-10-13
[Axis Tests] - Rework

bluong:
screen shot 2015-10-16 at 2 25 25 pm

### DIFF
--- a/test/axes/axisTests.ts
+++ b/test/axes/axisTests.ts
@@ -628,7 +628,7 @@ describe("Axis", () => {
     svg.remove();
   });
 
-  it("renders a duplicate tick in the first tier it can fit in", () => {
+  it("places the tick at the top most row even if it is a duplicated tick", () => {
     let annotatedTicks = [150, 150];
     let scale = new Plottable.Scales.Linear();
     scale.domain([100, 200]);
--- a/test/axes/axisTests.ts
+++ b/test/axes/axisTests.ts
@@ -1,6 +1,36 @@
 ///<reference path="../testReference.ts" />
 
 describe("Axis", () => {
+
+  const horizontalOrientations = ["top", "bottom"];
+  const verticalOrientations = ["left", "right"];
+  const orientations = horizontalOrientations.concat(verticalOrientations);
+
+  const isHorizOrient = (orientation: string) => horizontalOrientations.indexOf(orientation) > -1;
+  const numAttr = TestMethods.numAttr;
+
+  horizontalOrientations.forEach((horizontalOrientation) => {
+    let alignment = horizontalOrientation === "top" ? "bottom" : "top";
+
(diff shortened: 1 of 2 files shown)


test/axes/axisTests.ts
```diff
@@ -1,221 +1,237 @@
...
-    baseAxis.orientation("right");
-    assert.isNotNull(baseline.node(), "baseline was drawn");
-    assert.strictEqual(baseline.attr("x1"), "0");
-    assert.strictEqual(baseline.attr("x2"), "0");
-    assert.strictEqual(baseline.attr("y1"), "0");
-    assert.strictEqual(baseline.attr("y2"), String(SVG_HEIGHT));
-
-    svg.remove();
-  });
+    it("updates the layout when updating the margin after rendering", () => {
```
jtlan: Why an extra test for this `describe()` only?

test/axes/axisTests.ts
```diff
@@ -839,15 +871,16 @@ describe("BaseAxis", () => {
             assert.strictEqual(annotationRect.attr("visibility"), "hidden", `rect ${i} outside margin area should be not visible`);
           }
         });
+        axis.destroy();
         svg.remove();
       });
 
       it("shows annotation circles are not hidden regardless if rectangles are hidden", () => {
```
jtlan: This test name isn't quite a sentence.

jtlan:
Some of the `describe()` blocks for orientations don't seem symmetric with each other. Is there a reason for that? Maybe the unique tests could be pulled out into a different block.

bluong:
Employed the usage of wrapping `it` statements into a `forEach` for reusing a single test across orientations

test/axes/axisTests.ts
```diff
@@ -0,0 +1,652 @@
...
+      annotatedTickSet.add(annotatedTick);
+    });
+    assert.strictEqual(axis.content().selectAll(`.${Plottable.Axis.ANNOTATION_LABEL_CLASS}`).size(),
+      annotatedTickSet.size, "only unique annotations rendered");
+
+    axis.destroy();
+    svg.remove();
+  });
+
+  it("renders a duplicate tick in the first tier it can fit in", () => {
```
jtlan: This seems at odds with the previous test. What's the expected behavior?

bluong: The previous test makes sure that only the unique ones render.  However there was an earlier bug where duplicating an annotated tick would offset the tick into a different row.

Ex. Earlier bug had having annotated ticks be [50, 50, 50, 50] render single annotation for 50 but in the 4th row.

jtlan: I think this test should be renamed to make that clear then. "Does not leave extra space for duplicated ticks"?

PR #2878, opened 2015-10-13
interpolatedColorLegend padding

### LINKED ISSUE #2856: Padding on InterpolatedColorLegend
image
Here's an example: Slate would like to be able to place padding between the plot and legend. We don't have a way to do that right now (maybe legend.xAlignment("right"), but it isn't super customizable.

Ideally, they'd like to set a fixed amount of padding around the legend component (measured in pixels)

The API suggested is something like table.padding(row, col, 20);
However they'd like to be able to do just leftPadding, just rightPadding, top, bottom.

Maybe component.padding(pixelAmount, arrayOfDirections)
so you'd call it like:
component.padding(20, ["left", "bottom"]);

**Edit: the key ask behind this issue is that the legend is too close to the plot. We probably won't resolvve this with a padding API point, but just adjust the legend padding**

crmorford:
fixes #2856 

1. Make the padding as wide as the swatches
2. remove the padding on the ends (in the long direction)

before:
http://jsfiddle.net/n5prcpak/2/
http://jsfiddle.net/n5prcpak/3/
screen shot 2015-10-13 at 4 03 16 pm

after:
http://jsfiddle.net/n5prcpak/
http://jsfiddle.net/n5prcpak/1/
screen shot 2015-10-13 at 4 03 29 pm


src/components/interpolatedColorLegend.ts
```diff
@@ -221,55 +223,55 @@ export module Components {
...
         width: 0,
         height: 0
       };
 
       let numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES;
-      let textHeight = this._measurer.measure().height;
+
       if (this.expands() && textHeight > 0) {
-        let offset = this._isVertical() ? 2 * padding :  4 * padding - text0Width - text1Width;
+        let offset = this._isVertical() ? 0 :  2 * padding - text0Width - text1Width;
```
bluong: Why less offset here?

bluong: Why did this offset decrease?  It seems that now we can fit more swatches?

src/components/interpolatedColorLegend.ts
```diff
@@ -221,55 +223,55 @@ export module Components {
...
         numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), numSwatches);
       }
 
       if (this._isVertical()) {
         let longestTextWidth = Math.max(text0Width, text1Width);
-        swatchWidth = Math.max( (this.width() - 3 * padding - longestTextWidth), 0);
-        swatchHeight = Math.max( ((this.height() - 2 * padding) / numSwatches), 0);
-        swatchY = (d: any, i: number) => padding + (numSwatches - (i + 1)) * swatchHeight;
+        swatchWidth = Math.max( (this.width() - 2 * padding - textPadding - longestTextWidth), 0);
+        swatchHeight = Math.max( (this.height() / numSwatches), 0);
```
bluong: Remove precediting whitespace

src/components/interpolatedColorLegend.ts
```diff
@@ -221,55 +223,55 @@ export module Components {
...
-        swatchX = (d: any, i: number) => (padding + text0Width + padding) + i * swatchWidth;
+        swatchX = (d: any, i: number) => (text0Width + textPadding) + i * swatchWidth;
         swatchY = (d: any, i: number) => padding;
 
         upperWriteOptions.xAlign = "right";
-        upperLabelShift.x = -padding;
+        upperLabelShift.x = 0;
         lowerWriteOptions.xAlign = "left";
-        lowerLabelShift.x = padding;
+        lowerLabelShift.x = 0;
```
bluong: Why these changes?

test/components/interpolatedColorLegendTests.ts
```diff
@@ -279,4 +282,88 @@ describe("InterpolatedColorLegend", () => {
...
+
+  it("does not have padding on ends of vertical legends", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+    let orientations = ["left", "right"];
+    orientations.forEach((orientation) => {
+      legend.orientation(orientation).expands(true);
+      let height = (<Element> legend.content().select(".swatch-container").node()).getBoundingClientRect().height;
+      assert.closeTo(height, SVG_HEIGHT, window.Pixel_CloseTo_Requirement, "actual height is SVG_HEIGHT");
+
```
bluong: Remove the ending newline here

test/components/interpolatedColorLegendTests.ts
```diff
@@ -279,4 +282,88 @@ describe("InterpolatedColorLegend", () => {
...
   });
+
+  it("does not have padding on ends of vertical legends", () => {
+    let legend = new Plottable.Components.InterpolatedColorLegend(colorScale);
+    legend.renderTo(svg);
+    let orientations = ["left", "right"];
+    orientations.forEach((orientation) => {
+      legend.orientation(orientation).expands(true);
+      let height = (<Element> legend.content().select(".swatch-container").node()).getBoundingClientRect().height;
+      assert.closeTo(height, SVG_HEIGHT, window.Pixel_CloseTo_Requirement, "actual height is SVG_HEIGHT");
```
bluong: Can we include the orientation in the assertion message?

bluong:
Minor issues

plottable.js
```diff
@@ -5826,52 +5828,51 @@ var Plottable;
...
-                    y: padding,
+                    y: this._isVertical() ? 0 : padding,
                     width: 0,
                     height: 0
                 };
                 var numSwatches = InterpolatedColorLegend._DEFAULT_NUM_SWATCHES;
-                var textHeight = this._measurer.measure().height;
                 if (this.expands() && textHeight > 0) {
-                    var offset = this._isVertical() ? 2 * padding : 4 * padding - text0Width - text1Width;
+                    var offset = this._isVertical() ? 0 : 2 * padding - text0Width - text1Width;
```
crmorford: a vertical axis used to have padding on the top and bottom, so height - (2 * padding) was the height available to the swatches.

a horizontal axis used to have padding, text, padding, swatches, padding, text, padding. Now the outside padding is gone

that being said, i think that this should actually be 2 * textPadding - text0Width -text1Wdith, because textPadding is the (fixed) width between the text and swatches.

jtlan:
Is it possible to sacrifice padding first in constrained spaces, rather than shrinking the cells?
screen shot 2015-10-14 at 4 45 20 pm
http://jsfiddle.net/pqajfd51/

compare to a similar situation on **`develop`**:
screen shot 2015-10-14 at 4 45 47 pm

jtlan:
Also affects even more constrained scenarios:
screen shot 2015-10-14 at 4 55 20 pm

src/components/interpolatedColorLegend.ts
```diff
@@ -221,56 +222,61 @@ export module Components {
...
       if (this.expands() && textHeight > 0) {
-        let offset = this._isVertical() ? 2 * padding :  4 * padding - text0Width - text1Width;
+        let offset = this._isVertical() ? 0 :  2 * textPadding - text0Width - text1Width;
         let fullLength = this._isVertical() ? this.height() : this.width();
         numSwatches = Math.max(Math.floor((fullLength - offset) / textHeight), numSwatches);
       }
 
       if (this._isVertical()) {
+        padding = (this.width() - Math.max(text0Width, text1Width) - 2 * this._textPadding) / 2;
         let longestTextWidth = Math.max(text0Width, text1Width);
```
bluong: Let's move this calculation one line up and then use this variable in setting `padding`

bluong:
slight cleanup

jtlan:
There's still an edge case that occurs when the space is really constrained:
screen shot 2015-10-15 at 9 56 38 pm

... but in that scenario the layout is already totally busted, so I think this is probably fine for now. We should come up with another fail-safe in those cases (maybe with the help of #2887?)

crmorford:
http://jsfiddle.net/pqajfd51/3/
happens on develop too

PR #2881, opened 2015-10-14
enable downsampling in Stacked Area Plot

### LINKED ISSUE #2868: Downsampling create white space inside Stacked Area Plot
When downsampling is enabled, there are white spaces between two areas. 
Repro: http://jsfiddle.net/yxpq4uuv/36/ (You have to enable downsampling in stacked area plot first).
**Expected**: There should be no white spaces between two areas.

yunhaolucky:
**Issue**: 
When using downsampling, points with different x values are removed in different areas, which make the x value inconsistent among different dataset. This creates white spaces and overlappings between areas.

**Fix**:
The point drawn are the union of the downsampling result of each area. This means a point is removed only when all points at the same x position are removed by downsampling. This might affect the rendering performance.

Draw without white spaces and overlappings between areas: http://jsfiddle.net/yxpq4uuv/45/
Performance Test: http://jsfiddle.net/yxpq4uuv/46/

close #2868


src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
       return { x: pixelPoint.x, y: scaledYValue };
     }
 
+    protected _getDataToDraw() {
+      if (super.downsamplingEnabled() === false) {
+        return super._getDataToDraw();
+      }
+      let dataToDraw = new Utils.Map<Dataset, any[]> ();
+      if (this.datasets.length === 0) {
```
bluong: `this.datasets` returns the function, not the dataset array

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
       return { x: pixelPoint.x, y: scaledYValue };
     }
 
+    protected _getDataToDraw() {
+      if (super.downsamplingEnabled() === false) {
```
bluong: Why `super.downsamplingEnabled()` instead of `this.downSampling()`?

bluong: Change to
`if (!this.downsamplingEnabled()) {`?

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
...
+        return super._getDataToDraw();
+      }
+      let dataToDraw = new Utils.Map<Dataset, any[]> ();
+      if (this.datasets.length === 0) {
+        return dataToDraw;
+      }
+      let overallfilteredDataIndices: number[] = [];
```
bluong: correct the camelcase to `overallFilteredDataIndices`?

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
...
+      this.datasets().forEach((dataset, i) => {
+        let data = dataset.data();
+        let filteredDataIndices = data.map((d, i) => i);
+        if (super.croppedRenderingEnabled()) {
+          filteredDataIndices = super._filterCroppedRendering(dataset, filteredDataIndices);
+        }
+        if (super.downsamplingEnabled()) {
+          filteredDataIndices = super._filterDownsampling(dataset, filteredDataIndices);
```
bluong: Why all the `super` calls?

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
...
+        }
+        overallfilteredDataIndices = overallfilteredDataIndices.concat(filteredDataIndices
+          .filter((index) => {return overallfilteredDataIndices.indexOf(index) < 0; }));
+      });
+      this.datasets().forEach((dataset, i) => {
+        let data = dataset.data();
+        dataToDraw.set(dataset, [overallfilteredDataIndices.sort((a, b) => a - b).map((d, i) => data[d])]);
```
bluong: Why sort?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -396,7 +397,66 @@ describe("Plots", () => {
...
+
+      let renderer = new Plottable.Plots.StackedArea<number>();
+      renderer.y((d) => d.y, yScale);
+      renderer.x((d) => d.x, xScale);
+      renderer.addDataset(new Plottable.Dataset(data1));
+      renderer.addDataset(new Plottable.Dataset(data2));
+      renderer.downsamplingEnabled(true);
+      let xAxis = new Plottable.Axes.Numeric(xScale, "bottom");
+      new Plottable.Components.Table([[renderer], [xAxis]]).renderTo(svg);
+      let areas = (<any> renderer)._renderArea.selectAll(".area");
```
bluong: Please use `renderer.content().selectAll(".area")` instead

bluong:
Can you make the above changes to the source code?  I want to see the source code completed before moving to the test code as the test code seems a bit involved.

src/plots/stackedAreaPlot.ts
yunhaolucky: We want the indices to be sorted so that the result dataset of mapping function keep the same order as the original dataset.

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
...
+        let data = dataset.data();
+        let filteredDataIndices = data.map((d, i) => i);
+        if (this.croppedRenderingEnabled()) {
+          filteredDataIndices = this._filterCroppedRendering(dataset, filteredDataIndices);
+        }
+        if (this.downsamplingEnabled()) {
+          filteredDataIndices = this._filterDownsampling(dataset, filteredDataIndices);
+        }
+        overallFilteredDataIndices = overallFilteredDataIndices.concat(filteredDataIndices
+          .filter((index) => {return overallFilteredDataIndices.indexOf(index) < 0; }));
```
bluong: Why can't this idea be applied to normal `line` / `area`?

src/plots/stackedAreaPlot.ts
```diff
@@ -202,6 +181,33 @@ export module Plots {
...
+      });
+      this.datasets().forEach((dataset, i) => {
+        let data = dataset.data();
+        dataToDraw.set(dataset, [overallFilteredDataIndices.sort((a, b) => a - b).map((d, i) => data[d])]);
```
bluong: Remove the `i` parameter in the map.  Also replace the `d` parameter there.  It is incredibly missing as an `index` is actually being passed as the first argument.

test/plots/stackedAreaPlotTests.ts
```diff
@@ -396,7 +397,66 @@ describe("Plots", () => {
...
+      ];
 
+      let renderer = new Plottable.Plots.StackedArea<number>();
+      renderer.y((d) => d.y, yScale);
+      renderer.x((d) => d.x, xScale);
+      renderer.addDataset(new Plottable.Dataset(data1));
+      renderer.addDataset(new Plottable.Dataset(data2));
+      renderer.downsamplingEnabled(true);
+      let xAxis = new Plottable.Axes.Numeric(xScale, "bottom");
+      new Plottable.Components.Table([[renderer], [xAxis]]).renderTo(svg);
```
bluong: Why do we need the `xAxis` here?

test/plots/stackedAreaPlotTests.ts
```diff
@@ -396,7 +397,66 @@ describe("Plots", () => {
...
+      ];
+
+      let data2 = [
+        { x: 1, y: 1},
+        { x: 2, y: 2},
+        { x: 3, y: 3},
+        { x: 4, y: 2},
+      ];
 
+      let renderer = new Plottable.Plots.StackedArea<number>();
```
bluong: let's not use the name `renderer`.  It is an artifact of really old history.  Let's just call it a plot.

test/plots/stackedAreaPlotTests.ts
```diff
@@ -396,7 +397,66 @@ describe("Plots", () => {
...
+      let dataset = [data1, data2];
+      let EPSILON = 0.05;
+      assert.strictEqual(areas[0].length, dataset.length, "correct number of areas are drawn.");
+
+      let previousAreaValue = expectedRenderIndex.map((d) => 0);
+
+      dataset.forEach((d, i) => {
+        let area = d3.select(areas[0][i]);
+        let expectedAreaData = expectedRenderIndex.map((d) => dataset[i][d]);
+        let areaPoints = TestMethods.normalizePath(area.attr("d")).match(/(\-?\d+\.?\d*)(,|\s)(-?\d+\.?\d*)/g);
```
bluong: Where did you get this match from?

src/plots/stackedAreaPlot.ts
yunhaolucky: This should work in `line`/ `area`. But for `LinePlot` with several datasets, we do not need to make sure all datasets have the same indices. Therefore, maybe we'd better just keep this in `StackedAreaPlot`.

test/plots/stackedAreaPlotTests.ts
yunhaolucky: TestMethods.assertLinePathEqualToDataPoints

bluong:
The main thing I am concerned about is exposing `_filterCroppedRendering` and `_filterDownSampling` down to the children classes.  Based on your previous response, it seems that that is a differing step since `StackedArea` needs to maintain indices.  Wouldn't a better solution be to create a new small endpoint that exposes if the indices need to be the same and then switch on that endpoint?

adidahiya:
The codebase has changed a lot and this PR is stale, so I'm going to close this. We'd be open to revisiting this functionality in a fresh PR. Sorry for any inconvenience.

PR #2884, opened 2015-10-15
[TimeAxis Tests] - Rework

bluong:
screen shot 2015-10-16 at 4 49 21 pm

### DIFF
--- a/test/axes/timeAxisTests.ts
+++ b/test/axes/timeAxisTests.ts
@@ -2,6 +2,8 @@
 
 describe("TimeAxis", () => {
 
+  let orientations = ["top", "bottom"];
+
   describe("setting the orientation", () => {
     it("throws an error when setting a vertical orientation", () => {
       let scale = new Plottable.Scales.Time();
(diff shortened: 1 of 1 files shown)


test/axes/timeAxisTests.ts
```diff
@@ -205,212 +181,258 @@ describe("TimeAxis", () => {
...
 
-    let tiersToCreate = 15;
-    let configuration = Array.apply(null, Array(tiersToCreate)).map(() => {
-      return {interval: Plottable.TimeInterval.day, step: 2, formatter: Plottable.Formatters.time("%a %e") };
+      svg.remove();
     });
-    xAxis.axisConfigurations([configuration]);
 
-    xAxis.renderTo(svg);
+    it("shrinks in height if a newly set configuration provides less tiers", () => {
```
jtlan: "fewer tiers"

test/axes/timeAxisTests.ts
```diff
@@ -205,212 +181,258 @@ describe("TimeAxis", () => {
...
+    });
   });
 
-  it("three tier time axis should be possible", () => {
-    let svg = TestMethods.generateSVG();
-    let xScale = new Plottable.Scales.Time();
-    xScale.domain([new Date("2013-03-23 12:00"), new Date("2013-04-03 0:00")]);
-    let xAxis = new Plottable.Axes.Time(xScale, "bottom");
-    xAxis.margin(0);
+  describe("calculating space", () => {
```
jtlan: Probably "layout"?

test/axes/timeAxisTests.ts
```diff
@@ -205,212 +181,258 @@ describe("TimeAxis", () => {
...
-      xAxis.computeLayout({ x: 0, y: 0}, SVG_WIDTH, SVG_HEIGHT);
-      let coreHeight = xAxis.tickLabelPadding() + xAxis.innerTickLength();
-      let annotationHeight = xAxis.annotationTierCount() * (<any> xAxis)._annotationTierHeight();
-      let minimumHeight = coreHeight + xAxis.margin() + annotationHeight;
-      assert.operator(xAxis.height(), ">=", minimumHeight, "height includes all relevant pieces");
-      xAxis.destroy();
       svg.remove();
+      axis.destroy();
+
+      function getStyleInIE9(element: Element) {
```
jtlan: I think D3's `selection.style()` will produce the correct result.

jtlan:
Why are the "top" and "bottom" versions of the tests for tick positioning not symmetric?

test/axes/timeAxisTests.ts
```diff
@@ -1,416 +1,386 @@
...
-          for (let j = i + 1; j < numLabels; j++) {
-            clientRect1 = (<Element> visibleTickLabels[0][i]).getBoundingClientRect();
-            clientRect2 = (<Element> visibleTickLabels[0][j]).getBoundingClientRect();
-
-            assert.isFalse(Plottable.Utils.DOM.clientRectsOverlap(clientRect1, clientRect2), "tick labels don't overlap");
-          }
-        }
-      }
+      // very large time span
+      assert.doesNotThrow(() => scale.domain([new Date(0, 0, 1, 0, 0, 0, 0), new Date(5000, 0, 1, 0, 0, 0, 0)]));
```
jtlan: This was changed from `50000`? Wasn't the point to check a large domain?

bluong: When initially inspecting the unit tests, I also am noticing unit test times.  As we are growing our unit test count, we have to consider the time it takes for a unit test to complete.

To my understanding, having this value be 50000 put this test at a somewhat large test time.  Thus, I chose a value that would still count as a "very large time span" (10000 years) while pushing the test timer down.

jtlan: Do we know what bug this test was intended to solve? It looks like it was added in #716, when the Time Axis was first added.

bluong: It seems to be just a stress error-checking test on the tick calculation algorithm.  That said, what do you think we should do here?

jtlan: 5000 should be sufficient, I think?

PR #2890, opened 2015-10-16
Add sinon to test framework

nmiyake:


jtlan:
Looks good, but I'm wondering if maybe we shouldn't check in `bower_components` any more. @bluong, thoughts?

bluong:
The one thing that feels incredibly odd to me is the fact that we are checking in library dependencies in the first place.  It seems to be completely counter to the idea of version-control and also note the `node_modules` is not checked in.

I think what should happen is we have a script that people run that will run `bower install` and `npm install` / setup tasks.  I tried using this model in https://github.com/palantir/plottable-moment (no script yet, since i'm growing less confident in `npm install` usage)

jtlan:
Let's remove `bower_components` in a separate commit then.

nmiyake:
Going to wait to merge until this week's release has been cut. @crmorford , do you know when we're going to cut release?

nmiyake:
@jtlan , @bluong , FWIW, other projects I work on do check in bower_components to ensure that everyone's on the same version (even though they don't check in node_modules).

bluong:
@nmiyake feel free to merge this in now.  the release has already been cut.

nmiyake:
Close #2890

PR #2894, opened 2015-10-17
use nested modules

bluong:
Based on the last answer here https://stackoverflow.com/questions/12737942/does-typescript-support-namespace , we can nest our modules such that

```typescript
module A {
  export module B {
  }
}
```

can be rewritten as

```typescript
module A.B {
}
```

This is somewhat exactly what we have been looking for in our files to prevent the annoying indentation.

It also exposes a few awkward pieces of organization when placing exported types and interfaces.

QE: This should not affect anything from a user perspective, mainly taking advantage of one aspect of Typescript.  Regression pass please

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1,371 +1,337 @@
-declare module Plottable {
-    module Utils {
-        module Math {
-            /**
-             * Checks if x is between a and b.
-             *
-             * @param {number} x The value to test if in range
-             * @param {number} a The beginning of the (inclusive) range
-             * @param {number} b The ending of the (inclusive) range
-             * @return {boolean} Whether x is in [a, b]
-             */
-            function inRange(x: number, a: number, b: number): boolean;
-            /**
-             * Clamps x to the range [min, max].
-             *
-             * @param {number} x The value to be clamped.
-             * @param {number} min The minimum value.
-             * @param {number} max The maximum value.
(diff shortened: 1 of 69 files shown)


jtlan:
Isn't the nested version more semantically correct though? I note the non-nested version injects extra instances of "`Plottable.`" where it wasn't required before.

bluong:
Both versions of Plottable mean exactly the same thing semantically.  The main difference is compactness in writing (which was something we wanted for a while).

Also I suspect that the new version changes is because the according type is not declared in the same module declaration.

jtlan:
I'm just concerned that this increases the size of our JS file while not providing benefits other than allowing us to fix our module indenting, which usually isn't something I'm worried about unless I'm making a new file.

bluong:
You can see in the diff that this increases the size of our JS file by 12 lines, which is not big at all considering that the JS File is on the order of 6000 lines.  The two things that I find to be most relevant is that:
* It directly places the source code in the exact module that it should reside in (No need to do `module Plottable { export module Foo {`) to be in the correct module context.  Just do `module Plottable.Foo` which is much more concise.)
* By directly placing the source code in the right module, it also exposes a messiness in the code where if we did this before, we would have spotted more immediately and prevented.  Did you know that `Interaction`s placed their callback type signatures in the top level parent namespace?  This seems incorrect.

PR #2895, opened 2015-10-17
Fix sinon integration.

nmiyake:


bower.json
```diff
@@ -45,6 +45,6 @@
     "jQuery": "2.1.0",
     "jquery.simulate": "1.2.0",
     "requirejs": "2.1.18",
-    "sinon": "1.16.1"
+    "sinon": "http://sinonjs.org/releases/sinon-1.16.1.js"
```
bluong: This looks quite a bit odd to me.  Why this change?

aicioara: Any particular reason why not going for 1.17.1 (latest)

nmiyake: @aicioara , using 1.16.1 here because the "definitelyTyped" definitions are still at 1.16.0 and I don't want to introduce any incompatibilities.

@bluong , going this way because otherwise we don't actually get the release file and sinon doesn't actually work in the unit tests :(. Went with this method based on what I found online at https://muffinresearch.co.uk/using-sinon-js-with-bower-why-is-sinon-foo-undefined/, but if there's a better way to do this then I'm definitely open to it.

PR #2898, opened 2015-10-18
[Testing Audit] Table tests refactors

### LINKED ISSUE #2624: [Testing Audit] Table

jtlan:
The `Table` tests are very old, and some of them didn't make sense any more. The biggest issue was they weren't "unit" enough.

`Table` test battery after this refactor:
screen shot 2015-10-19 at 2 50 50 pm

[Before the refactor](https://cloud.githubusercontent.com/assets/1440449/10567337/2edcba54-75b6-11e5-85a4-f9ff2879d582.png)

Summary of changes:
* Tests reorganized and named in "it <VERB>" form.
* Layout tests broken out into separate sets for rows and columns.
* `_iterateLayout()` tests removed (because `iterateLayout()` is a private method), and its scenarios distributed across the layout tests.
* Use `componentAt()` in tests instead of `<any>`-casting.

Close #2624.


test/components/tableTests.ts
```diff
@@ -1,288 +1,112 @@
...
-    it("adding null to a table cell should throw an error", () => {
-      let c1 = new Plottable.Component();
-      let t = new Plottable.Components.Table([[c1]]);
-
-      assert.throw(() => t.add(null, 0, 0), "Cannot add null to a table cell");
+  function assertTableRows(table: Plottable.Components.Table, expectedRows: Plottable.Component[][], message: string) {
+    expectedRows.forEach((row, rowIndex) => {
+      row.forEach((component, columnIndex) => {
+        assert.strictEqual(table.componentAt(rowIndex, columnIndex), component,
+          message + ` contains the correct entry at [${rowIndex}, ${columnIndex}]`);
```
bluong: You have part of the string templated, might as well include `message` in the template

test/components/tableTests.ts
```diff
@@ -1,288 +1,112 @@
...
+      });
     });
+  }
 
-    it("add()-ing a Component to the Table should detach() it from its current location", () => {
-      let c1 = new Plottable.Component;
-      let svg = TestMethods.generateSVG();
-      c1.renderTo(svg);
+  describe("constructor behavior", () => {
+    it("adds the correct CSS class", () => {
```
bluong: When you say `correct`, do you mean that it adds the "table" CSS class?

test/components/tableTests.ts
```diff
@@ -1,288 +1,112 @@
...
 
-    it("add() works even if a component is added with a high column and low row index", () => {
-      // Solves #180, a weird bug
-      let t = new Plottable.Components.Table();
-      let svg = TestMethods.generateSVG();
-      t.add(new Plottable.Component(), 1, 0);
-      t.add(new Plottable.Component(), 0, 2);
-      t.renderTo(svg); // would throw an error without the fix (tested);
-      svg.remove();
+    it("can take a list of lists of Components in the constructor", () => {
```
bluong: Reword to "array of arrays"?

test/components/tableTests.ts
```diff
@@ -1,288 +1,112 @@
...
-    assert.deepEqual(translates[0], [0, 0], "first element is centered at origin");
-    assert.deepEqual(translates[1], [200, 0], "second element is located properly");
-    assert.deepEqual(translates[2], [0, 200], "third element is located properly");
-    assert.deepEqual(translates[3], [200, 200], "fourth element is located properly");
-    let bboxes = elements.map((e) => Plottable.Utils.DOM.elementBBox(e));
-    bboxes.forEach((b) => {
-      assert.strictEqual(b.width, 200, "bbox is 200 pixels wide");
-      assert.strictEqual(b.height, 200, "bbox is 200 pixels tall");
+  describe("checking Table contents", () => {
+    describe("has()", () => {
```
bluong: Remove the parens from the describe (or else cannot grep)

bluong: Is there any value to having a nested describe block here rather than either just an `it` or just a single `describe`?

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
     let c6 = new Plottable.Component();
     let table: Plottable.Components.Table;
-    it("works in basic case", () => {
+
+    it("removes the specified Component", () => {
       table = new Plottable.Components.Table([[c1, c2], [c3, c4], [c5, c6]]);
       table.remove(c4);
-      assert.deepEqual((<any> table)._rows, [[c1, c2], [c3, null], [c5, c6]], "remove one element");
+      assertTableRows(table, [[c1, c2], [c3, null], [c5, c6]], "the requested element was removed");
+      assert.isNull(c4.parent(), "Component disconnected from the Table");
```
bluong: factor out the variables?

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
     });
 
     it("does nothing when component is not found", () => {
       table = new Plottable.Components.Table([[c1, c2], [c3, c4]]);
       table.remove(c5);
-
-      assert.deepEqual((<any> table)._rows, [[c1, c2], [c3, c4]], "remove nonexistent component");
+      assertTableRows(table, [[c1, c2], [c3, c4]], "removing a nonexistent Component does not affect the table");
```
bluong: factor out the variables?

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
     });
 
-    it("removing component twice should have same effect as removing it once", () => {
+    it("has no further effect when called a second time with the same Component", () => {
       table = new Plottable.Components.Table([[c1, c2, c3], [c4, c5, c6]]);
 
-      table.remove(c1);
-      assert.deepEqual((<any> table)._rows, [[null, c2, c3], [c4, c5, c6]], "item twice");
+      let expectedRows = [[null, c2, c3], [c4, c5, c6]];
```
bluong: Can we factor out the variables and have `expectedRows` depend on the input?  Main concern is that let's say that someone were to add a new row to the input just to check behavior.  Then the test would fail unless that person changes `expectedRows` as well.

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+      const fixedNullsTable = new Plottable.Components.Table([
+        [new Mocks.FixedSizeComponent(), null],
+        [null, new Mocks.FixedSizeComponent()]
+      ]);
+      assert.isTrue(fixedNullsTable.fixedWidth(), "width is fixed if all columns contain fixed-width Components or null");
+
+      const someUnfixedTable = new Plottable.Components.Table([
+        [new Mocks.FixedSizeComponent(), new Mocks.FixedSizeComponent()],
+        [new Plottable.Component(), null]
+      ]);
```
bluong: The second column doesn't seem necessary?

test/components/tableTests.ts
```diff
@@ -1,288 +1,112 @@
...
-      let c2 = new Plottable.Component();
-      let c3 = new Plottable.Component();
-      let c4 = new Mocks.FixedSizeComponent(20, 10);
-      let table = new Plottable.Components.Table([
-        [c1, c2],
-        [c3, c4]
-      ]);
-      let result = (<any> table)._iterateLayout(500, 500);
-      verifyLayoutResult(result, [215, 215], [220, 220], [50, 20], [50, 10], false, false, "");
+  describe("add()", () => {
```
bluong: Remove parens?

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
       ]);
+      assert.isTrue(fixedNullsTable.fixedHeight(), "height is fixed if all columns contain fixed-height Components or null");
 
-      assert.strictEqual(table.componentAt(0, 0), c00, "retrieves the Component at [0, 0]");
-      assert.strictEqual(table.componentAt(0, 1), c01, "retrieves the Component at [0, 1]");
-      assert.strictEqual(table.componentAt(1, 0), c10, "retrieves the Component at [1, 0]");
-      assert.strictEqual(table.componentAt(1, 1), c11, "retrieves the Component at [1, 1]");
+      const someUnfixedTable = new Plottable.Components.Table([
+        [new Mocks.FixedSizeComponent(), new Plottable.Component()],
+        [new Mocks.FixedSizeComponent(), null]
```
bluong: Second row doesn't seem necessary?

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+      ]);
+      assert.isFalse(someUnfixedTable.fixedHeight(), "height is not fixed if any row contains a non-fixed-height Component");
     });
 
-    it("returns null when no Component exists at the specified row, column index", () => {
-      let c00 = new Plottable.Component();
-      let c11 = new Plottable.Component();
+    it("requests enough space for largest fixed-size Component in each row and column", () => {
+      const BIG_COMPONENT_SIZE = 50;
+      const SMALL_COMPONENT_SIZE = 20;
```
bluong: Unsure why this is cased like this

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+          TestMethods.assertPointsClose(component.origin(), expectedOrigin, 0, `Component at [${r}, ${c}] has the correct origin`);
+          expectedOrigin.x += component.width() + columnPadding;
+          maxHeight = maxHeight > component.height() ? maxHeight : component.height();
+        });
+        expectedOrigin.y += maxHeight + rowPadding;
+      });
+    }
+
+    it("divides available width evenly between non-fixed-width Components", () => {
+      const svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
```
bluong: Curious why we need this specific `SVG_WIDTH` and `SVG_HEIGHT`.  Since there is a default, I personally think it would be better to rely on the default behavior unless you absolutely need to take control over the width and height of the svg (constrained cases for example)

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+      verifyOrigins(tableRows);
+
+      table.destroy();
+      svg.remove();
+    });
+
+    it("gives width to fixed-width Components, then divides remainder between non-fixed-width Components", () => {
+      const svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+
+      const FIXED_COMPONENT_SIZE = 50;
```
bluong: Curious on casing

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+        "first non-fixed-width Component received half the remaining width");
+      assert.strictEqual(unfixedComponent2.width(), expectedUnfixedWidth,
+        "second non-fixed-width Component received half the remaining width");
+
+      table.destroy();
+      svg.remove();
+    });
+
+    it("divides available height evenly between non-fixed-height Components", () => {
+      const svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
```
bluong: Same as above.

test/components/tableTests.ts
```diff
@@ -294,77 +118,346 @@ describe("Tables", () => {
...
+      verifyOrigins(tableRows);
+
+      table.destroy();
+      svg.remove();
+    });
+
+    it("gives height to fixed-height Components, then divides remainder between non-fixed-height Components", () => {
+      const svg = TestMethods.generateSVG(SVG_WIDTH, SVG_HEIGHT);
+
+      const FIXED_COMPONENT_SIZE = 50;
```
bluong: Same as above.

bluong:
Three overarching main things that I can think of to fix:
* Remove the parens from the `describe` blocks.  I remember hearing that mocha does not work well with this style due to grep issues (From my standpoint, it also does not work when trying to use plain english).
* `SVG_WIDTH`, `SVG_HEIGHT` variables seem to be clutter in our test framework.  Currently our way of creating an SVG allows to have one of a default width and height.  Instead of making up this width / height on the fly, I think it's fine if we just use the default.  The assumption here is that the default width and height are non-constraining to the Components.  If a test wants to have them be constrained, then override at this point.
* Casing on variable names.  So far, I am unsure why certain local variables are camel cased and why certain ones are all uppercase.  Clarify here?

test/components/tableTests.ts
jtlan: Yes, although I'm wondering if that makes it fragile to potential changes to the class name (not that we'd be doing that).

test/components/tableTests.ts
jtlan: Which ones?

test/components/tableTests.ts
jtlan: Proving that `null` doesn't change the behavior.

test/components/tableTests.ts
jtlan: I was thinking I should do that, but I'm not sure how to name this `describe()` block then. Do I just do so without calling out the name of the method itself?

test/components/tableTests.ts
jtlan: It proves that having any one row non-fixed is sufficient to make the `Table` non-fixed, even if other rows are fixed.

test/components/tableTests.ts
jtlan: We use similar naming for things like `SVG_WIDTH`, so I was keeping with that.

test/components/tableTests.ts
jtlan: These values are divisible by 3, which keeps rounding/math errors from breaking the test.

bluong: This sounds that the width / height of the svg should be overridden in this test rather than have it be globalized.

test/components/tableTests.ts
bluong: I'm thinking of having a variable for the input array `[[c1, c2], [c3, c4], [c5, c6]]` and then have another variable that relates the input to the expected value `[[c1, c2], [c3, null], [c5, c6]]`

test/components/tableTests.ts
```diff
@@ -1,292 +1,112 @@
...
+      });
     });
+  }
 
-    it("add()-ing a Component to the Table should detach() it from its current location", () => {
-      let c1 = new Plottable.Component;
-      let svg = TestMethods.generateSVG();
-      c1.renderTo(svg);
+  describe("constructor behavior", () => {
+    it("adds the \"table\" CSS class", () => {
```
bluong: curious on does this `grep` correctly?

bluong: append "by default"?

test/components/tableTests.ts
bluong: Oh I just noticed that the variables are localized to the describe block which is a lot better.  Since you want the width / height to be divisible by 3, can you mark that in the variable name?
`divisibleByThreeWidth` maybe?

bluong: Would it make more sense to have the width / height be set per test, using some form of multiplier with the number of the rows and columns to prevent rounding errors?  The way it is now, if either `SVG_WIDTH` or if the number of rows change in a test, then the test would break unexpectedly.

test/components/tableTests.ts
jtlan: It's the `constructor`, so I think "by default" is probably unnecessary.

test/components/tableTests.ts
```diff
@@ -1,370 +1,472 @@
...
-      [c0, c1],
-      [c2, c3]
-    ]);
+  describe("removing Components", () => {
+    const c1 = new Plottable.Component();
+    const c2 = new Plottable.Component();
+    const c3 = new Plottable.Component();
+    const c4 = new Plottable.Component();
+    const c5 = new Plottable.Component();
+    const c6 = new Plottable.Component();
```
bluong: I am suspicious about defining these variables up here.  The problem is that if a test in the suite affects `c1`, then that affects any tests depending on `c1`

test/components/tableTests.ts
```diff
@@ -1,370 +1,472 @@
...
+    let table: Plottable.Components.Table;
```
bluong: Would it make more sense to wrap the initial setting of `table` in a `beforeEach()` or unset the setting of `table` in an `afterEach()`?  The concern here is that if a test did not reset the `table` variable to a new `table`, then they will use another test's `table`

test/components/tableTests.ts
```diff
@@ -1,370 +1,472 @@
...
 
-    it("iterateLayout works in the easy case where there is plenty of space and everything is satisfied on first go", () => {
-      let c1 = new Mocks.FixedSizeComponent(50, 50);
-      let c2 = new Plottable.Component();
-      let c3 = new Plottable.Component();
-      let c4 = new Mocks.FixedSizeComponent(20, 10);
-      let table = new Plottable.Components.Table([
-        [c1, c2],
-        [c3, c4]
+      const someUnfixedTable = new Plottable.Components.Table([
```
bluong: Rename to `unfixedTable`?  The `some` prefix sounds a bit arbitrary to me

test/components/tableTests.ts
```diff
@@ -1,370 +1,472 @@
...
-        [c1, c2],
-        [c3, c4]
+    it("is fixed-height if all rows contain only fixed-height Components, non-fixed otherwise", () => {
+      const fixedNullsTable = new Plottable.Components.Table([
+        [new Mocks.FixedSizeComponent(), null],
+        [null, new Mocks.FixedSizeComponent()]
+      ]);
+      assert.isTrue(fixedNullsTable.fixedHeight(), "height is fixed if all columns contain fixed-height Components or null");
+
+      const someUnfixedTable = new Plottable.Components.Table([
```
bluong: Rename to `unfixedTable`?  The `some` prefix sounds a bit arbitrary to me

test/components/tableTests.ts
jtlan: Only some, but not all of the columns are unfixed. Should I name it `someUnfixedColumnsTable`?

test/components/tableTests.ts
jtlan: I'll isolate them.

PR #2902, opened 2015-10-20
add title to swatches in interpolatedColorLegend

### LINKED ISSUE #2887: [InterpolatedColorLegend] Add <title>s to swatches
On `Legend`, there are `<title>` tags on the text so the text is still readable even if it gets cut off. We should add `<title>` tags to the swatches on `InterpolatedColorLegend` so the values associated with each swatch can be read.

Note that this should take `Configs.ADD_TITLE_ELEMENTS` into account.

ztsai:
add title to each swatch in `interpolatedColorLegend` if `Configs.ADD_TITLE_ELEMENTS` is true
demo: http://jsfiddle.net/ztsai/r3ozqxre/ http://jsfiddle.net/ztsai/r3ozqxre/3/
closes #2887

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -211,6 +211,7 @@ svg.plottable {
   fill: none;
   stroke: #CCC;
   stroke-width: 1px;
+  pointer-events: visiblestroke;
 }
 
 .plottable .waterfall-plot line.connector {
--- a/plottable.css
+++ b/plottable.css
@@ -211,7 +211,7 @@ svg.plottable {
   fill: none;
   stroke: #CCC;
   stroke-width: 1px;
-  pointer-events: visiblestroke;
+  pointer-events: none;
 }
 
 .plottable .waterfall-plot line.connector {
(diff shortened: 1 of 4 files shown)


jtlan:
Can we use the `Formatter` on the `InterpolatedColorLegend` to format the `<title>`s as well?

plottable.css
```diff
@@ -211,6 +211,7 @@ svg.plottable {
   fill: none;
   stroke: #CCC;
   stroke-width: 1px;
+  pointer-events: visiblestroke;
```
jtlan: Why is this change required?

ztsai: `.swatch-bounding-box` is drawn on top of swatches for outline, and without this css attribute, hover won't toggle tooltip

jtlan: Wouldn't `pointer-events: none;` be more semantically correct then?

crmorford:
tested on erry browser but not mobile because <title> doesn't work on mobile

Issue #1988, opened 2015-05-07
[Future] render(true) on ComponentContainer doesn't render children immediately


jtlan:
This bug currently happens with `_doRender()`, but I'm tagging it [Future] since it refers to the post-refactor API.

aicioara:
@jtlan Is this still the case?

PR #2904, opened 2015-10-20
[Component] - onRender callbacks

### LINKED ISSUE #1755: Adding beforeRender/afterRender hooks to the component
It would help a lot on the extensibility front to have these hooks for users that want to do stuff after a plot is rendered.

bluong:
A decently requested feature is the ability to take actions only after the DOM elements have been fully rendered.  Thus, in the vein of `onDetach()`, `onRender()` has been created to allow users to be able to do this.

API Changes:

``` typescript
/**
 * Adds a callback to be called on rendering the Component.
 *
 * @param {ComponentCallback} callback
 * @return {Component}
 */
onRender(callback: ComponentCallback): Component;
/**
 * Removes a callback that would be called on rendering the Component.
 * The callback is identified by reference equality.
 *
 * @param {ComponentCallback} callback
 * @return {Component}
 */
offRender(callback: ComponentCallback): Component;
protected _renderImmediately(): void;
protected _deferredTotalDrawTime(): number;
```

Fixes #1755


jtlan:
How does this interact with animations on `Plot`s?

bluong:
Due to animations using `setTimeout` and the nature of `setTimeout`, the callback will run during the animation (but after the raw elements are rendered on screen).

src/components/component.ts
```diff
@@ -236,9 +237,19 @@ export class Component {
     if (this._clipPathEnabled) {
       this._updateClipPath();
     }
+    this._renderImmediately();
```
ztsai: the 3 lines above were not being called in `Axis` and its children, and now they are. 
was it a conscious decision?

src/plots/plot.ts
```diff
@@ -452,6 +455,13 @@ export class Plot extends Component {
     this._additionalPaint(maxTime);
   }
 
+  protected _deferredTotalDrawTime() {
+    let drawSteps = this._paintedDrawSteps;
```
ztsai: why is this cached the not the `dataToDraw` and `drawers`?

bluong: The problem is that we need the draw steps in order to determine the time it takes to draw.  However, once a plot calls `_rendersImmediately()` then we cannot retrieve the time it takes to draw (Look at the `_getAnimator()` endpoint for more details).

This is one specific part that I feel a bit iffy about but I couldn't think of an alternative solution.

src/components/component.ts
bluong: I think I meant for the above 3 lines to actually exist in `_renderImmediately`  I will adjust

src/components/component.ts
```diff
@@ -233,10 +234,19 @@ export class Component {
    * Renders the Component without waiting for the next frame.
    */
   public renderImmediately() {
+    Utils.Window.setTimeout(() => this._onRenderCallbacks.callCallbacks(this), this._deferredTotalDrawTime());
```
ztsai: did you test if this does happen after all animation?

bluong: Testing for something happening after animation is decently difficult, mainly because we are forced to wait some timeout before the test can succeed.  There is definitely a way to accomplish this, however, this means that the test with the timeout is forced to consume the timeout.

This means that regardless of if the test were to succeed or fail, that test will take up valuable milliseconds to seconds of time.

ztsai: yeah, I don't mean adding an unit test, but I would just want to know if it's verified that this works as setTimeout is tricky

ztsai:
@crmorford, just a QE note, `onRender` should happen **after** animation is done for plots, please be sure to test with animations

crmorford:
Testing Plan:

- Table
- Group
-Plot
-Axis http://jsfiddle.net/qm03dkhf/
- call onRender() twice with two different callbacks
- test offRender() 
-Test for after animation 
-- maybe append something in each point's location
-- two plots with different animate speeds: put onRender on plot1, plot2, group
- Anonymous function http://jsfiddle.net/qm03dkhf/
- pass in function

crmorford:
http://jsfiddle.net/qm03dkhf/
In this, the alert from the callback pops up before the axis is actually showing up. I thought "rendered" meant that it was actually showing up on the svg.

crmorford:
callback isn't called when onRender is called on a table. http://jsfiddle.net/qm03dkhf/2/

jtlan:
>callback isn't called when onRender is called on a table. http://jsfiddle.net/qm03dkhf/2/

edit: group either http://jsfiddle.net/qm03dkhf/3/

... Is it because of #1988? or #2771?

bluong:
The latter is due to a mispositioning in the code.  I will remedy

bluong:
@crmorford For `Group` and `Table` do you expect the callbacks to be called once after all the components inside render?

crmorford:
I would expect the components inside group/table to render before the group/table is considered "done rendering"

jazzido:
FYI, I merged this PR into my fork. It worked perfectly for my use cases:

### Place labels on a `StackedBar` inside a `Table`
```javascript
        var chart = new Plottable.Components.Table([
            [yAxis, plot],
            [null, xAxis]
        ]);

        chart.onRender(function() {
          placeTotalLabels();
        });
```

### Place labels on a TreeMap implemented with `RectanglePlot`

```javascript
        var xScale = new Plottable.Scales.Linear(),
            yScale = new Plottable.Scales.Linear(),
            plot = new Plottable.Plots.Rectangle()
                .x(function(d) { return d.x; }, xScale)
                .y(function(d) { return d.y; }, yScale)
                .x2(function(d) { return d.x + d.dx; }, xScale)
                .y2(function(d) { return d.y + d.dy; }, yScale)
              .attr('fill', function(d) { return d._metadata.series.color; })
              .animated(true)
              .attr('stroke', '#fafafa');

        plot.onRender(function() {
          placeLabels(plot);
        });
```

jtlan:
>I would expect the components inside group/table to render before the group/table is considered "done rendering"

So effectively "`onRenderFinish`".

@jazzido, does the `chart.onRender()` call work as expected when animation is enabled on the `plot`?

jazzido:
> @jazzido, does the chart.onRender() call work as expected when animation is enabled on the plot?
We're not using animations, so I'm not sure.

MaestroJurko:
Is this feature already included in 2.0 rc2 version?

CalvinFernandez:
Currently without time to work on this feature. We can revisit another time

PR #2905, opened 2015-10-20
[Gitignore / Travis] - Ignore bower components

bluong:
The reason why `node_modules` is ignored while `bower_components` seems to be because Travis will willingly run `npm install` when we push our stuff to Travis.  Thus, if have Travis run `bower install`, we do not necessarily need `bower_components` to be checked in either.

Reference: http://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle


jtlan:
Any changes in workflow for developers?

bluong:
This should require no changes in workflow.  We just need to be more dutifully responsible when changing the `bower_components` in terms of communication.

jtlan:
As in, people will have to run `bower_install` in the future if packages are updated/changed?

bluong:
Yeap

jtlan:
This looks good, but I think we should check with similar projects here. Consider it a conditional +1.

bluong:
After discussion with other collaborators, it seems that removing the `bower_components` directory from the repo runs into 2 problems:
* Developers need to be kept up to date with this directory.
* Build systems need to import `bower_components` which can be flaky.

To conclude, we should not move forward with this until we have a really compelling reason for this.

PR #2907, opened 2015-10-21
Changes to animation framework

softwords:
Revised design for animators to take advantage of support for object constancy and produce richer animations.


softwords:
updated 11 November 2015 to reflect implementation using animateJoin (see #2930 )

**JoinResult**

A key change in this PR is to expose the D3 "enter / update / exit" pattern to the animators.
This introduces the type `JoinResult`, which has properties `enter` `exit` `update` and `merge` corresponding to these selections. `merge` is the combination of update and enter - in d3, once the `enter().append()` method is called on the bound data, that bound data is automatically updated to include the entering elements. `merge` is therefore in one-to-one correspondence with the incoming data.
The JoinResult  is created by Drawer in `_bindSelectionData`. It is then passed to the animators, by extending the `IAnimator` interface with a new method:

``` javascript
  animateJoin(joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
```

Note that `_bindSelectionData` no longer calls `.remove()` on the exit selection - otherwise you still could not do anything with it! `remove()`ing the `exit` selection is now a responsibility of the animator; see below.

**Initializer**

Usually, in the d3 "enter update exit" pattern, there is some initialization of attributes when the `enter()` elements are created; e.g.:

```
    selection.enter()
      .append(_svgElement)
      .attr(....)
```

`initializer` is a new property on Drawer that takes a function `()=> AttributeToProjector`. Drawer also gets the method `appliedInitializer(): AttributeToAppliedProjector` which is the result of the initializer function converted to an `AttributeToAppliedProjector`. 
This is applied to the `enter()` elements to initialize their state. 
The `initializer` is passed by the plot in its `_createDrawer` method, so the `initializer` may have access to the internals of the calling plot. In fact, a number of plots already have the code needed for the initializer - it's precisely the code in the `_animateOnNextRender` block. So this refactor does away with the pesky `_animateOnNextRender` logic, and makes the initialization of new elements more "mainstream" d3.

**Responsibilities of Animators**

Animators now get a lot of opportunities to work in detail with the "enter update exit" pattern, and in particular, to apply transitions to the `enter` and `exit` selections. Animators have two responsibilities: 
-- `remove()` the exit selection. 
-- update the `JoinResult` when transitions are applied.
For example, if the Drawer derives a transition from `joinResult.enter`, it should update `joinResult.enter` with that transition:

``` javascript
joinResult.enter = this.getTransition(joinResult.enter, 1000);
```

The point of this is that a subsequent transition derived from `joinResult.enter` will be a "subtransition", and (if you are careful not to interfere) d3 will handle scheduling this subtransition to chain from the first (see [here](https://github.com/mbostock/d3/wiki/Transitions#transition)).  This is one way to create more complex staged animations.

**Easing Functions**

This PR introduces some custom easing functions:
`atStart` - everything happens immediately at the start of the transition. ie this creates a "wait" after applying the attributes.
`atEnd` - everything happens in one go at the end of the transition - ie this creates a "wait" before applying the attributes.
`squEase` - (pronounced "squeeze") this applies the transition squeezed into a subinterval [a,b] where 0 <= a <= b <= 1. The actual easing to apply in that interval is passed as an argument. 
These are designed to help animators schedule animations between the enter update and exit selections - each can have a transition of the same duration, but the actual timing of activity will be determined by the easing function.

**Animator Hierarchy**

There are a number of new animator classes:
`Base` - basically equivalent to the current Easing, but with various protected helper methods available to subclasses.
`Attr` - apply a set of attributes to enter coming in - and to exit going out. 
`Opacity` - derived from Attr, fades incoming/outgoing in and out.
`Callback` - gets the actual `animate` function as a property. This allows the javascript client full access to the "enter update exit" selections. I think this is in keeping with the philosophy that plottable will always provide a way to get directly to d3 for occasions when you need something not provided in higher level code.  
`Reset` - applies the `Drawer`'s `appliedInitializer` to the entire `merge` selection. (this is a key reason for `Drawer` passing itself to the animator).
`Bar` - a staged animation specific to bar charts, that exits, then moves, then enters elements in turn. Demonstrates multiple transitions, custom easing functions, and the various helper functions in `Animators.Base`.

**Legacy behaviour**

Current animation resets all elements to a "start" state, and animates them back to their final state. To get this to happen in the new architecture, the `Drawer`'s `appliedIinitializer` has to be applied to all elements, not just the `enter()`ing elements. Two ways to do this:
-- use the `noConstancy` key. With `noConstancy`, every element is always in `enter`, so it gets initialised.
-- use the `Reset` animator. 

To support legacy custom animators, `Drawer` will invoke the `animate` method on any passed-in animator that does not have an implementation of `animateJoin`.  the animation is applied to `joinResult.merge` (see http://jsfiddle.net/oaskqta0/27/)

**Quicktests**

See the animation.html in the quicktests/overlay folder. Click on any of these animations to toggle between two datasets. The data items in these datasets have a `name` property, a new utility function `makeRandomNamedData()` generates names with some degree of overlap - so you get `enter` `update` and `exit` selections to observe. The dropdown bar lets you choose between key functions - useProperty("name") (ie constancy), noConstancy, and useIndex.

See also [bar chart](http://jsfiddle.net/oaskqta0/27/) and [scatter chart](http://jsfiddle.net/oaskqta0/28/) fiddles.

**//# sourceURL**

The new quicktests use the annotation
`//# sourceURL=<url>`
as the last line of the file. When chrome loads the file dynamically, the sourceURL annotation causes it to appear in the Source page list of files, which makes it much easier to examine and debug. I found it's better placed at the end rather than the start, because chrome removes the line, and thereafter seems to be one line out when you are stepping through the code.

softwords:
@bluong Perhaps you could merge develop into objectConstancy branch, which will make this PR much smaller?

bluong:
@softwords Can you merge / rebase from `objectConstancy`?  It seems that merge conflicts are preventing us from seeing the true diff?

bluong:
Also adding JSFiddles and the added API endpoints to the description would be greatly helpful.

softwords:
>> Can you merge / rebase from objectConstancy? It seems that merge conflicts are preventing us from seeing the true diff?

The branch from which I created the PR is up-to-date with develop, so that's where all the file changes are coming from. Wouldn't these go away if you merge develop into objectConstancy on your side?

bluong:
I have merged develop into objectConstancy it seems.  When you remerge objectConstancy to your branch on your end, what happens?

bluong:
Can you add a decently brief description of the items that are enabled by this specific pull request and JSFiddles and API changes brought upon by this specific pull request?

softwords:
@crmorford 's ealier fiddle : http://jsfiddle.net/oaskqta0/15/
with this pr now works:
http://jsfiddle.net/oaskqta0/18/

softwords:
This fiddle exercises the Reset Opacity and Bar animators, and also shows how the Easing Animator will behave in new ways with ObjectConstancy in place
http://jsfiddle.net/oaskqta0/19/

softwords:
Brief Summary:

Adds a new Animator, ```Bar``` for a customised staged animation for Bar Charts.
New Animator, ```Opacity```, respects object constancy by fading new elements in and old elements out.
```Easing``` animator will respect object constancy by transitioning updated elements from their original state to their new state.
```Reset``` animator preserves the legacy behaviour of Bar, Clustered Bar, Stacked Bar animations.,

**Comment by crmorford** [comment]
Hi @softwords!
I can tell from the jsfiddles that this is gonna be awesome :) I'm still having a little trouble digesting the changes, though.

**Is there any complete list of the public API points which have been added?**

More specific questions:
1. I'm not sure if the `Opacity` animator being added will work with `Easing`, or if they are mutually exclusive.
2. I'm also not quite sure how to set these. Do I set them by passing in the 'Opacity' or 'Reset' string to some function? Do they live somewhere in the Plottable namespace (Plottable.Animators.Reset, or something like that?)
3. What is the .rhythm method on Bar animators? Are there any other methods added, that aren't shown in this fiddle?

**Comment by softwords** [comment]
hello @crmorford 
>>this is gonna be awesome :) 

thanks! I hope it's turns out to be a worthwhile contribution. 
I will get on to summarising a list of API points today - meantime to answer your questions:
In plottable currently you can choose between two pre-built animators - ```Null``` or ```Easing```.  This PR adds a few more, they are all in the Plottable.Animators namespace: 
```Reset``` ```Bar``` ```Opacity``` are used in this fiddle.
There's also these new animators too (  they are more aimed at helping either ts or js users create custom animations, let's get back to those in a bit) 
```Attr``` and ```Callback```
All these inherit from a new
```Plottable.Animators.Base``` animator, which is not really intended to use on its own, it is basically the same as ```Easing```, but provides a lot of facilities the others use.

You assign them to a plot in the usual way; ie 
- create the animator
- configure its properties
- assign to the plot

``` javascript
       barAnimator = new Plottable.Animators.Bar()
            .stepDuration(2100)
            .yScale(yScale)
            .xScale(xScale)
        	.rhythm(.3);
    
        verticalBarPlot = new Plottable.Plots.Bar("vertical")
            .addDataset(ds)
            .animator(Plottable.Plots.Animator.MAIN, barAnimator)
```
So in that sense the animators are mutually exclusive, you assign one to your plot.

you can see them getting instantiated in the fiddle in the ```changeAnimator``` function
``` javascript
function changeAnimator() {
    
    var choice = d3.select("#animation").node().value;
    var animator;
    switch (choice) {
        case "reset":
            animator = new Plottable.Animators.Reset()
            	.stepDuration(500);
            break;
        case "opacity":
            var ine =  Plottable.Animators.EasingFunctions.squEase("linear", .5, 1);
            var oute =  Plottable.Animators.EasingFunctions.squEase("linear", 0,.5);
            animator = new Plottable.Animators.Opacity()
            	.stepDuration(2000)
            	.stepDelay(0)
            	.startOpacity(0)
            	.endOpacity(0)
            	.easingMode(ine)
            	.exitEasingMode(oute)
            break;
        case "easing":
            animator = new Plottable.Animators.Easing()
            	.stepDuration(500);
           break;
        case "bar":
        	animator = barAnimator;
    }
     verticalBarPlot.animator(Plottable.Plots.Animator.MAIN, animator);

    changedata(0); //replay
}
```
so the string is just a convenience to let you pick one from the select list.

```rhythm()``` is probably best understood by changing the values; try 0, 1, .1, .9 

If you watch the Bar animator carefully ( and maybe slow it down) you'll see that there is a "pause" between each step and the next - e.g. after the exiting bar shrink away, there is a short wait, then the updating bars resize - wait - then they move - wait - then the entering bars come in.
rhythm controls the relative length of these pauses, which creates a sense of jerkiness or smoothness in the animation. The smaller the value of ```rhythm```, the longer these pauses. 

```Bar``` also has ```xScale``` and ```yScale``` properties that it needs to share with the Plot (so it can figure out where y = 0 is )

**Comment by softwords** [comment]
Opacity animator API

```startOpacity``` - opacity to immediately apply to incoming elements - defaults to 0

```endOpacity``` - opacity to transition exiting elements to before they are removed - defaults to 0

You can also set ```startOpacity``` and ```endOpacity``` with the constructor

``` javascript
animator = new Plottable.Animators.Opacity(.1, .2);
```

easing modes

One key change in these new animators is that they can animate the entering, exiting and updating elements independently, as you can see clearly in the Opacity animator. These animations (d3 transitions) can potentially each have their own easing.

currently
```easingMode```   applies to the entering and updated elements
```exitEasingMode``` applies to the exiting elements

The main purpose for doing this is to schedule these animations to happen sequentially , rather than simultaneously. 

To do this, you create the easing function using the new in this PR 
```Plottable.EasingFunctions.squEase()``` method

``` javascript
            var ine =  Plottable.Animators.EasingFunctions.squEase("linear", .5, 1);
            var oute =  Plottable.Animators.EasingFunctions.squEase("linear", 0, .5);
            animator = new Plottable.Animators.Opacity()
            	.stepDuration(2000)
            	.stepDelay(0)
            	.startOpacity(0)
            	.endOpacity(0)
            	.easingMode(ine)
            	.exitEasingMode(oute)
```

squEase takes 3 arguments:
- a d3 easing function name
- a start position between 0 and 1
- end position, between 0 and 1

The animation is squeezed into the segment of the duration between start and end.
So in the code above, the easing function ```oute``` applies a linear easing in the first half of the duration, ```ine``` applies the linear easing in the second half of the duration.
with duration = 2000, you see the exiting elements fade in between t= 0 sec and t = 1 sec, then the updated elements move, and the incoming elements fad in, between t = 1 sec and t=2 sec

Try different start/end values in creating ```oute``` and ```ine``` and you'll see the effect. You could try even using a different d3 easing function as the first argument, the change will be subtle. The choices are listed [here](https://github.com/mbostock/d3/wiki/Transitions#d3_ease)
Notice that if start = end, the changes happens instantaneously at the specified point. In particular, if start = end = 0 the animation is applied immediately; if start=end=1 its applied instantly at the end of the duration.
you can get the same effects with the new items
```Plottable.EasingFunctions.atStart```
```Plottable.EasingFunctions.atEnd```

e.g. try
``` javascript
            var ine =  Plottable.Animators.EasingFunctions.atEnd
            var oute =  Plottable.Animators.EasingFunctions.atStart;
            animator = new Plottable.Animators.Opacity()
            	.stepDuration(2000)
            	.stepDelay(0)
            	.startOpacity(0)
            	.endOpacity(0)
            	.easingMode(ine)
            	.exitEasingMode(oute)
```

softwords:
Reset Animator API

```Reset``` animator sets all existing elements to an "initial" state, then animates them to their final position. This is the current behaviour of plots such as ```Bar``` and ```Scatter```. 
```Reset``` takes the same API as ```Easing```, in particular you can set ```stepDuration``` and ```stepDelay``` as currently. ```Reset``` will give the current behaviour regardless of the setting for constancy on the ```Dataset```; ```Easing``` will give the current behaviour only when ```noContancy``` is set on the ```Dataset```. These are the two ways to get the legacy behaviour, see my note above.

**Comment by softwords** [comment]
@crmorford also wanted to mention that I put quite a few quicktests based on different plot types in this PR, they are accessed from animation.html in quicktests/overlay. Are you able to see these from the PR? (I see now it would have been better to build these as fiddles...)

**Comment by bluong** [comment]
@softwords Just as an FYI.

While this pull request looks like it provides a bunch of sweet features, this pull request is just incredibly too large to parse.  I'm first going to review it so that I can extract the key pieces of context I need, and then it is incredibly likely that I'll ask you to split this pull request into pieces.  Is that alright with you?

**Comment by softwords** [comment]
@bluong , sure - I thought that may be the case. As I said before, putting this all up in one go at least lets you all see what it is about in its entirety, so you can decide how you want to proceed based on the bigger picture. There are some quite low-level changes in here, and I fully appreciate that you need to manage the risk when receiving that from an external contributor. 
I imagine you are also going to have to think about how to internalise this into your organisation - I notice you give great support to plottable through your gitter, and when this stuff gets released you'll probably get questions about it (like the questions @crmorford is raising now) . I would flag in advance I won't be putting my hand up to deal with any of that!
As a starting point for how this could be divided / reduced:
-- lot of the 48 files changed are actually in the quicktests - these could be discarded or held over? 
-- there's some administrative things:
--- .eslintrc this relates to the quicktests
--- .gitattributes this is the ultimate solution to my long-standing problem with plottable.d.ts
--- coverage.html --- inserts the polyfill for phantomjs .... but there may be a better way to get around this.

So the substantive changes are
-- Drawer.ts and Animator.ts - changes to the Animator interface and to Drawer are the crux of the design change.
-- changes to Null and Easing to accommodate the change
-- new animators
-- changes to plots - these are refactors to remove animateOnNextRender

That may be a starting point for getting to grips with what's in here...

**Comment by softwords** [comment]
Attr animator API

```Attr``` lets you specify the starting state for incoming elements, and the ending state for outgoing elements, as a set of attributes. 

The Size animation in [this fiddle](http://jsfiddle.net/oaskqta0/20/) uses ```Attr```

``` javascript
        case "size":
            // using the Attr animator
            
            // this is a circle size 16: M0,8A8,8 0 1,1 0,-8A8,8 0 1,1 0,8Z
        	// make a size 0 circle that can get animated
        	var proj = { d: "M0,0A0,0 0 1,1 0,-0A0,0 0 1,1 0,0Z" };
        	animator = new Plottable.Animators.Attr()
            	.stepDuration(1000)
        		.stepDelay(0)
        		.startAttrs(proj) // inisitlaise incoming elements
        		.endAttrs(proj)  // target of transition for outgoing elements
                .easingMode(ine)
            	.exitEasingMode(oute);
```

softwords:
@bluong - a further thought on "risk management" - because there is more work going on in Drawer._bindSelectionData, and because objectConstancy may mean more svg elements getting destroyed and recreated, it would be a good idea to test this PR against any very large datasets you have to flush out any adverse performance effects early...

softwords:
Callback Animator API

see http://jsfiddle.net/oaskqta0/21/

Callback allows the client to write their own animate handler, allowing full access to the d3 "enter update exit" pattern.

The ```callback()``` property sets the function to be used to do the animation. The function must have this signature:
```animate(selection, attrs, drawingTarget, drawer) ```

where:
```selection``` - not used in this context
```attrs``` - set of attributes (AttributeToAppliedProjector) to be applied
```drawingTarget``` - the drawingTarget object exposes the ```enter``` ```update``` ```exit``` and ```merge``` (ie ```enter + update```) selections.
```drawer``` - the calling drawer object

In the callback function, ```this``` is the ```Callback``` animator, so the callback function can use the methods exposed by ```Callback``` (inherited from ```Base```).

In particular, the callback may use ```this.getTransition()``` to derive a d3 transition from any of the drawingTarget selections.

```Base.getTransition(selection, duration, delay, easing)``` takes these arguments:
```selection``` the d3 selection or transition to dervie from
```duration``` the duration of the new transition - if 0, the selection is returned as-is.
```delay```  the step delay of the new transition. Note this is ignored if deriving a "subtransition" from a transition, since d3 will automatically calculate the delay to make the transitions sequential
```easing``` the easing function for the new transition. If not supplied, uses the ```easingMode()``` of the ```Callback``` animator

If the callback derives a transition from any of the selections in ```drawingTarget```, it must update ```drawingTarget``` with that transition. This ensures that any subsequent transition derived from that selection will be a "subtransition", and therefore scheduled sequentially by d3.

The callback function must also ensure that the ```exit``` selection is removed.

Example: this is the callback animate function used in the Explode! option in the [fiddle]( http://jsfiddle.net/oaskqta0/21/) . This sets exiting elements to a color from the category10 palette, then swells them and fades them. Incoming elements are initialized to a circle of size 0, then all elements are transitioned (using ```merge```) to their final state. The ```ine``` and ```oute``` squEase easing functions delay processing of the merge elements, and cause some overlap between the exit and merge processes.
Try with _NoConstancy_ option (causing all elements to exit with each refresh) for full pyrotechnic effect.

``` javascript
function cbAnimate(selection, attrs, drawingTarget, drawer) {
    var colors = d3.scale.category10();
    var proj = { d: "M0,48A48,48 0 1,1 0,-48A48,48 0 1,1 0,48Z",
                    opacity: 0
                   	};
    var ine =  Plottable.Animators.EasingFunctions.squEase("linear", .5, 1);
    var oute =  Plottable.Animators.EasingFunctions.squEase("linear", 0, .75);   
    var zeroCircle = "M0,0A0,0 0 1,1 0,-0A0,0 0 1,1 0,0Z";

   
    drawingTarget.exit
    	.attr("fill", function(d,i) {return colors(i); });
    drawingTarget.exit = this.getTransition(drawingTarget.exit, this.stepDuration(), this.stepDelay(), oute )
    	.attr(proj)
    	.remove();
    drawingTarget.enter
    	.attr({d: zeroCircle});
    drawingTarget.merge = this.getTransition(drawingTarget.merge, 
                                             this.stepDuration(), this.stepDelay(), ine)
    	.attr(attrs);

}
```

.gitattributes
```diff
@@ -0,0 +1 @@
+﻿*.d.ts text
```
bluong, line 1: Can your source a webpage or similar that says that this is indeed the fix?

bluong, line 1: Also if you could pull this single idea into a separate pull request into develop, I think that'd be one small step to breaking up this pull request.

bluong:
I defer if we want to add these quicktest changes to @crmorford .  In terms of splitting this pull request, can you make a pull request to develop with the quicktest additions (however let's stay away from the new animator changes if we can).

These quicktests seem useful to have just in general regardless of the happenings in this branch, so we can extract that piece out and then we can modify those quicktests in a later pull request.

src/animators/animator.ts
```diff
@@ -1,5 +1,39 @@
 module Plottable {
+export module Animators {
```
bluong: Typescript allows for declaring nested modules at the top level

Change to `module Plottable.Animators`

src/animators/animator.ts
```diff
@@ -1,5 +1,39 @@
...
+          return 0;
+        }
+        let tbar = (t - start) / (end - start);
+        if (typeof (easingFunction) === "string") {
+          easingFunction = d3.ease(<string>easingFunction);
+        }
+        return (<EasingFunction>easingFunction)(tbar);
+      };
+    }
+  }
```
bluong, line 38: All of these ideas seem to be relevant for the `easing` idea in `EasingAnimator` is this the case?

src/reference.ts
```diff
@@ -0,0 +1,100 @@
...
+/// <reference path="interactions/doubleClickInteraction.ts" />
+/// <reference path="interactions/keyInteraction.ts" />
+/// <reference path="interactions/pointerInteraction.ts" />
+/// <reference path="interactions/panZoomInteraction.ts" />
+/// <reference path="interactions/dragInteraction.ts" />
+
+/// <reference path="components/dragBoxLayer.ts" />
+/// <reference path="components/xDragBoxLayer.ts" />
+/// <reference path="components/yDragBoxLayer.ts" />
+/// <reference path="components/dragLineLayer.ts" />
```
bluong: This file is no longer needed

bluong:
As a heads up, I believe it is decently correct for each class addition to be a separate pull request.  This might amount to up to 5-7 pull requests, but I think this is the best way for us to iterate quickly with as least pain as possible.

From what I gather, most of the pull request is in these `Animator` classes, switching to use these `Animator`s on the appropriate `Plot`s, and finally an interesting change to `Drawer`.

With this I have a few questions before we split this up effectively:
* What exactly is an `Animator`?  Currently our `EasingAnimator` essentially takes a start and end state and uses an easing function to get from start to finish.  Thus, the current idea is a way to get from a start to an end state.  How does your ideas fit into this idea?  (When I see an `AttrAnimator` / `BarAnimator`, this somewhat does not fit into this idea.)  I am incredibly flexible in changing our idea of an `Animator` since it is ill-defined in Plottable currently, so I want to get your take on it.
* Can you find the one change that hallmarks your change the best?  Once we get acquainted with a single, solid idea, then I feel that we can iterate quicker on the other ideas.  From here we can start a pull request to get this one `Animator` through and then we can move from there. (The idea of a "minimum viable product" is incredibly useful here).

softwords:
>> What exactly is an ```Animator```? Currently our ```EasingAnimator``` essentially takes a start and end state and uses an easing function to get from start to finish. 

I'd be more general:

An _animator_ is responsible for moving the state of a selection to a new state reflecting new data. Each animator encapsulates a strategy for effecting this change.

Rather than _uses an easing function_, I'd characterise the current ```Easing``` animator behaviour as
 _uses a single d3 transition applied uniformly to all data-bound svg elements_. This is a special case of potential animators, in that only one transition is used. The broader definition allows for multiple transitions  used in sequence, and different transitions used on different selections.

>> If ```BarAnimator``` was meant to be an ```Animator``` that is incredibly specific to ```Bar`` ...

Yes it is, because it specifically works with x, y and height attributes. It will give reasonable results on ```ClusteredBar``` and ```StackedBar``` too.
 
>> Can you find the one change that hallmarks your change the best? 

Definitely! That change is:

_When effecting the change from one state to another, ```Animator```s can now access the "enter update exit" selections produced by data binding._

The implementation that makes this happen is
-- the introduction of the new type ```DrawingTarget``` that encapsulates the "enter update exit" selections, 
-- the construction of this object when data binding takes place, and 
-- having this object passed to each ```Animator``` when it is called.

src/animators/animator.ts
softwords, line 38: Yes, these EasingFunctions could be used with the `Easing` animator. An easing function is just a function e(t) from [0,1] to [0,1] that specifies how much of the animation is applied at time t.
d3 has a whole lot of these known by name, but more generally, you should be able to use a predefined one by name, or pass a function. 
This is why I added the types `EasingFunction` and `EasingFunctionSpecifier`.

bluong:
Ok @softwords I think I might enough context to run with.  Can you take one of your basic `Animator`s that can be immediately usable (not `BarAnimator` as I think that one is pretty heavily involved) and put that in a pull request with the pieces required to have that work?

Also, can you make a pull request for each piece that is required to have the above pull request work?  Essentially, one pull request for this new type `DrawingTarget` in its usage, one pull request for the `Animator` interface change / `EasingAnimator` change, one pull request for the `Drawer` change.  Does this make sense?  I essentially want to break this up as much as I can, while having a single new `Animator` implementation to rely on to see how all the pieces fit.

Again, thanks so much for being patient with us.  We are trying to iterate as quickly as possible so that we can include your changes.

.gitattributes
softwords, line 1: The clue was here:

http://stackoverflow.com/questions/6855712/why-does-git-treat-this-text-file-as-a-binary-file

If git thinks a file is a text file, it will convert the line ends to \n when storing, and convert back to the operating system default when extracting. Under Windows, the file gets created with \r\n line ends by the compiler. git doesn't know it's text so it stores those \r\n untranslated in the repository - hence every line is a change.

the git attirbutes entry tell git that *.d.ts is in fact a text file, so on storing, it translates the line ends back to \n like it should.

src/animators/animator.ts
softwords: This one doesn't quite correspond to the pattern, it's like Drawer.ts with some elements in Animators and one in Plottable.
I have done all the new classes this way though.

src/reference.ts
softwords: one down!...

src/animators/animator.ts
bluong: I'm more thinking of adding these types after the `Animator` interface so that essentially we have

``` typescript
module Plottable {
  export interface Animator {
  }
}

module Plottable.Animators {
  export stuff {
  }
}
```

I am more inclined in this way since I feel that it exposes namespaces a lot better.

src/animators/animator.ts
```diff
@@ -1,5 +1,39 @@
 module Plottable {
+export module Animators {
+  export type d3SelectionOrTransition = d3.Selection<any>|d3.Transition<any>;
 
+  export type EasingFunction = (t: number) => number;
+  export type EasingFunctionSpecifier = string|EasingFunction;
```
bluong: Instead of `Plottable` needing to export / declare these types, would it make more sense to submit a pull request to DefinitelyTyped for D3 to expose this?

crmorford:
>I defer if we want to add these quicktest changes to @crmorford . In terms of splitting this pull request, can you make a pull request to develop with the quicktest additions (however let's stay away from the new animator changes if we can).

>These quicktests seem useful to have just in general regardless of the happenings in this branch, so we can extract that piece out and then we can modify those quicktests in a later pull request.

Hi @softwords!
First, thank you for writing quicktests! So helpful :)

I agree with @bluong that it makes sense to push in the tests separately from the code changes. We could get those in first, and then I'll have them to look at as I look at the rest of the pull requests. We could even base that PR on develop and get them up for general use ASAP.

bluong:
I'm looking through your pull request again, and I'm just stating things to be more clear.  Let's take a concrete, simple `Animator` that people will use (Maybe `Opacity` or `Reset`) and then use this single `Animator` to be the start of a new, smaller pull request.  In accordance, only pieces relevant to that specific `Animator` need to be in that pull request (which might require some changes to `Drawer` and some changes to `Animator`).

With this I expect a pull request to something in the vein of only ("drawer.ts" changes, "animator.ts" changes, possibly "baseAnimator.ts"/"attrAnimator.ts"?, and your specific new animator changes)

softwords:
good morning @crmorford 
the quicktests are demonstrating the new animators, so I'm not sure how useful they will be to bring in in advance of the code changes. 
For more demonstration of what's in the API of the new animators I've put up a few more in the fiddle:
http://jsfiddle.net/oaskqta0/23/

So I'm not sure how you intend to move from this PR (or subsequent series of PR), into the Side Branch and then into develop - how do you see that working?

bluong:
Essentially there are 2 paths with the side branch that I currently see will work nicely for us.  One is to make a pull request into develop.  There are already a bunch of pieces here that are useful inherently without the need of the animators (world cup data? yes please!), and so even if we pull the `Animator` usage, I think there would be some use in develop.

The second is to make a pull request into objectConstancy.  With this, all the above applies except that we can keep the idea regarding KeyFunctions (which currently have no user impact).

In essence, yes without the animators, the full story of the quicktest is not well shown, but we can at least show pieces of it separately from the additional features you're providing.

softwords:
OK, well I'm chipping bits off into separate PRs on develop - so as I see it, when those get pulled, you merge develop into objectConstancy, and the number of files affected by the PR reduces.

softwords:
>> With this I expect a pull request to something in the vein of only "drawer.ts" changes, "animator.ts" changes, possibly "baseAnimator.ts"/"attrAnimator.ts"?, and your specific new animator changes

I'd suggest:

drawer.ts
animator.ts
nullAnimator.ts
easingAnimator.ts

Is the minimal start point, because once Animator interface changes Null and Easing need to change too.

Stage 2 would add :
Base
Attr
Reset
Opacity

bear in mind that even getting this far there won't be anything to see because of the animateOnNextRender in each plot type preventing any "proper" animation from happening

Stage 3:
barPlot
scatterPlot
areaPlot
linePlot

these are the refactors to get rid of animateOnNextRender - so now stuff will work as you see in the current PR.
At this point, there will be a need to think about defaults - ie when the code says:
``` javascript
dataset.data(dataset.data);
```
what do you _want_ to happen? (My opinion - nothing!)

When you have all this in the sidebranch - it can move to develop and provide some real new functionality

Stage 4: can probably be added straight into develop
Bar animator
Callback animator

>> and so even if we pull the Animator usage, I think there would be some use in develop

Can I ask then, that we take a step back, and you spend a bit of time at your end looking at the samples, the code, the documentation in the code and all notes in here, discussing in your team and decide whether you do want to incorporate this feature into plottable. 
I can work with you to go forward, but I need to be conscious of pressures on my time, and if you don't want to proceed, that's perfectly fine - I won't be offended :), but it would be better to know sooner rather than later!

bluong:
> I can work with you to go forward, but I need to be conscious of pressures on my time, and if you don't want to proceed, that's perfectly fine - I won't be offended :), but it would be better to know sooner rather than later!

As we are moving forward with this pull request, I want to be quite clear that I am getting more and more interested in pulling this into develop.  The statements made about having Animations be more relevant in Plottable definitely makes sense to me; I was just mainly curious in how everything fits in and also how developers will be able to read the code going forward.

The stages that you have described definitely seem to make sense to me.  Please pull the coverage / quicktest / other files into different pull requests  or revert and make your next pull request with Stage 1.  I think I have enough context to be able to review that somewhat expediently at this point.

bluong:
> and so even if we pull the Animator usage, I think there would be some use in develop

When I say this, I say this with essentially the worst case scenarios (I believe that this pull request will not run into the worst case scenario but better to be safe).

Pull requests should be kept as small and unitized as possible so that, if there is a problem with one pull request out of three, then we can still merge the other 2.  This is a lot better than the awkward scenario of having a massive pull request, where if there is a problem with any small piece of it, then the entirety cannot be merged.

Thus for this pull request, the absolute worst case scenario is that we cannot find a compromise in this new `Animator` architecture and we need to reject the pull request.  In this scenario, we probably still want to acquire the pieces that are still usable.  At this moment, I am near 100% positive this will not happen.  I think what you are providing here is definitely of value and somewhat nicely fits into our architecture.

crmorford:
You guys move too fast!
>the quicktests are demonstrating the new animators, so I'm not sure how useful they will be to bring in in advance of the code changes. 
@softwords the tests just won't run if they aren't ready, but they won't affect other tests. so we can just put them in now, and they'll start working when the API points are there :)

bluong:
I'm mainly leaving a comment here to show what API changes will occur in this pull request.  Collaborators want to take a look at the proposed API at an easy glance

```typescript
module Plottable.Drawers {
  /**
   * A JoinResult contains the enter, update and exit selections that are produced when
   * joining data to a collection of elements using the d3 data() method
   * (see http://bost.ocks.org/mike/join/ for detailed explanation)
   * JoinResult is contructed by Drawer, and passed to animators to allow animation of each selection
   */
  type JoinResult = {
    enter: d3.Selection<any> | d3.Transition<any>;
    update: d3.selection.Update<any> | d3.Transition<any>;
    exit: d3.Selection<any> | d3.Transition<any>;
    merge: d3.selection.Update<any> | d3.Transition<any>;
  };

  /**
   * Retieves a function that can supply initial settings to entering elements.
   * this function is typically supplied by the plot using the Drawer
   */
  initializer(): () => AttributeToProjector;
  /**
   * Sets the initializer function for the Drawer.
   * This function returns an AttributeToProjector that is applied
   * to the new elements appended to the enter() selection
   * Typically set from _createDrawer in the plot
   *
   * @param {() => AttributeToProjector} the function.
   * @returns {Drawer} The calling Drawer.
   */
  initializer(fnattrToProjector: () => AttributeToProjector): Drawer;
  appliedInitializer(): {
    [attr: string]: (datum: any, index: number) => any;
  };
}
module Plottable.Animators {
  type d3SelectionOrTransition = d3.Selection<any> | d3.Transition<any>;
  type EasingFunction = (t: number) => number;
  type EasingFunctionSpecifier = string | EasingFunction;
  /**
   * Custom easing functions. these are designed to co-ordinate the timing of
   * actvities on two or more transitions that are active simulataneously
   * (on different selections)
   * These are designed to help animators schedule animations between
   * the enter, update and exit selections - each can have a transition of the same duration,
   * but the actual timing of activity will be determined by the easing function.
   */
  class EasingFunctions {
    /**
     * atStart
     * An EasingFunction that applies all attributes at once at the start of the transition;
     * ie this creates a "wait" after applying the attributes.
     */
    static atStart: (t: number) => number;
    /**
     * atEnd
     * An EasingFunction that applies all attributes at once at the end of the transition;
     * ie this creates a "wait" before applying the attributes.
     */
    static atEnd: (t: number) => number;
    /**
     * squEase
     * squEase ("squeezing ease") applies the transition squeezed into a subinterval [a,b] where 0 <= a <= b <= 1.
     * The actual easing to apply in that interval is passed as an argument.
     * @param easingFunction {EasingFunctionSpecifier} an easing function, or the string descriptor
     * of a d3 built-in easing function ( as used by d3.ease())
     * @param start {number} value between 0 and 1 at which the animation starts
     * @param end  {end} value between start and 1 at which the animation ends
     * Note that if start == end, the are applied instanteously at that point and the
     * easingFunction parameter is not relevant. In particular start == end == 0 is the
     * same as atStart
     * @returns {EasingFunction} An easing function that can be applied to a d3 transition.
     */
    static squEase(easingFunction: EasingFunctionSpecifier, start: number, end: number): EasingFunction;
  }
}
module Plottable {
  interface Animator {
    /**
     * Applies attributes to the 'enter update exit' selections created when binding new data into
     * the selection of DOM elements bound to the plot's current data (if any).
     * (see http://bost.ocks.org/mike/join/)
     * This method is invoked on the animator assigned to the plot by Drawer when drawing the plot data.
     *
     * @param {JoinResult} joinResult JoinResult object - its enter, update, exit
     *     and merge (enter + update) properties provide access to the selections
     *     created by the data join.
     * @param {AttributeToAppliedProjector} attrToAppliedProjector The set of
     *     AppliedProjectors that we will use to set attributes on the selection.
     * @param {Drawer} drawer the calling Drawer.
     * An animator that derives a transition from any property of joinResult must update that
     * property with the resulting transition object.
     * Animators are responsible for removing exiting elements by ensuring that remove() is called on joinResult.exit
     */
    animateJoin(joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
  }
}
```

bluong:
The next pieces have not been looked at from me, so there will be pieces will need discussion.  I feel we can discuss these pieces after we narrow down the above API.

```typescript
declare module Plottable.Animators {
  /**
   * Base class for animators. Equivalent behaviour to Easing animator
   * Provides helper functions for subclassed animators
   */
  class Base implements Animator {
    /**
     * The default starting delay of the animation in milliseconds
     */
    private static _DEFAULT_START_DELAY_MILLISECONDS;
    /**
     * The default duration of one animation step in milliseconds
     */
    private static _DEFAULT_STEP_DURATION_MILLISECONDS;
    /**
     * The default maximum start delay between each step of an animation
     */
    private static _DEFAULT_ITERATIVE_DELAY_MILLISECONDS;
    /**
     * The default maximum total animation duration
     */
    private static _DEFAULT_MAX_TOTAL_DURATION_MILLISECONDS;
    /**
     * The default easing of the animation
     */
    private static _DEFAULT_EASING_MODE;
    private _startDelay;
    private _stepDuration;
    private _stepDelay;
    private _maxTotalDuration;
    private _easingMode;
    private _xScale;
    private _yScale;
    /**
     * Constructs the default animator
     *
     * @constructor
     */
    constructor();
    totalTime(numberOfSteps: number): number;
    animateJoin(joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
    animate(selection: d3.Selection<any>, attrToAppliedProjector: AttributeToAppliedProjector): d3.Selection<any> | d3.Transition<any>;
    /**
     * return a transition from the selection, with the requested duration
     * and (possibly) delay. As a convenience, this may return the selection itself
     * without applying a transition if duration = 0
     * If the selection passed in is a transition a subtransition is created. This subtransition
     * will use the default delay calculated by d3, so any delay passed to this function is ignored
     * @param { d3.Selection<any>|d3.Transition<any>|d3.selection.Update<any>} the d3 selection or transition
     * @param { number } duration The duration required for the transition. If 0, no transition is created
     * @param { number? } delay The delay to apply to the transition. If creating a subtransition, this is ignored.
     * @param { EasingFunctionSpecifier } easing An easing function, or the name of a predefined d3 easing function
     * to use on the transition. If not supplied, the easingMode of the calling Animator is used.
     * @return { any } The transition created, or , when duration = 0 the original selection is returned.
     */
    protected getTransition(selection: d3.Selection<any> | d3.Transition<any> | d3.selection.Update<any>, duration: number, delay?: (d: any, i: number) => number, easing?: EasingFunctionSpecifier): any;
    /**
     *  return true if the d3 object passed in is a transition
     */
    protected isTransition(selection: any): boolean;
    protected mergeAttrs(attr1: AttributeToAppliedProjector, attr2: AttributeToAppliedProjector): AttributeToAppliedProjector;
    /**
     * Return the AtributeToAppliedProjector comprising only those attributes listed in names
     * An Animator may use this function to create an intermediate collection of attributes to
     * use in a transition.
     * @param {AttributeToAppliedProjector} attr
     * @param {string[]} names The names of the required attributes
     *
     * @returns {AttributeToAppliedProjector} A new collection, compresing only thos attribuest listed in names[].
     */
    protected pluckAttrs(attr: AttributeToAppliedProjector, names: string[]): AttributeToAppliedProjector;
    /**
     * Return a delay function, using the current startDelay and stepDelay,
     * potentially adjusted to fit within the maxTotalDuration
     * @param selection selection on which the transition will be based
     * @returns delay function that may be passed to a transition
     */
    protected delay(selection: any): (d: any, i: number) => number;
    /**
     * Gets the start delay of the animation in milliseconds.
     *
     * @returns {number} The current start delay.
     */
    startDelay(): number;
    /**
     * Sets the start delay of the animation in milliseconds.
     *
     * @param {number} startDelay The start delay in milliseconds.
     * @returns {Base} The calling Animator.
     */
    startDelay(startDelay: number): Base;
    /**
     * Gets the duration of one animation step in milliseconds.
     *
     * @returns {number} The current duration.
     */
    stepDuration(): number;
    /**
     * Sets the duration of one animation step in milliseconds.
     *
     * @param {number} stepDuration The duration in milliseconds.
     * @returns {Base} The calling Animator.
     */
    stepDuration(stepDuration: number): Base;
    /**
     * Gets the maximum start delay between animation steps in milliseconds.
     *
     * @returns {number} The current maximum iterative delay.
     */
    stepDelay(): number;
    /**
     * Sets the maximum start delay between animation steps in milliseconds.
     *
     * @param {number} stepDelay The maximum iterative delay in milliseconds.
     * @returns {Base} The calling Animator.
     */
    stepDelay(stepDelay: number): Base;
    /**
     * Gets the maximum total animation duration constraint in milliseconds.
     *
     * If the animation time would exceed the specified time, the duration of each step
     * and the delay between each step will be reduced until the animation fits within
     * the specified time.
     *
     * @returns {number} The current maximum total animation duration.
     */
    maxTotalDuration(): number;
    /**
     * Sets the maximum total animation duration constraint in miliseconds.
     *
     * If the animation time would exceed the specified time, the duration of each step
     * and the delay between each step will be reduced until the animation fits within
     * the specified time.
     *
     * @param {number} maxTotalDuration The maximum total animation duration in milliseconds.
     * @returns {Base} The calling Animator.
     */
    maxTotalDuration(maxTotalDuration: number): Base;
    /**
     * Gets the current easing mode of the animation.
     *
     * @returns {string} the current easing mode.
     */
    easingMode(): EasingFunctionSpecifier;
    /**
     * Sets the easing mode of the animation.
     *
     * @param {EasingFunctionSpecifier} easingMode The desired easing mode.
     *   Can be any valid easing function, or the name of a d3 built-in easing function
     * @returns {Base} The calling Animator.
     */
    easingMode(easingMode: EasingFunctionSpecifier): Base;
    /**
     * xScale -- a reference to the xScale used by the owning plot
     * the animator can use this to calculate positions
     * @returns {Plottable.Scale<any, any>} the xScale.
     */
    xScale(): Plottable.Scale<any, any>;
    /**
     * Sets the easing mode of the animation.
     *
     * @param {Plottable.Scale<any, any} xScale.
     * @returns {Base} The calling Animator.
     */
    xScale(xScale: Plottable.Scale<any, any>): Base;
    /**
     * yScale - a reference to the yScale used by the owning plot
     * the animator can use this to calculate positions
     * @returns {Plottable.Scale<any, any>} the yScale.
     */
    yScale(): Plottable.Scale<any, any>;
    /**
     * a yScale available to the animator for rendering 'logical' y values.
     *
     * @param {Plottable.Scale<any, any} yScale.
     * @returns {Base} The calling Animator.
     */
    yScale(yScale: Plottable.Scale<any, any>): Base;
    /**
     * Adjust the iterative delay, such that it takes into account the maxTotalDuration constraint
     */
    protected _getAdjustedIterativeDelay(numberOfSteps: number): number;
  }
}
```

bluong:
```typescript
declare module Plottable.Animators {
  /**
   * Base for animators that animate specific attributes, such as Opacity, height... .
   */
  class Attr extends Base implements Animator {
    /**
     * Gets the attributes for entering elements. These are overlaid over
     * the target atr and
     * @returns {number} The current start delay.
     */
    startAttrs(): AttributeToAppliedProjector;
    /**
     * Sets the attributes for entering elements.
     *
     * @param {startAttrs} A collection of attribuets applied to entering elements.
     * These are applied over the top of the attributes pass to the animate method
     * Any attribute passed to startAttrs will transition to its final value
     * @returns {Attr} The calling Attr Animator.
     */
    startAttrs(startAttrs: AttributeToAppliedProjector): Attr;
    /**
     * Gets the attributes for exiting elements.
     * Exisitng elements will transition to these attribuest before being removed
     * @returns {AttributeToAppliedProjector} The current exiting attributes.
     */
    endAttrs(): AttributeToAppliedProjector;
    /**
     * Sets the attributes for entering elements.
     *
     * @param {endAttrs} A collection of attribuets applied to entering elements.
     * These are applied over the top of the attributes pass to the animate method
     * Any attribute passed to endAttrs will transition to its final value
     * @returns {Attr} The calling Attr Animator.
     */
    endAttrs(endAttrs: AttributeToAppliedProjector): Attr;
    /**
     * gets the easing mode for exiting elements.
     * Existing elements will transition out using this easing mode before being removed
     * if not specified, easingMode() is used
     * @returns {EasingFunctionSpecifier} The current exiting Easing mode.
     */
    exitEasingMode(): EasingFunctionSpecifier;
    /**
     * easing mode for exiting elements
     *
     * @param {EasingFunctionSpecifier} The easing function, or d3 easing function name.
     * Animations may separate the activity on entering and easing elements by using
     * the squEase easing function over different intervals.
     *
     * @returns {Attr} The calling Attr Animator.
     */
    exitEasingMode(exitEasingMode: EasingFunctionSpecifier): Attr;
  }
}
```

bluong:
A very specific one-time `Animator` for `Plots.Bar`
```typescript
module Plottable.Animators {
  /**
   * A "staged" animation specific for bar charts
   * Animates exit, update and enter in sequence
   */
  class Bar extends Attr implements Animator {
    constructor();
    /**
     * Sets the relationship between activity and waits in the major steps in the animation (0 to 1)
     * Applied using a squEase easing function
     *
     * @returns {number} The current rhythm.
     */
    rhythm(): number;
    /**
     * @param {number} rhythm the rhythm to apply to the major steps in the animation.
     * @returns {Bar} The calling Animator.
     */
    rhythm(rhythm: number): Bar;
    /**
     * Specify whether to split the rendering of the Update selection into two stages
     *
     * @returns {boolean} The current setting.
     */
    stageUpdate(): boolean;
    /**
     * @param {boolean} stageUpdate split the Update into two stages
     * @returns {Bar} The calling Animator.
     */
    stageUpdate(stageUpdate: boolean): Bar;
  }
}
```

bluong:
A class to reset attributes
```typescript
module Plottable.Animators {
  class Reset extends Base implements Animator;
}
```

bluong:
An AttrAnimator to affect opacities
```typescript
module Plottable.Animators {
  /**
   * Fade in - fade out the entering - exiting elements by transitioning opacity
   */
  class Opacity extends Attr implements Animator {
    /**
     *  @Constructor
     *  @param {number} startOpacity the initial opacity for entering elements
     *                   defaults to 0
     *
     *  @param {number} endOpacity the final opacity for exiting elements
     *                   defaults to 0
     */
    constructor(startOpacity?: number, endOpacity?: number);
    /**
     * Sets the starting opacity for entering elements.
     *
     * @returns {number} The current start opacity.
     */
    startOpacity(): number;
    /**
     * Sets the starting opacity for entering elements.
     *
     * @param {number} startOpacity The opacity to apply to new elements.
     * @returns {Opacity} The calling Animator.
     */
    startOpacity(startOpacity: number): Opacity;
    /**
     * Sets the ending opacity for exiting elements.
     *
     * @returns {number} The current end opacity.
     */
    endOpacity(): number;
    /**
     * Sets the ending opacity for entering elements.
     *
     * @param {number} endOpacity The opacity to transition to exiting elements.
     * @returns {Opacity} The calling Animator.
     */
    endOpacity(endOpacity: number): Opacity;
  }
}
```

bluong:
A class to aide custom animation
```typescript
module Plottable.Animators {
  type AnimateJoinCallback = (joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer) => void;
  /**
    * Allows the implementation of animate to be passed as a callback function
    * Provides javascript clients that build custom animations
    * with full access to the drawing target and properties of the Drawer (appliedIntializer)
    */
  class Callback extends Base implements Animator {
    constructor();
    /**
     * Gets the callback animate function.
     * @returns {Callback} The calling Callback Animator.
     */
    callback(): AnimateJoinCallback;
    /**
     * Sets the attributes for entering elements.
     *
     * @param {AnimateCallback} a function implementing Animator.animate
     * @returns {Callback} The calling Callback Animator.
     */
    callback(callback: AnimateJoinCallback): Callback;
    /**
     * Gets an inner animator.
     * callback functions have access to this animator, so callbacks may be designed to
     * wrap another animator
     * @returns {innerAnimator} The current innerAnimator.
     */
    innerAnimator(): Animator;
    /**
     * Sets the attributes for entering elements.
     *
     * @param {Animator} an Animator
     * @returns {Callback} The calling Animator.
     */
    innerAnimator(innerAnimator: Animator): Attr;
    innerAnimate(joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
  }
}
```

softwords:
travis-ci timed out :( otherwise pleased to find it wasn't too hard to get a clean build after merging with develop @bluong  when you have a chance could you restart this, and merge develop into objectconstancy. thanks!

softwords:
Hello @adidahiya , I understand you have taken over ownership of plottable development, previously @bluong   @jtlan @crmorford were looking at this.
So I'm keen to know if you intend to look again at this material. I have maintained it up-to-date with the develop stream since October, but this is becoming less easy - and I notice various breaking changes you have in the works. If these are committed to develop before this pull request, I assume the PR would need further rework to keep it compatible.
One small thing you could do would be merge develop into obectconstancy branch - this will give a better idea of what is actually in this PR.

PR #2914, opened 2015-10-22
Force git to recognize .d.ts files as text

softwords:
If git thinks a file is a text file, it will convert the line ends to \n when storing, and convert back to the operating system default when extracting. Under Windows, the file gets created with \r\n line ends by the compiler. git doesn't know it's text so it stores those \r\n untranslated in the repository - hence every line is a change.

the git attributes entry tell git that *.d.ts is in fact a text file, so on storing, it translates the line ends back to \n like it should.


jtlan:
Thanks for this. Apologies for the delay.

PR #2915, opened 2015-10-22
Add utility function to create random data with name, x, and y fields

softwords:
Add utility function to create random data with name, x, and y fields. Name is selected at random from the set "A" - "Z"
and will be unique in the returned array.
Supports consideration of object constancy in quicktests
.eslintrc modified so this function may be used without declaration without raising an error

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@
     "d3": true,
     "Plottable": true,
     "makeRandomData": true,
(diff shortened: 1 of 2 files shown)


softwords:
Time-out in the tests?

bluong:
I restarted the tests.

bluong:
Deferring to @crmorford .  Thanks for starting to break up the initial pull request!

PR #2921, opened 2015-10-23
Animation Stage1

softwords:
Define the DrawingTarget type, construct the DrawingTarget when databinding in Drawer, pass the DrawingTarget through Animator.animate 

Expects all animations to continue to work as currently.


softwords:
Note on Animator.animate signature:

In the plottable code base, there is only one instance of an animator being called outside of ```Drawer``` - to render the Bar Plot baseline. To support this case, as well as to support any javascript clients that may be using this, I have added the two new arguments to ```animate```, rather than change them. 
Maybe this could be reconsidered in future.
So ```Null``` and ```Easing``` both consider both cases: 
-- receiving a ```selection``` like before
-- receiving a ```DrawingTarget``` like now

```Drawer``` is passed as a 4th argument, because it will give the animator access to some useful things: especially the ```appliedInitializer()```  (```Reset``` depends on this), and also the ```Dataset```, and particularly its metadata, which could come in handy. 

I considered making ```DrawingTarget``` a read-only public property of ```Drawer``` in which case the ```animate``` interface could omit the ```drawingTarget``` argument ( which could still be accessed as ```drawer.DrawingTarget()```), but opted for this formulation - one less public method.

**Comment by softwords** [comment]
DrawingTarget

the terminology ```merge``` - to represent the combined ```update``` and ```enter``` selections, I have adopted from [d3.chart](https://github.com/misoproject/d3.chart/wiki/available-lifecycle-events).

Note that as soon as you call ```selection.enter().append()``` those appended elements are added to the selection, converting it from ```update``` to ```merge```. This is why ```Drawer``` clones the ```update``` selection before it calls ```enter()``` - to preserve the set of elements which constitute ```update```.

File: src/animators/animator.ts [hunk 1]
```diff
@@ -1,4 +1,39 @@
 module Plottable {
+export module Animators {
+  export type d3SelectionOrTransition = d3.Selection<any>|d3.Transition<any>;
```
bluong: Is there a better name for the union here?

bluong: I think Typescript is slowly moving away from having the union types be bunched up and it also looks a bit cluttered.

Can you add a whitespace around `|`

src/animators/animator.ts
```diff
@@ -1,4 +1,39 @@
...
+
+  export type EasingFunction = (t: number) => number;
+  export type EasingFunctionSpecifier = string|EasingFunction;
```
bluong: I think Typescript is slowly moving away from having the union types be bunched up and it also looks a bit cluttered.

Can you add a whitespace around `|`

src/animators/animator.ts
```diff
@@ -1,4 +1,39 @@
...
+        if (t <= start) {
+          return 0;
+        }
+        let tbar = (t - start) / (end - start);
+        if (typeof (easingFunction) === "string") {
+          easingFunction = d3.ease(<string>easingFunction);
+        }
+        return (<EasingFunction>easingFunction)(tbar);
+      };
+    }
```
bluong: Can we have documentation for the above functions?

src/animators/animator.ts
```diff
@@ -1,4 +1,39 @@
...
+        if (t >= end) {
+            return 1;
+        }
+        if (t <= start) {
+          return 0;
+        }
+        let tbar = (t - start) / (end - start);
+        if (typeof (easingFunction) === "string") {
+          easingFunction = d3.ease(<string>easingFunction);
+        }
```
bluong: I believe there is a way to use Typescript's type guarding so that you do not have to cast

Can you try

```typescript
if (typeof easingFunction === "string") {
  easingFunction = d3.ease(<easingFunction);
}
```

src/animators/animator.ts
```diff
@@ -11,7 +46,8 @@ export interface Animator {
    *     transition object so that plots may chain the transitions between
    *     animators.
    */
-  animate(selection: d3.Selection<any>, attrToAppliedProjector: AttributeToAppliedProjector): d3.Selection<any> | d3.Transition<any>;
+  animate(selection: d3.Selection<any>|d3.Transition<any>, attrToAppliedProjector: AttributeToAppliedProjector,
+    drawingTarget?: Drawers.DrawingTarget, drawer?: Drawer): d3.Selection<any> | d3.Transition<any>;
```
bluong: This is specifically the part that I think I will have questions with, mainly because it seems now `Drawer`s are leaking into the `Animator` and because `animate` is being much more complicated.

src/animators/nullAnimator.ts
```diff
@@ -1,17 +1,26 @@
 module Plottable.Animators {
 
-  /**
-   * An animator implementation with no animation. The attributes are
-   * immediately set on the selection.
-   */
+   /**
+    * An animator implementation with no animation. The attributes are
+    * immediately set on the selection.
+    */
```
bluong: Fix indentation

src/drawers/drawer.ts
```diff
@@ -74,17 +119,29 @@ export class Drawer {
    * @param{any[]} data The data to be drawn
    */
   private _bindSelectionData(data: any[]) {
-    // if the dataset has a key, use it when binding the data   
+    // if the dataset has a key, use it when binding the data
     let dataElements: d3.selection.Update<any>;
+    let selection = this.renderArea().selectAll(this.selector());
```
bluong: What is incorrect about `this.selection()`?

src/drawers/drawer.ts
```diff
@@ -60,6 +74,37 @@ export class Drawer {
...
+   * @returns {Drawer} The calling Drawer.
+   */
+  public initializer(fnattrToProjector: () => AttributeToProjector): Drawer;
+  public initializer(fnattrToProjector?: () => AttributeToProjector): any {
+    if (fnattrToProjector == null) {
+      return this._initializer;
+    }
+    this._initializer = fnattrToProjector;
+    return this;
+  }
```
bluong: Mainly curious: why are we setting up the initial state here instead of using a null animator to set up the initial state at the plot level?

src/drawers/drawer.ts
```diff
@@ -60,6 +74,37 @@ export class Drawer {
...
+
+  /*
+   * Return the AttrToAppliedProjector generated from the initializer
+   *
+   * @returns {AttrToAppliedProjector} .
```
bluong: Remove the period here, and probably give mmore documentation

bluong:
This is a great start I think.  I can review this pretty expediently I hope and we should hopefully be able to merge this step in soon.

bluong:
If possible, can you modify your pull request description to describe the API Additions that this pull request has added, any change in behavior (We still have no new animators but can the behavior still be changed?), and a JSFiddle of any changes in behavior?

bluong:
In terms of the fact that baseline is being animated, I can do a query on my end to see if it is ok to remove the animation idea from baseline if this causes less API strain.  I believe that animations are never used for the baseline (note that it uses an `Animators.Null`) even though we might have thought about doing so (which is why some animation idea is attached to it).

In this case, can we change the API for `animate` or is it forced to stay the same?

The main problem I have with `animate` is that it has 2 optional arguments now which increases the argument count to 4, which is borderline a complicated endpoint.  (By the way, I think using overloads instead of optional arguments may make your signature more clear?).  Based on the above idea regarding baseline, is there a different idea that we can use?

bluong:
My main concerns are below.  As said before, I think this is great work being put here, and I mainly have a few questions and few coding concerns

* Please add documentation about the types surrounding the `Easing` idea.  Specifically the functions are not 100% clear to me what they are.  To be even more specific the name `squEase` is pretty unclear.  Is this an easing function that progresses at a quadratic rate?
* The `animate` API has gotten significantly more complicated.  I need to look this over a few more times to see if we can find a simpler API.  If it must be this complicated, then this is ok but I would much rather prefer a much more usable API.  I put some notes above regarding this.
* The rest of the concerns should be listed above.

Just want to again say thanks for sticking with us so far.  I think we finally have a workflow that works out for us, and hopefully we can process this quickly.

softwords:
>> To be even more specific the name squEase is pretty unclear. Is this an easing function that progresses at a quadratic rate?

```squEase``` I intended would be pronounced _squeeze_, which is what it does - squeeze the animation into a specified subset of the duration. Remember that the rate that it progresses within that interval is just determined by its first argument.

File: src/animators/animator.ts (hunk 4 above)
Comment (softwords) [inline src/animators/animator.ts]: interestingly, type gaurding doesn't apply in this case:

```javascript
easingFunction = d3.ease(easingFunction);
```
because the value of ```easingFunction``` gets modified

changing it to 
```javascript
return = d3.ease(easingFunction)(tbar);
```

does recognize ```easingFunction``` as a string

From the typescript language spec:

>> •	In the true expression of a conditional expression, the type of a variable or parameter is narrowed by a type guard in the condition when true, __provided the true expression contains no assignments to the variable or parameter__.

softwords:
>> The animate API has gotten significantly more complicated. I need to look this over a few more times to see if we can find a simpler API. If it must be this complicated, then this is ok but I would much rather prefer a much more usable API. I put some notes above regarding this.

I think this is a list of the possibilities:

```javascript
animate(Drawer, AttributeToAppliedProjector): void
```

FOR: smallest structure that works with the new format
AGAINST: doesn't support previous usage
strictly speaking, is this a semver change to 2.0.0?
requires DrawingTarget to be added as a public method of Drawer
consumers of this api will _always_ have to get drawingTarget out of Drawer as the first thing they do:

```javascript
let drawingTarget = drawer.DrawingTarget();
```

```javascript
animate(DrawingTarget, AttributeToAppliedProjector, Drawer?): void
```

FOR: friendliest format for new structure
Drawer doesn't need to be considered if not required 
doesn't required Drawer.DrawingTarget() method

AGAINST
doesn't support previous usage
strictly speaking, is this a semver change to 2.0.0?


```javascript
animate(DrawingTarget | d3.Selection<any>, AttributeToAppliedProjector, Drawer?): d3.Selection<any>
```

FOR: not a breaking change
AGAINST: bit more complicated
animators will have to test the type of the incoming variable (bear in mind d3.Selection is ultimately just an array)


```javascript
animate(d3.Selection<any>, AttributeToAppliedProjector, DrawingTarget?, Drawer?): d3.Selection<any>
```

FOR: Not a breaking change for existing users
Drawer can be ignored if not needed
old and new usage distinguished without a type check
AGAINST: longest signature

I don't think it is a bad thing to have an API point that allows a client to apply a transition to a random selection, theoretically there could be javascript clients doing that now. (as mentioned there is one such use in the plottable code base itself: bar lot baseline). 

So I'd propose - leave it like I have it for now, it doesn't break anything existing, and let's keep it as an open issue as we move forward with actual implementations when you will get a better feel for it.

src/drawers/drawer.ts
softwords: Because doing it at the plot level can't distinguish between entering and updated elements - so what happens now is that existing elements get overridden with initial state. This refactor applies this initial state only to entering elements:

```javascript
    this._drawingTarget.enter = dataElements.enter()
      .append(this._svgElementName)
      .attr(this.appliedInitializer());
```

softwords:
Confirming there are no behaviour changes in this version - you can verify with the quicktest sequence, in particular all the animation tests behave as before.

softwords:
@bluong can i request when you can you continue to merge develop into objectConstancy - I'll do the same with my Stage1 branch, and that may save us ugly surprises down the track!

bluong:
I can merge in develop on demand, unless you want me to merge in develop at a set pace.  Up to you.

bluong:
We also do have the option of overloading the `animate` API to have 2 completely different signatures.  One that uses the new `Animator` structure and one that does not.  As long as there is a third signature that is essentially an implementation signature, Typescript will allow for this (look at the getter / setter signatures as context)

bluong:
Note that we only have to move to 2.0 via semver if we do not support the API in earlier versions.  In this regard, overloading the signature will route around the different signature issue.

bluong:
Looking through the proposed API and the above points of utilizing overloading, I am interested in the first two ideas.

For the first one, what do you mean by

> consumers of this api will always have to get drawingTarget out of Drawer as the first thing they do:

It seems that a consumer does not need to interact with a `DrawingTarget` at all based on the API.  Can you elaborate?

For the second one, why do we need the `Drawer` argument?

softwords:
>> We also do have the option of overloading the animate API to have 2 completely different signatures.

That's a very good thought - I'll experiment with that...

src/animators/animator.ts
```diff
@@ -1,25 +1,91 @@
...
+        }
+        let tbar = (t - start) / (end - start);
+        if (typeof easingFunction === "string") {
+          return d3.ease(easingFunction)(tbar);
+        }
+        return (<EasingFunction>easingFunction)(tbar);
+      };
+    }
+  }
 }
```
bluong: The above looks like it should be moved to an easingFunctions.ts file / module

softwords:
>> It seems that a consumer does not need to interact with a DrawingTarget at all based on the API. Can you elaborate?

The ```drawingTarget``` is what holds the selections (enter, update, exit, merge) that we want the animator to act on. So any animator called by ```Drawer``` in this new architecture will, as a minimum :
get ```exit```, then remove() it
get ```merge``` and change it:

``` javascript
drawingTarget.exit
   .remove();
drawingTarget.merge
.attr(attrToAppliedProjector);
```

Have a look at what happens in ```nullAnimator``` for example. This is the minimal implementation of ```animate```  note that it is treating the special case where drawingTarget is not supplied just to support any legacy uses.

It may help to revisit the big pull request with everything in to get a better feel for this : 
take a look at the ```animate``` implementation in ```Base`` and ```Attr``` for example.

src/animators/animator.ts
softwords: agreed

softwords:
>> For the second one, why do we need the Drawer argument?

For immediate use, to get access to the ```appliedInitializer()``` This is the function that describes how the incoming elements are initialized - and ```Drawer``` will take care of that. But - if the animator has access to that, it can apply it to _all_ elements (ie ```merge```) - this is how the ```Reset``` animator provides the legacy behaviour. - see the big PR.

Having the ```appliedInitializer()``` is also useful to an animtor that wants to transition elements out - ie provide a transition on the ```exit``` selection. Instead of _starting_ at this state, the transition will _end_ at this state:

``` javascript
drawingTarget.exit = this.getTransition(drawingTarget.exit, 1000)
 .attr(drawer.appliedInitializer())
 .remove();
```

Because this ```appliedInitializer``` has been supplied originally by the plot, it can e.g. shrink bars down to the baseline - _knowing where that baseline is_. again, see the big PR.

So faced with the need to get ```appliedInitializer``` to the ```animator``` I chose to send the reference to the ```drawer``` - because it could be potentially useful to an elaborate customised animator to have access to the ```dataset``` , and even to the dataset's metadata.

In the big Pr there is also the ```Callback``` animator which will make all this available to the javascript clients, so... end-users could get creative with their custom animations.  see the Explode! sample in 
http://jsfiddle.net/oaskqta0/23/

bluong:
Mainly curious.  Is there a reason to have the final argument other than to have an `appliedInitializer`?

bluong:
> `animate(DrawingTarget, AttributeToAppliedProjector, Drawer?): void`

If we decide to go with this signature. I believe the `Drawer` argument cannot be optional.  If it was optional, then users of say an `Animators.Reset` might not accidentally pass in a `Drawer` and then they would error.  Instead, we should enforce that a `Drawer` is always going to be input and then implementations can just ignore the parameter.

softwords:
>>  If it was optional, then users of say an ```Animators.Reset``` might not accidentally pass in a Drawer and then they would error. 

Remember that, even now, end-users of an animator are not responsible for calling the ```animate``` method - it is called on their behalf by ```Drawer.drawStep()```  So they don't get a chance to mess up.

That said, with overloaded call signatures there is no reason for this to be optional - I was thinking about accommodating that one pesky call for the bar Plot baseline.

softwords:
Here is ```Animator``` with the overloaded ```animate``` signature . Technically this works and I've tried it out on the still-to-come ```Animator``` implementations as well. 
Each implementation has to figure out which version it is working with - it does that by testing if the first argument is an array (d3 selections are arrays decorated with various extra properties). 
So this works, but..... it feels like we are shoe-horning two methods into one. 
Perhaps the other alternative to preserve the semantic versioning is to introduce a whole new method into Animator. 

```javascript
animateJoin(drawingTarget: Drawers.DrawingTarget, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
```
_join_ is used here in the widely used d3 sense - see http://bost.ocks.org/mike/join/ , so this name is a good description of what it does. This is the function that would be called in ```drawer._drawStep()```:

```javascript
    step.animator.animateJoin(this._drawingTarget, step.attrToAppliedProjector, this);
```
The ```animate``` method can continue, but its uses would be rare - to apply a set of attributes to a d3 selection or transition.  So the use on the bar plot baseline would be quite valid, and javascript clients may continue to use this if they already do. 

Anyone that has built a custom animator would need to change that animator - you will have a better idea from user feedback as to how far clients push the API like this. But on the other hand, such (hypothetical) users will be big winners from these changes, because their scope to build animators will be very much improved.

So - this is now my preferred approach... What do you think?

PRO: 
- cleaner, more meaningful interface by separating two distinct signatures
- implementations are cleaner - do not need to test arguments to decide which signature has been used

AGAINST: 
- will break any custom animators built by end users ( are there any??)
- may have implications for tests?
- ```animate``` method is no longer the most "important" part of animator ( ```animateJoin``` is).

bluong:
I'll need to discuss with other collaborators to see if this makes sense (likely won't have an elaborate response until Monday unfortunately!), but at first glance...

If we can find another name for this idea then ideally we would use that name instead of overloading, but the problem is that it seems like `animate` is actually the correct name.  `animateJoin` seems a bit off to me, especially if starts becoming the mainstay endpoint in the `Animator` class.  That said, it also has some bit of history within its naming, so I'm not completely against the naming.

In terms of having the mainstay endpoint, what I am slowly alluding to is possibly the deprecation of the current `animate` function.  This means that we still cannot take it out, but it seems more than likely that `animate` will stop being used since it seems this new functionality is superior / encompasses it.

Another point regarding

> Remember that, even now, end-users of an animator are not responsible for calling the animate method - it is called on their behalf by Drawer.drawStep() So they don't get a chance to mess up.

While this is somewhat true, it would be nice if users can use the animator as somewhat standalone.  It currently is a standalone item, which is pretty nice for isolated testing.  It just has little meaning in production usage by itself.

softwords:
Just on the renaming - I have also thought that, if we go with ```animateJoin```, then a more descriptive name for the type currently called ```DrawingTarget``` would be ```JoinResult```  . That is, it is the set of selections created by the _join_ in the d3 sense. I'd like to think someone new to plottable with some d3 behind them would quickly "get" this.

```javascript
animateJoin(joinResult: Drawers.JoinResult, attrToAppliedProjector: AttributeToAppliedProjector, drawer: Drawer): void;
```

>> which is pretty nice for isolated testing.

I have another branch where I've fleshed out this idea of ```animateJoin```, and in there I've been able to simply extend the current drawerTests ( there's a ```MockAnimator``` in there) to create new tests of this binding logic.  The ```MockAnimator``` seemed to me harder to build with the overloaded interface.
I you want to have a look at this, I can push it as a new PR, just to have the two side by side.

bluong:
Hello @softwords !  I'm still thinking about how best to go about this and I was curious about the below...

What happens if we do 2 things that might make things easier for animator extension and animator usage?  One is that I am still thinking of using this signature
```typescript
animate(Drawer, AttributeToAppliedProjector)
```

The way I am thinking of the above signature is that an `Animator` will animate by taking the current state of the `Drawer` and transitioning the attributes to the `AttributeToAppliedProjector`.  There are problems with this as you've highlighted, and I want to explore those problems.

One problem is that a `drawingTarget()` or so endpoint would have to be exposed.  I am thinking about it now, and I am somewhat aligned with if `drawingTarget()` only acts as a getter then this exposes "what is the last data transition that the Drawer did".  In this regard this seems ok to me to expose.  It is exposing a somewhat valuable piece to how the `Drawer` operates.

The other problem is that even if the above is done, then there is redundant code in the implementation of each `Animator`.  I actually think this redundant code can be avoided if we do either 1 of 2 things.  If we extract this idea into a function at the super class, then animators could just be calling the same function per implementation instead of a whole block (this would ease implementation).  Another idea is to actually make use of inheritance and implement this block once at the super class, and then each of the subclasses calls `super.animate` to use the implemented block.

How does the above sound to you?

bluong:
On an extreme note, one thing that could completely ease usage but might bloat the `Drawer` class is to actually put the `AttributeToAppliedProjector` as a getter endpoint on the `Drawer` essentially saying what the attributes of the element will be after complete drawing.

With the above idea, this would drop the endpoint to be...

```typescript
animate(Drawer);
```

The above sounds incredibly nice in theory, but it requires `Drawer` to be a lot more exposed about the information that it is processing.  Considering that all of these ideas are ported to the `Drawer` in the first place, and essentially just carried on to the `Animator`, maybe this idea can work.

What do you think about this idea?

softwords:
good morning (!) @bluong 
before we go too much further on this conversation, I'll I'd like to show you the most recent refactor, that 
separates animate into ```animate``` and ```animateJoin```
renames ```DrawingTarget``` to ```JoinResult```
updates the documentation around this paradigm

I have this in a separate branch, so i could
- create a new, temporary PR in parallel to this one, so you can compare them
- merge into this PR and we can roll it back later if we need to...

What would you prefer?

bluong:
If you have a separate branch containing all of this, feel free to make another pull request.  However, if you are just extending the current branch and you are confident in your Git-Fu (to revert back commits) then we can just work with this pull request.

bluong:
Right now, we are investigating the proper API for this thing, so right now I am personally reading less context to each of the surroundings except for this `animate` / `animateJoin` API

softwords:
>>  and you are confident in your Git-Fu...

Implicit warning heeded and I have created a separate PR!

3 strands of this issue to untangle I think:
a)  what should this interface look like?
b) what should it be called?
c) what to do with the legacy interface - overload / separate / drop ?

 I will use ```JoinResult``` in place of ```DrawingTarget``` below:

a)  what should this interface look like?

The choice comes down to 
``` javascript
xxxxx(Drawer, AttributeToAppliedProjector)
```
or
``` javascript
xxxxx(JoinResult, AttributeToAppliedProjector, Drawer)
```
or
``` javascript
xxxxx(Drawer)
```
The first requires ```JoinResult()``` Read-only on Drawer as you note - not a bad thing , then animator implementations will get the ```JoinResult``` from there. 
The second is perhaps most obvious what you are expected to do with it. 
The last also needs ```AttributeToAppliedProjector``` exposed on ```Drawer```- I don't like this because that object potentially comes _on each element_ of the DrawStep[] array; Drawer would need to take each one in turn and set its public property before invoking the ```Animator```, this seems a fussy way to pass this argument.

b) what should it be called?

Ideally, I agree you'd want to call this ```animate``` obviously. But if that's not possible for legacy reasons, ```animateJoin``` works for me if the documentation reinforces this:
_animate_ animates a selection
_animateJoin_ animates a data join (ie a ```JoinResult```)

c) what to do with the legacy interface - overload / separate / drop ?

I've tried the overloading, and you've seen this PR - I think it's clumsy and confusing ultimately.
Separating is in the other PR #2930 - this is my preferred approach now - leaves us with the naming question which is the lesser evil to my mind.
Dropping the current interface is the most risky I suppose, but does leave the name ```animate``` free for the new signature - semver issues aside.


So my votes:

a)  what should this interface look like?
b) what should it be called?

``` javascript
animateJoin(JoinResult, AttributeToAppliedProjector, Drawer)
```
c) what to do with the legacy interface - overload / separate / drop ?
separate (ie preserve)

Note that all this is transparent to most users who will configure and assign an ```animator``` to a plot - one thing that struck me as I worked on some development code ("dogfooding" !) is that you never see or mention these interfaces anyway.

softwords:
>> Another idea is to actually make use of inheritance and implement this block once at the super class, and then each of the subclasses calls super.animate to use the implemented block.

Back in the original big PR #2907 there is a ```Base``` animator from which the new animators derive:

Base --> Attr Callback Bar

Attr --> Opacity

bluong:
Talked with collaborators and after quite a bit of discussion, we came up with an alternative API that hopefully aims to be clean and easy to use for the user, and we are wondering if this makes sense with you.

The overall objective is to create “enter” / “exit” / “merge” animators to configure animation for each piece.  To do this, it seems we need to…
* Create keyFunctions so that D3 knows which pieces of data are “enter” data, “exit” data, and “merge” data (You have already completed this part!)
* Add a mechanism to associate Animators with enter/update/exit selections
 Remove the `_animateOnNextRender()` blocks

The above approach has its own ambiguity, specifically regarding creating a way to associate “enter” data with the “enter” `Animator`, etc, so we will need to sort that piece out.

At this point, we theoretically have all the pieces to do meaningful animation (at this stage we can probably merge into develop).  At this point we can think about other pieces (pull requests directly to develop now).
* OpacityAnimator
* MultiStageAnimator
* ResetAnimator

Does the above sound like an idea we can work with?  I greatly apologize for informing you of the switch in approach so late in the pull request.  I hope you are still willing to continue with your work as I still think this code provides a ton of value, and I will continue to iterate if you intend to do so (even more expediently I hope!).

softwords:
good morning @bluong 

So if I am understanding correctly, your proposal here is that there are separate animators applied to each of the selections enter / update / exit. ?

This has some similarities to some of the approaches I considered when I first start to look into this, in #2512 ; e.g like adding a member ```appliesTo``` to the ```drawStep``` object.
I can see the appeal of this in that it suggests the interface on ```Animator``` can remain unchanged. It can continue to accept a sinle ```Selection``` rather the the new ```JoinResult``` object. As you say this presents some housekeeping issues to marry up each selection with its animator, but I think a deeper problem that you will hit when you try to build sophisticated animation sequences using this approach will be co-ordination of the activities of these animators which are operating in their own silo. 
To give an example, consider the ```Bar``` animator in #2907 This animator
- removes the ```exit``` elements, then
- moves the ```update``` elements, then
- adds in the ```enter``` elements

Each of these 3 steps is allocated a time slice of the overall duration.
However, it's smart enough to realise that, if there are _no_ exiting elements, it should not sit doing nothing, but instead moves the ```update``` process forward. Similarly, if there is no ```update```, it moves the ```enter``` process forward. 
What this means is that: the animation on ```enter``` needs to know about ```exit``` and ```update```. I think this will pose a problem using the model you are suggesting.

bluong:
> However, it's smart enough to realise that, if there are no exiting elements, it should not sit doing nothing, but instead moves the update process forward. Similarly, if there is no update, it moves the enter process forward.

Do you mean that we need to have 3 animation steps ("exit", then "update", then "enter") that need to happen in sequence every time?  (If no "exit" animation needs to occur then immediately move to "update" then immediately move to "enter")

If the above is the case, can we consider taking advantage of the `_generateDrawSteps()` to make a draw step for each of the animation steps?  Note that this is currently done in order to work with the "main" and "reset" animation modes, which does "reset" -> "main" ("reset" is only added if "reset" needs to happen).  What we can do instead of the current world is to have "exit" -> "update" -> "enter" (And as before, only add the "exit" and "update" steps if they are necessary).

Does the above approach make sense?  Also, this somewhat pinpoints to me where the change needs to happen for the 2nd step (We need to change `_generateDrawSteps()`) so that we associate the right data with the right animation step.

softwords:
>> can we consider taking advantage of the _generateDrawSteps() to make a draw step for each of the animation steps? 

In the very earliest hacks I did directly into plottable.js ( the fiddles are still there in #2512) I did take this approach, adding ```appliesTo``` to each ```drawStep``` created. I put new methods on the plot ```enter()``` and ```exit()``` that allowed you to specify what those animations should do.

@jtlan  and @aicioara  gave valuable feedback on this, and I changed my thinking to the current approach.

One thing I would say about ```Drawer.draw``` is that I think the scheduling of animations by using ```setTimeout``` is a bit ... suspect. #2707 shows why. It doesn't look like any d3 samples you'll see. I've tried to move away from this by using chained transitions and the ```squEase``` easing function as two more d3-friendly ways to manage sequencing.  By eliminating a draw step for reset, you are close to a point where you could consider getting rid of this ```setTimeout``` step altogether. (On that note, one consequence of logic in the ```Bar``` animator in my implementation - its ability to discard steps -  is that the calculated duration of animator may not be accurate.)
If you start to pass more complex arrays of ```drawStep```s, I have a feeling you may see more of artefacts like #2707.

bluong:
Yes, I do believe that using chained transitions is better than `setTimeout()`.

Accordingly, it seems like we want to adjust our rendering so we can more rely on chained transitioning for animation.

Based on the above, I currently see 2 decently sized issues.
* Find a way to associate data selections with the right transition (as previously talked about)
* Shift our architecture to be more based on D3's timing logic

Based on the above 2 issues, I currently see this:  The first one is required to have animations work for each piece of enter/exit/merge, while the second one is a definite upgrade in architecture, likely solving bugs along the way.  Does it make sense to solve the second issue separate from the first?

I feel that if we can remedy the first issue, then we have a much more direct path to Plottable operating like the JSFiddles that you have provided.  However, if you think that solving the second issue will greatly affect how the first option will be implemented, we can try to solve that second issue along with the first one.

softwords:
>> ... associate “enter” data with the “enter” Animator, etc, so we will need to sort that piece out.

Stepping back a bit - I think this becomes the crux of the issue:

my approach: 
- one animator gets _all_ the selections ( enter/update/exit))

your approach:
- 3 animators get one selection each

I think I am mounting a robust defence for the approach I have taken (!) here and further refined in the alternative #2930. And please do remember that there is concrete implementations of all of this in the 'big PR' #2907 and that the fiddles http://jsfiddle.net/oaskqta0/23/ and http://jsfiddle.net/oaskqta0/19/ are built from that PR. So you really can see what it would look like in both coding and functionality.

I don't believe you have spelled out the case for why your approach would be preferable...  over to you! :)

bluong:
A few advantages of using separate `Animator`s are that:
* The API would not have to drastically change as much, meaning that discussions surrounding changing the API can be less had.  That said, if the resultant behavior has a good architecture, then I am for deprecating the old API or similar.
* There seems to be more modularity in the separate `Animator` approach.  What if we wanted to do some form of `Attr` animation for entering data, an `Attr` animation for merging data, and an `Opacity` animation for exiting data?  To my gathering, separate `Animator`s would work nicely for this approach.

What do you think about this?

softwords:
>> There seems to be more modularity in the separate Animator approach. 

I think the _granularity_ may be more of a hindrance than a help in restricting the amount of information available to each animator.

That said, the animations you describe are very easy to build with the ```Attr``` animator, which takes an attribute collection to apply to the entering elements on entry, and an attribute collection to transition the exiting elements to before removing.

See _Size-in / Fade-out_  and _Fade-in / Size-out_ options in updated fiddle:
http://jsfiddle.net/oaskqta0/24/

_Bubble_ and _Raindrops_ work the same way.

christinaforney:
Hi @softwords - I just wanted to let you know that this PR is still on our radar, and that the team is temporarily focused on some other things, so we need to delay working on this with you for a little while. 

We really appreciate your contribution, and will pick up iterating with you as soon as we can.

softwords:
Hello @christinaforney  - thanks for the heads-up on that. This month is awkward for me anyway - I'm working on-site in other (developing) countries; as well of the usual time-zone issues (GMT+11 for me) I'll have limited time and less reliable internet - so out conversations may have longer turn-arounds. I'll do my best!

Issue #2630, opened 2015-08-17
[Testing Audit] Plot and XYPlot (not descendants)


crmorford:

aicioara:
Blocking on this until children are merged?

PR #2923, opened 2015-10-23
[PlotTests] - Rework

bluong:
Test screenshot:
screen shot 2015-10-26 at 2 25 24 pm

`Plot` is incredibly odd in that it cannot render on its own.  Thus, situations that directly deal with a rendered product are not tested.  However, other cases regarding anchoring, dataset mangement, etc have been reworked to follow the new guidellines.

Part of #2630

### DIFF
--- a/test/plots/plotTests.ts
+++ b/test/plots/plotTests.ts
@@ -17,6 +17,85 @@ class CountingPlot extends Plottable.Plot {
 
 describe("Plots", () => {
   describe("Plot", () => {
+
+    it("adds a \"plot\" css class by default", () => {
+      const plot = new Plottable.Plot();
+      assert.isTrue(plot.hasClass("plot"), "plot class added by default");
+    });
+
+    describe("managing datasets", () => {
+      let plot: Plottable.Plot;
+
+      beforeEach(() => {
+        plot = new Plottable.Plot();
+      });
+
+      it("can add a dataset", () => {
+        const dataset = new Plottable.Dataset();
+        plot.addDataset(dataset);
+
+        assert.deepEqual(plot.datasets(), [dataset], "dataset has been added");
+      });
+
+      it("can remove a dataset", () => {
+        const dataset = new Plottable.Dataset();
+        plot.addDataset(dataset);
+        plot.removeDataset(dataset);
+        assert.deepEqual(plot.datasets(), [], "dataset has been removed");
+      });
+
+      it("can set the datasets", () => {
+        const datasetCount = 5;
+        const datasets = Plottable.Utils.Math.range(0, datasetCount).map(() => new Plottable.Dataset());
+        plot.datasets(datasets);
+        assert.deepEqual(plot.datasets(), datasets, "datasets have been set");
+      });
+
+      it("adds a g element for each dataset to the render area", () => {
+        const datasetCount = 5;
+        const datasets = Plottable.Utils.Math.range(0, datasetCount).map(() => new Plottable.Dataset());
+        plot.datasets(datasets);
+
+        const svg = TestMethods.generateSVG();
+        plot.anchor(svg);
+
+        assert.strictEqual(plot.content().select(".render-area").selectAll("g").size(), datasetCount, "g for each dataset");
+
+        svg.remove();
+      });
+
+      it("updates the scales extents when the datasets get updated", () => {
+        const scale = new Plottable.Scales.Linear();
+
+        const data = [5, -5, 10];
+        const dataset = new Plottable.Dataset(data);
+
+        plot.attr("foo", (d) => d, scale);
+        plot.addDataset(dataset);
+
+        const oldDomain = scale.domain();
+
+        const svg = TestMethods.generateSVG();
+        plot.anchor(svg);
+
+        assert.operator(scale.domainMin(), "<=", Math.min.apply(null, data), "domainMin extended to at least minimum");
+        assert.operator(scale.domainMax(), ">=", Math.max.apply(null, data), "domainMax extended to at least maximum");
+
+        plot.removeDataset(dataset);
+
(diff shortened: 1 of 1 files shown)


jtlan:
`"dataset"` should be uppercase if it refers to the `Dataset` object type.

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
 
-      let dataPointConverter = (datum: any, index: number) => dataPoints[index];
+        svg.remove();
+      });
 
-      // Create mock drawer with already drawn items
-      let mockDrawer = new Plottable.Drawer(dataset);
-      (<any> mockDrawer)._svgElementName = "circle";
-      mockDrawer.selector = () => "circle";
+      it("updates the scale extents in dataset order", () => {
```
jtlan: I wonder if this is more of a `Scales.Category` test?

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
-      (<any> plot)._additionalPaint = additionalPaint;
-      plot.animator(Plottable.Plots.Animator.MAIN, animator);
-      let svg = TestMethods.generateSVG();
-      plot.x((d: any) => d.x, x);
-      plot.y((d: any) => d.y, y);
-      plot.renderTo(svg);
-      assert.strictEqual(recordedTime, 20, "additionalPaint passed appropriate time argument");
-      svg.remove();
-    });
+    it("disconnects the data extents from the scales when destroyed", () => {
```
jtlan: Wrong title.

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
+            "the clipPath reference was updated");
+          svg.remove();
+        });
+      }
+    });
 
-        let svg = TestMethods.generateSVG();
-        let plot = new Plottable.Plot();
-        plot.renderTo(svg);
+    describe("setting the attribute of plot elements", () => {
```
jtlan: You could label this `"setting attributes with attr()"` if you want, since parens in the name of a `describe()` can be found by `grep`.

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
+        attrAccessor = plot.attr("foo").accessor;
+        plot.datasets().forEach((dataset, datasetIndex) => {
+          dataset.data().forEach((datum, index) => {
+            assert.strictEqual(attrAccessor(datum, index, dataset), stringAccessor(datum, index),
+              `can set attribute for string datum ${index} in dataset ${datasetIndex}`);
+          });
+        });
+      });
+
+      it("can set the attribute based on the scaled input data", () => {
```
jtlan: Might be clearer to say something like `"can apply a Scale to the returned values"`.

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
+        const data = [1, 2, 3, 4, 5];
+        const dataset = new Plottable.Dataset(data);
+        plot.addDataset(dataset);
+
+        const numberAccessor = (d: any, i: number) => d + i * 10;
```
jtlan: What's the point of also adding the index?

test/plots/plotTests.ts
```diff
@@ -1,584 +1,302 @@
...
+        attrScale = plot.attr("foo").scale;
+        plot.datasets().forEach((dataset, datasetIndex) => {
+          dataset.data().forEach((datum, index) => {
+            assert.strictEqual(attrScale.scale(attrAccessor(datum, index, dataset)), categoryScale.scale(stringAccessor(datum, index)),
+              `can set based on scaled version of string datum ${index} in dataset ${datasetIndex}`);
+          });
+        });
+      });
+
+      it("updates the scales extents when an attribute is set", () => {
```
jtlan: Might want to specify that it updates the extent of the `Scale` associated with that attribute.

test/plots/plotTests.ts
bluong: Testing the index portion of the accessor

test/plots/plotTests.ts
bluong: This test should exist at the plot level since plots install extents onto the scale.

Scales however should be testing their own extents, so tests surrounding this idea should be added there as well.

jtlan: What I'm getting that is that for most scales, the order of extent updating isn't relevant.

test/plots/plotTests.ts
jtlan: That should go in a separate test, right? With an `Accessor` that includes `assert` statements.

test/plots/plotTests.ts
bluong: Remove this test then?  Seems like it isn't useful

test/plots/plotTests.ts
bluong: Should I add another test for this idea or can we assume it's covered by the previous test and remove the index accessor here?

jtlan: Can you add another test that asserts that the index is passed in by using some sort of `Accessor` that includes a counter and `assert`s?

test/plots/plotTests.ts
jtlan: Yeah.

test/plots/plotTests.ts
jtlan: Example:
```Typescript
it("passes the correct index to the Accessor", () => {
  const data = ["A", "B", "C"];
  const dataset = new Plottable.Dataset(data);
  plot.addDataset(dataset);
  const indexCheckAccessor = (datum: any, index:number) => {
    assert.strictEqual(index, data.indexOf(datum), "was passed correct index";
  };
  plot.attr("foo", indexCheckAccessor);
});
```

bluong: Done

PR #2930, opened 2015-10-26
Animation Stage1 Alternative - animateJoin method

softwords:
Alternative version of Stage 1 that splits of `animateJoin`  into a separate method of `Animator`.


softwords:
@bluong alternative PR ...

src/animators/easingFunctions.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+        let tbar = (t - start) / (end - start);
+        if (typeof easingFunction === "string") {
+          return d3.ease(easingFunction)(tbar);
+        }
+        return (<EasingFunction>easingFunction)(tbar);
+      };
+    }
+  }
+}
```
bluong, line 66: Can we extract this out (possibly into a pull request into develop if you can attach it to `Animators.Easing` or into `objectConstancy` and then we can connect the pieces later)?

src/drawers/drawer.ts
```diff
@@ -106,9 +168,20 @@ export class Drawer {
...
+    // (using the 'legacy' api point 'animate') continue to work as before
+    // this test should be removed if and when semver = 2.0.0
+    if (step.animator.animateJoin) {
+      step.animator.animateJoin(this._joinResult, step.attrToAppliedProjector, this);
+    } else {
+      // an old animator won't remove exit
+      this._joinResult.exit
+        .remove();
+      step.animator.animate(selection, step.attrToAppliedProjector);
+    }
```
softwords, line 181: good morning @bluong 
I have thought further on the issue of the change to animator interface, specifically adding and using the `animateJoin` method as in here.

First point to note is that for the majority of cases, the javascript consumer never calls the `animator` directly, it is called on their behalf by the `Drawer` when they assign the animator to their plot.

So, the only case that presents a problem is if a javascript client has built a custom animator and implemented `animate` - `animate` would not be called, and in fact there would be an error at this point because `animateJoin` is not found.

This (unattractive) change will defend against this.

I think the only circumstance now where any consumer of plottable will see a difference is that if you are authoring a custom animator in typescript, you'll get a compile warning about the missing member.

softwords:
@bluong could you merge develop into objectConstancy please and I'll update all the pull requests accordingly

softwords:
Perhaps we should now close #2921 - (I think the overloaded interface is out of consideration?)  and make this the target for future deliberation... 
Meantime I have updated the big PR #2907 with all these changes, so it will continue to represent the complete, current state of this development thread, and be the source of any future smaller pull requests.
#2907 is also the source of the fiddles http://jsfiddle.net/oaskqta0/24/  and http://jsfiddle.net/oaskqta0/27/
so we keep that verification that it all still works!

In http://jsfiddle.net/oaskqta0/27/ I have added a "legacy JS animator", that is, a class built in JS that implements the legacy animate function, to verify that this is gracefully suppported in this build:

```javascript

function myLegacyAnimator() {
    this.animate = function(selection, attrs) {
        alert("hello from the legacy JS custom animator");
        selection.attr(attrs);
    }
}
```

src/animators/easingFunctions.ts
softwords, line 66: Not really sure what you mean by "attach to to Animators.Easing" - did you mean a namespace?
Animators.Easing is already in use by the class.

bluong, line 66: Oh I mean that `Animators.Easing` can make use of these easing functions, however the signatures need to be modified to account for this.

bluong:
I've updated the objectContancy branch

src/animators/easingFunctions.ts
bluong, line 66: To be more specific, I think this `easingFunctions.ts` file specifically can be made as a pull request to develop, as long as `Animators.Easing`'s `easing()` endpoint is modified to make use of the above types.

softwords:
fixed fiddle for callback option 'explode' after change to new interface
http://jsfiddle.net/oaskqta0/28/

softwords:
Hi @bluong you guys have clearly been deep in thought! Can you merge develop into objectConstancy at some point and I can keep my PR up to date with develop too.

softwords:
Hi @bluong noticing that you are working towards a version 2? If you have a moment to give me a quick heads-up on whether you are considering using any of this animation stuff? 
As time goes by I'm seeing conflicts inevitably appearing between these PRs and develop.... if you don't see yourselves using this PR then I won't try to keep up - I'll stay with my own fork which does what I need it to.

PR #2938, opened 2015-10-29
enable downsampling in Stacked Area Plot

### LINKED ISSUE #2868: Downsampling create white space inside Stacked Area Plot

yunhaolucky:
**Issue**: 
When using downsampling, points with different x values are removed in different areas, which make the x value inconsistent among different dataset. This creates white spaces and overlappings between areas.

**Fix**:
The point drawn are the union of the downsampling result of each area. This means a point is removed only when all points at the same x position are removed by downsampling. This might affect the rendering performance.

Draw without white spaces and overlappings between areas: http://jsfiddle.net/yxpq4uuv/50/
Performance Test: http://jsfiddle.net/yxpq4uuv/51/

close #2868

### DIFF
--- a/test/plots/plotTests.ts
+++ b/test/plots/plotTests.ts
@@ -1,584 +1,318 @@
 ///<reference path="../testReference.ts" />
 
-class CountingPlot extends Plottable.Plot {
-  public renders: number = 0;
-
-  public render() {
-    ++this.renders;
-    return super.render();
-  }
-
-  protected _createDrawer(dataset: Plottable.Dataset) {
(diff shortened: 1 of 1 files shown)


bluong:
Please remedy the "plottable.js" additions.  Your change should not warrant an 8000 line change in plottable.js

PR #2942, opened 2015-10-29
[DO NOT MERGE] Bar Plot tests for review

jtlan:
#2940 contains all the commits that actually create this change, but the diff is too large for Github to view.

Since basically every line was replaced, this branch was created to show the effect of the change.

# DO NOT MERGE.

### DIFF
--- a/test/plots/barPlotTests.ts
+++ b/test/plots/barPlotTests.ts
@@ -8,15 +8,15 @@ describe("Plots", () => {
       });
 
       it("defaults to vertical", () => {
-        let defaultPlot = new Plottable.Plots.Bar<number, number>();
+        const defaultPlot = new Plottable.Plots.Bar<number, number>();
         assert.strictEqual(defaultPlot.orientation(), "vertical", "default Plots.Bar() are vertical");
       });
 
       it("sets orientation on construction", () => {
-        let verticalPlot = new Plottable.Plots.Bar<number, number>("vertical");
+        const verticalPlot = new Plottable.Plots.Bar<number, number>("vertical");
         assert.strictEqual(verticalPlot.orientation(), "vertical", "vertical Plots.Bar()");
 
-        let horizontalPlot = new Plottable.Plots.Bar<number, number>("horizontal");
+        const horizontalPlot = new Plottable.Plots.Bar<number, number>("horizontal");
         assert.strictEqual(horizontalPlot.orientation(), "horizontal", "horizontal Plots.Bar()");
       });
     });
@@ -60,14 +60,14 @@ describe("Plots", () => {
         function assertCorrectRendering() {
           const baseline = barPlot.content().select(".baseline");
           const scaledBaselineValue = valueScale.scale(<number> barPlot.baselineValue());
-          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "1"), scaledBaselineValue,
-            `baseline ${valuePositionAttr + "1"} is correct`);
-          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "2"), scaledBaselineValue,
-            `baseline ${valuePositionAttr + "2"} is correct`);
-          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "1"), 0,
-            `baseline ${basePositionAttr + "1"} is correct`);
-          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "2"), TestMethods.numAttr(svg, baseSizeAttr),
-            `baseline ${basePositionAttr + "2"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, `${valuePositionAttr}1`), scaledBaselineValue,
+            `baseline ${valuePositionAttr}1 is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, `${valuePositionAttr}2`), scaledBaselineValue,
+            `baseline ${valuePositionAttr}2 is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, `${basePositionAttr}1`), 0,
+            `baseline ${basePositionAttr}1 is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, `${basePositionAttr}2`), TestMethods.numAttr(svg, baseSizeAttr),
+            `baseline ${basePositionAttr}2 is correct`);
 
           const bars = barPlot.content().selectAll("rect");
           assert.strictEqual(bars.size(), data.length, "One bar was created per data point");
@@ -87,7 +87,7 @@ describe("Plots", () => {
 
             const valuePosition = TestMethods.numAttr(bar, valuePositionAttr);
             const isShifted = isVertical ? (datum.value > barPlot.baselineValue()) : (datum.value < barPlot.baselineValue());
-            const expectedValuePosition = (isShifted) ? scaledBaselineValue - valueSize : scaledBaselineValue;
+            const expectedValuePosition = isShifted ? scaledBaselineValue - valueSize : scaledBaselineValue;
             assert.closeTo(valuePosition, expectedValuePosition, window.Pixel_CloseTo_Requirement,
               `bar ${valuePositionAttr} is correct (index ${index})`);
           });
@@ -115,7 +115,7 @@ describe("Plots", () => {
         });
 
         it("can autorange value scale based on visible points on base scale", () => {
-          const firstTwoBaseValues = [data[0].base, data[1].base ];
+          const firstTwoBaseValues = [ data[0].base, data[1].base ];
           valueScale.padProportion(0);
           baseScale.domain(firstTwoBaseValues);
           barPlot.addDataset(dataset);
@@ -230,7 +230,7 @@ describe("Plots", () => {
               }
               valueScale = new Plottable.Scales.Linear();
 
-              const baseAccessor = (scaleType === "Time") ? (d: any) => new Date(d.base) : (d: any) => d.base;
+              const baseAccessor = scaleType === "Time" ? (d: any) => new Date(d.base) : (d: any) => d.base;
               const valueAccessor = (d: any) => d.value;
               if (orientation === Plottable.Plots.Bar.ORIENTATION_VERTICAL) {
                 barPlot.x(baseAccessor, baseScale);
@@ -304,7 +304,7 @@ describe("Plots", () => {
               const barSize = TestMethods.numAttr(bar, baseSizeAttr);
               const svgSize = TestMethods.numAttr(svg, baseSizeAttr);
               assert.operator(barSize, ">=", svgSize / 4, "bar is larger than 1/4 of the available space");
-              assert.operator(barSize, "<=", svgSize / 2, "bar is smaller than half the available space");
+              assert.operator(barSize, "<=", svgSize / 2, "bar is smaller than 1/2 of the available space");
             });
 
             it("computes a sensible width when given repeated base value", () => {
@@ -321,7 +321,7 @@ describe("Plots", () => {
                 const bar = d3.select(this);
                 const barSize = TestMethods.numAttr(bar, baseSizeAttr);
                 assert.operator(barSize, ">=", svgSize / 4, "bar is larger than 1/4 of the available space");
-                assert.operator(barSize, "<=", svgSize / 2, "bar is smaller than half the available space");
+                assert.operator(barSize, "<=", svgSize / 2, "bar is smaller than 1/2 of the available space");
               });
             });
 
@@ -406,7 +406,6 @@ describe("Plots", () => {
         });
 
         it("draws one label per datum", () => {
-          barPlot.labelsEnabled(true);
           barPlot.labelsEnabled(true);
           const texts = barPlot.content().selectAll("text");
           assert.strictEqual(texts.size(), data.length, "one label drawn per datum");
--- a/test/plots/barPlotTests.ts
+++ b/test/plots/barPlotTests.ts
@@ -1,1 +1,783 @@
 ///<reference path="../testReference.ts" />
+
+describe("Plots", () => {
+  describe("Bar Plot", () => {
+    describe ("orientations", () => {
+      it("rejects invalid orientations", () => {
+        assert.throws(() => new Plottable.Plots.Bar("diagonal"), Error);
+      });
+
+      it("defaults to vertical", () => {
+        let defaultPlot = new Plottable.Plots.Bar<number, number>();
+        assert.strictEqual(defaultPlot.orientation(), "vertical", "default Plots.Bar() are vertical");
+      });
+
+      it("sets orientation on construction", () => {
+        let verticalPlot = new Plottable.Plots.Bar<number, number>("vertical");
+        assert.strictEqual(verticalPlot.orientation(), "vertical", "vertical Plots.Bar()");
+
+        let horizontalPlot = new Plottable.Plots.Bar<number, number>("horizontal");
+        assert.strictEqual(horizontalPlot.orientation(), "horizontal", "horizontal Plots.Bar()");
+      });
+    });
+
+    const orientations = [Plottable.Plots.Bar.ORIENTATION_VERTICAL, Plottable.Plots.Bar.ORIENTATION_HORIZONTAL];
+    orientations.forEach((orientation) => {
+      const isVertical = orientation === Plottable.Plots.Bar.ORIENTATION_VERTICAL;
+      const basePositionAttr = isVertical ? "x" : "y";
+      const baseSizeAttr = isVertical ? "width" : "height";
+      const valuePositionAttr = isVertical ? "y" : "x";
+      const valueSizeAttr = isVertical ? "height" : "width";
+
+      describe(`rendering when ${orientation}`, () => {
+        const data = [
+          { base: "A", value: 1 },
+          { base: "B", value: 0 },
+          { base: "C", value: -1 }
+        ];
+
+        let svg: d3.Selection<void>;
+        let barPlot: Plottable.Plots.Bar<string | number, number | string>;
+        let baseScale: Plottable.Scales.Category;
+        let valueScale: Plottable.Scales.Linear;
+        let dataset: Plottable.Dataset;
+
+        beforeEach(() => {
+          svg = TestMethods.generateSVG();
+          barPlot = new Plottable.Plots.Bar<string | number, number | string>(orientation);
+          baseScale = new Plottable.Scales.Category();
+          valueScale = new Plottable.Scales.Linear();
+          if (orientation === Plottable.Plots.Bar.ORIENTATION_VERTICAL) {
+            barPlot.x((d: any) => d.base, baseScale);
+            barPlot.y((d: any) => d.value, valueScale);
+          } else {
+            barPlot.y((d: any) => d.base, baseScale);
+            barPlot.x((d: any) => d.value, valueScale);
+          }
+          dataset = new Plottable.Dataset(data);
+        });
+
+        function assertCorrectRendering() {
+          const baseline = barPlot.content().select(".baseline");
+          const scaledBaselineValue = valueScale.scale(<number> barPlot.baselineValue());
+          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "1"), scaledBaselineValue,
+            `baseline ${valuePositionAttr + "1"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "2"), scaledBaselineValue,
+            `baseline ${valuePositionAttr + "2"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "1"), 0,
+            `baseline ${basePositionAttr + "1"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "2"), TestMethods.numAttr(svg, baseSizeAttr),
+            `baseline ${basePositionAttr + "2"} is correct`);
+
+          const bars = barPlot.content().selectAll("rect");
+          assert.strictEqual(bars.size(), data.length, "One bar was created per data point");
+          bars.each(function(datum, index) {
+            const bar = d3.select(this);
+            const baseSize = TestMethods.numAttr(bar, baseSizeAttr);
+            assert.closeTo(baseSize, baseScale.rangeBand(), window.Pixel_CloseTo_Requirement,
+              `bar ${baseSizeAttr} is correct (index ${index})`);
+
+            const valueSize = TestMethods.numAttr(bar, valueSizeAttr);
+            assert.closeTo(valueSize, Math.abs(valueScale.scale(datum.value) - scaledBaselineValue),
+              window.Pixel_CloseTo_Requirement, `bar ${valueSizeAttr} is correct (index ${index})`);
+
+            const basePosition = TestMethods.numAttr(bar, basePositionAttr);
+            assert.closeTo(basePosition, baseScale.scale(datum.base) - 0.5 * baseSize, window.Pixel_CloseTo_Requirement,
+              `bar ${basePositionAttr} is correct (index ${index})`);
+
+            const valuePosition = TestMethods.numAttr(bar, valuePositionAttr);
+            const isShifted = isVertical ? (datum.value > barPlot.baselineValue()) : (datum.value < barPlot.baselineValue());
+            const expectedValuePosition = (isShifted) ? scaledBaselineValue - valueSize : scaledBaselineValue;
+            assert.closeTo(valuePosition, expectedValuePosition, window.Pixel_CloseTo_Requirement,
+              `bar ${valuePositionAttr} is correct (index ${index})`);
+          });
+        }
+
+        it("renders with no data", () => {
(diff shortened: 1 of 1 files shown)


bluong:
Since this is aiming at a side branch, we can actually be ok with merging this in I think.  We'll just need another simple pull request to remove the old tests

jtlan:
The reason I don't want to do it, is that we lose the individual commits and their commentary.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
 ///<reference path="../testReference.ts" />
+
+describe("Plots", () => {
+  describe("Bar Plot", () => {
+    describe ("orientations", () => {
```
bluong: I prefer either using an adjective or using the endpoint name for the describe message.  Note that if a test were to fail, we would get this message...

"Plots Bar Plot orientations ..."

Ex. "Plots Bar Plot orientations rejects invalid orientations"

Lots of broken English here.  To respond to this I think the first two pieces can be remedied using our module structure.  However, the third piece just breaks the flow completely.  An endpoint here or an adjective would continue the sentence flow pretty seamlessly into the the `it`

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+      it("rejects invalid orientations", () => {
+        assert.throws(() => new Plottable.Plots.Bar("diagonal"), Error);
```
bluong: Do we want to report a message here?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+        let defaultPlot = new Plottable.Plots.Bar<number, number>();
+        assert.strictEqual(defaultPlot.orientation(), "vertical", "default Plots.Bar() are vertical");
+      });
+
+      it("sets orientation on construction", () => {
+        let verticalPlot = new Plottable.Plots.Bar<number, number>("vertical");
+        assert.strictEqual(verticalPlot.orientation(), "vertical", "vertical Plots.Bar()");
+
+        let horizontalPlot = new Plottable.Plots.Bar<number, number>("horizontal");
+        assert.strictEqual(horizontalPlot.orientation(), "horizontal", "horizontal Plots.Bar()");
```
bluong: `const` all the things

bluong: you can do a `orientations.forEach` here

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+      });
+    });
+
+    const orientations = [Plottable.Plots.Bar.ORIENTATION_VERTICAL, Plottable.Plots.Bar.ORIENTATION_HORIZONTAL];
+    orientations.forEach((orientation) => {
+      const isVertical = orientation === Plottable.Plots.Bar.ORIENTATION_VERTICAL;
```
bluong: This sounds useful to have outside the `forEach`?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+      const baseSizeAttr = isVertical ? "width" : "height";
+      const valuePositionAttr = isVertical ? "y" : "x";
+      const valueSizeAttr = isVertical ? "height" : "width";
+
+      describe(`rendering when ${orientation}`, () => {
+        const data = [
+          { base: "A", value: 1 },
+          { base: "B", value: 0 },
+          { base: "C", value: -1 }
+        ];
```
bluong: We don't necessarily need the `data` variable here since it's stored inside `dataset`

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+        let svg: d3.Selection<void>;
+        let barPlot: Plottable.Plots.Bar<string | number, number | string>;
+        let baseScale: Plottable.Scales.Category;
+        let valueScale: Plottable.Scales.Linear;
+        let dataset: Plottable.Dataset;
```
bluong: We don't necessarily need the `dataset` variable here since it's stored inside `barPlot`

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+          assert.strictEqual(barPlot.content().selectAll("rect").size(), 0, "draws no bars when the domain contains no data points");
+        });
+
+        afterEach(function() {
+          if (this.currentTest.state === "passed") {
+            barPlot.destroy();
+            svg.remove();
+          }
+        });
```
bluong: I usually see the `afterEach` blocks before the test code to have an association between `beforeEach` and `afterEach` (One should cancel out the other / one should reset etc).  Which structure is more preferable?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+        function assertCorrectRendering() {
+          const baseline = barPlot.content().select(".baseline");
+          const scaledBaselineValue = valueScale.scale(<number> barPlot.baselineValue());
+          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "1"), scaledBaselineValue,
+            `baseline ${valuePositionAttr + "1"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, valuePositionAttr + "2"), scaledBaselineValue,
+            `baseline ${valuePositionAttr + "2"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "1"), 0,
+            `baseline ${basePositionAttr + "1"} is correct`);
+          assert.strictEqual(TestMethods.numAttr(baseline, basePositionAttr + "2"), TestMethods.numAttr(svg, baseSizeAttr),
```
bluong: Template the `numAttr` second argument

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+            `baseline ${basePositionAttr + "2"} is correct`);
```
bluong: you can place the 2 outside the templating

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+            assert.closeTo(valueSize, Math.abs(valueScale.scale(datum.value) - scaledBaselineValue),
+              window.Pixel_CloseTo_Requirement, `bar ${valueSizeAttr} is correct (index ${index})`);
+
+            const basePosition = TestMethods.numAttr(bar, basePositionAttr);
+            assert.closeTo(basePosition, baseScale.scale(datum.base) - 0.5 * baseSize, window.Pixel_CloseTo_Requirement,
+              `bar ${basePositionAttr} is correct (index ${index})`);
+
+            const valuePosition = TestMethods.numAttr(bar, valuePositionAttr);
+            const isShifted = isVertical ? (datum.value > barPlot.baselineValue()) : (datum.value < barPlot.baselineValue());
+            const expectedValuePosition = (isShifted) ? scaledBaselineValue - valueSize : scaledBaselineValue;
```
bluong: parens around `isShifted` unnecessary

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+            assert.closeTo(valuePosition, expectedValuePosition, window.Pixel_CloseTo_Requirement,
+              `bar ${valuePositionAttr} is correct (index ${index})`);
+          });
+        }
+
+        it("renders with no data", () => {
+          assert.doesNotThrow(() => barPlot.renderTo(svg), Error);
```
bluong: Do we want a message?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+          const firstTwoBaseValues = [data[0].base, data[1].base ];
+          valueScale.padProportion(0);
+          baseScale.domain(firstTwoBaseValues);
+          barPlot.addDataset(dataset);
+          barPlot.autorangeMode(valuePositionAttr);
+          barPlot.renderTo(svg);
+
+          const valueScaleDomain = valueScale.domain();
+          const expectedValueDomainMin = Math.min(data[0].value, data[1].value);
+          const expectedValueDomainMax = Math.max(data[0].value, data[1].value);
```
bluong: Can't we use `valueScale.domainMin()` / `domainMax()`?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+          assert.strictEqual(valueScaleDomain[0], expectedValueDomainMin, "lower bound of domain set based on visible points");
+          assert.strictEqual(valueScaleDomain[1], expectedValueDomainMax, "upper bound of domain set based on visible points");
+        });
+
+        it("doesn't show values from outside the base scale's domain", () => {
+          baseScale.domain(["-A"]);
+          barPlot.addDataset(dataset);
+          barPlot.renderTo(svg);
+
+          assert.strictEqual(barPlot.content().selectAll("rect").size(), 0, "draws no bars when the domain contains no data points");
```
bluong: `assert.isTrue(barPlot.content().selectAll("rect").empty())`?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+        afterEach(function() {
+          if (this.currentTest.state === "passed") {
+            barPlot.destroy();
+            svg.remove();
+          }
+        });
+      });
+
+      describe(`auto bar width calculation when ${orientation}`, () => {
+        const scaleTypes = ["Linear", "ModifiedLog", "Time"];
```
bluong: Why not
`[Plottable.Scales.Linear, Plottable.Scales.ModifiedLog, Plottable.Scales.Time]`?

test/plots/barPlotTests.ts
jtlan: Not worth doing a `forEach()` since the rest of the orientation tests don't need it.

test/plots/barPlotTests.ts
jtlan: It's a property not a function, so I wanted to leave it that way.

test/plots/barPlotTests.ts
jtlan: Easier to access it directly.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+            it("computes a sensible width", () => {
+              const data = [
+                { base: 1, value: 5 },
+                { base: 10, value: 2 },
+                { base: 100, value: 4 }
+              ];
+              dataset.data(data);
+
+              const closestSeparation = Math.abs(baseScale.scale(data[1].base) - baseScale.scale(data[0].base));
```
bluong: Would it less fragile to loop over and find the closest?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+              assert.strictEqual(bars.size(), data.length, "one bar was drawn per datum");
+              const svgSize = TestMethods.numAttr(svg, baseSizeAttr);
+              bars.each(function() {
+                const bar = d3.select(this);
+                const barPosition = TestMethods.numAttr(bar, basePositionAttr);
+                const barSize = TestMethods.numAttr(bar, baseSizeAttr);
+                assert.operator(barPosition, ">=", 0, `bar is within visible area (${basePositionAttr})`);
+                assert.operator(barPosition, "<=", svgSize, `bar is within visible area (${basePositionAttr})`);
+                assert.operator(barPosition + barSize, ">=", 0, `bar is within visible area (${baseSizeAttr})`);
+                assert.operator(barPosition + barSize, "<=", svgSize, `bar is within visible area (${baseSizeAttr})`);
```
bluong: All of the four test messages essentially have the same string content.  If one of them were to fail, it'd be hard to detect which one of them failed.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+              });
+            });
+
+            it("does not crash when given bad data", () => {
+              const badData: any = [
```
bluong: Did you need to `any`-cast here?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+              const singleDatumData = [
+                { base: 1, value: 5 }
+              ];
+              dataset.data(singleDatumData);
+
+              const bar = barPlot.content().select("rect");
+              const barSize = TestMethods.numAttr(bar, baseSizeAttr);
+              const svgSize = TestMethods.numAttr(svg, baseSizeAttr);
+              assert.operator(barSize, ">=", svgSize / 4, "bar is larger than 1/4 of the available space");
+              assert.operator(barSize, "<=", svgSize / 2, "bar is smaller than half the available space");
```
bluong: Very minor, why is one "1/4" and the other "half"

test/plots/barPlotTests.ts
jtlan: I think it reads better to have a direct reference to it.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+          return {
+            x: (labelBoundingClientRect.left + labelBoundingClientRect.right) / 2 - plotBoundingClientRect.left,
+            y: (labelBoundingClientRect.top + labelBoundingClientRect.bottom) / 2 - plotBoundingClientRect.top
+          };
+        };
+
+        it("does not show labels by default", () => {
+          const texts = barPlot.content().selectAll("text");
+          assert.strictEqual(texts.size(), 0, "by default, no texts are drawn");
```
bluong: `assert.isTrue(texts.empty())`?

test/plots/barPlotTests.ts
jtlan: I noticed that in other files I looked at just now. I'll move it up, since it's easy to get lost at the bottom.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+        });
+
+        it("draws one label per datum", () => {
+          barPlot.labelsEnabled(true);
+          barPlot.labelsEnabled(true);
```
bluong: Why double set?

test/plots/barPlotTests.ts
jtlan: It's the only `assert` in the `it()`, so probably not necessary.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+          const texts = barPlot.content().selectAll("text");
+          assert.strictEqual(texts.size(), data.length, "one label drawn per datum");
+          texts.each(function(d, i) {
+            assert.strictEqual(d3.select(this).text(), data[i].value.toString(), `by default, label text is the bar's value (index ${i})`);
+          });
+        });
+
+        it("hides the labels if bars are too thin to show them", () => {
+          svg.attr(baseSizeAttr, TestMethods.numAttr(svg, baseSizeAttr) / 10);
```
bluong: Let's extract the second argument to a var to show that we are forcible constraining the space

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+          barPlot.redraw();
+          barPlot.labelsEnabled(true);
+
+          const texts = barPlot.content().selectAll("text");
+          assert.strictEqual(texts.size(), 0, "no labels drawn");
+        });
+
+        it("can apply a formatter to the labels", () => {
+          barPlot.labelsEnabled(true);
+          const formatter = (n: number) => `${n}%`;
```
bluong: Not that it really matters, but why not `percentageFormatter`?

test/plots/barPlotTests.ts
jtlan: Not for the expected values :)

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+        it("shows labels inside or outside the bar as appropriate", () => {
+          barPlot.labelsEnabled(true);
+
+          const labels = barPlot.content().selectAll(".on-bar-label, .off-bar-label");
+          assert.strictEqual(labels.size(), data.length, "one label drawn per datum");
+
+          const bars = barPlot.content().select(".bar-area").selectAll("rect");
+          labels.each((d, i) => {
+            const labelBoundingClientRect = <any> (<SVGElement> labels[0][i]).getBoundingClientRect();
+            const barBoundingClientRect = <any> (<SVGElement> bars[0][i]).getBoundingClientRect();
```
bluong: Why `any`-casting?

test/plots/barPlotTests.ts
jtlan: `strictEqual()` gives a better error message because it will say how many actually rendered.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+        it("hides labels cut off by lower end of base scale", () => {
+          barPlot.labelsEnabled(true);
+          data.forEach((d, i) => {
+            let texts = barPlot.content().selectAll("text");
+            const centerOfText = getCenterOfText(<SVGElement> texts[0][i]);
+            const centerValue = baseScale.invert(isVertical ? centerOfText.x : centerOfText.y);
+            baseScale.domain([centerValue, centerValue + (DEFAULT_DOMAIN[1] - DEFAULT_DOMAIN[0])]);
+
+            texts = barPlot.content().selectAll("text"); // re-select after rendering
```
bluong: Place the comment above the code?  I'll need to look into tslint to see if enforcable.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+            const labelPosition = labelBoundingRect.left + window.Pixel_CloseTo_Requirement;
+            const linePosition = lineBoundingRect.right;
+            assert.operator(labelPosition, ">=", linePosition, "label with value = baseline is drawn to the right of the baseline");
+          }
+        });
+
+        it("hides labels cut off by lower end of base scale", () => {
+          barPlot.labelsEnabled(true);
+          data.forEach((d, i) => {
+            let texts = barPlot.content().selectAll("text");
```
bluong: Would it be better to have 2 `texts` vars, one for before domain is changed, one for after

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+            const centerOfText = getCenterOfText(<SVGElement> texts[0][i]);
+            const centerValue = baseScale.invert(isVertical ? centerOfText.x : centerOfText.y);
+            baseScale.domain([centerValue - (DEFAULT_DOMAIN[1] - DEFAULT_DOMAIN[0]), centerValue]);
+
+            texts = barPlot.content().selectAll("text"); // re-select after rendering
+            assert.strictEqual(d3.select(texts[0][i]).style("visibility"), "hidden", `label for bar with index ${i} is hidden`);
+          });
```
bluong: Why not loop over the `texts`?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+              texts = barPlot.content().selectAll("text");
+              assert.strictEqual(texts.size(), dataset.data().length, "texts were repopulated by drawLabels after the update");
+              called = true; // for some reason, in phantomJS, `done` was being called multiple times and this caused the test to fail.
+              done();
+            }
+          };
+          dataset.data(dataset.data());
+          texts = barPlot.content().selectAll("text");
+          assert.strictEqual(texts.size(), 0, "texts were immediately removed");
+        });
```
bluong: Alternatively you can check that the bug did not happen (old labels are not on the DOM)

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+      });
+
+      describe(`retrieving Entities when ${orientation}`, () => {
+        const data = [
+          { base: -1, value: -1 },
+          { base: 0, value: 0 },
+          { base: 1, value: 1 }
+        ];
+        const DEFAULT_DOMAIN = [-2, 2];
```
bluong: Is the value here associated with the above `data` var?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+          it("considers bars cut off by the value scale", () => {
+            valueScale.domain([-0.5, 0.5]);
+            testEntityNearest();
+          });
+
+          it("considers bars cut off by the base scale", () => {
+            baseScale.domain([-0.8, 0.8]);
+            testEntityNearest();
+          });
```
bluong: How do the above cut off the bars?

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+              x: barPlot.width() / 2,
+              y: barPlot.height() / 2
+            };
+            const nearestEntity = barPlot.entityNearest(centerOfPlot);
+            assert.isUndefined(nearestEntity, "returns undefined when no bars are in view");
+          });
+        });
+
+        describe("retrieving the Entities at a given point", () => {
+          it("returns the Entity at a given point, or an empty array if no bars are there", () => {
```
bluong: Consider 2 different tests

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+
+        afterEach(function() {
+          if (this.currentTest.state === "passed") {
+            barPlot.destroy();
+            svg.remove();
+          }
+        });
+      });
+    });
+
```
bluong: Remove this newline.  Not a big deal.

test/plots/barPlotTests.ts
bluong: Oh I see

test/plots/barPlotTests.ts
bluong: Yes but if we make the assumption that `valueScale` works properly, then `domainMin` / `domainMax` can be assumed to be correct right?

test/plots/barPlotTests.ts
bluong: Do you think that number of rectangles will help?

test/plots/barPlotTests.ts
jtlan: It's the only `assert()` in the test, so it'd be a bit redundant with the test title itself.

test/plots/barPlotTests.ts
jtlan: But the point of this test is to check that the autoranging takes place based on the visible data.

test/plots/barPlotTests.ts
jtlan: Probably not very much, but this way checks for no elements and checks for *n* elements look the same.

test/plots/barPlotTests.ts
jtlan: Hmm, didn't think of that. That's much easier :)

test/plots/barPlotTests.ts
jtlan: The attribute is appended to the end to help distinguish.

test/plots/barPlotTests.ts
jtlan: Yeah, because the two values don't correctly identify as data values of the right type. In fact, they *aren't* of the right type, which is the entire point of the test -- this sort of scenario could arise easily in Javascript.

test/plots/barPlotTests.ts
jtlan: Made a mistake here.

test/plots/barPlotTests.ts
jtlan: Not sure, I found the test like this. If I wrote it I'd probably do something like `${n} foo`.

test/plots/barPlotTests.ts
jtlan: The later lines with `labelBoundingClientRect[valueSizeAttr]` won't work unless I do this, since Typescript doesn't consider that `valueSizeAttr` will always be either `"width"` or `"height"` and therefore equivalent to `labelBoundingClientRect.width`/`labelBoundingClientRect.height`.

test/plots/barPlotTests.ts
jtlan: Hmm, that might be easier to read.

test/plots/barPlotTests.ts
jtlan: Yeah, larger and smaller than the min and max values. I could write code to generate it programmatically, but then a reader would have to verify that *that* code is correct too, which is harder than just looking at the values and seeing that they're bigger than the data values.

test/plots/barPlotTests.ts
jtlan: Cuts the bars off partially since their positions are at `-1` and `1`. Again, possible to programmatically generate the domain, but it might be hard to follow.

test/plots/barPlotTests.ts
```diff
@@ -1 +1,783 @@
...
+          }
+        });
+      });
+
+      describe(`retrieving Entities when ${orientation}`, () => {
+        const data = [
+          { base: -1, value: -1 },
+          { base: 0, value: 0 },
+          { base: 1, value: 1 }
+        ];
```
bluong: Consider initializing variables in `describe` blocks only in `beforeEach` unless it is a non-object `const`.  This more secures the independency of tests

test/plots/barPlotTests.ts
jtlan: Didn't work due to weirdness in **coverage.html** involving multiple `Plottable` objects :(

test/plots/barPlotTests.ts
jtlan: Done. Not sure why this comment thread appears in a weird spot on the main PR page.

jtlan:
Closing because main PR merged.

PR #2952, opened 2015-11-10
XDragBoxLayerTests refactor

ztsai:
screen shot 2015-11-09 at 7 03 10 pm

part of #2623 
second pass; mostly `let` to `const`
add a test for setting `yExtent`

### DIFF
--- a/test/components/xDragBoxLayerTests.ts
+++ b/test/components/xDragBoxLayerTests.ts
@@ -1,23 +1,14 @@
 ///<reference path="../testReference.ts" />
 
-describe("Layer Components", () => {
+describe("SelectionBoxLayer", () => {
   describe("XDragBoxLayer", () => {
     describe("Basic Usage", () => {
-      let SVG_WIDTH = 400;
-      let SVG_HEIGHT = 300;
+      const SVG_WIDTH = 400;
+      const SVG_HEIGHT = 300;
 
-      let quarterTopLeftPoint = {
-        x: SVG_WIDTH / 4,
-        y: SVG_HEIGHT / 4
-      };
-      let middlePoint = {
-        x: SVG_WIDTH / 2,
-        y: SVG_HEIGHT / 2
-      };
-      let quarterBottomRightPoint = {
-        x: SVG_WIDTH * 3 / 4,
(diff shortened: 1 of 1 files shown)


test/components/xDragBoxLayerTests.ts
```diff
@@ -1,23 +1,14 @@
 ///<reference path="../testReference.ts" />
 
-describe("Layer Components", () => {
+describe("SelectionBoxLayer", () => {
```
jtlan: Not sure this makes sense? What are the `SelectionBoxLayer` tests called then?

I notice that the `SelectionBoxLayer` tests are grouped under `"Interactive Components"`; ironically, that's a good grouping title for these tests, but not the `SelectionBoxLayer` ones!

test/components/xDragBoxLayerTests.ts
```diff
@@ -1,23 +1,14 @@
...
-      let quarterBottomRightPoint = {
-        x: SVG_WIDTH * 3 / 4,
-        y: SVG_HEIGHT * 3 / 4
-      };
+      const SVG_WIDTH = 400;
+      const SVG_HEIGHT = 300;
+
+      const quarterTopLeftPoint = { x: SVG_WIDTH / 4, y: SVG_HEIGHT / 4 };
+      const middlePoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 2 };
+      const quarterBottomRightPoint = { x: SVG_WIDTH * 3 / 4, y: SVG_HEIGHT * 3 / 4 };
```
jtlan: I think these are easier to read non-inlined.

test/components/xDragBoxLayerTests.ts
ztsai: I thought we layer describe blocks by class inheritance, if not I will change everything to `Interactive Components`

jtlan: Doesn't seem like we have a consistent rule for this. Let's go with "reasonable common ancestor" for now. For example, if you change `SelectionBoxLayer` you'll want to rerun the tests for its subclasses.

PR #2953, opened 2015-11-10
refactor YDragBoxLayer test

ztsai:
screen shot 2015-11-10 at 11 09 10 am

part of #2623 
second pass; mostly let to const
add a test for setting xExtent

### DIFF
--- a/test/components/yDragBoxLayerTests.ts
+++ b/test/components/yDragBoxLayerTests.ts
@@ -6,9 +6,18 @@ describe("SelectionBoxLayer", () => {
       const SVG_WIDTH = 400;
       const SVG_HEIGHT = 300;
 
-      const quarterTopLeftPoint = { x: SVG_WIDTH / 4, y: SVG_HEIGHT / 4 };
-      const middlePoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 2 };
-      const quarterBottomRightPoint = { x: SVG_WIDTH * 3 / 4, y: SVG_HEIGHT * 3 / 4 };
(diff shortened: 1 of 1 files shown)


test/components/yDragBoxLayerTests.ts
```diff
@@ -1,23 +1,14 @@
 ///<reference path="../testReference.ts" />
 
-describe("Layer Components", () => {
+describe("SelectionBoxLayer", () => {
```
jtlan: Not sure this makes sense? What are the `SelectionBoxLayer` tests called then?

I notice that the `SelectionBoxLayer` tests are grouped under `"Interactive Components"`; ironically, that's a good grouping title for these tests, but not the `SelectionBoxLayer` ones!

test/components/yDragBoxLayerTests.ts
```diff
@@ -1,23 +1,14 @@
...
-      let quarterBottomRightPoint = {
-        x: SVG_WIDTH * 3 / 4,
-        y: SVG_HEIGHT * 3 / 4
-      };
+      const SVG_WIDTH = 400;
+      const SVG_HEIGHT = 300;
+
+      const quarterTopLeftPoint = { x: SVG_WIDTH / 4, y: SVG_HEIGHT / 4 };
+      const middlePoint = { x: SVG_WIDTH / 2, y: SVG_HEIGHT / 2 };
+      const quarterBottomRightPoint = { x: SVG_WIDTH * 3 / 4, y: SVG_HEIGHT * 3 / 4 };
```
jtlan: I think these are easier to read non-inlined.

Issue #1998, opened 2015-05-08
Legend text is truncated when given a trailing newline


istarnx:
![screen shot 2015-05-07 at 7 48 29 pm](https://cloud.githubusercontent.com/assets/11776100/7529769/6d77a872-f4f2-11e4-94eb-37db5fa89d90.png)

In the screenshot, the legend entry's text is "dataset1\n". I would expect trailing whitespace to be ignored.

crmorford:
Testing Notes:
* Have 3 legend entries all with newlines, make sure top/middle/bottom all show up the same
* Have 3 legend entries, bottom with newline, make sure it shows up correctly
* Works as expected for horizontal and vertical legends
* Works for short words and long words / sentences
* Works with other types of whitespace at end (space/tab/etc)

aicioara:
Repro:
http://jsfiddle.net/f2L94hg2/1/
http://jsfiddle.net/f2L94hg2/

Note that whitespace `\s` `\t` is trimmed. `\n` is not.

aicioara:
Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

CalvinFernandez:
#3172 should provide a means of resolution

PR #2959, opened 2015-11-11
Sanitize newlines in category scale domains

### LINKED ISSUE #2958: [Axes.Category] Blows up when passed newlines
`Axes.Category` fails to render when its `Scale` has domain values that include newline characters (`\n`).

**repro**: http://jsfiddle.net/fw39ca6q/
**actual**: Nothing is drawn, and `Uncaught Error: SingleLineWrapper is designed to work only on single line` appears in the console.
**expected**: `Axes.Category` either:
* Ignores the newlines, replacing them with spaces
* respects the newlines if possible

TheBeruriahIncident:
**Issue:** Newlines in the domain of category scale would cause an exception, as the category axis uses SingleLineWrapper.
**Fix:** The more general Wrapper (not SingleLineWrapper) is used.

**Before:** https://jsfiddle.net/ux9w72do/
**After:** https://jsfiddle.net/yvaayb38/

Close #2958

### DIFF
--- a/src/scales/categoryScale.ts
+++ b/src/scales/categoryScale.ts
@@ -142,7 +142,14 @@ module Plottable.Scales {
     }
 
     protected _getDomain() {
-      return this._d3Scale.domain();
+      let newlineRemover = function(domain) {
+        if (typeof domain !== "string") {
+          return domain;
+        }
+        return domain.replace(/\n/g, " ");
+      };
+
+      return this._d3Scale.domain().map(newlineRemover);
     }
 
     protected _setBackingScaleDomain(values: string[]) {
--- a/src/scales/categoryScale.ts
+++ b/src/scales/categoryScale.ts
@@ -142,14 +142,7 @@ module Plottable.Scales {
     }
 
     protected _getDomain() {
-      let newlineRemover = function(domain) {
-        if (typeof domain !== "string") {
-          return domain;
-        }
-        return domain.replace(/\n/g, " ");
-      };
-
-      return this._d3Scale.domain().map(newlineRemover);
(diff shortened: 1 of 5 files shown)


src/scales/categoryScale.ts
```diff
@@ -142,7 +142,14 @@ module Plottable.Scales {
...
     protected _getDomain() {
-      return this._d3Scale.domain();
+      let newlineRemover = function(domain) {
+        if (typeof domain !== "string") {
+          return domain;
+        }
+        return domain.replace(/\n/g, " ");
+      };
+
+      return this._d3Scale.domain().map(newlineRemover);
```
jtlan: I'm not sure it's a good idea to modify the domain. This change would produce some unexpected behavior:
```Javascript
var originalDomain = ["two\nlines"];
categoryScale.domain(originalDomain);
assert.deepEqual(categoryScale.domain(), originalDomain); // fails
```

jtlan:
It seems like an easier way to fix this is to change `_wrapper` in `Axes.Category` to a `SVGTypewriter.Wrappers.Wrapper` instead of a `SVGTypewriter.Wrappers.SingeLineWrapper`, which also has the benefit of preserving the user-inserted newlines.

crmorford:
i wonder if this fixes #1998 in addition to #2958 (will test)

jtlan:
Probably won't fix #1998, because that would require a change to `Legend`. Maybe `Legend` is best served by the original fix (replace newlines with spaces)?

crmorford:
some examples: http://jsfiddle.net/fw39ca6q/1/

PR #2960, opened 2015-11-11
Row/Column + Padding/Weight don't accept Infinity or negative numbers

### LINKED ISSUE #2896: [Table] rowPadding() and columnPadding() don't reject bad values
Specifically, `Infinity` and negative values can be put in, resulting in odd behavior. It's possible that users might want negative padding for some edge cases, but I think that's unlikely.

### LINKED ISSUE #2897: [Table] rowWeight() and columnWeight() don't reject bad values
In particular, they accept negative and infinite values, which result in bad behavior and errors.

crmorford:
fixes #2896 
fixes #2897 

http://jsfiddle.net/9f87y5hm/
uncomment in window.onload() to test all things i'm claiming to fix with this PR

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -6106,7 +6106,7 @@ var Plottable;
                 if (rowPadding == null) {
                     return this._rowPadding;
                 }
-                if (rowPadding == Infinity || rowPadding < 0) {
+                if (rowPadding === Infinity || rowPadding < 0) {
                     throw Error("rowPadding must be a positive finite value");
                 }
                 this._rowPadding = rowPadding;
@@ -6117,7 +6117,7 @@ var Plottable;
                 if (columnPadding == null) {
                     return this._columnPadding;
                 }
-                if (columnPadding == Infinity || columnPadding < 0) {
+                if (columnPadding === Infinity || columnPadding < 0) {
                     throw Error("columnPadding must be a positive finite value");
                 }
                 this._columnPadding = columnPadding;
@@ -6128,7 +6128,7 @@ var Plottable;
                 if (weight == null) {
                     return this._rowWeights[index];
                 }
-                if (weight == Infinity || weight < 0) {
+                if (weight === Infinity || weight < 0) {
                     throw Error("rowWeight must be a positive finite value");
                 }
                 this._rowWeights[index] = weight;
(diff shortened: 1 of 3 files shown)


plottable.js
```diff
@@ -6106,6 +6106,9 @@ var Plottable;
                 if (rowPadding == null) {
                     return this._rowPadding;
                 }
+                if (rowPadding === Infinity || rowPadding < 0) {
```
ztsai: what about `NaN` and number strings?
there's a method called `Utils.Math.isValidNumber` that checks for invalid/ non-finite numbers

plottable.js
```diff
@@ -6106,6 +6106,9 @@ var Plottable;
...
+                    throw Error("rowPadding must be a positive finite value");
```
ztsai: nitpicking, but 0 is a valid input, so `non-negative` is a better description than `positive`

test/components/tableTests.ts
```diff
@@ -471,6 +471,38 @@ describe("Tables", () => {
         table.destroy();
         svg.remove();
       });
+
+      it("rowPadding and columnPadding error on invalid input", () => {
+        const svg = TestMethods.generateSVG();
+        const component0 = new Plottable.Component();
+        const component1 = new Plottable.Component();
+        const component2 = new Plottable.Component();
+        const component3 = new Plottable.Component();
```
ztsai: variables are not used other than in table declaration, maybe just create components inline?

test/components/tableTests.ts
```diff
@@ -471,6 +471,30 @@ describe("Tables", () => {
         table.destroy();
         svg.remove();
       });
+
+      it("rowPadding and columnPadding error on invalid input", () => {
+        const svg = TestMethods.generateSVG();
+        const table = new Plottable.Components.Table([[new Plottable.Component(), new Plottable.Component()],
+                                                      [new Plottable.Component(), new Plottable.Component()]]);
+        assert.throws(() => table.columnPadding(-1), "columnPadding must be a non-negative finite value");
```
ztsai: sorry, should have caught this earlier but the proper way to assert error is
`(<any> assert).throws(() => c.foo(), Error, "Error message", "The reason");`
with a hackhack comment referring #2661

also, can you test for `NaN` and number string?

jtlan:
Looking good so far; I tested with `0` too, which should be valid for both weights and paddings.

PR #2961, opened 2015-11-11
Passing null to InterpolatedColorLegend formatter returns old formatter

### LINKED ISSUE #2802: InterpolatedColor.formatter can be set to null
`InterpolatedColor.formatter` can be set to null by using `InterpolatedColor.formatter(null)`. A null formatter raises errors when drawing labels in InterpolatedColor.

repro: http://jsfiddle.net/yunhaolucky/qnmcsmx7/
**Actual**: `InterpolatedColor.formatter` is set to null and an error is thrown when drawing labels.
**Expected**: Do not allow user to set formatter to null. If user try to set formatter to null by running  `InterpolatedColor.formatter(null)`, do not change the formatter and return the current formatter.

agnanachandran:
(fixes #2802)


agnanachandran:
Woops accident; meant to make PRs to our fork only (no plans to make PRs to this project)

jtlan:
Why not? This looks like a good fix.

agnanachandran:
We're doing this as part of a school project; we may consider merging in fixes later and whatnot, but we need permission from our prof. to do so.

Issue #2625, opened 2015-08-17
[Testing Audit] Gridline, Legend, InterpolatedColorLegend, Labels


crmorford:

ztsai:
- [ ] Gridline
- [ ] Legend
- [ ] InterpolatedColorLegend
- [ ] Label

PR #2967, opened 2015-11-12
refactor gridlinesTests

ztsai:
screen shot 2015-11-12 at 1 24 15 pm

test clean up
part of #2625

### DIFF
--- a/test/components/gridlinesTests.ts
+++ b/test/components/gridlinesTests.ts
@@ -1,62 +1,62 @@
 ///<reference path="../testReference.ts" />
 
 describe("Gridlines", () => {
+  const SVG_WIDTH = 640;
+  const SVG_HEIGHT = 480;
+  let svg: d3.Selection<void>;
+  let xScale: Plottable.Scales.Linear;
+  let yScale: Plottable.Scales.Linear;
+  let gridlines: Plottable.Components.Gridlines;
 
-  it("Scale ranges are set to the Gridlines dimensions when layout is computed", () => {
-    let svg = TestMethods.generateSVG(640, 480);
-    let xScale = new Plottable.Scales.Linear();
+  beforeEach(() => {
+    svg = TestMethods.generateSVG(640, 480);
+    xScale = new Plottable.Scales.Linear();
     xScale.domain([0, 10]);
 
-    let yScale = new Plottable.Scales.Linear();
+    yScale = new Plottable.Scales.Linear();
     yScale.domain([0, 10]);
+    gridlines = new Plottable.Components.Gridlines(xScale, yScale);
+  });
 
-    let gridlines = new Plottable.Components.Gridlines(xScale, yScale);
+  it("sets ranges of scales to the Gridlines dimensions when layout is computed", () => {
     gridlines.renderTo(svg);
 
-    assert.deepEqual(xScale.range(), [0, 640], "x scale range extends to the width of the svg");
-    assert.deepEqual(yScale.range(), [480, 0], "y scale range extends to the height of the svg");
+    assert.deepEqual(xScale.range(), [0, SVG_WIDTH], "x scale range extends to the width of the svg");
+    assert.deepEqual(yScale.range(), [SVG_HEIGHT, 0], "y scale range extends to the height of the svg");
 
     svg.remove();
   });
 
-  it("Gridlines and axis tick marks align", () => {
-    let svg = TestMethods.generateSVG(640, 480);
-    let xScale = new Plottable.Scales.Linear();
-    xScale.domain([0, 10]); // manually set domain since we won't have a renderer
+  it("aligns Gridlines and axis tick marks", () => {
     let xAxis = new Plottable.Axes.Numeric(xScale, "bottom");
-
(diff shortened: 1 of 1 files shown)


test/components/gridlinesTests.ts
```diff
@@ -1,62 +1,62 @@
...
-    let xScale = new Plottable.Scales.Linear();
+  const SVG_WIDTH = 640;
+  const SVG_HEIGHT = 480;
+  let svg: d3.Selection<void>;
+  let xScale: Plottable.Scales.Linear;
+  let yScale: Plottable.Scales.Linear;
+  let gridlines: Plottable.Components.Gridlines;
+
+  beforeEach(() => {
+    svg = TestMethods.generateSVG(640, 480);
```
jtlan: Should probably use the `const`s here, but I don't think this test is even dependent on the dimensions, so we can probably go with the defaults.

test/components/gridlinesTests.ts
```diff
@@ -1,62 +1,62 @@
...
+  });
 
-    let gridlines = new Plottable.Components.Gridlines(xScale, yScale);
+  it("sets ranges of scales to the Gridlines dimensions when layout is computed", () => {
     gridlines.renderTo(svg);
 
-    assert.deepEqual(xScale.range(), [0, 640], "x scale range extends to the width of the svg");
-    assert.deepEqual(yScale.range(), [480, 0], "y scale range extends to the height of the svg");
+    assert.deepEqual(xScale.range(), [0, SVG_WIDTH], "x scale range extends to the width of the svg");
+    assert.deepEqual(yScale.range(), [SVG_HEIGHT, 0], "y scale range extends to the height of the svg");
```
jtlan: If we get rid of `SVG_WIDTH` and `SVG_HEIGHT`, probably use `numAttr(svg, ...)` here.

test/components/gridlinesTests.ts
```diff
@@ -1,62 +1,62 @@
...
 
     svg.remove();
   });
 
-  it("Gridlines and axis tick marks align", () => {
-    let svg = TestMethods.generateSVG(640, 480);
-    let xScale = new Plottable.Scales.Linear();
-    xScale.domain([0, 10]); // manually set domain since we won't have a renderer
+  it("aligns Gridlines and axis tick marks", () => {
```
jtlan: This test is probably more accurately "uses the TickGenerators of its Scales" -- that's what's actually going on, and that way we don't need to make Axes.

jtlan:
Can we add some tests that the `Gridlines` update when its `Scale`s update?

test/components/gridlinesTests.ts
```diff
@@ -1,63 +1,116 @@
...
+    assert.deepEqual(yScale.range(), [TestMethods.numAttr(svg, "height"), 0], "y scale range extends to the height of the svg");
 
     svg.remove();
   });
 
-  it("Gridlines and axis tick marks align", () => {
-    let svg = TestMethods.generateSVG(640, 480);
-    let xScale = new Plottable.Scales.Linear();
-    xScale.domain([0, 10]); // manually set domain since we won't have a renderer
+  it("uses the TickGenerators of its Scales", () => {
```
jtlan: Can you change this test to not use `Axes`? Basically, we should be able to render the `Gridlines`, then check that the lines are in the places specified by `xScale.ticks()` and `yScale.ticks()`.

test/components/gridlinesTests.ts
```diff
@@ -1,63 +1,116 @@
...
+      assert.closeTo(y, yScale.scale(yTicks[i]), 1, "y gridline is updated");
+    });
+    svg.remove();
+  });
+
+  it("throws error on non-Quantitative Scales", () => {
+    const categoryScale = new Plottable.Scales.Category();
+    // HACKHACK #2661: Cannot assert errors being thrown with description
+    (<any> assert).throw(() => new Plottable.Components.Gridlines(<any> categoryScale, null), Error,
+      "xScale needs to inherit from Scale.QuantitativeScale", "can't not set xScale to category scale");
```
jtlan: Probably "can't" instead of "can't not"?

ztsai: LOL

test/components/gridlinesTests.ts
```diff
@@ -1,63 +1,116 @@
...
+
+  it("updates gridlines when scales update", () => {
+    gridlines.renderTo(svg);
+
+    let xGridlines = gridlines.content().select(".x-gridlines").selectAll("line");
+    let xTicks = xScale.ticks();
+    assert.strictEqual(xGridlines.size(), xTicks.length, "There is an x gridline for each x tick");
+    xGridlines.each(function(gridline, i) {
+      const x = TestMethods.numAttr(d3.select(this), "x1");
+      assert.closeTo(x, xScale.scale(xTicks[i]), 1, "x gridline drawn on ticks");
```
jtlan: This is the logic that should be in the "uses tick generators" test.

Also, we should use either `strictEqual`, or `closeTo` with `window.Pixel_CloseTo_Requirement`.

Issue #2968, opened 2015-11-13
[Dispatcher] internal fields misleadingly named


jtlan:
In `Dispatcher` there is this field:
```Typescript
protected _eventToCallback: { [eventName: string]: (e: Event) => any; } = {};
```

But also these:
```Typescript
protected _callbacks: Utils.CallbackSet<Function>[] = [];

protected _setCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function);
protected _unsetCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function);
```

The naming here is misleading since there are actually *two* kinds of callbacks involved.`_eventToCallback` is closer to "event to processing-function-*of*-`Dispatcher`", and `_set`/`_unsetCallback` is closer to "set/unset callbacks to be *invoked by* `Dispatcher`.

Proposed renamings:
`_eventToCallback` --> `_eventToProcessingFunction`
`_callbacks` --> `_callbackSets`
`_setCallback` --> `_addCallback`
`_unsetCallback` --> `_removeCallback`

It's also likely that `_callbackSets` should be privatized and the signatures of `_set`/`_unsetCallback` changed:
```Typescript
private _callbackSets: { [eventName: string]: Utils.CallbackSet<Function>; }

protected _setCallback(eventName: string, callback: Function);
protected _unsetCallback(eventName: string, callback: Function);
```

This seems more consistent.

PR #2969, opened 2015-11-13
[Dispatcher tests] Clean up parent-level Dispatcher tests.

jtlan:
A more thorough cleaning (removing `<any>`s, etc) probably depends on https://github.com/palantir/plottable/issues/2968.

Last checkbox of https://github.com/palantir/plottable/issues/2628, at least until #2968 gets done.

### DIFF
--- a/test/dispatchers/dispatcherTests.ts
+++ b/test/dispatchers/dispatcherTests.ts
@@ -2,63 +2,65 @@
 
 describe("Dispatchers", () => {
   describe("Dispatcher", () => {
-    it("_connect() and _disconnect()", () => {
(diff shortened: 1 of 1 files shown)


ztsai:
looks good
do you think that unsetting one callback doesn't disconnect other callbacks on the same event should be tested here?

jtlan:
Seems worthwhile, actually. I'll add that.

jtlan:
Actually, because of the weird architecture in `Dispatcher`, it's hard to write a useful test for that. I'll add a note to #2968 to add a test at that time.

ztsai:
make sense

PR #2972, opened 2015-11-16
refactor InterpolatedColorLegendTests

ztsai:
part of #2625 
- break down tests into subgroups
- change all `_content` to `.content()`
- refactor similar tests into loop
- update test names
- `let` -> `const`

screen shot 2015-11-16 at 3 11 15 pm


jtlan:
Possibly worth refining `"Basic Usage"` into `Input validation` and `rendering`, but that can be done later on.

Issue #2336, opened 2015-06-15
Domain for Linear Scale with stringy values is computed badly


aicioara:
**Expected**: http://jsfiddle.net/fttawduw/2/
**Actual**: http://jsfiddle.net/fttawduw/1/

jtlan:
Same issue from #2103.

aicioara:
@jtlan duplicate?

jtlan:
Duplicate, but this is a narrower repro.

PR #2974, opened 2015-11-17
[Testing Audit] Clean up Scale and QuantitativeScale tests

jtlan:
**Summary**:
* Split tests for `Scale` and `QuantitativeScale` into two files.
* Added mock-`Scale` class, allowing addition of tests for more methods.
* Added additional tests cases for `QuantitativeScale`.
* Corrected a test on `QuantitativeScale`, then skipped it because it doesn't pass (due to #2336).

Current test battery for `Scale`:
screen shot 2015-11-17 at 12 25 03 am

Current test battery for `QuantitativeScale`:
screen shot 2015-11-17 at 12 25 08 am

Should finish off #2640.


test/scales/scaleTests.ts
```diff
@@ -1,58 +1,184 @@
...
+        scale.addIncludedValuesProvider(provider);
+
+        const expectedIncludedValues = [providerValue];
+        assert.deepEqual(scale.includedValues(), expectedIncludedValues, "only includes the value once");
+      });
     });
 
+    describe("autodomaining", () => {
+      it("overrides a set domain if explicitly autodomained", () => {
+        const expectedDomain = [{}];
```
ztsai: just curious, why `[{}]`?

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,68 @@
...
+      it("ignores infinite values when computing the extent", () => {
+        const values = [Infinity, -Infinity, 1, 11];
+        const scale = new Plottable.QuantitativeScale();
+        const computedExtent = scale.extentOfValues(values);
+
+        assert.deepEqual(computedExtent, [1, 11], "the extent is the minimum and the maximum values");
+      });
+    });
+
+    describe("padding", () => {
```
ztsai: make sense

test/scales/scaleTests.ts
jtlan: I wanted an object since two objects can't be strict equal to each other, so the guards against a case where the logic just happens to produce the correct values. But I think this is too confusing, so I'll change it to just declare the expected auto and explicitly set values.

test/scales/quantitativeScaleTests.ts
```diff
@@ -0,0 +1,68 @@
...
+      it("rejects negative values for padding", () => {
+        const scale = new Plottable.QuantitativeScale();
+        assert.throws(() => scale.padProportion(-0.05), Error);
```
jtlan: I'll check the emitted message, but since this is the only `assert` in this `it()` having a message for the `assert` itself seems redundant.

Issue #2622, opened 2015-08-17
[Testing Audit] Drawers and Animators


crmorford:

jtlan:
Probably going to do this as two PRs, one for `Drawer`s and one for `Animator`s.

jtlan:
Completed by #2976, #2978, and #2987.

PR #2976, opened 2015-11-18
[Testing Audit] parent Drawer tests

jtlan:
Only the refactor for the parent `Drawer` class' tests (I'll do the subclasses in a separate PR).

Once again, a number of methods were not being tested, and are now being tested using the power of a mock class.

screen shot 2015-11-18 at 12 28 15 am

Part of #2622.

### DIFF
--- a/test/drawers/drawerTests.ts
+++ b/test/drawers/drawerTests.ts
@@ -62,6 +62,26 @@ describe("Drawers", () => {
       assert.isFalse(document.body.contains(<any> svg.node()), "renderArea was removed from the DOM");
     });
 
+    it("correctly computes the total draw time", () => {
+      function makeFixedTimeAnimator(totalTime: number) {
+        return <Plottable.Animator> {
+          animate: () => null,
+          totalTime: () => totalTime
+        };
+      }
+
+      const animationTimes = [10, 20];
+      const drawSteps = animationTimes.map((time) => {
+        return {
+          attrToProjector: <Plottable.AttributeToProjector> {},
+          animator: makeFixedTimeAnimator(time)
+        };
+      });
+      const totalTime = drawer.totalDrawTime([], drawSteps);
+      const expectedTotalTime = d3.sum(animationTimes);
+      assert.strictEqual(totalTime, expectedTotalTime, "returned the total time taken by all Animators");
+    });
+
     it("draws elements accoding to a specified drawStep", () => {
       const svg = TestMethods.generateSVG();
       const data = ["A", "B", "C"];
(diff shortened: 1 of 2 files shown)


test/drawers/drawerTests.ts
```diff
@@ -1,137 +1,206 @@
...
+          return oldTimeout(f, time, args);
+        };
+      });
+
+      after(() => {
+        Plottable.Utils.Window.setTimeout = oldTimeout;
+      });
+
+      beforeEach(() => {
+        timings = [];
```
ztsai: isn't `timngs` initialized at line 123 already?

test/drawers/drawerTests.ts
```diff
@@ -1,137 +1,206 @@
...
+            attrToProjector: <Plottable.AttributeToProjector> {},
+            animator: makeFixedTimeAnimator(time)
+          };
+        });
+        const totalTime = drawer.totalDrawTime([], drawSteps);
+        const expectedTotalTime = d3.sum(animationTimes);
+        assert.strictEqual(totalTime, expectedTotalTime, "returned the total time taken by all Animators");
+      });
+
+      it("drawer timing works as expected for null animators", () => {
```
ztsai: test name doesn't read semantically with `it`

test/drawers/drawerTests.ts
```diff
@@ -1,137 +1,206 @@
...
+        const a1 = new Plottable.Animators.Null();
+        const a2 = new Plottable.Animators.Null();
+        const ds1: Plottable.Drawers.DrawStep = {attrToProjector: {}, animator: a1};
+        const ds2: Plottable.Drawers.DrawStep = {attrToProjector: {}, animator: a2};
+        const steps = [ds1, ds2];
+        drawer.draw([], steps);
+        assert.deepEqual(timings, [0, 0], "setTimeout called twice with 0 time both times");
+      });
+
+      it("drawer timing works for non-null animators", (done) => {
```
ztsai: test name

test/drawers/drawerTests.ts
```diff
@@ -1,137 +1,206 @@
...
+        };
+        const callback2 = () => {
+          assert.isTrue(callback1Called, "callback2 called after callback 1");
+          callback2Called = true;
+        };
+        const callback3 = () => {
+          assert.isTrue(callback2Called, "callback3 called after callback 2");
+          done();
+        };
+        const a1 = new MockAnimator(20, callback1);
```
ztsai: factor out `20`, `10`, and `0` into variables?

test/scales/linearScaleTests.ts
```diff
@@ -349,6 +349,12 @@ describe("Scales", () => {
     });
 
     describe("Plot interaction", () => {
+      function createMockDrawer(dataset: Plottable.Dataset) {
```
ztsai: is this a part of this commit?

test/drawers/drawerTests.ts
jtlan: You're right. It probably shouldn't be initialized there, since it needs to get reset before each run.

test/scales/linearScaleTests.ts
jtlan: Had to move it here, because `createMockDrawer` used to be declared in the global namespace. Previously these tests referenced it from there, so when I deleted it from the global namespace I had to move it inside this `describe()`.

test/drawers/drawerTests.ts
jtlan: Done.

PR #2982, opened 2015-11-19
remove wheel plot

### LINKED ISSUE #2957: [Future] Remove Plots.Wheel
Development of it was stopped halfway, and it appears that the users who initially asked for it wound up using `Plots.Rectangle` for its use case instead. If we go to 2.0.0, we should roll it back until it is ready.

ztsai:
closes #2957 
roll back wheel plot


jtlan:
Let's track all the 2.0.0 changes on this page: https://github.com/palantir/plottable/wiki/2.0.0-Changes.

PR #2910, opened 2015-10-21
revert insideSVG check for pointer interaction

ztsai:
revert #2416(#2821) to unblock slate

before 2416: http://jsfiddle.net/crjz5531/8/
after 2416: http://jsfiddle.net/crjz5531/9/
now: http://jsfiddle.net/crjz5531/10/

PR #2983, opened 2015-11-19
consider overlay outside of component for pointerInteractions

ztsai:
pointerIneraction part of #2416
revert #2910


ztsai:
wrong base branch

PR #2989, opened 2015-11-20
refactor pointerInteractionTests

ztsai:
#2628 
- loop through mouse event and pointer event
- use const when possible

screen shot 2015-11-19 at 8 10 39 pm


jtlan:
Usually I prefer to put the `describe()` inside the `forEach()`, and not the other way around, so we get this:
```
describe
  describe Mode-A
    test1
    test2
  describe mode-B
    test1
    test2
```

PR #2996, opened 2015-11-23
add preventDefault API point on keyInteraction

### LINKED ISSUE #2601: `Interactions.Key` should prevent default action.
After playing around with `Interactions.Key`, I found that if I map the "backspace" key and then use it, the browser will go back to the previous page.  This is likely because the default action is not being prevented.

ztsai:
demo: http://jsfiddle.net/ztsai/4mac484s/
command + p would open a printing dialog for example, after enabling preventDefault, command + p does not trigger printing dialog

`preventDefault(): boolean` gets the correct setting
`preventDefault(preventDefault: boolean): Key` sets preventDefault setting and returns the calling object

closes #2601

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -4486,10 +4486,19 @@ declare module Plottable {
             private _downedKeys;
             private _keyDownCallback;
             private _keyUpCallback;
+            private _preventDefault;
             protected _anchor(component: Component): void;
             protected _unanchor(): void;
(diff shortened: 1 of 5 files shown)


jtlan:
Fixed version pointing to correct remote: http://jsfiddle.net/L6byud3y/

jtlan:
Closing for now to get the 2.0 RC out.

PR #2997, opened 2015-11-23
set return types for Scale._getDomain() and Scale._getRange()

### LINKED ISSUE #2971: [Scale] Return types of _getDomain() and _getRange() left unspecified
`_getDomain()` should return `D[]`, and `_getRange()` should return `R[]`. This shouldn't be a breaking change because if they didn't return those values, the `Scale` wouldn't work anyway.

ztsai:
closes #2971


jtlan:
We shouldn't have to explicitly specify the return types on the `Scale` subclasses, since the generics should propagate.

PR #2999, opened 2015-11-24
Fix implicit-any in keyInteractionTests

jtlan:


ztsai:
actually, the `option` parameter is not used in this test, do you just want to remove it?
(the event calls that used `options` is moved out of that describe block)

PR #3000, opened 2015-11-24
[Dispatchers] Refactor architecture for clarity

### LINKED ISSUE #2968: [Dispatcher] internal fields misleadingly named

jtlan:
Refactored how `Dispatcher` registers events. Previously it used these methods:
```Typescript
// OLD DESIGN
protected _setCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function): void;
protected _unsetCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function): void;
```
This seemed kind of odd, since there was a method that modified non-instance fields, and subclasses had to define a new `CallbackSet` for each type of event they wanted to add listeners for, and then call `.callCallbacks()` manually. This seemed silly.

The refactored version:
```Typescript
// NEW DESIGN
protected _addCallbackForEvent(eventName: string, callback: Function): void;
protected _removeCallbackForEvent(eventName: string, callback: Function): void;
protected _callCallbacksForEvent(eventName: string, ...args: any[]): void;
```
Now, subclasses register callbacks using a unique event-name, and the root `Dispatcher` class handles creating `CallbackSet`s. This change also meant we could remove the `protected` (and [misleadingly named](https://github.com/palantir/plottable/issues/2968)!) `_callbacks` array.

Summary of changes:
* ` _eventToCallback` --> `_eventToProcessingFunction`
* `_setCallback()` and `_unsetCallback()` replaced by `_addCallbackForEvent()` and `_removeCallbackForEvent()`.
* `_callCallbacksForEvent()` added.
* `protected` `_callbacks` array removed.
* `private` `CallbackSet`s removed and replaced with registration-strings.

Close #2968.

### DIFF
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -1,6 +1,6 @@
 module Plottable {
 export class Dispatcher {
-  protected _eventToCallback: { [eventName: string]: (e: Event) => any; } = {};
+  protected _eventToProcessingFunction: { [eventName: string]: (e: Event) => any; } = {};
   private _eventNameToCallbackSet: { [eventName: string]: Utils.CallbackSet<Function>; } = {};
   private _connected = false;
 
@@ -18,8 +18,8 @@ export class Dispatcher {
     if (this._connected) {
       return;
     }
-    Object.keys(this._eventToCallback).forEach((event: string) => {
-      let callback = this._eventToCallback[event];
+    Object.keys(this._eventToProcessingFunction).forEach((event: string) => {
+      let callback = this._eventToProcessingFunction[event];
       document.addEventListener(event, callback);
     });
     this._connected = true;
@@ -27,8 +27,8 @@ export class Dispatcher {
 
   private _disconnect() {
     if (this._connected && this._hasNoCallbacks()) {
-      Object.keys(this._eventToCallback).forEach((event: string) => {
-        let callback = this._eventToCallback[event];
+      Object.keys(this._eventToProcessingFunction).forEach((event: string) => {
+        let callback = this._eventToProcessingFunction[event];
         document.removeEventListener(event, callback);
       });
       this._connected = false;
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -1,14 +1,10 @@
 module Plottable {
 export class Dispatcher {
   protected _eventToCallback: { [eventName: string]: (e: Event) => any; } = {};
-  protected _callbacks: Utils.CallbackSet<Function>[] = [];
   private _eventNameToCallbackSet: { [eventName: string]: Utils.CallbackSet<Function>; } = {};
   private _connected = false;
 
   private _hasNoCallbacks() {
-    if (this._callbacks.some((callbackSet) => callbackSet.size > 0)) {
-      return false;
-    }
     const eventNames = Object.keys(this._eventNameToCallbackSet);
     for (let i = 0; i < eventNames.length; i++) { // for-loop so return can break out
       if (this._eventNameToCallbackSet[eventNames[i]].size !== 0) {
@@ -59,16 +55,6 @@ export class Dispatcher {
     if (callbackSet != null) {
       callbackSet.callCallbacks.apply(callbackSet, args);
     }
-  }
-
-  protected _setCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function) {
-    callbackSet.add(callback);
-    this._connect();
-  }
-
-  protected _unsetCallback(callbackSet: Utils.CallbackSet<Function>, callback: Function) {
-    callbackSet.delete(callback);
-    this._disconnect();
   }
 }
 }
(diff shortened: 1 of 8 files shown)


src/dispatchers/mouseDispatcher.ts
```diff
@@ -183,7 +180,7 @@ module Plottable.Dispatchers {
         let newMousePosition = this._translator.computePosition(event.clientX, event.clientY);
         if (newMousePosition != null) {
           this._lastMousePosition = newMousePosition;
-          callbackSet.callCallbacks(this.lastMousePosition(), event);
+          this._callCallbacksForEvent(eventName, this._lastMousePosition, event);
```
ztsai: either use `this.lastMousePosition()` or remove that function

src/dispatchers/touchDispatcher.ts
```diff
@@ -38,16 +38,14 @@ module Plottable.Dispatchers {
...
-      this._moveCallbacks = new Utils.CallbackSet<TouchCallback>();
-      this._endCallbacks = new Utils.CallbackSet<TouchCallback>();
-      this._cancelCallbacks = new Utils.CallbackSet<TouchCallback>();
-      this._callbacks = [this._moveCallbacks, this._startCallbacks, this._endCallbacks, this._cancelCallbacks];
-
-      this._eventToCallback["touchstart"] = (e: TouchEvent) => this._measureAndDispatch(e, this._startCallbacks, "page");
-      this._eventToCallback["touchmove"] = (e: TouchEvent) => this._measureAndDispatch(e, this._moveCallbacks, "page");
-      this._eventToCallback["touchend"] = (e: TouchEvent) => this._measureAndDispatch(e, this._endCallbacks, "page");
-      this._eventToCallback["touchcancel"] = (e: TouchEvent) => this._measureAndDispatch(e, this._cancelCallbacks, "page");
+      this._eventToProcessingFunction[Touch._TOUCHSTART_EVENT_NAME] =
```
ztsai: probably in a different commit, but all these events are `page` events so do we still need that extra parameter?

src/dispatchers/mouseDispatcher.ts
jtlan: ... strange, I don't remember changing this, but I'll put it back the way it was.

src/dispatchers/touchDispatcher.ts
jtlan: Which extra parameter? `_measureAndDispatch()` defaults to `"element"` so we need to pass `"page"`.

src/dispatchers/dispatcher.ts
```diff
@@ -1,42 +1,60 @@
...
   }
 
   private _connect() {
     if (this._connected) {
       return;
     }
-    Object.keys(this._eventToCallback).forEach((event: string) => {
-      let callback = this._eventToCallback[event];
+    Object.keys(this._eventToProcessingFunction).forEach((event: string) => {
+      let callback = this._eventToProcessingFunction[event];
```
ztsai: nitpicking, but processingFunction is a better variable name here

src/dispatchers/dispatcher.ts
```diff
@@ -1,42 +1,60 @@
...
     this._connected = true;
   }
 
   private _disconnect() {
-    if (this._connected && this._hasNoListeners()) {
-      Object.keys(this._eventToCallback).forEach((event: string) => {
-        let callback = this._eventToCallback[event];
+    if (this._connected && this._hasNoCallbacks()) {
+      Object.keys(this._eventToProcessingFunction).forEach((event: string) => {
+        let callback = this._eventToProcessingFunction[event];
```
ztsai: same here

src/dispatchers/touchDispatcher.ts
ztsai: I don't think `scope` parameter is necessary in  `_measureAndDispatch ` as for current implementation, it's doing additional check for nothing; but I am also okay with leaving it there to be consistent with the Mouse Dispatcher

jtlan: Seems like a refactor for later.

PR #2984, opened 2015-11-19
pointerInteraction should not trigger under overlay

ztsai:
pointerIneraction part of #2416 
revert #2910

PR #3009, opened 2015-12-01
Set _isTopLevelComponent based on whether the root node is an SVG

### LINKED ISSUE #3004: [Component] Top-level Components don't reset to non-top-level
`Component`s that are anchored directly under the root `<svg>` are considered "top-level". However, this "top-level"-ness isn't reset when the `Component` is then moved under a `Table` or `Group`:

**repro**: https://jsfiddle.net/m7t0a682/1/
**Actual**: Previously top-level label acts as though it's still anchored and centers itself in the SVG
**Expected**: `Component` acts the same as though it hadn't previously been anchored top-level

jtlan:
Previously was set to true and never reset!

Blocking changes on #2984.

Close #3004.

### DIFF
--- a/test/components/componentTests.ts
+++ b/test/components/componentTests.ts
@@ -182,7 +182,7 @@ describe("Component", () => {
       svg.remove();
     });
 
-    it("removes the parent when detaching", () => {
+    it("sets its parent to null when detaching", () => {
       let parent = new Plottable.Components.Group([c]);
       c.detach();
       assert.isNull(c.parent(), "parent removed upon detaching");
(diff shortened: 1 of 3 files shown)


test/components/componentTests.ts
```diff
@@ -397,8 +397,19 @@ describe("Component", () => {
       svg.remove();
     });
 
-    it("throws an error when computing layout when attached to non-root node using default arguments", () => {
-      let g = svg.append("g");
+    it("requires arguments when not anchored directly under the svg", () => {
+      const g = svg.append("g");
```
ztsai: lines 401 to 411 seem to be duplicate code?

jtlan: Needs to have been previously anchored to trigger the bug.

ztsai: Oh, I must have misread something, never mind

PR #3010, opened 2015-12-02
upgrade to ts 1.7.3

ztsai:


src/animators/easingAnimator.ts
```diff
@@ -72,7 +72,7 @@ module Plottable.Animators {
      * @param {number} startDelay The start delay in milliseconds.
      * @returns {Easing} The calling Easing Animator.
      */
-    public startDelay(startDelay: number): Easing;
+    public startDelay(startDelay: number): this;
     public startDelay(startDelay?: number): any {
```
adidahiya: why are so many method return types annotated as `any`? shouldn't it be `Easing | number` here (or `this | number` with TS 1.7.3)? was this code never updated to use union types?

jtlan: This last signature is the *implementation*, and has to be assignable to any of the previous signatures (called "overloads"), according to the spec:

>When a function has overload declarations, the overloads determine the call signatures of the type given to the function object and the function implementation signature (if any) must be assignable to that type.

https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#61-function-declarations

The problem is that `number | this` isn't assignable to either `number` or `this` according to the casting rules, so you'll get a compiler error ("`Overload signature is not compatible with function implementation.`"). We can't simply leave out the `any` either, because that causes the error "`No best common type exists among return expressions.`" Typescript's [handbook](http://www.typescriptlang.org/Handbook#functions-overloads) includes an example where the implementation signature returns `any`.

adidahiya: cool, makes sense. looks like a TypeScript flaw if we have to use `any` so much to achieve the nice typings for getter/setter methods like this.

jtlan:
@ztsai There's some that were left out:
* `Axes.Numeric`: `usesTextWidthApproximation()`
* `Interactions.PanZoom`: `xScales()`, `yScales()`, `minDomainExtent()`, and `maxDomainExtent()`
* `Plots.Line`: `autorangeSmooth()`, `downsamplingEnabled()`, `croppedRenderingEnabled()`
* `Plots.Pie`: `sectorValue()`, `innerRadius()`, `outerRadius()`
* `Plot`: `attr()`, `animated()`
* `Plots.Scatter`: `size()`, `symbol()`
* `Plots.StackedArea`: `downsamplingEnabled()` (which currently returns `Plots.Line`!)


I'd do a global search for "`): any`" afterwards just to make sure.

PR #3012, opened 2015-12-03
Add backing <rect> to catch events on Safari.

jtlan:
As referenced in https://github.com/palantir/plottable/pull/2984.

The top-level `Component` now adds an `opacity: 0` `<rect>` to the `<svg>` before anchoring, and removes it when detaching.

On **`develop`**: https://jsfiddle.net/j6c01b1e/
On **`develop-2.0`**: https://jsfiddle.net/xwpvdrns/
On **`safariEvents`**: https://jsfiddle.net/zs8tuctx/

Repro with all three event types:
* On **`develop`**: https://jsfiddle.net/324pawue/
* On **`develop-2.0`**: https://jsfiddle.net/5fhzscr9/
* On **`safariEvents`**: https://jsfiddle.net/yen85gLp/

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1368,6 +1368,7 @@ declare module Plottable {
         private _clipPathID;
         private _onAnchorCallbacks;
         private _onDetachCallbacks;
+        private static _SAFARI_EVENT_BACKING_CLASS;
         constructor();
         /**
          * Attaches the Component as a child of a given d3 Selection.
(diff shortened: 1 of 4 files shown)


ztsai:
looks good to me
but can you also test for the case where root Component changes?

jtlan:
I'll add that test.

jtlan:
Cross-checked on both mobile and desktop safari.

PR #3015, opened 2015-12-03
[Axis] Restore _computedWidth and _computedHeight

jtlan:
These fields were removed in #3011, but were not actually deprecated and so should not have been removed.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -1673,6 +1673,8 @@ declare module Plottable {
         protected _scale: Scale<D, number>;
         private _formatter;
         private _orientation;
+        protected _computedWidth: number;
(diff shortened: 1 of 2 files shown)


ztsai:
it's actually deprecated
it could be removed completely in `_computedHeight` and `_computedWeight`

jtlan:
Hmm, it actually looks like the values are cached, but never used. We should remove them.

jtlan:
Also marked as deprecated in https://github.com/palantir/plottable/releases/tag/v1.15.0.

jtlan:
This PR superseded by https://github.com/palantir/plottable/pull/3016.

PR #3020, opened 2015-12-05
[PiePlot] bug fix in pieplot where scale changes are not reflected upon dataset…

jimexist:
Fix a bug where the pieplot will render `NaN`s when the scale is included and then dataset is updated. The cause is that the `_updatePieAngles` function is called before `super. _onDatasetUpdate` which updated the scale with new data items.

However I cannot find a simple fix to this issue easily, because i believe `_updatePieAngles` should be called between `_updateExtends` and `render` and they are both in the same `super. _onDatasetUpdate` call. The current pull request includes a hack to call `render` twice, and it only serves a purpose to show the replicating test code setup.


jtlan:
@Jimexist Can you provide a repro (via JSFiddle, etc) for the bug?

Also, can you fill out the [contributor license agreement](https://github.com/palantir/plottable/blob/develop/Plottable_Personal_Contributor_License_Agreement.pdf?raw=true) and send it to opensource@palantir.com? Thanks!

Jimexist:
@jtlan OK the jsFiddle link is https://jsfiddle.net/etareduce/p2Lab5cv/2/, and I'll send over agreement later.

jtlan:
@Jimexist, we've received the agreement. Thanks!

test/plots/piePlotTests.ts
```diff
@@ -143,6 +143,46 @@ describe("Plots", () => {
       });
     });
 
+    describe("Rendering with scale", () => {
```
jtlan: How come this is a separate `describe()` instead of just an `it()` under an existing `describe()`?

It should be possible to override the old `sectorValue()` call with the new one.

jimexist: i updated the branch to not introduce a new `describe` block

adidahiya:
lgtm, thanks @Jimexist!

adidahiya:
regarding double-rendering: it is not ideal and @endrjuskr was looking into alternative approaches. if either of you find a suitable alternative we can update the implementation.

PR #3029, opened 2015-12-08
[QuantitativeScale] Compute padded domain based off of candidate doma…

### LINKED ISSUE #3019: [QuantitativeScale] Locks up if both ends of domain set to same value
Specific conditions that cause this:
* `QuantitativeScale` has more than one distinct included value (doesn't matter if they come from the same `IncludedValuesProvider` or not)
* `QuantitativeScale` has a nonzero `padProportion()`.

Under these conditions, if both ends of the `QuantitativeScale`'s domain are set to the same value, it will be unable to `autoDomain()` out of it.

**repro**: https://jsfiddle.net/sg0zqqwp/

jtlan:
…in, not current domain

Since scale() and invert() calls are used to compute the padded domain, the padded size
depends on the current domain and range, which causes problems if the domain or range
have been set to span zero width. This change temporarily sets the domain when computing
the padded domain.

Also took the opportunity to fix a test for snappingDomainEnabled to use less fragile numbers.

Close #3019.

### DIFF
--- a/test/scales/timeScaleTests.ts
+++ b/test/scales/timeScaleTests.ts
@@ -128,7 +128,7 @@ describe("Scales", () => {
       });
 
       it("doesn't lock up if a zero-width domain is set while there are value providers", () => {
-        let scale = new Plottable.Scales.Time();
+        const scale = new Plottable.Scales.Time();
         scale.padProportion(0.1);
         const provider = () => [new Date(2000, 5, 5), new Date(2000, 5, 6)];
         scale.addIncludedValuesProvider(provider);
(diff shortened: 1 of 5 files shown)


test/scales/timeScaleTests.ts
```diff
@@ -126,6 +126,23 @@ describe("Scales", () => {
         assert.strictEqual(domain[0].getTime(), dayBefore.getTime(), "left side of domain was expaded by one day");
         assert.strictEqual(domain[1].getTime(), dayAfter.getTime(), "right side of domain was expaded by one day");
       });
+
+      it("doesn't lock up if a zero-width domain is set while there are value providers", () => {
+        let scale = new Plottable.Scales.Time();
```
ztsai: why not const?

jtlan: Should be const :/

jtlan:
Original repro of locking: https://jsfiddle.net/p2gr8u60/
Repro as the bug manifests in Slate: https://jsfiddle.net/g3vgypdv/

Issue #3032, opened 2015-12-09
Stacked area plot + domainMin + croppedRendering = incorrect rendering


istarnx:
Segments of stacked area charts whose values are less than its scale's domainMin() setting will not render if croppedRendering is enabled (which is the default behavior).

Suppose x = [1,2,3] and stack one's y = [3,3,3] and stack two's y = [1,1,1]. If we do scale.domainMin(2), stack two disappears because that segment's height (1) is less than the domain min.

jtlan:
**Repro**: https://jsfiddle.net/15a30ybm/

PR #3033, opened 2015-12-10
[StackedArea Plot] Default to no cropping by default

jtlan:
Cropped rendering on `StackedArea` is currently broken (as per #3032), so disabling it by default for now until the algorithm can be fixed.

### DIFF
--- a/src/plots/stackedAreaPlot.ts
+++ b/src/plots/stackedAreaPlot.ts
@@ -29,7 +29,7 @@ module Plottable.Plots {
 
       if (croppedRendering === true) {
         // HACKHACK #3032: cropped rendering doesn't currently work correctly on StackedArea
-        Utils.Window.warn("Warning: Stacked Area Plot does not support downsampling.");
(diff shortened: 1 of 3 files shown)


ztsai:
if `croppedRenderingEnabled` is broken, do we want to disable it like how we disabled `downsamplingEnabled`?

src/plots/stackedAreaPlot.ts
```diff
@@ -17,6 +17,23 @@ module Plottable.Plots {
...
+    public croppedRenderingEnabled(): boolean;
+    public croppedRenderingEnabled(croppedRendering: boolean): this;
+    public croppedRenderingEnabled(croppedRendering?: boolean): any {
+      if (croppedRendering == null) {
+        return super.croppedRenderingEnabled();
+      }
+
+      if (croppedRendering === true) {
+        // HACKHACK #3032: cropped rendering doesn't currently work correctly on StackedArea
+        Utils.Window.warn("Warning: Stacked Area Plot does not support downsampling.");
```
ztsai: croppedRendering

PR #2903, opened 2015-10-20
Use a gradient for InterpolatedColorLegend instead of swatches

crmorford:
...but it's actually just a large amount of very skinny swatches.

before: http://jsfiddle.net/j5b9f60c/57/

screen shot 2015-10-20 at 1 48 30 pm

after http://jsfiddle.net/j5b9f60c/58/
screen shot 2015-10-20 at 1 48 17 pm

PR #3039, opened 2015-12-18
[Plots.Scatter] "reset" path attribute is no longer empty string

### LINKED ISSUE #3037: [Plots.Scatter] Points don't animate correctly
on `Plots.Scatter`, points no longer "inflate" to their full size. This is particularly evident if the step duration is increased:

``` Javascript
var animator = new Plottable.Animators.Easing();
animator.stepDuration(1000);
scatterPlot.animator(Plottable.Plots.Animator.MAIN, animator);
```

**Repro**: https://jsfiddle.net/kejm7xuy/

**Actual**: Points simply appear
**Expected**: Points grow to their full size from nothing.

jtlan:
Instead, it uses the `SymbolFactory` with a size of `0`. This means that animation now work properly, with the points "growing" to their final size.

**Before**: https://jsfiddle.net/kejm7xuy/
**After**: https://jsfiddle.net/7vb5ozv2/

Close #3037.

### DIFF
--- a/test/plots/scatterPlotTests.ts
+++ b/test/plots/scatterPlotTests.ts
@@ -57,10 +57,11 @@ describe("Plots", () => {
 
         plot.renderTo(svg);
 
-        const expectedPathString = TestMethods.normalizePath(symbolFactory(0));
-        const path = plot.content().select("path");
-        const normalizedActualPath = TestMethods.normalizePath(path.attr("d"));
-        assert.strictEqual(normalizedActualPath, expectedPathString, "path string is initialized with the correct symbol and a size of 0");
+        const pathString = plot.content().select("path").attr("d");
+        const zeroSizePathString = symbolFactory(0);
+        const normalizingPath = svg.append("path"); // generated and normalized paths don't align on all browsers
+        const expectedPathString = normalizingPath.attr("d", zeroSizePathString).attr("d");
+        assert.strictEqual(pathString, expectedPathString, "path string is initialized with the correct symbol and a size of 0");
       });
 
       it.skip("correctly handles NaN, undefined, Infinity, and non-number x and y values", () => {
--- a/test/plots/scatterPlotTests.ts
+++ b/test/plots/scatterPlotTests.ts
@@ -19,11 +19,17 @@ describe("Plots", () => {
             .y((d: any) => d.y, yScale);
       });
 
+      afterEach(function() {
+        if (this.currentTest.state === "passed") {
+          plot.destroy();
+          svg.remove();
+        }
+      });
+
       it("renders correctly with no data", () => {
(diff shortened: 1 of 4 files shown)


softwords:
This treats a specific case handled by more general solutions in #2907, various examples are at http://jsfiddle.net/oaskqta0/28/

 #2907 and its subset #2903 are both still up-to-date with develop - but because they are targetted at a very out-of-date side-branch, they appear to affect hundreds more files than they actually do. 
 In the new year perhaps we can resume discussions on whether you want to try to use this and if so, how.

test/plots/scatterPlotTests.ts
```diff
@@ -19,11 +19,17 @@ describe("Plots", () => {
             .y((d: any) => d.y, yScale);
       });
 
+      afterEach(function() {
```
adidahiya, line 22: `() => {`

test/plots/scatterPlotTests.ts
```diff
@@ -19,11 +19,17 @@ describe("Plots", () => {
...
+        if (this.currentTest.state === "passed") {
```
adidahiya, line 23: this this kind of test reflection common? is there a harm in always destroying the plot and removing the svg?

test/plots/scatterPlotTests.ts
jtlan, line 22: Has to use `function` because of the need to use a specific `this`-context.

test/plots/scatterPlotTests.ts
jtlan, line 23: Yeah, if the test fails, you won't be able to inspect the DOM and understand what went wrong.

test/plots/scatterPlotTests.ts
```diff
@@ -34,8 +40,28 @@ describe("Plots", () => {
...
+
+        plot.animator(Plottable.Plots.Animator.MAIN, new Mocks.NoOpAnimator());
+        plot.animated(true);
+
+        plot.renderTo(svg);
+
+        const pathString = plot.content().select("path").attr("d");
+        const zeroSizePathString = symbolFactory(0);
+        const normalizingPath = svg.append("path"); // generated and normalized paths don't align on all browsers
+        const expectedPathString = normalizingPath.attr("d", zeroSizePathString).attr("d");
```
adidahiya, line 63: strange that you have to create the `<path>`, add it to the DOM, set an attr, and read that attr back -- is this for cross-browser consistency?

adidahiya:
looks good overall, feel free to merge

test/plots/scatterPlotTests.ts
jtlan, line 63: Yeah. I tried to use the normalization function in `TestMethods`, but it doesn't always work correctly due to different choices regarding repeating path commands.

PR #3041, opened 2015-12-29
Make plottable more useable via npm

jhutchins:


adidahiya:
Hey @jhutchins thanks for the PR. This is on the right track and I plan on making Plottable usable via NPM soon. This is not an ideal solution however; there are more steps required to get the typings working properly. Also, Plottable depends on svg-typewriter, which is not available on NPM, so this would  fail at runtime.

PR #3055, opened 2016-02-08
Add ambient external module declaration

tomshen:
This allows you to import Plottable as an external module.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -5044,3 +5044,7 @@ declare module Plottable.Components {
         destroy(): void;
     }
 }
+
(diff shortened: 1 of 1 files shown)


adidahiya:
this is an incomplete solution; although the external module exists at runtime in the plottable.js UMD bundle, it does not pull in its dependencies correctly and is thus not usable in a proper NPM workflow. to get this external module working you need additional build configuration like webpack. there was a similar declined PR recently (#3041). the proper solution involves moving plottable and svg-typewriter to external modules; I'll open an issue for that

PR #3064, opened 2016-03-16
Category Scale - Minor Refactor

m4dc4p:
- Make use of scale interface for setting/getting range.
- Use _range member consistently

### DIFF
--- a/src/scales/categoryScale.ts
+++ b/src/scales/categoryScale.ts
@@ -33,23 +33,10 @@ module Plottable.Scales {
       return super.domain(values);
     }
 
-    protected _setDomain(values: string[]) {
-      super._setDomain(values);
-      this.range(this.range()); // update range
-    }
-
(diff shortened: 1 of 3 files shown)


src/scales/categoryScale.ts
```diff
@@ -33,23 +33,10 @@ module Plottable.Scales {
       return super.domain(values);
     }
 
-    protected _setDomain(values: string[]) {
-      super._setDomain(values);
-      this.range(this.range()); // update range
```
endrjuskr, line 38: Why we don't need to update `range` any more?

m4dc4p, line 38: As the range didn't change due to this method, I assumed this call was only in place to cause an update to `_d3Scale.rangeBands` (in the `range()` method).

That same effect now occurs in `_setBackingDomain`.

PR #3065, opened 2016-03-18
Bugfix/bar chart modified log axis

### LINKED ISSUE #3045: Bar chart with modified log axis is broken
Create a bar chart with a modified log y-axis. Ensure a y-value of at least 100. Note that the axis doesn't render.

endrjuskr:
`ModifiedScale._getDomain` was returning untransformed domain, which is different than backing domain.

I introduce `Scale._backingDomain`, which returns and sets backing domain.

JsFiddle to verify fix - https://jsfiddle.net/0vrx17uu/1/

Closes #3045.

### DIFF
--- a/src/scales/modifiedLogScale.ts
+++ b/src/scales/modifiedLogScale.ts
@@ -19,7 +19,7 @@ module Plottable.Scales {
      *
      * For negative values, scale(-x) = -scale(x).
      *
-     * The range and domain for the scale should also be set, using the 
+     * The range and domain for the scale should also be set, using the
      * range() and domain() accessors, respectively.
      *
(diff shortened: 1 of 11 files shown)


src/scales/modifiedLogScale.ts
```diff
@@ -98,8 +98,15 @@ module Plottable.Scales {
       super._setDomain(transformedDomain);
     }
```
themadcreator, line 100: So, how exactly is this different than the previous behavior? As far as I can tell, modifiedLogScale still returns _untransformedDomain, which was indicated as the source of the bug. I can see from your fiddle (thank you for that!) that the modified log scale is rendering, but what about this change set fixes it?

Issue #3066, opened 2016-03-18
Allow Gridlines with Categories


m4dc4p:
The Gridlines component only works with scales that extend QuantitiveScale. Gridlines with category scales would be nice.

PR #3067, opened 2016-03-18
Gridlines & Category Scales

m4dc4p:
Enhance Gridlines component so it can work with Category scale objects,
in addition to QuantitiveScale objects. (See #3066.)

### DIFF
--- a/test/components/gridlinesTests.ts
+++ b/test/components/gridlinesTests.ts
@@ -64,27 +64,29 @@ describe("Gridlines", () => {
   });
 
   it("draws gridlines on category ticks and updates when scale update", () => {
-    let categoryScale: Plottable.Scales.Category = new Plottable.Scales.Category().domain(["a", "b", "c"]);
-    let categoryGrid: Plottable.Components.Gridlines = new Plottable.Components.Gridlines(null, categoryScale);
+    let check = function(lines: d3.Selection<any>, scale: Plottable.Scales.Category, axis: string) {
+      let ticks = scale.domain();
+      assert.strictEqual(lines.size(), ticks.length, "There is a " + axis + " gridline for each " + axis + " tick");
+      lines.each(function(gridline: any, i: number) {
+        const v = TestMethods.numAttr(d3.select(this), axis + "1");
+        assert.closeTo(v, scale.scale(ticks[i]), window.Pixel_CloseTo_Requirement, axis + " gridline drawn on ticks");
+      });
+    };
+    let ys: Plottable.Scales.Category = new Plottable.Scales.Category().domain(["a", "b", "c"]);
+    let xs: Plottable.Scales.Category = new Plottable.Scales.Category().domain(["x", "y", "z"]);
+    let categoryGrid: Plottable.Components.Gridlines = new Plottable.Components.Gridlines(xs, ys);
 
     categoryGrid.renderTo(svg);
 
-    let yGridlines = categoryGrid.content().select(".y-gridlines").selectAll("line");
-    let yTicks = categoryScale.domain();
-    assert.strictEqual(yGridlines.size(), yTicks.length, "There is a y gridline for each y tick");
-    yGridlines.each(function(gridline, i) {
-      const y = TestMethods.numAttr(d3.select(this), "y1");
-      assert.closeTo(y, categoryScale.scale(yTicks[i]), window.Pixel_CloseTo_Requirement, "y gridline drawn on ticks");
-    });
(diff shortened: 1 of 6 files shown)


test/components/gridlinesTests.ts
```diff
@@ -63,14 +63,33 @@ describe("Gridlines", () => {
...
   });
 
-  it("throws error on non-Quantitative Scales", () => {
-    const categoryScale = new Plottable.Scales.Category();
-    // HACKHACK #2661: Cannot assert errors being thrown with description
-    (<any> assert).throw(() => new Plottable.Components.Gridlines(<any> categoryScale, null), Error,
-      "xScale needs to inherit from Scale.QuantitativeScale", "can't set xScale to category scale");
-    (<any> assert).throw(() => new Plottable.Components.Gridlines(null, <any> categoryScale), Error,
-      "yScale needs to inherit from Scale.QuantitativeScale", "can't set yScale to category scale");
+  it("draws gridlines on category ticks and updates when scale update", () => {
```
crmorford, line 66: Can you test in the x-direction also?

crmorford:
@adidahiya can you check out this PR? It looks pretty small.

test/components/gridlinesTests.ts
m4dc4p, line 66: Done! See my latest commit (55e136d).

m4dc4p:
@adidahiya ping!

adidahiya:
@m4dc4p can you please post a jsfiddle to demonstrate the change? In a bar chart with bars for different categories, do the gridlines draw on the location of the category or _between_ categories?

m4dc4p:
@adidahiya The grid lines go through the middle of each category. Here's a sample:

  http://m4dc4p.github.io/plottable/gridLines.html

(I tried using jsfiddle but coudlnt get it to work with my branch.)

adidahiya:
thanks for the example.

This behavior is counter to what I expected. I expected the gridlines to appear between categories. A quick google search for "grid lines category chart" confirms that as the more common use case. What do you think about changing the position to match something like this?

image

m4dc4p:
I used this in a scatter chart, where its not so obvious which label applies to which data point. Here are some counter examples - the lack of no gridlines make it hard to tell values a given point has (all from http://superuser.com/questions/485883/how-to-create-dynamic-scatter-plot-matrix-with-labels-and-categories-on-both-axi):





How could we address both use cases w/o too much additional work?

adidahiya:
Ah, I see, those are good scatter plot examples. I can see a case for allowing both kinds of positioning for grid lines :+1:. I don't have a whole lot of time to investigate the details right now -- could you write up a short proposal for the API?

m4dc4p:
Two alternatives come to mind:
- An "offset" amount (in pixels) that shifts the gridlines from alignment with ticks. Because labels are all equal width, grid lines could be made to line up wiht the edge of each label pretty easily.
- A higher-level "alignment" setting (left/right/center) which aligns the grid lines with the appropriate edge of each label. (For horizontal gridlines, that would translate to top/bottom/center.)

Thoughts?

adidahiya:
I like the second approach, but don't we only need two alignment settings? Something like:
- "label" positions it to center on the category label
- "tick" positions gridlines on axis ticks

cc @crmorford, @jtlan for API feedback

m4dc4p:
Can you clarify how "label" and "tick" alignment would be different? Don't ticks always line up with the center of labels?

adidahiya:
ah, my mistake. it probably looks more like: `tickAlignment: "center" | "between"`? open to more naming ideas

m4dc4p:
Maybe `on` instead of `center`? I'm :+1:  either way.

I don't have time to do this work on this PR - are you comfortable merging this PR as is? I'm happy to write up an issue (and probably implement it but can't say for sure 100%).

jtlan:
> ah, my mistake. it probably looks more like: `tickAlignment: "center" | "between"`? open to more naming ideas

Seconding `"center"`/`"between"` for consistency with `Axes.Time`. Would this be an API on `Scales.Category`?

m4dc4p:
@crmorford @jtlan Created an issue to finish this discussion in - https://github.com/palantir/plottable/issues/3075.

I agree with jtlan, this should probably be a property on either the Scales.Category class or the Scale base class.

themadcreator:
@m4dc4p Thanks for this PR! I'm taking over the revisions requested (line/tick position). This required creating a new PR since yours came from a fork and I can't push to it. So, i'm closing this PR and continuing the work in #3189

m4dc4p:
Glad this work lives on. Enjoy! 👍

PR #3069, opened 2016-03-22
Nearest entity on group

### LINKED ISSUE #3068: Expose closest data point for plot group interactions
when adding pointer interactions to a plot group, users currently have to manually do some `nearestEntity` tie-breaking to figure out the closest plot entity for an interaction. let's make this use case easier.

endrjuskr:
Closes #3068.


src/plots/plotGroup.ts
```diff
@@ -0,0 +1,33 @@
...
+        }
+      });
+
+      return closestPlotEntity;
+    }
+    /**
+     * Adds a Plot to this Plot Group.
+     * The added Plot will be rendered above Plots already in the Group.
+     */
+    public append(component: Component): any {
```
jtlan: Shouldn't this take in a `Plot` and not a `Component`, then?
It should also return `this` and not `any`.

src/plots/plotGroup.ts
```diff
@@ -0,0 +1,33 @@
+module Plottable.Plots {
```
jtlan: `Plottable.Components` seems like the more appropriate module?

plottable.d.ts
```diff
@@ -3897,6 +3897,16 @@ declare module Plottable.Plots {
...
     }
 }
+declare module Plottable.Plots {
+    class PlotGroup extends Components.Group {
+        entityNearest(point: Point): PlotEntity;
+        /**
+         * Adds a Plot to this Plot Group.
+         * The added Plot will be rendered above Plots already in the Group.
+         */
+        append(component: Component): any;
```
themadcreator: Components.Group already defines .append, doesn't it? Does this definition here add anything other than the comment above it? If not, remove it.

src/plots/plotGroup.ts
```diff
@@ -0,0 +1,33 @@
...
+      let closestPlotEntity: PlotEntity;
+      let minDistSquared = Infinity;
+      this.components().forEach((plot: Plot) => {
+        let candidatePlotEntity = plot.entityNearest(point);
+        if (candidatePlotEntity == null) {
+            return;
+        }
+
+        let position = candidatePlotEntity.position;
+        let distSquared = Math.pow(position.x - point.x, 2) + Math.pow(position.y - point.y, 2);
```
themadcreator: This position is the centroid of the entity or is it the top-left point?

In either case, we should be considering the full bounds of the entity since the nearest entity might not be the entity with the nearest centroid.

src/plots/plotGroup.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      return closestPlotEntity;
+    }
+    /**
+     * Adds a Plot to this Plot Group.
+     * The added Plot will be rendered above Plots already in the Group.
+     */
+    public append(component: Component): any {
+      if (component != null && !(component instanceof Plot)) {
+        throw new Error("Plot Group only accepts plots");
```
themadcreator: I think jtlan's typing suggestions would obviate this.

src/plots/plotGroup.ts
themadcreator: Looking at the existing plot.entityNearest definition, it looks like we only use the entity position, so this is fine. However, switch this to use the utility method: Utils.Math.distanceSquared

jtlan: The position of an `Entity` is defined by the `Plot` it comes from -- for example, it's the middle of points on `Plots.Scatter`, but is at one of the edges of the bars on `Plot.Bar`.

jtlan: However, this means that this code could result in counterintuitive results. Take this example:
pointeroverlay

In this scenario, the algorithm will return the circled orange `Entity` rather than the green one because it's closer, even though the query point (cursor) is inside the bar.

themadcreator: Yeah, preferably, our entityNearest algorithm would use the nearest point on the entity not just "entity.position". However, for the purposes of this PR, I think it's fine to use the same standard as the library.

src/components/plotGroup.ts
```diff
@@ -0,0 +1,34 @@
...
+
+      return closestPlotEntity;
+    }
+
+    /**
+     * Adds a Plot to this Plot Group.
+     * The added Plot will be rendered above Plots already in the Group.
+     */
+    public append(plot: Plot): this {
+      if (plot != null && !(plot instanceof Plot)) {
```
endrjuskr, line 27: Kept error because JS does not check types.

plottable.d.ts
adidahiya: Now that the argument type has been narrowed from `Component` to `Plot`, this override seems somewhat useful.

PR #3071, opened 2016-03-30
Feature/euclidean distance on linecharts

### LINKED ISSUE #3070: entityNearest on line plots should use Euclidean distance
On http://plottablejs.org/components/plots/line/ , "entityNearest" is described as:

"Returns the PlotEntity nearest to the query point by X then by Y, or undefined if no PlotEntity can be found."

On charts with multiple lines, this will not return the nearest entity to the query point. Users expect the function "entityNearest" to use Euclidean distance, like it does on scatter plots.

According the the API documentation on "Plot":

entityNearest(queryPoint: Point): PlotEntity
Defined in plottable.d.ts:2906

Returns the PlotEntity nearest to the query point by the Euclidian norm, or undefined if no PlotEntity can be found.

http://plottablejs.org/docs/classes/plottable.plot.html#entitynearest

galenmarchetti:
Implemented a euclidean distance metric on linechart nearestEntity() rather than relying on only x-distance.

Closes #3070.

### DIFF
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3588,7 +3588,7 @@ declare module Plottable.Plots {
             [attr: string]: (datum: any, index: number, dataset: Dataset) => any;
         };
         /**
-         * Returns the PlotEntity nearest to the query point by X then by Y, or undefined if no PlotEntity can be found.
+         * Returns the PlotEntity nearest to the query point by Euclidean distance, or undefined if no PlotEntity can be found.
          *
          * @param {Point} queryPoint
          * @returns {PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
--- a/plottable.d.ts
+++ b/plottable.d.ts
@@ -3587,13 +3587,6 @@ declare module Plottable.Plots {
         protected _generateAttrToProjector(): {
             [attr: string]: (datum: any, index: number, dataset: Dataset) => any;
         };
-        /**
-         * Returns the PlotEntity nearest to the query point by Euclidean distance, or undefined if no PlotEntity can be found.
-         *
-         * @param {Point} queryPoint
-         * @returns {PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
-         */
-        entityNearest(queryPoint: Point): PlotEntity;
         protected _propertyProjectors(): AttributeToProjector;
         protected _constructLineProjector(xProjector: Projector, yProjector: Projector): (datum: any, index: number, dataset: Dataset) => string;
         protected _getDataToDraw(): Utils.Map<Dataset, any[]>;
--- a/plottable.js
+++ b/plottable.js
@@ -8928,30 +8928,6 @@ var Plottable;
                 });
                 return attrToProjector;
             };
-            /**
-             * Returns the PlotEntity nearest to the query point by Euclidean distance, or undefined if no PlotEntity can be found.
-             *
-             * @param {Point} queryPoint
-             * @returns {PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
-             */
-            Line.prototype.entityNearest = function (queryPoint) {
-                var _this = this;
-                var minDist = Infinity;
-                var closest;
-                this.entities().forEach(function (entity) {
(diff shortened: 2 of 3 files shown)


jtlan:
@galenmarchetti The X-then-Y algorithm was initially implemented at the request of Design -- can you explain the motivation for this change?

themadcreator:
@jtlan - quoting from the linked issue:

> On charts with multiple lines, this will not return the nearest entity to the query point. Users expect the function "entityNearest" to use Euclidean distance, like it does on scatter plots.

I can see an argument for the euclidean distance being more intuitive. Do you recall why designed wanted this X-then-Y method?

Also @galenmarchetti , since the base class Plot already uses the euclidean distance for entityNearest, I think this PR should simply remove the override in linePlot.ts

galenmarchetti:
@jtlan with multiple lines on one chart, proper tooltip behavior will only occur if lines are composed of the same x-axis points. if you have lines that are derived from different calculations along the x values, then the tooltip behavior will rapidly jump between lines even though the cursor is directly above a second line.

one example is displaying multiple regression lines within a scatter plot. the motivating case had a scatter plot with points of four different colors, displaying four different regression lines. Because the regression lines are each calculated from different sets, they aren't each generated from the same set of "x" value points. this leads to strange tooltip behavior.

I am curious about the design motivation for deciding against a Euclidean distance metric. if there's a convincing case for line plots to be treated this way, then that's fine, but in this case it seems to me that the API specs for the base class Plot should not specify that distance is Euclidean.

@themadcreator that does make more sense. i'll submit a commit implementing that

jtlan:
@galenmarchetti, @themadcreator:
I think the general idea was that the user's cursor would follow the "flow" of the line -- consider something like a stock chart. If the line oscillates a lot the user has to move their cursor up and down to get tooltips follow it. I could go either way on this one so I'd be interested to hear a designer's take.

galenmarchetti:
@jtlan +1 on the designers take. I now understand the use case that motivated the X-distance, then Y-distance decision, but I'm uncertain whether the negative effects of that design are worth it. 

A user can intuit what's happening when its very clear where the points that make up the line are. However, when the line doesn't oscillate quickly enough for the points to be obvious, it then becomes quite confusing when the user is trying to produce a tooltip on a line but a tooltip is instead appearing on a line thats far away.

My 2 cents. I agree I'd like to hear a designers take

adidahiya:
@galenmarchetti would you be able to post two jsfiddles here demonstrating the interaction behaviors? If not I can try to get something up later today

galenmarchetti:
@adidahiya @jtlan I think the motivation for this pull request is obvious in these two jsfiddles:

X-then-Y algorithm: http://jsfiddle.net/galenmarchetti/pj00hv0g/5/

Euclidian distance algorithm: http://jsfiddle.net/galenmarchetti/63t0o9d6/24/

jtlan:
> X-then-Y algorithm: http://jsfiddle.net/galenmarchetti/pj00hv0g/5/

Okay, I can see how this could be annoying (flickers up and down between the two lines), although it also has other behavior that's a bit odd (snapping between troughs or between peaks). Were we able to get some design input?

> with multiple lines on one chart, proper tooltip behavior will only occur if lines are composed of the same x-axis points.

I expect that this is the most common use case (same X-values), but this is good to think about.

If we switch to Euclidian might be good to put the old logic somewhere for people who are interested in replicating it.

adidahiya:
I'm on board with the results of this PR.

@jtlan what's a good way to preserve the old logic? a utility helper method? not sure about keeping dead code around... I could be explicit in the release notes about this behavior change and how to get the old functionality back with a subclass

jtlan:
@adidahiya It's best to explicitly call out behavioral changes in the release notes. I would either:
- Offer the original source in a code block for users to copy if they wanted the old functionality (by applying it to the results of `entities()`, although this will be slower)
- Move the original logic to another method, `entityNearestByXThenY()`

I think the second option is better.

adidahiya:
cool, renaming the method to `entityNearestByXThenY()` sounds good to me.

adidahiya:
abandoning in favor of #3079.

Issue #3072, opened 2016-04-04
Align label text on Category Axis


m4dc4p:
The alignment of tick labels on category axis depends on the orientation of the axis. I would like more direct control over label alignment (e.g., "left", "right", or "center"). 
- If not set, labels should render according to existing behavior.
- Alignment should be applied before rotating text.
- Alignment also implies that a "padding" setting must be present, otherwise labels will be rendered next to the outer edge of the axis in certain cases.

PR #3073, opened 2016-04-04
Support for alignment & padding in category axis

m4dc4p:
- Added 'tickTextAlignment' property, which specifies alignment for tick
  labels.
- Added 'tickTextPadding' property, which specifies padding between tick
  text and the outer edge of the axis. Works for all orientation of axis
  and rotated text.

See http://m4dc4p.github.io/plottable/textAlign.html for a page demonstrating all the textAlignment/textPadding options.

### DIFF
--- a/src/axes/categoryAxis.ts
+++ b/src/axes/categoryAxis.ts
@@ -4,6 +4,8 @@ module Plottable.Axes {
     private _measurer: SVGTypewriter.Measurers.CacheCharacterMeasurer;
     private _wrapper: SVGTypewriter.Wrappers.Wrapper;
     private _writer: SVGTypewriter.Writers.Writer;
+    private _tickTextAlignment: string = null;
+    private _tickTextPadding: number = 0;
 
     /**
      * Constructs a Category Axis.
@@ -74,6 +76,80 @@ module Plottable.Axes {
     }
 
     /**
+     * Gets the current alignment. If not set, null will be returned.
+     */
+    public tickTextAlignment(): string;
+    /**
+     * Set alignment of tick labels. Must be one of "left", "right",
+     * or "center", or a falsy value. Setting to a falsy value will
+     * reset tick label alignment to the default alignment.
+     *
+     * Note that alignment is applied before rotation (see "{@link
+     * tickLabelAngle}").
+     * 
+     * @param {String} tickTextAlignment One of "left", "right", "center",
+     * or a falsy value.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextAlignment(tickTextAlignment: string): this;
+    public tickTextAlignment(tickTextAlignment?: string): any {
+      if (arguments.length === 0) {
+        return this._tickTextAlignment;
+      }
+
+      if (! tickTextAlignment) {
+        this._tickTextAlignment = null;
+      } else {
+        let v = tickTextAlignment.toLowerCase();
+        if (v !== "left" && v !== "right" && v !== "center") {
+          throw new Error("tickTextAlignment '" + tickTextAlignment + "' not supported. Must be left, right, or center.");
+        }
+        this._tickTextAlignment = v;
+      }
+
+      return this;
+    }
+
+    /**
+     * Returns the current padding between tick labels and the
+     * opposite edge of the axis.
+     */
+    public tickTextPadding(): number;
+    /**
+     * Sets Padding between labels and outer edge of the axis (i.e.,
+     * the edge opposite the plot, as defined by the axis' {@link
+     * orientation}).
+     *
+     * Padding will only be applied when {@link tickTextAlignment} is
+     * set and either of the following conditions hold:
+     *
+     *   * Orientation is "left" or "right"; labels are not rotated;
+     *     for "left" orientation, {@link tickTextAlignment} must be
+     *     "left". For "right" orientation, {@link tickTextAlignment}
+     *     must be "right".
+     *
+     *   * Orientation is "top" or "bottom"; labels are rotated. For
+     *     "top" orientation and 90 rotation, alignment must "left";
+     *     for -90 rotation, alignment must be "right". For "bottom"
+     *     orientation and 90 rotation, alignment must "right"; For
+     *     -90 rotation, alignment must be "left".
+     *
+     * @param {Number} tickTextPadding Padding in pixels.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextPadding(tickTextPadding: number): this;
+    public tickTextPadding(tickTextPadding?: number): any {
+      if (arguments.length === 0) {
+        return this._tickTextPadding;
+      }
+
+      this._tickTextPadding = tickTextPadding ? tickTextPadding : 0;
+      return this;
+    }
+
+    /**
      * Gets the tick label angle in degrees.
      */
     public tickLabelAngle(): number;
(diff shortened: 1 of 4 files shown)


m4dc4p:
For #3072.

adidahiya:
hey @m4dc4p we'll take a look at this within the next few days. thanks for your patience.

src/axes/categoryAxis.ts
```diff
@@ -119,17 +246,24 @@ module Plottable.Axes {
...
+          xAlign: self.tickTextAlignment() || xAlign[self.orientation()],
           yAlign: yAlign[self.orientation()],
           textRotation: self.tickLabelAngle()
         };
+
         self._writer.write(self.formatter()(d), width, height, writeOptions);
+
+        if (translate.x !== 0 || translate.y !== 0) {
+          let text = writeOptions.selection;
+          text.attr("transform", "translate(" + translate.x + ", " + translate.y + ") " + text.attr("transform"));
```
adidahiya: please use a template string

src/axes/categoryAxis.ts
```diff
@@ -105,6 +228,10 @@ module Plottable.Axes {
       let self = this;
       let xAlign: {[s: string]: string};
       let yAlign: {[s: string]: string};
+      let result = this._calcTextPadding();
+      let pad = result.pad;
+      let translate = result.translate;
```
adidahiya: simplify these two lines to

``` ts
const { pad, translate } = result;
```

and prefer using `const` wherever possible

src/axes/categoryAxis.ts
```diff
@@ -97,6 +173,53 @@ module Plottable.Axes {
       return this;
     }
 
+    private _calcTextPadding(): { translate: { x: number, y: number }, pad: { x: number, y: number }} {
+      // Padding always moves the label *away* from the outer edge of
+      // the axis (tickLabelPadding moves labels away from the inner
+      // edge of the axis already).
```
adidahiya: I think these 3 lines can be a JSDoc comment on the method?

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,80 @@ module Plottable.Axes {
...
+      if (arguments.length === 0) {
+        return this._tickTextAlignment;
+      }
+
+      if (! tickTextAlignment) {
+        this._tickTextAlignment = null;
+      } else {
+        let v = tickTextAlignment.toLowerCase();
+        if (v !== "left" && v !== "right" && v !== "center") {
+          throw new Error("tickTextAlignment '" + tickTextAlignment + "' not supported. Must be left, right, or center.");
```
adidahiya: please use a template string

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,80 @@ module Plottable.Axes {
...
+     *
+     * Note that alignment is applied before rotation (see "{@link
+     * tickLabelAngle}").
+     * 
+     * @param {String} tickTextAlignment One of "left", "right", "center",
+     * or a falsy value.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextAlignment(tickTextAlignment: string): this;
```
adidahiya, line 95: since we're on TS 1.8, we could theoretically use string literal types, but that might break consumers who are using TS 1.7, so we should probably hold off.

cc @JKillian for thoughts on the implications of this

src/axes/categoryAxis.ts
```diff
@@ -145,26 +279,27 @@ module Plottable.Axes {
       let totalInnerPaddingRatio = (ticks.length - 1) * scale.innerPadding();
       let expectedRangeBand = axisSpace / (totalOuterPaddingRatio + totalInnerPaddingRatio + ticks.length);
       let stepWidth = expectedRangeBand * (1 + scale.innerPadding());
+      let tickTextPadding = this._calcTextPadding().pad;
```
adidahiya: `const`?

adidahiya:
@m4dc4p overall the idea sounds good and the PR could just use some small revisions

m4dc4p:
@adidahiya @JKillian I updated the PR based on your comments. Thanks for taking a look!

src/axes/categoryAxis.ts
jkillian, line 95: Yep, for now let's avoid using them to avoid `<=1.7.x` breakage

src/axes/categoryAxis.ts
```diff
@@ -98,13 +175,67 @@ module Plottable.Axes {
...
+     * subtracted from overall space when rendering text (important for
+     * proper line breaking).
+     * 
+     * The translate value gives the amount the label should be moved
+     * to give the proper padding.
+     * 
+     * Pad will always be a positive amount, but translate can be
+     * negative depending on rotation of labels and axis orientation.
+     */
+    private _calcTextPadding(): { translate: { x: number, y: number }, pad: { x: number, y: number }} {
```
jkillian, line 188: It would be more consistent to make this `padding` instead of `pad` I think

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,81 @@ module Plottable.Axes {
...
+     *     "left". For "right" orientation, {@link tickTextAlignment}
+     *     must be "right".
+     *
+     *   * Orientation is "top" or "bottom"; labels are rotated. For
+     *     "top" orientation and 90 rotation, alignment must "left";
+     *     for -90 rotation, alignment must be "right". For "bottom"
+     *     orientation and 90 rotation, alignment must "right"; For
+     *     -90 rotation, alignment must be "left".
+     *
+     * @param {Number} tickTextPadding Padding in pixels.
```
jkillian, line 139: `number` instead of `Number`

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,81 @@ module Plottable.Axes {
...
+     *
+     *   * Orientation is "left" or "right"; labels are not rotated;
+     *     for "left" orientation, {@link tickTextAlignment} must be
+     *     "left". For "right" orientation, {@link tickTextAlignment}
+     *     must be "right".
+     *
+     *   * Orientation is "top" or "bottom"; labels are rotated. For
+     *     "top" orientation and 90 rotation, alignment must "left";
+     *     for -90 rotation, alignment must be "right". For "bottom"
+     *     orientation and 90 rotation, alignment must "right"; For
```
jkillian, line 136: "must be" instead of "must"

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,81 @@ module Plottable.Axes {
...
+     * Padding will only be applied when {@link tickTextAlignment} is
+     * set and either of the following conditions hold:
+     *
+     *   * Orientation is "left" or "right"; labels are not rotated;
+     *     for "left" orientation, {@link tickTextAlignment} must be
+     *     "left". For "right" orientation, {@link tickTextAlignment}
+     *     must be "right".
+     *
+     *   * Orientation is "top" or "bottom"; labels are rotated. For
+     *     "top" orientation and 90 rotation, alignment must "left";
```
jkillian, line 134: "must be" instead of "must"

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,81 @@ module Plottable.Axes {
...
+     * tickLabelAngle}").
+     * 
+     * @param {String} tickTextAlignment One of "left", "right", "center",
+     * or a falsy value.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextAlignment(tickTextAlignment: string): this;
+    public tickTextAlignment(tickTextAlignment?: string): any {
+      if (arguments.length === 0) {
```
jkillian, line 97: Instead of checking `arguments.length`, let's use the idiom used elsewhere in the codebase and check if the `tickTextAlignment` param is specified or not. ([Similar to here](https://github.com/palantir/plottable/blob/8e736c4367f9e2ca56b6edb15731a16efe014dc8/src/plots/stackedAreaPlot.ts#L52) for example.)

src/axes/categoryAxis.ts
```diff
@@ -74,6 +76,81 @@ module Plottable.Axes {
...
+    public tickTextAlignment(): string;
+    /**
+     * Set alignment of tick labels. Must be one of "left", "right",
+     * or "center", or a falsy value. Setting to a falsy value will
+     * reset tick label alignment to the default alignment.
+     *
+     * Note that alignment is applied before rotation (see "{@link
+     * tickLabelAngle}").
+     * 
+     * @param {String} tickTextAlignment One of "left", "right", "center",
```
jkillian, line 90: should be `string` as opposed to `String`

src/axes/categoryAxis.ts
m4dc4p, line 97: That can't distinguish between null values and absence of an argument. "Falsy" is required to reset alignment, and null is "falsy." It's not consistent but I think it's the better way.

m4dc4p:
@JKillian @adidahiya Appreciate your comments. I don't have time to do any more updates to this PR right now. Please merge or decline as-is (or update if you wish).

src/axes/categoryAxis.ts
jkillian, line 97: I see what you're saying, seems good to me.

jkillian:
>  I don't have time to do any more updates to this PR right now. Please merge or decline as-is (or update if you wish).

Thanks for the heads-up, most likely we can just merge this as is and make the small documentation tweaks in a follow-up. Thanks for the contribution @m4dc4p!

m4dc4p:
No problem, thanks for your understanding! :)

hellochar:
This hasn't moved in too long, feel free to submit a new PR if this is still valuable

Issue #3051, opened 2016-01-22
Remove quicktest for Plots.Wheel


adidahiya:
It is deprecated as of 2.0.

PR #3076, opened 2016-04-14
delete bad test, delete colors from dropdown, fix broken tests

crmorford:
something is still wrong with realistic/add_remove_dataset.js, i might pull in justin.


adidahiya:
fixes #3051

adidahiya:
lgtm 👍

PR #3079, opened 2016-04-29
Rename LinePlot#entityNearest to entityNearestByXThenY

### LINKED ISSUE #3070: entityNearest on line plots should use Euclidean distance

adidahiya:
Fixes #3070.

LinePlot now uses the default euclidean distance algorithm for nearest entities.
The old X-then-Y algorithm is preserved in this newly renamed method.

### DIFF
--- a/plottable-npm.d.ts
+++ b/plottable-npm.d.ts
@@ -3611,7 +3611,7 @@ declare namespace Plottable.Plots {
          * @param {Point} queryPoint
          * @returns {PlotEntity} The nearest PlotEntity, or undefined if no PlotEntity can be found.
          */
-        entityNearest(queryPoint: Point): PlotEntity;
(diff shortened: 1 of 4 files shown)


adidahiya:
1 sauce labs failure is a flake.

Issue #3086, opened 2016-07-14
Enable events API for Pan and Zoom


SamMorrowDrums:
The pan and zoom interaction is really nice, but it's also an excellent opportunity for developers to enable dynamic loading of data.

I suggest adding event listeners, that are triggered at the end of Pan events and Zoom events, so it is possible to fill in the missing data. The user can then, for example, check the scale domains on callback, and make a request for the data they haven't downloaded yet.

yocontra:
This can be closed (#3087 landed)

Issue #2651, opened 2015-08-18
Event callbacks for PanZoomInteraction


istarnx:
We have some stuff we'd like to disable upon pan/zoom. Can you guys add callbacks to signal panZoom start/end?

christinaforney:
Are waiting on clarification of need before continuing to work on this.

christinaforney:
e

jtlan:
@christinaforney What was the eventual clarification?

MadsMadsDk:
Any news on this? I have a situation where I want to do things after panning has ended. Currently I'm listening for timescale-updates, but it creates a great deal of overhead.

PR #3087, opened 2016-07-14
Add onPan and onZoom to the PanZoom interaction

SamMorrowDrums:
As mentioned in #3086 I am proposing to add events for Pan and Zoom, so that users can automatically retrieve more data for charts.

The proposed API is such:

``` javascript
panZoomInteraction = new Plottable.Interactions.PanZoom();

panZoomInteraction.onPan(callback);
panZoomInteraction.offPan(callback);

panZoomInteraction.onZoom(callback);
panZoomInteraction.offZoom(callback);

```

Callbacks occur on the most applicable end event (touchend, dragend or scroll), in an attempt to minimise the number of calls where possible, as this will most likely be used for retrieving server data.

### DIFF
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -1,4 +1,8 @@
 namespace Plottable.Interactions {
+
+  export type PanCallback = (e: Event) => void;
+  export type ZoomCallback = (e: Event) => void;
+
   export class PanZoom extends Interaction {
     /**
      * The number of pixels occupied in a line.
@@ -21,6 +25,9 @@ namespace Plottable.Interactions {
 
     private _minDomainExtents: Utils.Map<QuantitativeScale<any>, any>;
     private _maxDomainExtents: Utils.Map<QuantitativeScale<any>, any>;
+
+    private _panCallbacks = new Utils.CallbackSet<PanCallback>();
+    private _zoomCallbacks = new Utils.CallbackSet<ZoomCallback>();
 
     /**
      * A PanZoom Interaction updates the domains of an x-scale and/or a y-scale
@@ -171,6 +178,10 @@ namespace Plottable.Interactions {
       ids.forEach((id) => {
         this._touchIds.remove(id.toString());
       });
+
+      if (this._touchIds.size() > 0) {
+        this._zoomCallbacks.callCallbacks(e);
+      }
     }
 
     private _magnifyScale<D>(scale: QuantitativeScale<D>, magnifyAmount: number, centerValue: number) {
@@ -205,6 +216,7 @@ namespace Plottable.Interactions {
         this.yScales().forEach((yScale) => {
           this._magnifyScale(yScale, zoomAmount, translatedP.y);
         });
+        this._zoomCallbacks.callCallbacks(e);
       }
     }
 
@@ -242,6 +254,7 @@ namespace Plottable.Interactions {
         });
         lastDragPoint = endPoint;
       });
+      this._dragInteraction.onDragEnd((e) => this._panCallbacks.callCallbacks(e));
     }
 
     private _nonLinearScaleWithExtents(scale: QuantitativeScale<any>) {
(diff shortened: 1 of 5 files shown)


adidahiya:
Nice, this looks clean & well-tested. Is it easy to tell what new data you should retrieve from a server on these events? What might that code look like for Plottable users?

adidahiya:
Actually, can you change the API to `onPanEnd` and `onZoomEnd`? Looking at https://github.com/palantir/plottable/issues/2651, it seems like we'd also like to have hooks into the _start_ of these interactions. Naming the ones in this PR more precisely will make it easier to add those hooks later.

// cc @jtlan for thoughts

SamMorrowDrums:
Yes, that makes sense. I'll update the naming and push again.

I'm actually working on code right now that would fire on this event as we speak! Naturally there are a lot of race conditions that can occur getting async data. Also, there is no actual scroll-end event, so a lot of people implement a timer, and don't fire the true callback until they stop upping the timer.

I'm not sure that the timer code belongs in Plottable codebase though. For now I'll just update the names, and work on implementation of the calling code!

SamMorrowDrums:
Also, for reference I'm using an intermediate callback that gets the x-axis domain for my data gathering:

``` javascript
panZoomInteraction.onZoomEnd(function () {
        const d = that.xScale.domain();
        callback(d[0], d[1]);
});
```

jtlan:
> Is it easy to tell what new data you should retrieve from a server on these events? What might that code look like for Plottable users?

I would expect that users would invoke `domain()` calls to figure out the view window at that time. Alternatively the callbacks could return the domain of the `Scale`s at that time, but there wouldn't be any type information because `PanZoom` interaction isn't genericized.

src/interactions/panZoomInteraction.ts
```diff
@@ -424,5 +437,49 @@ namespace Plottable.Interactions {
       this._maxDomainExtents.set(quantitativeScale, maxDomainExtent);
       return this;
     }
+
+    /**
+     * Adds a callback to be called when panning ends.
+     *
+     * @param {PanCallback} callback
+     * @returns {Event} The calling PanZoom Interaction.
```
jtlan: This and other registration calls should return `{this}` not `{Event}`.

src/interactions/panZoomInteraction.ts
```diff
@@ -1,4 +1,8 @@
 namespace Plottable.Interactions {
+
+  export type PanCallback = (e: Event) => void;
+  export type ZoomCallback = (e: Event) => void;
```
jtlan: Not sure these callbacks should take `Event`s -- that's sort of atypical if you compare to the callbacks passed to `Click`, `Drag`, and `Pointer` `Interaction`s. I would expect the final domains of the `Scale`s or something like that.

SamMorrowDrums: Yeah, I was a little unsure as to what to return, but you can take multiple
scales right, so would it be an array of domains returned?

jtlan: Ideally it would be something like:

```
{
  xDomains: Map<Scale, X[]>;
  yDomains: Map<Scale, Y[]>;
}
```

but the typing on `PanZoom` doesn't allow for that since it's not generic on `<X, Y>` (also object-keyed `Map` isn't standard yet). For now we might want the callback to pass nothing (but still keep it type-defined), since the users can retrieve the domain information from the `Scale`s directly.

SamMorrowDrums:
OK @adidahiya & @jtlan I've fixed the documentation and removed the event arguments from the types, as it sounds like we are not yet decided on best approach for passing the domains (although I agree that is certainly major use case, and users will likely do this manually for now)

I think, if we want a more realistic Zoom event from scroll wheel, we need to implement an intermediate callback that sets a timer (of say 100ms), and keeps cancelling timer and extending until no wheel events occur in time, and callback fires.

Should this work on the event timer be completed now, or are we better progressing with the events, and then building on them (as part of the Zoom and Pan start event work)?

SamMorrowDrums:
```

**Comment by SamMorrowDrums** [comment]
I'm doing the timeout in my own code, and its behaviour is pretty solid. Here is roughly what I'm doing in my own codebase:

``` javascript
import Plottable from "plottable";

class PanZoom extends Plottable.Interactions.PanZoom {
    constructor (cb) {
        super();
        this.timer = null;
        this.onDateChange = cb;
        if (cb) {
            this._dragInteraction.onDragEnd(cb);
        }
    }

    _handleWheelEvent (p, e) {
        const that = this;
        super._handleWheelEvent(p, e);
        if (this.onDateChange) {
            if (this.timer) {
                clearTimeout(this.timer);
            }
            this.timer = setTimeout(function () {
                that.onDateChange();
            }, 100);
        }
    }

    _handleTouchEnd (ids, idToPoint, e) {
        super._handleTouchEnd(ids, idToPoint, e);
        if (this.onDateChange && this._touchIds.size() > 0) {
            this.onDateChange();
        }
    }
}

export default PanZoom;
```

SamMorrowDrums:
@adidahiya you asked how I was planning to implement the calling code, current process is:
- Callback fires
- Pass changed domain into an array
- Check if it's within existing data
- Send data request (and empty array)
- Wait until data gathered
- If more callbacks have fired during data gathering, get the min and max of the domains from callbacks, and then if more data required data, send request (clear array again) and repeat.

SamMorrowDrums:
I have it working perfectly for me, with the data gathering. Really nice feel to scroll eternally through time series and zoom in and out, without any restriction. Obviously, there are some limits, for example you might constrain the array length, and start to pull out data that's no longer visible, but I'm very happy with this behaviour. It really lifts the experience!

SamMorrowDrums:
@adidahiya is there anything further I should do to finish this?

adidahiya:
@SamMorrowDrums sorry for the delay, this looks great!

A5048:
> As mentioned in #3086 I am proposing to add events for Pan and Zoom, so that users can automatically retrieve more data for charts.
> 
> The proposed API is such:
> 
> ``` javascript
> panZoomInteraction = new Plottable.Interactions.PanZoom();
> 
> panZoomInteraction.onPan(callback);
> panZoomInteraction.offPan(callback);
> 
> panZoomInteraction.onZoom(callback);
> panZoomInteraction.offZoom(callback);
> 
> ```
> 
> Callbacks occur on the most applicable end event (touchend, dragend or scroll), in an attempt to minimise the number of calls where possible, as this will most likely be used for retrieving server data.
>

A5048 (commented):
@arash_sharifi_basht

PR #3109, opened 2016-11-16
Expose simpler API for selecting chart entities

themadcreator:
Add .entitiesAt method to base Plot class, which most plots
already implement.

Add missing .entitiesAt method using already existing "nearest"
methods for line and segment plots.

Add datasetIndex to Plot.PlotEntity object to easily determine
which series was selected.


CalvinFernandez (approved):
still looks good

PR #3111, opened 2016-11-17
Constrain pan/zoom to min/max domain values

themadcreator:
Previously it was only possible to set the min/max
domain EXTENT. This was only respected by the total
zoom level.

For example, if min/max extent was set to [0, 10],
then it was possible to pan to [10, 20] or zoom to
[-5, 5].

This change adds domain VALUE constraints to both pan
and zoom to respect the absolute domain boundaries.

The implementation for zoom constraints must consider
that the zoom interaction may violate the constraints
for min, max, or both.

We handle each of these cases seperately:
min only: We adjust the "center" coordinate of the zoom.
max only: We adjust the "center" coordinate of the zoom.
min and max: We adjust both the "center" and "zoom" values.

This also supports partial constraints (i.e. only min
or max set).

### DIFF
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -26,6 +26,9 @@ namespace Plottable.Interactions {
     private _minDomainExtents: Utils.Map<QuantitativeScale<any>, any>;
     private _maxDomainExtents: Utils.Map<QuantitativeScale<any>, any>;
 
+    private _minDomainValues: Utils.Map<QuantitativeScale<any>, any>;
+    private _maxDomainValues: Utils.Map<QuantitativeScale<any>, any>;
+
     private _panEndCallbacks = new Utils.CallbackSet<PanCallback>();
     private _zoomEndCallbacks = new Utils.CallbackSet<ZoomCallback>();
 
@@ -46,6 +49,8 @@ namespace Plottable.Interactions {
       this._touchIds = d3.map<Point>();
       this._minDomainExtents = new Utils.Map<QuantitativeScale<any>, number>();
       this._maxDomainExtents = new Utils.Map<QuantitativeScale<any>, number>();
+      this._minDomainValues = new Utils.Map<QuantitativeScale<any>, number>();
+      this._maxDomainValues = new Utils.Map<QuantitativeScale<any>, number>();
       if (xScale != null) {
         this.addXScale(xScale);
       }
@@ -126,12 +131,20 @@ namespace Plottable.Interactions {
         return { x: (point.x - oldPoints[i].x) / magnifyAmount, y: (point.y - oldPoints[i].y) / magnifyAmount };
       });
 
+      let oldCenterPoint = PanZoom._centerPoint(oldPoints[0], oldPoints[1]);
+      let centerX = oldCenterPoint.x;
+      let centerY = oldCenterPoint.y;
+
       this.xScales().forEach((xScale) => {
-        magnifyAmount = this._constrainedZoomAmount(xScale, magnifyAmount);
+        const constrained = this._constrainedZoom(xScale, magnifyAmount, centerX);
+        centerX = constrained.centerPoint;
+        magnifyAmount = constrained.zoomAmount;
       });
 
       this.yScales().forEach((yScale) => {
-        magnifyAmount = this._constrainedZoomAmount(yScale, magnifyAmount);
+        const constrained = this._constrainedZoom(yScale, magnifyAmount, centerY);
+        centerY = constrained.centerPoint;
+        magnifyAmount = constrained.zoomAmount;
       });
 
       let constrainedPoints = oldPoints.map((oldPoint, i) => {
@@ -141,17 +154,15 @@ namespace Plottable.Interactions {
         };
       });
 
-      let oldCenterPoint = PanZoom._centerPoint(oldPoints[0], oldPoints[1]);
-
-      let translateAmountX = oldCenterPoint.x - ((constrainedPoints[0].x + constrainedPoints[1].x) / 2);
+      let translateAmountX = centerX - ((constrainedPoints[0].x + constrainedPoints[1].x) / 2);
       this.xScales().forEach((xScale) => {
-        this._magnifyScale(xScale, magnifyAmount, oldCenterPoint.x);
+        this._magnifyScale(xScale, magnifyAmount, centerX);
         this._translateScale(xScale, translateAmountX);
       });
 
-      let translateAmountY = oldCenterPoint.y - ((constrainedPoints[0].y + constrainedPoints[1].y) / 2);
+      let translateAmountY = centerY - ((constrainedPoints[0].y + constrainedPoints[1].y) / 2);
       this.yScales().forEach((yScale) => {
-        this._magnifyScale(yScale, magnifyAmount, oldCenterPoint.y);
+        this._magnifyScale(yScale, magnifyAmount, centerY);
         this._translateScale(yScale, translateAmountY);
       });
     }
@@ -185,7 +196,7 @@ namespace Plottable.Interactions {
     }
 
     private _magnifyScale<D>(scale: QuantitativeScale<D>, magnifyAmount: number, centerValue: number) {
-      let magnifyTransform = (rangeValue: number) => scale.invert(centerValue - (centerValue - rangeValue) * magnifyAmount);
+      let magnifyTransform = (rangeValue: number) => scale.invert(this._zoomAt(rangeValue, magnifyAmount, centerValue));
       scale.domain(scale.range().map(magnifyTransform));
     }
 
@@ -201,26 +212,38 @@ namespace Plottable.Interactions {
 
         let deltaPixelAmount = e.deltaY * (e.deltaMode ? PanZoom._PIXELS_PER_LINE : 1);
         let zoomAmount = Math.pow(2, deltaPixelAmount * .002);
+        let centerX = translatedP.x;
+        let centerY = translatedP.y;
 
         this.xScales().forEach((xScale) => {
-          zoomAmount = this._constrainedZoomAmount(xScale, zoomAmount);
+          const constrained = this._constrainedZoom(xScale, zoomAmount, centerX);
+          centerX = constrained.centerPoint;
+          zoomAmount = constrained.zoomAmount;
         });
 
         this.yScales().forEach((yScale) => {
-          zoomAmount = this._constrainedZoomAmount(yScale, zoomAmount);
+          const constrained = this._constrainedZoom(yScale, zoomAmount, centerY);
+          centerY = constrained.centerPoint;
+          zoomAmount = constrained.zoomAmount;
         });
 
         this.xScales().forEach((xScale) => {
-          this._magnifyScale(xScale, zoomAmount, translatedP.x);
+          this._magnifyScale(xScale, zoomAmount, centerX);
         });
         this.yScales().forEach((yScale) => {
-          this._magnifyScale(yScale, zoomAmount, translatedP.y);
+          this._magnifyScale(yScale, zoomAmount, centerY);
         });
+
         this._zoomEndCallbacks.callCallbacks();
       }
     }
 
-    private _constrainedZoomAmount(scale: QuantitativeScale<any>, zoomAmount: number) {
+    private _constrainedZoom(scale: QuantitativeScale<number>, zoomAmount: number, centerPoint: number) {
+      zoomAmount = this._constrainZoomExtents(scale, zoomAmount);
+      return this._constrainZoomValues(scale, zoomAmount, centerPoint);
+    }
+
+    private _constrainZoomExtents(scale: QuantitativeScale<any>, zoomAmount: number) {
       let extentIncreasing = zoomAmount > 1;
 
       let boundingDomainExtent = extentIncreasing ? this.maxDomainExtent(scale) : this.minDomainExtent(scale);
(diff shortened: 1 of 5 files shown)


src/interactions/panZoomInteraction.ts
```diff
@@ -232,6 +255,92 @@ namespace Plottable.Interactions {
...
+
+      const minDomain = this.minDomainValue(scale);
+      const maxDomain = this.maxDomainValue(scale);
+
+      // if no constraints set, we're done
+      if (minDomain == null && maxDomain == null) {
+        return { centerPoint, zoomAmount };
+      }
+
+      const scaleDomain = scale.domain();
```
CalvinFernandez: nit: just for added clarity can we do `const [scaleDomainMin, scaleDomainMax] = scale.domain();`

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+     *
+     * Note that this differs from minDomainExtent/maxDomainExtent, in that
+     * those methods provide constraints such as showing at least 2 but no more
+     * than 5 values at a time.
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @returns {D} The minimum domain extent for the scale.
```
CalvinFernandez: "The minimum domain _value_ for the scale"

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+     *
+     * Note that this differs from minDomainExtent/maxDomainExtent, in that
+     * those methods provide constraints such as showing at least 2 but no more
+     * than 5 values at a time.
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @param {D} minDomainExtent The minimum domain extent for the scale.
```
CalvinFernandez: copy-paste bugz

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
+    public minDomainValue<D>(quantitativeScale: QuantitativeScale<D>, minDomainValue: D): this;
+    public minDomainValue<D>(quantitativeScale: QuantitativeScale<D>, minDomainValue?: D): any {
+      if (minDomainValue == null) {
+        return this._minDomainValues.get(quantitativeScale);
+      }
+      let maxValueForScale = this.maxDomainValue(quantitativeScale);
+      if (maxValueForScale != null && maxValueForScale.valueOf() < minDomainValue.valueOf()) {
+        throw new Error("minDomainValue must be smaller than maxDomainValue for the same Scale");
```
CalvinFernandez, line 616: should we care if they're equal? that also sounds invalid to me

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+     *
+     * Note that this differs from minDomainExtent/maxDomainExtent, in that
+     * those methods provide constraints such as showing at least 2 but no more
+     * than 5 values at a time.
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @returns {D} The maximum domain extent for the scale.
```
CalvinFernandez: "The maximum domain _value_ for the scale."

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+     *
+     * Note that this differs from minDomainExtent/maxDomainExtent, in that
+     * those methods provide constraints such as showing at least 2 but no more
+     * than 5 values at a time.
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @param {D} maxDomainExtent The maximum domain extent for the scale.
```
CalvinFernandez: copy-paste bug

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,117 @@ namespace Plottable.Interactions {
...
+    }
+
+    /**
+     * Uses the current domain of each scale to apply a minimum and maximum
+     * domain value for that scale.
+     *
+     * This constrains the pan/zoom interaction to show no more than the domain
+     * of the scale.
+     */
+    public constrainToScaleDomainValues() {
```
CalvinFernandez: is this method used anywhere?

src/interactions/panZoomInteraction.ts
```diff
@@ -244,19 +353,44 @@ namespace Plottable.Interactions {
...
+    /**
+     * Returns a new translation value that respects domain min/max value
+     * constraints.
+     */
+    private _constrainedTranslation(scale: QuantitativeScale<number>, translation: number) {
+      const scaleDomain = scale.domain();
+      if (translation > 0) {
+        const bound = this.maxDomainValue(scale);
+        if (bound != null) {
+          const currentRange = scale.scale(scaleDomain[1]);
```
CalvinFernandez: nit `currentMaxRange`

src/interactions/panZoomInteraction.ts
```diff
@@ -244,19 +353,44 @@ namespace Plottable.Interactions {
...
+          const maxRange = scale.scale(bound);
+          translation = Math.min(currentRange + translation, maxRange) - currentRange;
```
CalvinFernandez: this is so clean 👍

src/interactions/panZoomInteraction.ts
```diff
@@ -232,6 +255,92 @@ namespace Plottable.Interactions {
...
+
+      // add fallback to prevent overflowing both min and max
+      if (maxDomain != null && maxDomain != null) {
+        const maxRange = scale.scale(maxDomain);
+        const currentMaxRange = scale.scale(scaleDomain[1]);
+        const testMaxRange = this._zoomAt(currentMaxRange, zoomAmount, centerPoint);
+
+        const minRange = scale.scale(minDomain);
+        const currentMinRange = scale.scale(scaleDomain[0]);
+        const testMinRange = this._zoomAt(currentMinRange, zoomAmount, centerPoint);
```
CalvinFernandez, line 306: could probably DRY this if you're feeling particularly motivated

CalvinFernandez (commented):
looking good! needs some tests I think

src/interactions/panZoomInteraction.ts
themadcreator: it's a new public method that we'll use in the blueprint-chart to easily attach constraints using the existing scale

src/interactions/panZoomInteraction.ts
themadcreator: Yeah, there is some duplication here, but I couldn't think of a way to do that without losing some of the clarity of flow of control especially with all the null checking required for each of these cases. Also these calculations are super fast, so it's not a perf concern.

src/interactions/panZoomInteraction.ts
CalvinFernandez: ah okay -- the name of the method doesn't make it super clear what it does; this is subjective of course. @hellochar thoughts on this ? does it make sense to you?

src/interactions/panZoomInteraction.ts
```diff
@@ -232,6 +255,92 @@ namespace Plottable.Interactions {
       return compareF(zoomAmount, boundingDomainExtent / domainExtent);
     }
 
+    private _constrainZoomValues(scale: QuantitativeScale<number>, zoomAmount: number, centerPoint: number) {
```
hellochar, line 258: I didn't think through the math here but just want to make sure this will work for logarithmic scales as well?

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,112 @@ namespace Plottable.Interactions {
...
+      this.yScales().forEach((scale) => this._constrainToScaleDomainValues(scale));
+    }
+
+    private _constrainToScaleDomainValues<D>(scale: QuantitativeScale<D>) {
+      this._minDomainValues.delete(scale);
+      this._maxDomainValues.delete(scale);
+      const [ domainMin, domainMax ] = scale.domain();
+      this.minDomainValue(scale, domainMin);
+      this.maxDomainValue(scale, domainMax);
+    }
```
hellochar, line 678: this file is getting really large - over 700 lines. We should refactor this soon (would be good to revisit the codebase as a whole here)

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,112 @@ namespace Plottable.Interactions {
...
+     * those methods provide constraints such as showing at least 2 but no more
+     * than 5 values at a time.
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @returns {D} The minimum domain value for the scale.
+     */
+    public minDomainValue<D>(quantitativeScale: QuantitativeScale<D>): D;
```
hellochar, line 593: Is it worth calling out that the user must first `.add` the scale?

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,112 @@ namespace Plottable.Interactions {
...
+     * @param {D} maxDomainExtent The maximum domain value for the scale.
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
+    public maxDomainValue<D>(quantitativeScale: QuantitativeScale<D>, maxDomainValue: D): this;
+    public maxDomainValue<D>(quantitativeScale: QuantitativeScale<D>, maxDomainValue?: D): any {
+      if (maxDomainValue == null) {
+        return this._maxDomainValues.get(quantitativeScale);
+      }
+      let minValueForScale = this.minDomainValue(quantitativeScale);
+      if (minValueForScale != null && maxDomainValue.valueOf() <= minValueForScale.valueOf()) {
```
hellochar, line 658: why do we `.valueOf`?

src/interactions/panZoomInteraction.ts
```diff
@@ -439,6 +577,112 @@ namespace Plottable.Interactions {
...
+     *
+     * By contrast, minDomainValue/maxDomainValue set a boundary beyond which
+     * the user cannot pan/zoom.
+     *
+     * @param {QuantitativeScale<any>} quantitativeScale The scale to query
+     * @param {D} maxDomainExtent The maximum domain value for the scale.
+     * @returns {Interactions.PanZoom} The calling PanZoom Interaction.
+     */
+    public maxDomainValue<D>(quantitativeScale: QuantitativeScale<D>, maxDomainValue: D): this;
+    public maxDomainValue<D>(quantitativeScale: QuantitativeScale<D>, maxDomainValue?: D): any {
```
hellochar: nit: return type should be `D | this`

Aside: how do you guys feel about this fluent method style for settable properties? The method naming feels confusing to me. Perhaps changing these to getter/setters would help differentiate between "properties that you can set" and "methods that act" (e.g. `attachTo` or `computeLayout`)?

src/interactions/panZoomInteraction.ts
hellochar: I didn't know what this method did until I read your comment. Can we just implement this on the blueprint-chart side?

hellochar (commented):
Looks fine; would like to check log scales just to ensure this works before we merge.

src/interactions/panZoomInteraction.ts
CalvinFernandez: I'm also for letting end-users implement this piece.

src/interactions/panZoomInteraction.ts
themadcreator: As with minDomainExtent/maxDomainExtent, you don't actually have to add the scale to the interaction before this. You could also add it after.

src/interactions/panZoomInteraction.ts
themadcreator: Okay, the type is easy enough to change.

As to your other comment, if we really wanted to differentiate these, why not use actual getter/setter properties? Obviously it exposes the possibility of side effects when doing things as simple assignment. Nevertheless, it seems like a more modern approach.

src/interactions/panZoomInteraction.ts
themadcreator: I don't know, I was following the maxDomainExtent function as a guide. Figure I should be consistent with the code I find there since I'm still getting to know it.

src/interactions/panZoomInteraction.ts
themadcreator: your point about constrainToScaleDomainValues, I think we should keep it since it's bundles up a few lines of functionality. At the very least, we should have a method to simultaneously set the min/max.

Otherwise, we can get into situations where the domain goes from [0, 100] -> [400, 500] and the first call to `minDomainValue(scale, 400)` would fail because `maxDomainValue(scale)` is still set to 100

hellochar (commented):
Did we test this with log scales? Looks fine otherwise

hellochar (changes requested):
Oh sorry @themadcreator you'll need to add the built plottable.js and plottable-npm.d.ts files as well (see https://github.com/palantir/plottable#contributing)

hellochar (approved):

PR #3112, opened 2016-11-18
[slightly breaking] Fix ts2.0 errors

hellochar:
We were on typescript 2.0 but `grunt-ts` wasn't updated and didn't support the new features. This change upgrades grunt-ts to understand ts2.0 errors and fixes the ones that existed.

The main change is that Axis orientations previously were any `"string"` type, but now we've defined an alias `type AxisOrientation = "left" | "top" | "right" | "bottom";` that `.orientation()` sets/gets. 

This will break existing ts users who pass a `"string"` type into orientation, although it's unclear how many users will actually be affected (how many are even using typescript?). Alternatively we could push back that change into our next major rev. Thoughts?

Also: can someone point me to how to generate docs locally?

### DIFF
--- a/plottable-npm.d.ts
+++ b/plottable-npm.d.ts
@@ -1665,6 +1665,7 @@ declare namespace Plottable.Components {
     }
 }
 declare namespace Plottable {
+    type AxisOrientation = "left" | "top" | "right" | "bottom";
     class Axis<D> extends Component {
         /**
          * The css class applied to each end tick mark (the line on the end tick).
@@ -1720,9 +1721,9 @@ declare namespace Plottable {
          *
          * @constructor
          * @param {Scale} scale
-         * @param {string} orientation One of "top"/"bottom"/"left"/"right".
+         * @param {AxisOrientation} orientation Orientation of this Axis.
          */
-        constructor(scale: Scale<D, number>, orientation: string);
+        constructor(scale: Scale<D, number>, orientation: AxisOrientation);
         destroy(): void;
         protected _isHorizontal(): boolean;
         protected _computeWidth(): number;
@@ -1858,14 +1859,14 @@ declare namespace Plottable {
         /**
          * Gets the orientation of the Axis.
          */
-        orientation(): string;
+        orientation(): AxisOrientation;
         /**
          * Sets the orientation of the Axis.
          *
-         * @param {number} orientation One of "top"/"bottom"/"left"/"right".
+         * @param {AxisOrientation} orientation The orientation to apply to this axis.
          * @returns {Axis} The calling Axis.
          */
-        orientation(orientation: string): this;
+        orientation(orientation: AxisOrientation): this;
         /**
          * Gets whether the Axis shows the end tick labels.
          */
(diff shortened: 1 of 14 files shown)


CalvinFernandez:
whoops of course

src/axes/numericAxis.ts
```diff
@@ -13,9 +13,9 @@ namespace Plottable.Axes {
      *
      * @constructor
      * @param {QuantitativeScale} scale
-     * @param {string} orientation One of "top"/"bottom"/"left"/"right".
+     * @param {AxisOrientaiton} orientation Orientation of this Numeric Axis.
```
CalvinFernandez: `AxisOrientaiton` => `AxisOrientation`

src/axes/timeAxis.ts
```diff
@@ -258,9 +264,9 @@ namespace Plottable.Axes {
       return mostPreciseIndex;
     }
 
-    public orientation(): string;
-    public orientation(orientation: string): this;
-    public orientation(orientation?: string): any {
+    public orientation(): TimeAxisOrientation;
+    public orientation(orientation: TimeAxisOrientation): this;
+    public orientation(orientation?: TimeAxisOrientation): any {
```
CalvinFernandez: does this still return `any`?

src/core/formatters.ts
```diff
@@ -216,6 +216,7 @@ namespace Plottable.Formatters {
           return d3.time.format(timeFormat[i].format)(d);
         }
       }
+      return undefined;
```
CalvinFernandez: can you change timeFormat[7] to the default case and return this at the end instead of `undefined` functionally the same but makes the default case easier to find

src/plots/segmentPlot.ts
```diff
@@ -245,6 +245,7 @@ namespace Plottable.Plots {
             return this._lineIntersectsSegment(startPoint, endPoint, point, corners[index - 1]) &&
                    this._lineIntersectsSegment(point, corners[index - 1], startPoint, endPoint) ;
           }
+          return undefined;
```
CalvinFernandez: whats the difference between returning undefined and returning false?

src/plots/segmentPlot.ts
```diff
@@ -245,6 +245,7 @@ namespace Plottable.Plots {
             return this._lineIntersectsSegment(startPoint, endPoint, point, corners[index - 1]) &&
                    this._lineIntersectsSegment(point, corners[index - 1], startPoint, endPoint) ;
```
CalvinFernandez: nit: space at the end for some reason, `) ;`

plottable-npm.d.ts
```diff
@@ -1665,6 +1665,7 @@ declare namespace Plottable.Components {
     }
 }
 declare namespace Plottable {
+    type AxisOrientation = "left" | "top" | "right" | "bottom";
```
CalvinFernandez: not currently the standard in this library so perhaps not worth it but I think it would be worth alphabetizing things   like this

CalvinFernandez (changes requested):
few small things, looks good overall

src/plots/segmentPlot.ts
hellochar: good catch, false is more appropriate

CalvinFernandez (approved):

PR #3115, opened 2016-11-22
Bugfix: Make SVGTypewriter work with browserify

### LINKED ISSUE #3114: v2.3.0 NPM distribution
When building a JS bundle with Browserify, I get the error`SVGTypewiter is not defined` in the browser console, when trying to draw charts. I've tried bundling without minifying etc. and the issue does not occur with 2.2.0, so I am assuming that there is a genuine issue with the NPM distribution of Plottable 2.3.0.

CalvinFernandez:
Make svg-typewriter explicit npm dependency
Remove concat task from Grunt
Add svg-typewriter as dependency in UMD config

fixes #3114


giladgray (approved):
looks reasonable to me but this is my first comment on plottable...

package.json
```diff
@@ -47,11 +47,11 @@
...
     "requirejs": "2.1.18",
     "sinon": "1.16.1",
-    "svg-typewriter": "0.3.0-beta.0",
     "tslint": "3.15.1",
     "typescript": "2.0.6"
   },
   "dependencies": {
-    "d3": "^3.5.16"
+    "d3": "^3.5.16",
+    "svg-typewriter": "^0.3.0-beta.0"
```
adidahiya, line 55: we should make a real non-beta release of this very soon

adidahiya (approved):

PR #3119, opened 2016-11-30
Infrastructure: Re-add SVGTypewriter as bower dep

### LINKED ISSUE #3118: plottable.js breaks when used standalone
Prior to 2.3.0 we bundled SVGTypewriter into plottable.js directly; this is no longer the case and is breaking existing users who pull in plottable standalone through a script tag. This is also breaking bower users. We should unbreak people for the time being and have a better dependency story in 3.0

CalvinFernandez:
Add SVGTypewriter to bower components
Remove SVGTypewriter as npm dependency
Change references to library from `node_modules` to `bower_components`
Add svg typewriter concat task to Grunt

fixes #3118


adidahiya:
This isn't quite what I meant in https://github.com/palantir/plottable/issues/3118#issuecomment-263894796 -- I believe you can bundle svg typewriter from node_modules, you don't need to check it in as a bower component.

hellochar:
So, to recap:

(a) users pulling the code in with npm, bower, or referencing with a script tag directly will all get the same behavior, which is just that plottable.js contains the source code of svgtypewriter itself (and as such, users only need to pull in d3 separately, which was the contract before).
(b) users can use plottable.js in a requireJS environment, or in a "use window.Plottable directly" environment, which was the contract before

Is that correct?

hellochar (approved):
Looks good. Can we get a ticket tracking removing this when 3.0 comes around?

adidahiya (approved):

PR #3121, opened 2016-12-01
tickLabelMaxWidth on Category Axes

hellochar:
Adds a new property `tickLabelMaxWidth` on Categorical axes. When set, tick labels will be truncated with ellipsis to be at most `tickLabelMaxWidth()` pixels wide, and will also never span more than one line. This ensures the axis doesn't grow to an undesirable width (or, through wrapping, grow to an undesirable height).


src/axes/categoryAxis.ts
```diff
@@ -97,11 +128,42 @@ namespace Plottable.Axes {
...
-     * Measures the size of the ticks while also writing them to the DOM.
-     * @param {d3.Selection} ticks The tick elements to be written to.
+     * Set or get the tick label's max width on this axis. When set, tick labels will be truncated with ellipsis to be
+     * at most `tickLabelMaxWidth()` pixels wide, and will also never span more than one line. This ensures the axis
+     * doesn't grow to an undesirable width (or, through wrapping, grow to an undesirable height).
+     *
+     * Passing no arguments retrieves the value, while passing a number sets the value. Pass undefined to un-set the max
+     * width.
+     * @param maxWidth
+     * @returns {any}
```
CalvinFernandez: returns `number | this` ?

CalvinFernandez (approved):
looks fine

CalvinFernandez:
@hellochar no tickLabelMaxWidth for other axis types?

hellochar:
@CalvinFernandez linear axes use a different codepath for drawing ticks that doesn't use measurers/wrappers at all so it'd be more work to implement. Lets release just for categorical and see what the need is for numeric (I suspect many people can get away with using formatters for numeric to reduce tick label length anyways)

PR #3123, opened 2016-12-01
Prepare release v2.4.0

hellochar:


hellochar:
# Release notes preview

## Features

### tickLabelMaxWidth

Adds a new property tickLabelMaxWidth on Categorical axes. When set, tick labels will be truncated with ellipsis to be at most tickLabelMaxWidth() pixels wide, and will also never span more than one line. This ensures the axis doesn't grow to an undesirable width (or, through wrapping, grow to an undesirable height).

### Typescript API Breaks

* Typescript 1.8 and above is now required since we are now using string literals in our codebase.
* Axis orientations previously were any "string" type, but now we've defined a string literal type `AxisOrientation = "left" | "top" | "right" | "bottom"` that `Axis.orientation()` sets/gets.

## Upgrade Notes

*  Existing users who pass a "string" type into orientation may get an "Argument of type 'string' is not assignable to parameter of type 'AxisOrientation'." error. Ensure the string you're passing in really is only one of the four orientations (`"left" | "top" | "right" | "bottom"`), and then type your variable/parameter as an AxisOrientation appropriately.

themadcreator (approved):

adidahiya (approved):

PR #3130, opened 2016-12-05
Interactions: Expose raw click event

CalvinFernandez:
Expose raw click or touch event on mouse down / mouse up
actions.


hellochar (approved):
Code looks fine. What's the main use case for this again?

CalvinFernandez:
We're using it to determine if the metaKey was pressed while clicking in order to implement multiselect

PR #3136, opened 2016-12-09
Drag zoom

CalvinFernandez:
This is something we still want -- rebasing @danmane's original pr

### DIFF
--- a/src/components/dragZoomLayer.ts
+++ b/src/components/dragZoomLayer.ts
@@ -1,5 +1,3 @@
-///<reference path="../reference.ts" />
-
 module Plottable {
 export module Components {
   export class DragZoomLayer extends Components.SelectionBoxLayer {
@@ -96,8 +94,8 @@ export module Components {
 
       if (!this.isZoomed) {
         this.isZoomed = true;
-        this.xDomainToRestore = this._xScale.domain();
-        this.yDomainToRestore = this._yScale.domain();
+        this.xDomainToRestore = this.xScale().domain();
+        this.yDomainToRestore = this.yScale().domain();
       }
       this.interpolateZoom(x0, x1, y0, y1);
     }
@@ -119,10 +117,10 @@ export module Components {
     }
 
     private interpolateZoom(x0f: Numeric, x1f: Numeric, y0f: Numeric, y1f: Numeric) {
-      let x0s: number = this._xScale.domain()[0].valueOf();
-      let x1s: number = this._xScale.domain()[1].valueOf();
-      let y0s: number = this._yScale.domain()[0].valueOf();
-      let y1s: number = this._yScale.domain()[1].valueOf();
+      let x0s: number = this.xScale().domain()[0].valueOf();
+      let x1s: number = this.xScale().domain()[1].valueOf();
+      let y0s: number = this.yScale().domain()[0].valueOf();
+      let y1s: number = this.yScale().domain()[1].valueOf();
 
       // Copy a ref to the ease fn, so that changing ease wont affect zooms in progress
       let ease = this.easeFn;
@@ -139,8 +137,8 @@ export module Components {
         let x1 = interpolator(x1s, x1f, p);
         let y0 = interpolator(y0s, y0f, p);
         let y1 = interpolator(y1s, y1f, p);
-        this._xScale.domain([x0, x1]);
-        this._yScale.domain([y0, y1]);
+        this.xScale().domain([x0, x1]);
+        this.yScale().domain([y0, y1]);
         if (p < 1) {
           Utils.DOM.requestAnimationFramePolyfill(draw);
         } else {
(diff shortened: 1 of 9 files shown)


src/core/interfaces.ts
```diff
@@ -72,4 +72,7 @@ export interface Entity<C extends Component> {
   selection: d3.Selection<any>;
   component: C;
 }
+
+export type Numeric = number | {valueOf(): number};
```
hellochar, line 76: I'm guessing the point of having this type is to accept number or Date objects? I dug around a lot but couldn't find a clear answer.

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+      this._dragInteraction.onDragEnd((startPoint, endPoint) => {
+        this.boxVisible(false);
+        this.bounds({topLeft: startPoint, bottomRight: endPoint});
+        if (dragging) {
+          this.zoom();
+        }
+        dragging = false;
+      });
+
+      this._doubleClickInteraction.onDoubleClick(() => this.unzoom());
```
hellochar, line 53: this adds double-click to unzoom; we should call that out

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+      let y1s: number = this.yScale().domain()[1].valueOf();
+
+      // Copy a ref to the ease fn, so that changing ease wont affect zooms in progress
+      let ease = this.easeFn;
+      let interpolator = (a: Numeric, b: Numeric, p: number): number => {
+        return d3.interpolateNumber(a.valueOf(), b.valueOf())(ease(p));
+      };
+      this.isZooming(true);
+      let start = Date.now();
+      let draw = () => {
```
hellochar, line 132: Wonder if we should be implementing our own animation logic here instead of using a Plottable.Animator

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+      let y1: number = this.yExtent()[0].valueOf();
+
+      if (x0 === x1 || y0 === y1) {
+        return;
+      }
+
+      if (!this.isZoomed) {
+        this.isZoomed = true;
+        this.xDomainToRestore = this.xScale().domain();
+        this.yDomainToRestore = this.yScale().domain();
```
themadcreator, line 98: Is it possible the domain will change after the scale is zoomed? These may become out of date

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+    private unzoom() {
+      if (!this.isZoomed) {
+        return;
+      }
+      this.isZoomed = false;
+      this.interpolateZoom(this.xDomainToRestore[0], this.xDomainToRestore[1],
+                           this.yDomainToRestore[0], this.yDomainToRestore[1]);
+    }
+
+    // If we are zooming, disable interactions, to avoid contention
```
themadcreator, line 113: Cool, but will other interactions contend as well?

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+      let draw = () => {
+        let now = Date.now();
+        let passed = now - start;
+        let p = this._animationTime === 0 ? 1 : Math.min(1, passed / this._animationTime);
+        let x0 = interpolator(x0s, x0f, p);
+        let x1 = interpolator(x1s, x1f, p);
+        let y0 = interpolator(y0s, y0f, p);
+        let y1 = interpolator(y1s, y1f, p);
+        this.xScale().domain([x0, x1]);
+        this.yScale().domain([y0, y1]);
```
themadcreator, line 141: This doesn't respect max zoom extents. Could we instead use the PanZoomInteraction now that we expose the .zoom method?

src/components/dragZoomLayer.ts
```diff
@@ -0,0 +1,152 @@
...
+    private _animationTime = 750;
+
+    /* Constructs a SelectionBoxLayer with an attached DragInteraction and DoubleClickInteraction.
+     * On drag, it triggers an animated zoom into the box that was dragged.
+     * On double click, it zooms back out to the original view, before any zooming.
+     * The zoom animation uses an easing function (default d3.ease("cubic-in-out")) and is customizable.
+     * Usage: Construct the selection box layer and attach x and y scales, and then add the layer
+     * over the plot you are zooming on using a Component Group.
+     */
+    constructor(xScale: QuantitativeScale<Numeric>,
```
themadcreator, line 19: now that Category scales support pan/zoom maybe we could instead pass a panzoominteraction here? Or pass in TransformableScale types?

themadcreator (changes requested):

adidahiya:
Is this still relevant? If not, please close as this is getting pretty stale.

PR #3137, opened 2016-12-13
Enable pan/zoom on categorical scales

themadcreator:
Introduces new interface `TransformableScale`. This interface abstracts the
necessary API for implementing pan/zoom on various types of scales.

Due to some limitations of d3, for example category scales can't invert screen
space coordinates, we nevertheless allow the scale types to support pan/zoom if
they implement this interface.

Implemented `TransformableScale` interface on `Linear`, `Log`, `Time` and
`Category` Scales. The three `Quantitative` scales have fairly trivial
implementations, but the `Category` scale requires special handling.

Added an intermediate coordinate space called the *Transformation Space* to the
`Category` scale. The conversion from data to screen coordinates is as follows:

Data Space -> _d3Scale -> Transformation Space -> _d3TransformationScale -> Screen Space.

The *Transformation Space* coordinates are initialized to [0, 1].

Notably, range band calculations are internally computed in *Transformation
Space* and transformed to screen space in methods like `rangeBand()` and
`stepWidth()`.


themadcreator:
lol 
image

src/axes/categoryAxis.ts
```diff
@@ -300,6 +300,12 @@ namespace Plottable.Axes {
       let xTranslate = this.orientation() === "right" ? this._tickSpaceRequired() : 0;
       let yTranslate = this.orientation() === "bottom" ? this._tickSpaceRequired() : 0;
       Utils.DOM.translate(this._tickLabelContainer, xTranslate, yTranslate);
+
+      // hide ticks and labels that overflow the axis
+      this._showAllTickMarks();
+      this._showAllTickLabels();
+      this._hideOverflowingTickLabels();
+      this._hideTickMarksWithoutLabel();
```
hellochar, line 308: Sometimes if the space for a tick label is too small, it won't get drawn: https://jsfiddle.net/hellochar/t8j5d7vt/ Will these "non-labelled tick-marks" get removed now?

src/scales/linearScale.ts
```diff
@@ -25,6 +25,14 @@ namespace Plottable.Scales {
       return this._d3Scale(value);
     }
 
+    public scaleTransformation(value: number) {
+      return this.scale(value);
+    }
+
+    public getTransformationDomain() {
+      return this.domain() as [number, number];
```
hellochar, line 33: nit: might make sense to have this.domain() return a `[number, number]` type directly

src/scales/quantitativeScale.ts
```diff
@@ -256,6 +257,24 @@ export class QuantitativeScale<D> extends Scale<D, number> {
     }
   }
 
+  public magnify(magnifyAmount: number, centerValue: number) {
```
hellochar, line 260: Would it make sense for the `Plottable.Scale` class to declare abstract methods for `magnify`/`translate`/`scaleTransformation`/`getTransformationDomain` ? And then make quantitativeScale abstract, to pass down scaleTransformation and getTransformationDomain?

src/scales/categoryScale.ts
```diff
@@ -130,8 +154,34 @@ namespace Plottable.Scales {
...
     public scale(value: string): number {
-      // scale it to the middle
-      return this._d3Scale(value) + this.rangeBand() / 2;
+      // Determine the middle of the range band for the value
+      let untransformed = this._d3Scale(value) + this._d3Scale.rangeBand() / 2;
+      // Convert to screen space
+      return this._d3TransformationScale(untransformed);
+    }
+
+    public magnify(magnifyAmount: number, centerValue: number) {
```
hellochar, line 163: Is this the "single source of truth" for panning/zooming a scale? If so, PanZoomInteraction should use these

hellochar (changes requested):
Looks good in general; mainly wondering about the hiding tick marks possible regression

src/scales/quantitativeScale.ts
CalvinFernandez: yes also wondering this

src/scales/categoryScale.ts
themadcreator: Yeah, it is and PanZoomInteraction does use these.

src/scales/quantitativeScale.ts
themadcreator: Not all `Plottable.Scale`s are transformable (e.g. color scales). So, I added the `TransformableScale` interface to the particular scales that implement it. Quantitative scales implements this interface and are essentially abstract. However, Plottable doesn't use true abstract types, but instead implements error-throwing methods, so i followed that pattern.

CalvinFernandez: 👍

src/scales/linearScale.ts
themadcreator: Scale's `this.domain` returns `D[]` where `D` is the templated domain type. It should probably return `[D, D]`, but I don't want to redo all the other instances.

src/interactions/panZoomInteraction.ts
```diff
@@ -3,14 +3,74 @@ namespace Plottable.Interactions {
...
+    return center - (center - value) * zoom;
+  }
+
+  /**
+   * Type guarded function to check if the scale implements the
+   * `TransformableScale` interface. Unfortunately, there is no way to do
+   * runtime interface typechecking, so we have to explicitly list all classes
+   * that implement the interface.
+   */
+  export function isTransformable(scale: any): scale is TransformableScale {
```
hellochar: Do we use this method anywhere?

src/interactions/panZoomInteraction.ts
```diff
@@ -3,14 +3,74 @@ namespace Plottable.Interactions {
...
+  /**
+   * This interface abstracts the necessary API for implementing pan/zoom on
+   * various types of scales.
+   *
+   * Due to some limitations of d3, for example category scales can't invert
+   * screen space coordinates, we nevertheless allow the scale types to support
+   * pan/zoom if they implement this interface. See `Category`'s
+   * `_d3TransformationScale` for more info.
+   */
+  export interface TransformableScale {
```
hellochar: Ohhh my god Github hid this whole file from me... Okay this makes a lot more sense. Imo this interface belongs in `Plottable.Scales` instead of `Plottable.Interactions` (and you can `type TransformableScale = Plottable.Scales.TransformableScale` in this file to prevent needing to type it out everywhere)

hellochar (approved):
Tested this branch on my jsfiddle and it was fine; looks good!

src/axes/categoryAxis.ts
themadcreator: This won't be an issue because those labels aren't hidden with the CSS visibility rule. Instead the SVGTypewriter just doesn't draw the text because there isn't enough space. But the label is still technically "visible".

src/interactions/panZoomInteraction.ts
themadcreator: We need it in the internal library to detect which scales can be pan/zoomed.

src/interactions/panZoomInteraction.ts
themadcreator: Sure, I'll move it.

Issue #3122, opened 2016-12-01
Add `entitiesIn` to all XY Plots


CalvinFernandez:
I would like to be able to determine which entities lie along a given x/y dimension. entities in as implemented for bar chart works very well for this case, however I need it to work for all of my xy plot types.

PR #3138, opened 2016-12-14
Interaction: add `entitiesIn` to line plot

CalvinFernandez:
a component of #3122

### DIFF
--- a/src/plots/linePlot.ts
+++ b/src/plots/linePlot.ts
@@ -348,6 +348,44 @@ namespace Plottable.Plots {
     }
 
     /**
+     * Gets the Entities that intersect the Bounds.
+     *
+     * @param {Bounds} bounds
+     * @returns {PlotEntity[]}
+     */
+    public entitiesIn(bounds: Bounds): PlotEntity[];
+    /**
+     * Gets the Entities that intersect the area defined by the ranges.
+     *
+     * @param {Range} xRange
+     * @param {Range} yRange
+     * @returns {PlotEntity[]}
+     */
+    public entitiesIn(xRange: Range, yRange: Range): PlotEntity[];
+    public entitiesIn(xRangeOrBounds: Range | Bounds, yRange?: Range): PlotEntity[] {
+      let dataXRange: Range;
+      let dataYRange: Range;
+      if (yRange == null) {
+        let bounds = (<Bounds> xRangeOrBounds);
+        dataXRange = { min: bounds.topLeft.x, max: bounds.bottomRight.x };
+        dataYRange = { min: bounds.topLeft.y, max: bounds.bottomRight.y };
+      } else {
+        dataXRange = (<Range> xRangeOrBounds);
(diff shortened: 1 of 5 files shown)


src/plots/linePlot.ts
```diff
@@ -348,6 +348,44 @@ namespace Plottable.Plots {
...
+        dataXRange = { min: bounds.topLeft.x, max: bounds.bottomRight.x };
+        dataYRange = { min: bounds.topLeft.y, max: bounds.bottomRight.y };
+      } else {
+        dataXRange = (<Range> xRangeOrBounds);
+        dataYRange = yRange;
+      }
+
+      const xProjector = Plot._scaledAccessor(this.x());
+      const yProjector = Plot._scaledAccessor(this.y());
+      return this.entities().filter((entity) => {
```
themadcreator, line 379: perf note. I noticed that plot's `.entities` method does a couple linear time passes on the data. Notably, it uses `_lightweightEntities` then maps them to "heavyweight" entities, which as far as i can tell, just adds a d3 selection. For perf, we should maybe consider using lightweight entities when filtering, then convert to `PlotEntities` as a post process. Lots of other methods use `this.entities()` so, probably not worth in this PR, but for the future.

themadcreator: Filed as #3145

test/plots/linePlotTests.ts
```diff
@@ -199,6 +199,60 @@ describe("Plots", () => {
         linePlot.renderTo(svg);
       });
 
+      afterEach(() => {
+        svg.remove();
```
themadcreator, line 203: heh, no brainer

themadcreator (approved):
Looks good. Keep em coming'

src/plots/linePlot.ts
CalvinFernandez: 👍

src/plots/linePlot.ts
```diff
@@ -348,6 +348,44 @@ namespace Plottable.Plots {
...
+    public entitiesIn(bounds: Bounds): PlotEntity[];
+    /**
+     * Gets the Entities that intersect the area defined by the ranges.
+     *
+     * @param {Range} xRange
+     * @param {Range} yRange
+     * @returns {PlotEntity[]}
+     */
+    public entitiesIn(xRange: Range, yRange: Range): PlotEntity[];
+    public entitiesIn(xRangeOrBounds: Range | Bounds, yRange?: Range): PlotEntity[] {
```
hellochar, line 365: nit: should we declare an abstract `entitiesIn` method on XYPlot?

src/plots/linePlot.ts
```diff
@@ -348,6 +348,44 @@ namespace Plottable.Plots {
...
+      let dataXRange: Range;
```
hellochar, line 366: nit: we use this Range type to specify a range of numbers, but I've also seen `[number, number]` in some places

PR #3151, opened 2017-01-09
Component API: Expose Resize Event

CalvinFernandez:
Expose a resize event on `Component` to aid in knowing when a size
has been set on a component.

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -11,6 +11,9 @@ export namespace Components {
     static CENTER = "center";
   }
 }
+
+export type IResizeHandler = (size: Plottable.Size) => void;
+
 export class Component {
   private _element: d3.Selection<void>;
   private _content: d3.Selection<void>;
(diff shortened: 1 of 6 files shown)


src/components/component.ts
```diff
@@ -11,13 +11,17 @@ export namespace Components {
     static CENTER = "center";
   }
 }
+
+export type IResizeHandler = (size: Plottable.Size) => void;
```
hellochar: nit: would prefer to not introduce a new Plottable.Size interface and just pass two width/height numbers directly. Introducing a new interface can cause confusion later on since some parts of the codebase use it while other parts don't

src/components/component.ts
```diff
@@ -214,10 +220,15 @@ export class Component {
     };
     this._element.attr("transform", "translate(" + this._origin.x + "," + this._origin.y + ")");
     this._boxes.forEach((b: d3.Selection<void>) => b.attr("width", this.width()).attr("height", this.height()));
+
+    if (this._resizeHandler !== undefined) {
```
hellochar: I'd `!= null` here; we want to be lenient on many different types

hellochar (changes requested):

hellochar (approved):

PR #3155, opened 2017-01-11
add tickLabelMaxLines property to category scale

hellochar:
Allows fine grained control over how many lines the tick label should be.

**Warning**: This modifies the existing behavior of `tickLabelMaxWidth` -- when set, maxLines no longer defaults to 1. Instead, maxLines is fully controlled by `tickLabelMaxLines`.

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -5029,8 +5029,8 @@ var Plottable;
                  */
                 get: function () {
                     var wrapper = new SVGTypewriter.Wrappers.Wrapper();
-                    if (this._tickLabelMaxWidth != null) {
-                        wrapper.maxLines(1);
+                    if (this._tickLabelMaxLines != null) {
+                        wrapper.maxLines(this._tickLabelMaxLines);
                     }
                     return wrapper;
                 },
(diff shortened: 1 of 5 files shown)


plottable.js
```diff
@@ -5029,8 +5029,8 @@ var Plottable;
                  */
                 get: function () {
                     var wrapper = new SVGTypewriter.Wrappers.Wrapper();
-                    if (this._tickLabelMaxWidth != null) {
```
themadcreator, line 5032: We default back to 1 if _tickLabelMaxLines isn't set, right?

src/axes/categoryAxis.ts
```diff
@@ -13,8 +20,8 @@ namespace Plottable.Axes {
      */
     private get _wrapper() {
       const wrapper = new SVGTypewriter.Wrappers.Wrapper();
-      if (this._tickLabelMaxWidth != null) {
-        wrapper.maxLines(1);
+      if (this._tickLabelMaxLines != null) {
+        wrapper.maxLines(this._tickLabelMaxLines)
```
themadcreator, line 24: Same

themadcreator (changes requested):

plottable.js
hellochar: No, we now don't have any default maxLines. I felt it was simpler from an API perspective. Thoughts?

src/axes/categoryAxis.ts
hellochar: This is the same code as the one in plottable.js; I committed the built file as well

themadcreator: Ah, right. Okay well I'm fine so long as it's on purpose.

themadcreator (approved):

Issue #1041, opened 2014-09-15
[Pie] angle-setting API


jtlan:
- [ ] startAngle api point
  - [ ] Default "startAngle" (0)
- [ ] endAngle api point
  - [ ] Default "endAngle" (2 \* pi)

CalvinFernandez:
completed in #3156

PR #3156, opened 2017-01-11
Add alignment and start/end angle to pie plots

carlodx:
Addresses issue [#1041](https://github.com/palantir/plottable/issues/1041)

Adding to pie plot: 
- `startAngle(angle: number)`:  sets the start angle, so pies do not need to be a complete circle
- `endAngle(angle: number)` : sets the end angle, so pies do not need to be a complete circle
- `hAlign(alignment: string)`: the alignment is by default in the center of the render area. using `"left"` or `"right"` you can align the horizontal axis to the left or the right
- `vAlign(alignment: string)`: the alignment is by default in the center of the render area. using `"top"` or `"bottom"` you can align the vertical axis to the top or the bottom.

### DIFF
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -229,7 +229,7 @@ namespace Plottable.Plots {
     }
 
     /**
-     * Gets whether slice labels are enabled.
+     * Gets the end angle of the Pie Plot.
      *
      * @returns {number} Returns the end angle
      */
@@ -374,10 +374,10 @@ namespace Plottable.Plots {
       let sectorValueAccessor = Plot._scaledAccessor(this.sectorValue());
       let dataset = this.datasets()[0];
       let data = this._getDataToDraw().get(dataset);
-      let pie = d3.layout.pie().sort(null).value((d, i) => sectorValueAccessor(d, i, dataset))(data);
-      let factor = (this.endAngle() - this.startAngle()) / (2 * Math.PI);
-      this._startAngles = pie.map((slice) => slice.startAngle * factor + this.startAngle());
-      this._endAngles = pie.map((slice) => slice.endAngle * factor + this.startAngle());
+      let pie = d3.layout.pie().sort(null).startAngle(this._startAngle).endAngle(this._endAngle)
+        .value((d, i) => sectorValueAccessor(d, i, dataset))(data);
+      this._startAngles = pie.map((slice) => slice.startAngle);
+      this._endAngles = pie.map((slice) => slice.endAngle);
     }
 
     protected _getDataToDraw() {
@@ -410,7 +410,7 @@ namespace Plottable.Plots {
                          .value((d: any, i: number) => {
                            let value = scaledValueAccessor(d, i, dataset);
                            return Pie._isValidData(value) ? value : 0;
-                         })(dataset.data());
+                         }).startAngle(this._startAngle).endAngle(this._endAngle)(dataset.data());
       let startAngle = pie[index].startAngle;
       let endAngle = pie[index].endAngle;
       let avgAngle = (startAngle + endAngle) / 2;
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -17,8 +17,11 @@ namespace Plottable.Plots {
      */
     constructor() {
       super();
-      this.innerRadius(0);
-      this.outerRadius(() => Math.min(this.width(), this.height()) / 2);
+      this.innerRadius(0)
+      this.outerRadius(() => {
+        let pieCenter = this._pieCenter();
+        return Math.min(Math.max(this.width() - pieCenter.x, pieCenter.x), Math.max(this.height() - pieCenter.y, pieCenter.y));
+      });
       this.addClass("pie-plot");
       this.attr("fill", (d, i) => String(i), new Scales.Color());
 
@@ -36,10 +39,7 @@ namespace Plottable.Plots {
       let pieCenter = this._pieCenter()
       this._renderArea.attr("transform", "translate(" + pieCenter.x + "," + pieCenter.y + ")");
       
-      let radiusLimit = Math.max(
-        Math.sqrt(Math.pow(pieCenter.x, 2) + Math.pow(pieCenter.y, 2)),
-        Math.sqrt(Math.pow(this.height() - pieCenter.y, 2) + Math.pow(this.width() - pieCenter.y, 2))
-      );
+      let radiusLimit = Math.min(Math.max(this.width() - pieCenter.x, pieCenter.x), Math.max(this.height() - pieCenter.y, pieCenter.y));
 
       if (this.innerRadius().scale != null) {
         this.innerRadius().scale.range([0, radiusLimit]);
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @carlosdiazh! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/plots/piePlotTests.ts
```diff
@@ -142,6 +142,140 @@ describe("Plots", () => {
...
+        assert.strictEqual(piePlot.startAngle(expectedStartAngle), piePlot, "setter returns the calling pie plot");
+        let arcPaths = piePlot.content().selectAll(".arc.fill");
+        assert.strictEqual(arcPaths.size(), 2, "has two sectors");
+
+        let pathString0 = d3.select(arcPaths[0][0]).attr("d");
+        let decomposedPath0 = TestMethods.decomposePath(TestMethods.normalizePath(pathString0));
+
+        let moveCommand0 = decomposedPath0[0];
+        assert.closeTo(moveCommand0.arguments[0], 0, 1, "starts first arc at the bottom of the circle (x)");
+        assert.operator(moveCommand0.arguments[1], ">", 0, "starts first arc at the bottom of circle (y)");
```
hellochar: nit: I find assert.isAbove and assert.isBelow slightly more readable

src/plots/piePlot.ts
```diff
@@ -198,6 +204,53 @@ namespace Plottable.Plots {
...
+      } else {
+        this._startAngle = angle;
+        this._updatePieAngles();
+        this.render();
+        return this;
+      }
+    }
+
+    /**
+     * Gets whether slice labels are enabled.
```
hellochar: update docs here

src/plots/piePlot.ts
```diff
@@ -276,8 +375,9 @@ namespace Plottable.Plots {
       let dataset = this.datasets()[0];
       let data = this._getDataToDraw().get(dataset);
       let pie = d3.layout.pie().sort(null).value((d, i) => sectorValueAccessor(d, i, dataset))(data);
-      this._startAngles = pie.map((slice) => slice.startAngle);
-      this._endAngles = pie.map((slice) => slice.endAngle);
+      let factor = (this.endAngle() - this.startAngle()) / (2 * Math.PI);
```
hellochar: Looks like d3.layout.pie has something similar already: see https://github.com/d3/d3-3.x-api-reference/blob/master/Pie-Layout.md#startAngle . Can we take advantage of that?

hellochar (changes requested):
Thanks for the PR @carlosdiazh ! 

Code looks pretty straightforward, but I'm not sure I understand the purpose of hAlign/vAlign. Pie plots don't have vertical or horizontal axes. The code looks like it just shifts the entire pie within the rendering area?

carlodx:
@hellochar The purpose of `hAlign` and `vAlign` is to be able to place the center of the pie in one of the sides of the render area, so that when you are drawing a half pie, half of the render area isn't left out blank. This is the only thing I could think of, but if there is something else I could use I'd be happy to change it.

Also, I've addressed the rest of the change requests

hellochar:
Ah I see - so you mean when setting startAngle/endAngle by itself, you'll get pies that only cover half the area: 
image

In order to fix that, users can do `vAlign("top")` to push it to the top: 
image

But won't that have the same issue where the second half of the chart is completely empty? Furthermore, you never want to use `hAlign/vAlign` by itself (otherwise the pie would escape its drawing boundary); you only want to set them to certain values in combination with `startAngle`/`endAngle`, is that right? I'm worried that exposing this API surface area will confuse users since it's not obvious what it does/what it's used for. 

How would you feel about auto-scaling the pie chart to fill the plot width/height depending on the start/endAngle?

@CalvinFernandez @themadcreator thoughts?

src/plots/piePlot.ts
```diff
@@ -198,6 +204,53 @@ namespace Plottable.Plots {
...
       return this;
     }
+    /**
+     * Gets the start angle of the Pie Plot
+     *
+     * @returns {number} Returns the start angle
+     */
+    public startAngle(): number;
+    /**
+     * Sets the end angle of the Pie Plot.
```
CalvinFernandez: Sets the ~end~ start angle of the Pie Plot

src/plots/piePlot.ts
```diff
@@ -198,6 +204,53 @@ namespace Plottable.Plots {
...
+     *
+     * @param {number} endAngle
```
CalvinFernandez: ~endAngle~ start angle

src/plots/piePlot.ts
```diff
@@ -222,6 +275,52 @@ namespace Plottable.Plots {
       }
     }
 
+   /**
+     * Get whether slice labels are enabled.
+     *
+     * @returns {boolean} Whether slices should display labels or not.
```
CalvinFernandez: string

CalvinFernandez (commented):
This is great! Thanks for the PR @carlosdiazh! re: @hellochar's comment, I think one solution could be to adjust the size of the component during the space negotiation phase to ensure that the chart always correctly fills the plot area. Also not a huge fan of `hAlign/vAlign` .  I'm also fine with just leaving the second half of the chart completely empty and revisiting this as a follow up commit.

src/plots/piePlot.ts
```diff
@@ -30,8 +32,15 @@ namespace Plottable.Plots {
 
     public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
       super.computeLayout(origin, availableWidth, availableHeight);
-      this._renderArea.attr("transform", "translate(" + this.width() / 2 + "," + this.height() / 2 + ")");
-      let radiusLimit = Math.min(this.width(), this.height()) / 2;
+      
+      let pieCenter = this._pieCenter()
+      this._renderArea.attr("transform", "translate(" + pieCenter.x + "," + pieCenter.y + ")");
```
hellochar, line 40: We should update the transform in the minAngle/maxAngle setter, not just in `computeLayout`

src/plots/piePlot.ts
```diff
@@ -275,11 +331,54 @@ namespace Plottable.Plots {
...
       let data = this._getDataToDraw().get(dataset);
-      let pie = d3.layout.pie().sort(null).value((d, i) => sectorValueAccessor(d, i, dataset))(data);
+      let pie = d3.layout.pie().sort(null).startAngle(this._startAngle).endAngle(this._endAngle)
+        .value((d, i) => sectorValueAccessor(d, i, dataset))(data);
       this._startAngles = pie.map((slice) => slice.startAngle);
       this._endAngles = pie.map((slice) => slice.endAngle);
     }
 
+    private _pieCenter(): Point {
+      let a = this._startAngle < this._endAngle ? this._startAngle: this._endAngle;
```
hellochar, line 341: Holy cow this is impressive. Can we add comments to describe what's going on? Is there a way to reduce the number of if/elses?

hellochar (changes requested):
Thanks for removing hAlign/vAlign. Feels close now; just some polish on the new pie centering logic

src/plots/piePlot.ts
carlodx: Sure, I'll add some comments. 
I've been trying to come up with a more generic way to compute the center but I haven't been able to find anything that doesn't leave out some possibilities potentially cutting out part of the pie.

src/plots/piePlot.ts
carlodx: I tried doing this, but `this._renderArea` is still undefined

src/plots/piePlot.ts
carlodx: I've added a general comment, above the if/elses, it's a bit hard to explain exactly whats going on on each condition without drawing. Hopefully this'll do. Let me know though. 

We could also get rid of the last if on each block but I feel like stating the condition makes it easier to understand so one can picture in their head the image of the area of the whole circle the pie should take.

hellochar (approved):

hellochar:
Looks great, thanks @carlosdiazh !

hellochar:
Oops, I let https://github.com/palantir/plottable/issues/3168 through... @carlosdiazh if you're hitting that feel free to submit another PR to fix. If not, us Plottable folks will triage and deal with it eventually

PR #3159, opened 2017-01-12
Interaction: LightweightEntity Caching

CalvinFernandez:
Implements LightweightEntity caching in base plot in
order to speed up various chart interactions. The entity
cache is invalidated whenever underlying data changes. The
cache is constructed lazily on the first new call to `entities()`
from the time of invalidate (or initialization).

Entity positioning is stored as Points in the Data space, rather
than the screen space in order to avoid having to rebuild the
cache on each pan/zoom interaction. Comparison is done on chart
and cursor positions which are all converted to data space first.

-- notable changes

* Base plot now implements entity caching.
* `LightweightPlotEntity` is now exported publicly.
* Adds `TransformableAccessorScaleBinding`.
* Moves `TransformableScale` from PanZoomInteraction
  to `Plottable.TransformableScale` namespace.
* Quantitative Scales accessors x(), y(), x1(), y1() now
  return an instance of `TransformableScale` instead of
  simply `Scale`
* Adds `utils/entityIndex.ts`, which defines
  and implements a simple entity storage API
* New public method `bounds` on component returns
  the {Bounds} of the component
* Adds `invertTransformation` method to `TransformableScale`
  which maps pixel coordinates to Transformation Space

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -527,6 +527,23 @@ export class Component {
   }
 
   /**
+   * @returns {Bounds} for the component in pixel space, where the topLeft
+   * represents the component's minimum x and y values and the bottomRight represents
+   * the component's maximum x and y values.
+   */
+  public bounds(): Bounds {
+    const topLeft = this.origin();
+
+    return {
+      topLeft,
+      bottomRight: {
+        x: topLeft.x + this.width(),
+        y: topLeft.y + this.height()
+      },
+    }
+  }
+
+  /**
    * Removes a Component from the DOM and disconnects all listeners.
    */
   public destroy() {
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -11,16 +11,14 @@ namespace Plottable.Interactions {
     return center - (center - value) * zoom;
   }
 
-  export type TransformableScale = Plottable.Scale<any, number> & Plottable.Scales.TransformableScale;
-
   export class PanZoom extends Interaction {
     /**
      * The number of pixels occupied in a line.
      */
     private static _PIXELS_PER_LINE = 120;
 
-    private _xScales: Utils.Set<TransformableScale>;
-    private _yScales: Utils.Set<TransformableScale>;
+    private _xScales: Utils.Set<TransformableScale<any, number>>;
+    private _yScales: Utils.Set<TransformableScale<any, number>>;
     private _dragInteraction: Interactions.Drag;
     private _mouseDispatcher: Dispatchers.Mouse;
     private _touchDispatcher: Dispatchers.Touch;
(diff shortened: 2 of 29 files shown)


src/components/component.ts
```diff
@@ -527,6 +527,23 @@ export class Component {
   }
 
   /**
+   * @returns {Bounds} for the component in pixel space, where the topLeft
+   * represents the component's minimum x and y values and the bottomRight represents
+   * the component's maximum x and y values.
+   */
+  public bounds(): Bounds {
```
hellochar, line 534: should specify this is relative to its parent (since it's based off this.origin())

src/interactions/panZoomInteraction.ts
```diff
@@ -11,16 +11,14 @@ namespace Plottable.Interactions {
...
-
   export class PanZoom extends Interaction {
     /**
      * The number of pixels occupied in a line.
      */
     private static _PIXELS_PER_LINE = 120;
 
-    private _xScales: Utils.Set<TransformableScale>;
-    private _yScales: Utils.Set<TransformableScale>;
+    private _xScales: Utils.Set<TransformableScale<any, number>>;
```
hellochar, line 20: nit: is `any` the right type here? Would it be valuable to give PanZoom a type parameter for this instead?

src/plots/plot.ts
```diff
@@ -11,32 +28,29 @@ namespace Plottable.Plots {
     scale?: Scale<D, R>;
   }
 
+  export interface TransformableAccessorScaleBinding<D, R> {
```
hellochar, line 36: Can you add some docs for this interface? It feels very important but I don't know what it's used for

src/utils/entityIndex.ts
```diff
@@ -0,0 +1,80 @@
+namespace Plottable.Utils {
+
+  /**
+   * EntityIndex stores entities and makes them searchable.
+   * Valid entities must be positioned in Cartesian space.
+   */
+  export interface EntityIndex<T extends PositionedEntity> {
```
hellochar: nit: "Index" in the name reminds me of "index in an array"; can we use a different name? Maybe EntityStore, EntityCache?

src/plots/plot.ts
```diff
@@ -594,6 +634,42 @@ export class Plot extends Component {
              (d: any, i: number, ds: Dataset) => binding.scale.scale(binding.accessor(d, i, ds));
   }
 
+  /**
+   * _invertPixelPoint converts a point in pixel coordinates to a point in data coordinates
+   * @param {Point} point Representation of the point in pixel coordinates
+   * @return {Point} Returns the point represented in data coordinates
+   */
+  protected _invertPixelPoint(point: Point): Point {
```
hellochar: Is this implemented for Pie and Wheel?

src/scales/scale.ts
```diff
@@ -61,6 +61,11 @@ namespace Plottable.Scales {
      * `scaleTransformation`.
      */
     getTransformationDomain(): [number, number];
+
+    /**
+     * Returns value in *Transformation Space* for the provided *screen space*.
+     */
+    invertTransformation(value: number): number;
```
hellochar: nit: call this `inverted` to signify this applies the already-inverted transformation (rather than saying this method inverts the transformation itself)

src/utils/entityIndex.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  export interface PositionedEntity {
+    position: Point;
+  }
+
+  /**
+   * Array-backed implementation of {EntityIndex}
+   */
+  export class EntityArray<T extends PositionedEntity> implements EntityIndex<T> {
```
hellochar: I smell a quadtree in the future!!!!

hellochar (approved):
Looks pretty straightforward. I'm a little worried that we keep increasing the API surface area and adding more methods; feel like there's some overlapping responsibilities/a better way to structure all this stuff

src/plots/plot.ts
CalvinFernandez: nope :p

src/interactions/panZoomInteraction.ts
CalvinFernandez: just copied the typings, @themadcreator would be a better person to ask

CalvinFernandez:
@hellochar Yea, thanks for being vigilant about keeping the API minimal. In this PR the only API expansion that leaves a bad taste in my mouth is the addition of `LightweightPlotEntity` . Can you think of anything else ?

hellochar:
Nothing concrete, but I get the feeling that invertedPixelSize and invertedTransform overlap with other TransformableScale methods

hellochar:
@CalvinFernandez Want to merge?

CalvinFernandez:
Need to write tests -- then we can merge

hellochar:
LGTM

PR #3164, opened 2017-01-16
Update svg-typewriter

themadcreator:
* Switch to npm package
* Update build scripts
* Use CacheMeasurer everywhere for PERFORMANCE!


themadcreator:
Check out this buttery smooth pan/zoom

smooth-zoom

CalvinFernandez (dismissed):
Good idea! Unfortunately we need to keep svg-typewriter bundled for now as it will break all standalone builds. We tried this a month or so ago and reverted

adidahiya:
this is the relevant discussion btw https://github.com/palantir/plottable/pull/3119

Gruntfile.js
```diff
@@ -64,15 +64,11 @@ module.exports = function(grunt) {
       src: ["license_header.txt", "plottable.js"],
       dest: "plottable.js"
     },
-    svgtypewriter: {
```
hellochar, line 67: We need to keep concatting this. Imo until we Plottable 3.0.0 and convert everything to modules, we need to build an svgtypewriter bundle file and concat it (or use webpack and create a plottable.js bundle, which is equivalent)

src/axes/categoryAxis.ts
```diff
@@ -50,7 +50,7 @@ namespace Plottable.Axes {
 
     protected _setup() {
       super._setup();
-      this._measurer = new SVGTypewriter.Measurers.CacheCharacterMeasurer(this._tickLabelContainer);
+      this._measurer = new SVGTypewriter.CacheMeasurer(this._tickLabelContainer);
```
hellochar, line 53: Later on in this file we call _measurer.reset() whenever the chart is redrawn:

https://github.com/palantir/plottable/blob/bd/update-svg-typewriter/src/axes/categoryAxis.ts#L345

Will this blow away our caching improvements?

hellochar (changes requested):
I tried running tests but it errored from not being able to find svg-typewriter :(

src/axes/categoryAxis.ts
themadcreator: Yes, it clears the cache

themadcreator: But the cache is still useful because it can still be reused for common substrings while measuring

hellochar (changes requested):
Just devtested this and had to add `@types/d3` as a dependency before it could get picked up by a client. Additionally the client had to remove `@types/d3` from its dependencies (or else typescript would see two typings). I'm okay with just pulling in `@types/d3` as a dependency, although there's some discussion over where it belongs (cc @adidahiya )

Other than that it looked good and actually fixed a long standing (been there since v1.0.0) bug where labels sometimes disappeared.

adidahiya:
`@types/d3` should be a dependency in package.json with a floating version range. I think an ideal implementation of Plottable would keep the API surface small and not leak that type to consumers in .d.ts files, but it doesn't seem like we're at a point where we can do that.

themadcreator:
`@types/d3` is already a dependency of svg-typewriter, so it should have been brought in.

CalvinFernandez (approved):

hellochar (approved):

PR #3165, opened 2017-01-17
Add downsampling for number of category axis tick labels shown

### LINKED ISSUE #3160: when there are too many Category labels, all of them get hidden
image

Related to #2420 

Explore downsampling the # of shown labels so we at least show something

hellochar:
downsampling

Category axes with too many labels compared to the space its given would
previously degenerate into rendering all "..."s or no text at all.

We now drop ticks at regular intervals when rendering the ticks would
otherwise be too tight (set at 15px per tick). This allows at least
some labels to be drawn - a strict improvement over the previous
behavior.

Fixes #3160

### DIFF
--- a/src/axes/categoryAxis.ts
+++ b/src/axes/categoryAxis.ts
@@ -1,9 +1,14 @@
 namespace Plottable.Axes {
+  export interface DownsampleInfo {
+    domain: string[];
+    stepWidth: number;
+  }
+
   export class Category extends Axis<string> {
     /**
      * How many pixels to give labels at minimum before downsampling takes effect.
      */
-    public static MINIMUM_WIDTH_PER_LABEL = 15;
+    private static _MINIMUM_WIDTH_PER_LABEL_PX = 15;
 
     private _tickLabelAngle = 0;
 
@@ -112,18 +117,22 @@ namespace Plottable.Axes {
     }
 
     protected _getTickValues() {
-      return this.getDownsampleInfo()[0];
+      return this.getDownsampleInfo().domain;
     }
 
     /**
      * Take the scale and drop ticks at regular intervals such that the resultant ticks are all a reasonable minimum
      * distance apart. Return the resultant ticks to render, as well as the new stepWidth between them.
      *
-     * @return [downsampled domain, new stepWidth]
+     * @param scale the scale being downsampled. Defaults to this Axis' scale.
+     * @return {DownsampleInfo} an object holding the resultant domain and new stepWidth.
      */
-    public getDownsampleInfo(scale = <Scales.Category> this._scale): [string[], number] {
-      const downsampleRatio = Math.ceil(Category.MINIMUM_WIDTH_PER_LABEL / scale.stepWidth());
-      return [scale.domain().filter((d, i) => i % downsampleRatio === 0), downsampleRatio * scale.stepWidth()];
+    public getDownsampleInfo(scale = <Scales.Category> this._scale): DownsampleInfo {
+      const downsampleRatio = Math.ceil(Category._MINIMUM_WIDTH_PER_LABEL_PX / scale.stepWidth());
+      return {
+        domain: scale.domain().filter((d, i) => i % downsampleRatio === 0),
+        stepWidth: downsampleRatio * scale.stepWidth(),
+      };
     }
 
     /**
@@ -266,7 +275,7 @@ namespace Plottable.Axes {
         .outerPadding(thisScale.outerPadding())
         .range([0, this._isHorizontal() ? axisWidth : axisHeight]);
 
-      const [ticks, stepWidth] = this.getDownsampleInfo(scale);
+      const { domain, stepWidth } = this.getDownsampleInfo(scale);
 
       // HACKHACK: https://github.com/palantir/svg-typewriter/issues/25
       // the width (x-axis specific) available to a single tick label.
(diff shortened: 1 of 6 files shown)


src/axes/categoryAxis.ts
```diff
@@ -1,5 +1,10 @@
 namespace Plottable.Axes {
   export class Category extends Axis<string> {
+    /**
+     * How many pixels to give labels at minimum before downsampling takes effect.
+     */
+    public static MINIMUM_WIDTH_PER_LABEL = 15;
```
CalvinFernandez: nit:` MINIMUM_WIDTH_PER_LABEL_PX`

src/axes/categoryAxis.ts
```diff
@@ -108,7 +112,18 @@ namespace Plottable.Axes {
...
 
     protected _getTickValues() {
-      return this._scale.domain();
+      return this.getDownsampleInfo()[0];
+    }
+
+    /**
+     * Take the scale and drop ticks at regular intervals such that the resultant ticks are all a reasonable minimum
+     * distance apart. Return the resultant ticks to render, as well as the new stepWidth between them.
+     *
```
CalvinFernandez, line 126: Include `@param` info as well

src/axes/categoryAxis.ts
```diff
@@ -108,7 +112,18 @@ namespace Plottable.Axes {
...
+     * @return [downsampled domain, new stepWidth]
```
CalvinFernandez: ooc why did you choose the array over something slightly more explicit like `{ tickDomain: string[], stepWidth: number }` ?

src/axes/categoryAxis.ts
CalvinFernandez: also why is this property public?

src/axes/categoryAxis.ts
```diff
@@ -303,28 +324,28 @@ namespace Plottable.Axes {
     public renderImmediately() {
       super.renderImmediately();
       let catScale = <Scales.Category> this._scale;
-      let tickLabels = this._tickLabelContainer.selectAll("." + Axis.TICK_LABEL_CLASS).data(this._scale.domain(), (d) => d);
+      const [downsampledDomain, stepWidth] = this.getDownsampleInfo();
+      let tickLabels = this._tickLabelContainer.selectAll("." + Axis.TICK_LABEL_CLASS).data(downsampledDomain, (d) => d);
+      // Give each tick a stepWidth of space which will partition the entire axis evenly
+      let availableTextWidth = stepWidth;
+      if (this._isHorizontal() && this._tickLabelMaxWidth != null) {
```
CalvinFernandez, line 340: do we not care about vertical?

src/axes/categoryAxis.ts
```diff
@@ -242,40 +256,47 @@ namespace Plottable.Axes {
...
-            height = axisWidth - this._tickSpaceRequired();
-          }
-          // HACKHACK: Wrapper fails under negative circumstances
-          height = Math.max(height, 0);
+      // HACKHACK: https://github.com/palantir/svg-typewriter/issues/25
+      // the width (x-axis specific) available to a single tick label.
+      let width = axisWidth - this._tickSpaceRequired(); // default for left/right
+      if (this._isHorizontal()) { // case for top/bottom
+        width = stepWidth; // defaults to the band width
+        if (this._tickLabelAngle !== 0) { // rotated label
```
CalvinFernandez, line 285: can we make this explicitly 90 or -90 ? those are the two cases that we care about right?

CalvinFernandez (changes requested):
a few small nits -- also looks like the tests are failing

src/axes/categoryAxis.ts
hellochar: out of laziness ;) yeah an object with names would be better; I'll change it to that

src/axes/categoryAxis.ts
hellochar: Imo that's just more verbose but with no gain. Regardless I want to rework large parts of this since it's so hacky so might make more sense to tackle things then

src/axes/categoryAxis.ts
hellochar: This variable is named incorrectly; for vertical axes `availableTextWidth` actually represents the vertical height between successive ticks (aka the step width). tickLabelMaxWidth should only affect the width available so when it's vertical, nothing should happen. I've gone and renamed with `availableTextSpace`. I actually also found a subtle bug here, so fixed that too

src/axes/categoryAxis.ts
```diff
@@ -112,18 +117,22 @@ namespace Plottable.Axes {
...
-      return this.getDownsampleInfo()[0];
+      return this.getDownsampleInfo().domain;
     }
 
     /**
      * Take the scale and drop ticks at regular intervals such that the resultant ticks are all a reasonable minimum
      * distance apart. Return the resultant ticks to render, as well as the new stepWidth between them.
      *
-     * @return [downsampled domain, new stepWidth]
+     * @param scale the scale being downsampled. Defaults to this Axis' scale.
```
CalvinFernandez: Sorry to nitpick on this, but remember this is a public method so let's keep comments more formal
`@param {Scales.Category} scale - The scale being downsampled. Defaults to this Axis' scale.`

CalvinFernandez (changes requested):
Can you change the comment format! Only thing left, sorry  :p

CalvinFernandez (approved):

PR #3167, opened 2017-01-17
Stacked Bar: Summed Labels

CalvinFernandez:
image
image
image

### DIFF
--- a/src/plots/stackedBarPlot.ts
+++ b/src/plots/stackedBarPlot.ts
@@ -1,5 +1,11 @@
 namespace Plottable.Plots {
   export class StackedBar<X, Y> extends Bar<X, Y> {
+    private static _STACKED_BAR_LABEL_AREA_CLASS = "stacked-bar-label-text-area"
+    protected static _STACKED_BAR_LABEL_PADDING = 5;
+
+    private _labelArea: d3.Selection<void>;
+    private _measurer: SVGTypewriter.Measurers.CacheCharacterMeasurer;
+    private _writer: SVGTypewriter.Writers.Writer;
     private _stackingResult: Utils.Stacking.StackingResult;
     private _stackedExtent: number[];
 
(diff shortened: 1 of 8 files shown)


src/utils/stackingUtils.ts
```diff
@@ -46,6 +46,33 @@ namespace Plottable.Utils.Stacking {
...
   }
 
+  export function stackedExtents(stackingResult: StackingResult) {
+    const maximumExtents: Utils.Map<string, number> = new Utils.Map<string, number>();
+    const minimumExtents: Utils.Map<string, number> = new Utils.Map<string, number>();
+
+    stackingResult.forEach((stack) => {
+      stack.forEach((datum: StackedDatum, key: string) => {
+        // correctly handle negative bar stacks
+        const maximalValue = Math.max([datum.offset + datum.value, datum.offset], datum.offset);
```
hellochar, line 71: How does this work? It looks like we're passing an array argument to Math.max, which shouldn't work

src/utils/stackingUtils.ts
```diff
@@ -46,6 +46,33 @@ namespace Plottable.Utils.Stacking {
     return datasetToKeyToStackedDatum;
   }
 
+  export function stackedExtents(stackingResult: StackingResult) {
+    const maximumExtents: Utils.Map<string, number> = new Utils.Map<string, number>();
+    const minimumExtents: Utils.Map<string, number> = new Utils.Map<string, number>();
+
+    stackingResult.forEach((stack) => {
+      stack.forEach((datum: StackedDatum, key: string) => {
```
hellochar: nit: keys could be numbers or Dates, (e.g. Numeric x scale on histograms)

src/plots/stackedBarPlot.ts
```diff
@@ -54,6 +60,92 @@ namespace Plottable.Plots {
...
+      this._writer = new SVGTypewriter.Writers.Writer(this._measurer);
+    }
+
+    protected _drawLabels() {
+      super._drawLabels();
+
+      // remove all current labels before redrawing
+      this._labelArea.selectAll("g").remove();
+
+      const baselineValue = +this.baselineValue();
```
hellochar, line 75: Why do we convert to number here

hellochar (approved):
* can/should we DRY this with the barPlot label drawing code at all?

* would it be difficult to draw labels (or truncate) if the area is too small? I've seen some other charts where the bar stack itself is small (~10px) but the label just bleeds a bit into the other stacks and it looks fine

src/utils/stackingUtils.ts
CalvinFernandez: oops good point

src/utils/stackingUtils.ts
CalvinFernandez: This is the `Math.max` from Utils, not the window's Math.max

src/plots/stackedBarPlot.ts
CalvinFernandez: the baseline value isn't _technically_ numeric, but I'm operating under the assumption that you need a numeric secondary scale in order to have a stacked bar that works correctly

pkwi:
With font size 12pt - it might fit for every individual bar that is minimum 10px high (See Sprint bar). Otherwise, text might not be readable (See Winter bar).

screen shot 2017-01-19 at 2 33 18 pm

CalvinFernandez:
> Should we DRY? 

Yea, but I'm too scared to touch that code here. 

> would it be difficult to draw labels (or truncate) if the area is too small?

Probably wouldn't be too hard but truncated numbers lose more meaning than truncated strings so its really a question of whether it makes sense or not @pkwi thoughts? 

> I've seen some other charts where the bar stack itself is small (~10px) but the label just bleeds a bit into the other stacks and it looks fine

Yea we can probably finagle the width to consume 100% of the space. @pkwi thoughts?  Down to do this in a follow-up commit.

CalvinFernandez:
image

pkwi:
I would avoid allowing labels to stick outside the stack. Otherwise, we might have a situation like this:

screen shot 2017-01-19 at 6 04 28 pm

Where labels are overlapping each other.

hellochar:
Are overlapping labels that bad? In the example you gave I can at least still see all the numbers and the bars they represent, which does convey more information

hellochar (approved):

ccolgrove:
@hellochar Re: overlapping labels. This is bad because it _looks_ broken, regardless of whether or not it actually is broken

Issue #3117, opened 2016-11-30
pan/zoom performance on bar charts is poor


hellochar:
pan/zoom on a bar chart forces the chart to redraw; a trivial stacked bar of 4 elements w/ 5 stacks each took 250ms to flush:

![image](https://cloud.githubusercontent.com/assets/1079495/20736718/4d134f90-b676-11e6-8ba7-8123e418c58e.png)

Few things to fix:

- [ ] Bar.render() is called twice, once from plottable.js line 1093:
    ```js
    // Top level render; Containers will put their children in the toRender queue
    _componentsNeedingRender.forEach(function (component) { return component.render(); });
    ```
    and again in plottable.js line 3371, in ComponentContainer.render (probably the grid + plots group)

- [x] Bar.render() itself is dominated by the scales changing -> axes must redraw -> text gets remeasured -> svgtypewriter calls getBBox() (which forces browser to reflow layout), as well as many calls to `appendChild()` and `removeChild()`. I suspect this is causing thrashing and hurting performance significantly. Another part of this is unnecessary calls to Plot._computeExtent (#2495)

- [x] Axes and Legends are also causing reflow. The following methods likely also cause thrashing: Axis.requestedSpace, Legend.requestedSpace, Axis.renderImmediately, Legend.renderImmediately

hellochar:
The second two points will hopefully be addressed by #3164

Issue #2495, opened 2015-07-22
Plot._computeExtent is called too many times. 


aicioara:
![screen shot 2015-07-22 at 2 57 02 pm](https://cloud.githubusercontent.com/assets/3248682/8838535/07a3b24a-3082-11e5-9774-acccda61c8e7.png)

It seems like there are a lot of redundant calls to `Plot._computeExtent`, and possibly other similar methods. 

Not flagged as a performance issue yet, but assuming we have 100k data points, this method iterates over them all and applies a filter, which results in a lot of garbage. Maybe we should operate into a more factory pattern way, gather the requirements in a lightweight mode and calculate the necessary pieces only on `.render()`. This will change a good amounts of tests though (in the sense that we have to start rendering components to svg to test them, or hook into appropriate methods)

As an aside, during my perf testing, I found out that a good amount of time was spent on garbage collection (can't remember exact numbers, but it was probably 10% of the time)

PR #3171, opened 2017-01-18
Reduce unnecessary call to computeLayout in renderTo()

hellochar:
register and then flush the component to renderTo it. Previously,
computeLayout and render were both invoked two times per renderTo().
This should have no regression in behavior but charts will render faster.

Addresses first point of #3117


CalvinFernandez (changes requested):
Looks good! Agree with these changes for the most part.

1) This feels like render whack-a-mole. Did you investigate queueing renders and canceling multiple requested renders within the same render loop? Not necessarily asking for change here, just wondering what your thought process was.

2)  We can't remove exported functions without a major rev. Can you deprecate `registerToComputeLayout` and file an issue to remove in 3.x instead of totally removing it? 

3) Can we get at least 1 test proving this reduces the traffic?

hellochar:
1. Requesting a render puts it in a Set of components-to-render-on-the-next-flush (which is usually in a requestAnimationFrame), so multiple renders are already a no-op since it's just adding to a queue. The whack-a-mole is in "requesting a render (this is calling `.render()`)" in the first place which could happen many times. This shouldn't matter because requesting a render should be a trivial operation, but Bar plots do computation as part of requesting a render in the first place. This is tracked in #2495 and I tried fixing this for a few hours but couldn't understand exactly what was happening so gave up on it for now.

2. Will do (it's just a rename anyways)

3. Yeah I'll write a test

CalvinFernandez:
👍 cool just the rename and we're good

test/components/componentTests.ts
```diff
@@ -735,6 +739,34 @@ describe("Component", () => {
...
+      const legend = new Plottable.Components.Legend(colorScale);
+      const table = new Plottable.Components.Table([[barPlot, legend]]);
+      const computeLayoutSpy = sinon.spy(table, "computeLayout");
+
+      table.renderTo(svg);
+
+      assert.strictEqual(computeLayoutSpy.callCount, 1, "component only computes layout once");
+      table.destroy();
+      svg.remove();
+      done();
```
CalvinFernandez, line 767: you only need `done` for async things

CalvinFernandez (approved):

Issue #3161, opened 2017-01-16
maxWidth for Series Legend


hellochar:
This would prevent legends from growing overly large and dominating the table layout's space request algorithm.

CalvinFernandez:
closed by #3172

Issue #3162, opened 2017-01-16
configurable maxLines for series legend labels


hellochar:
Series legends may get really long e.g. "count_of_sum_of_xs_compared_to_intrinsic_y_value". Expose a configuration option on legends to allow line wrapping on each series legend.

helenkg:
@hellochar let's do that this week, worst case next? as well as #3161?

CalvinFernandez:
closed by #3172

PR #3172, opened 2017-01-19
Legend: maxWidth and maxLinesPerEntry

CalvinFernandez:
Adds `maxWidth` property to legend
 - When defined, maxWidth restricts legend growth to at most `maxWidth` pixels.

Adds `maxLinesPerEntry` property to legend
- When defined, `maxLinesPerEntry` specifies number of times a single entry may wrap before being truncated

Refactors legend to include `LegendTable` to aid with layout calculation.

### DIFF
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -290,9 +290,9 @@ namespace Plottable.Components {
       super._setup();
       let fakeLegendRow = this.content().append("g").classed(Legend.LEGEND_ROW_CLASS, true);
       let fakeLegendEntry = fakeLegendRow.append("g").classed(Legend.LEGEND_ENTRY_CLASS, true);
-      fakeLegendEntry.append("text");
+      fakeLegendEntry.append("text"); 
       this._measurer = new SVGTypewriter.Measurer(fakeLegendRow);
-      this._wrapper = new SVGTypewriter.Wrappers.Wrapper().maxLines(this.maxLinesPerEntry());
+      this._wrapper = new SVGTypewriter.Wrapper().maxLines(this.maxLinesPerEntry());
       this._writer = new SVGTypewriter.Writer(this._measurer, this._wrapper).addTitleElement(Configs.ADD_TITLE_ELEMENTS);
     }
 
@@ -494,7 +494,10 @@ namespace Plottable.Components {
     public requestedSpace(offeredWidth: number, offeredHeight: number): SpaceRequest {
       // if max width is set, the table is guaranteed to be at most maxWidth wide.
       // if max width is not set, the table will be as wide as the longest untruncated row
-      const table = this._buildLegendTable(Utils.Math.min([this.maxWidth(), Infinity], Infinity), offeredHeight);
+      const table = this._buildLegendTable(
+        Utils.Math.min([this.maxWidth(), offeredWidth], offeredWidth),
+        offeredHeight
+      );
 
       return {
         minHeight: table.getHeight(),
--- a/src/components/legend.ts
+++ b/src/components/legend.ts
@@ -1,4 +1,233 @@
 namespace Plottable.Components {
+  /**
+   * The Legend's column representation. Stores position information
+   * as well as data
+   */
+  class LegendColumn<T> {
+    constructor(private _width = 0, private _height = 0, private _data: T) {}
+
+    /**
+     * @return {T} data Returns any data stored in the column
+     */
+    public getData() {
+      return this._data;
+    }
+
+    /**
+     * @return {number} height The height of the column.
+     */
+    public getHeight() {
+      return this._height;
+    }
+
+    /**
+     * @return {number} width The width of the column.
+     */
+    public getWidth() {
+      return this._width;
+    }
+
+    /**
+     * @param {number} width Sets the width of the column
+     */
+    public setWidth(width: number) {
+      return this._width = width;
+    }
+  }
+
+  /**
+   * The Legend's row representations. Stores positioning information
+   * and column data.
+   */
+  class LegendRow {
+    constructor(
+        private _columns: LegendColumn<any>[] = [],
+        private _bottomPadding = 0,
+        private _maxWidth = Infinity) {}
+
+    public addColumn(column: LegendColumn<any>) {
+      const desiredColumnWidth = column.getWidth();
+      // choose the smaller of 1) remaining space, 2) desired width
+      const widthRemaining = this.getWidthAvailable();
+      column.setWidth(Math.min(widthRemaining, desiredColumnWidth));
+      this._columns.push(column);
+    }
+
+    public forEach(callback: (column: LegendColumn<any>, index?: number) => void) {
+      this._columns.forEach(callback);
+    }
+
+    public reduce<U>(callbackfn: (
+      previousValue: U,
+      currentValue: LegendColumn<any>,
+      currentIndex: number,
+      array: LegendColumn<any>[]) => U, initialValue: U): U {
+      return this._columns.reduce(callbackfn, initialValue);
+    }
+
+    /**
+     * Returns the bounds the column, relative to the row.
+     * @param {LegendColumn<any>} column The column in question
+     * @returns {Bounds} bounds
+     */
+    public getBounds(column: LegendColumn<any>) {
+      const indexOfColumn = this._columns.indexOf(column);
+      const columnXOffset = this._columns.slice(0, indexOfColumn)
+        .reduce((sum, column) => sum + column.getWidth(), 0);
+
+      return {
+        topLeft: { x: columnXOffset, y: 0 },
+        bottomRight: {
+          x: columnXOffset + column.getWidth(),
+          y: column.getHeight(),
+        }
+      };
+    }
+
+    public getColumns() {
+      return this._columns;
+    }
+
+    /**
+     * Returns the height of the row, including the bottomPadding.
+     * @return {number} height
+     */
+    public getHeight() {
+      return Utils.Math.max(this._columns.map((column) => column.getHeight()), 0)
+        + this._bottomPadding;
+    }
+
+    /**
+     * Padding applied below the row. Affects the spacing between rows. Defaults to 0.
+     * @param {bottomPadding} number
+     */
+    public setBottomPadding(bottomPadding: number) {
+      this._bottomPadding = bottomPadding;
+    }
+
+    /**
+     * Sets the maximum allowable width of this column.
+     * @param {number} maxWidth
+     */
+    public setMaxWidth(maxWidth: number) {
+      this._maxWidth = maxWidth;
+    }
+
+    /**
+     * Returns the current width of the row constrained by maxWidth, if set.
+     * @returns {number} width
+     */
+    public getWidth() {
+      return Math.min(
+        this._columns.reduce((sum, column) => sum + column.getWidth(), 0),
+        this._maxWidth
+      );
+    }
+
+    /**
+     * Returns the remaining width available in the row based on the maximum
+     * width of this row.
+     * @returns {number} widthRemaining
+     */
+    public getWidthAvailable() {
+      const widthConsumed = this.getWidth();
+      return Math.max(this._maxWidth - widthConsumed, 0);
+    }
+  }
+
+  /**
+   * Stores LegendRows. Useful for calculating and maintaining
+   * positioning information about the Legend.
+   */
+  class LegendTable {
+    constructor(
+        private _maxWidth = Infinity,
+        private _maxHeight = Infinity,
+        private _padding = 0,
+        private _rows: LegendRow[] = []) {}
+
+    public addRow(row: LegendRow) {
+      row.setMaxWidth(this._maxWidth - this._padding * 2);
+      this._rows.push(row);
+    }
+
+    public forEach(callback: (row: LegendRow, index?: number) => void) {
+      this._rows.forEach(callback);
+    }
+
+    /**
+     * Returns the bounds relative to the parent and siblings of the row or
+     * column. If column is not specified, bounds will be calculate for the row only.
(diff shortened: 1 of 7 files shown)


CalvinFernandez:
closes #3161 #3162

jchomicki:
@CalvinFernandez What is the status on this? A downstream ticket is blocking on this.

CalvinFernandez:
@jchomicki waiting on reviewers. ping @hellochar @themadcreator

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
-        let entriesInRow = d3.select(this).selectAll("g." + Legend.LEGEND_ENTRY_CLASS);
-        entriesInRow.attr("transform", (value: string, i: number) => {
-          let translateString = "translate(" + xShift + ", 0)";
-          xShift += layout.entryLengths.get(value);
-          return translateString;
+      const table = this._buildLegendTable(this.width(), this.height());
+      const entryNames = this._colorScale.domain().slice().sort((a, b) => this._comparator(this._formatter(a), this._formatter(b)));
+
+      // clear content from previous renders
+      this.content().selectAll("*").remove();
```
themadcreator, line 530: This isn't really the idiomatic way to do d3. Removing and re-adding everything on every render causes lots of DOM thrashing. You can see from the code that this replaces that it applies the rows to the DOM using the `.data` on the selection then using `.enter`, `.exit` and the selection itself to update the stuff that's there.

src/components/legend.ts
```diff
@@ -159,70 +438,68 @@ namespace Plottable.Components {
...
-          currentRow = [];
-          spaceLeft = availableWidth;
-        }
-        currentRow.push(e);
-        spaceLeft -= entryLength;
-      });
+    public requestedSpace(offeredWidth: number, offeredHeight: number): SpaceRequest {
+      // if max width is set, the table is guaranteed to be at most maxWidth wide.
+      // if max width is not set, the table will be as wide as the longest untruncated row
+      const table = this._buildLegendTable(Utils.Math.min([this.maxWidth(), Infinity], Infinity), offeredHeight);
```
themadcreator: Why the extra `Infinity`?

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
+      table.forEach((row: LegendRow, rowIndex: number) => {
+        const rowBounds = table.getBounds(row);
+        const rowRoot = tableRoot.append("g").classed(Legend.LEGEND_ROW_CLASS, true)
+          .attr("transform", `translate(${rowBounds.topLeft.x}, ${rowBounds.topLeft.y})`);
+
+        let rowEntry: d3.Selection<any>;
+        const rowElement = row.forEach((column: LegendColumn<string>, index: number) => {
+          const columnBounds = table.getBounds(row, column);
+
+          if (index % 2  === 0) {
```
themadcreator: even/odd index detection for symbol/text determination seems pretty fragile

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
-              highY >= p.y && lowY <= p.y) {
-            let entrySelection = d3.select(this);
-            let datum = entrySelection.datum();
-            entities.push({
-              datum: datum,
-              position: { x: symbolX, y: symbolY },
-              selection: entrySelection,
-              component: legend,
-            });
+      function within(p: Point, bounds: Bounds) {
```
themadcreator: don't declare this function every call. Instead make it a private static method on `Legend` or add it to `Utils`

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
+        return bounds.topLeft.x <= p.x
+          && bounds.bottomRight.x >= p.x
+          && bounds.topLeft.y <= p.y
+          && bounds.bottomRight.y >= p.y;
+      };
+
+      const table = this._buildLegendTable(this.width(), this.height());
```
themadcreator, line 478: is `_calculateLayoutInfo` the conventional name for this in other components? If so, we should still use it. If not, whatever.

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
+          // we've already found the nearest entity; just return it.
+          return entity;
+        }
+
+        const rowBounds = table.getBounds(row);
+        const withinRow = within(p, rowBounds);
+
+        if (!withinRow) {
+          // the nearest entity isn't within this row, continue;
+          return entity;
```
themadcreator, line 490: since we've gotten here without modifying `entity` and given the first conditional, this is always going to be an empty array right? Using `reduce` in this way seems a little odd to me. Why even use an array as opposed to value/null?

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
+            // HACKHACK The 2.x API returns the center {x, y} of the symbol as the position.
+            const rowTranslate = d3.transform(d3.select(rowElement).attr("transform")).translate;
+            const symbolTranslate = d3.transform(symbolElement.attr("transform")).translate;
+
+            return [{
+              datum: column.getData(),
+              position: { x: rowTranslate[0] + symbolTranslate[0], y: rowTranslate[1] + symbolTranslate[1] },
+              selection: d3.select(entryElement),
+              component: this,
+            }];
```
themadcreator, line 514: and here instead of pushing to `entity`, we just return a statically declared single item array, which seems weird that we've tried to pass around the initial empty array through all the reduce calls.

src/components/legend.ts
```diff
@@ -1,4 +1,233 @@
 namespace Plottable.Components {
+  /**
+   * The Legend's column representation. Stores position information
+   * as well as data
+   */
+  class LegendColumn<T> {
```
themadcreator: This seems like a struct, and since you're not exporting it, why not just make it one instead of all the extra getter setter mess?

src/components/legend.ts
```diff
@@ -1,4 +1,233 @@
...
+    public setWidth(width: number) {
+      return this._width = width;
+    }
+  }
+
+  /**
+   * The Legend's row representations. Stores positioning information
+   * and column data.
+   */
+  class LegendRow {
```
themadcreator, line 25: Will "row" and "column" always make sense? Will we ever rotate this text? Will we ever present non-grid-aligned (what about center aligned) legends?

src/components/legend.ts
```diff
@@ -1,4 +1,233 @@
...
+    }
+
+    public forEach(callback: (row: LegendRow, index?: number) => void) {
+      this._rows.forEach(callback);
+    }
+
+    /**
+     * Returns the bounds relative to the parent and siblings of the row or
+     * column. If column is not specified, bounds will be calculate for the row only.
+     * If the column is specified, the bounds will be calculate for the column only.
```
themadcreator: wait, so to calculate the bounds for a column i also have to pass a row?

src/components/legend.ts
```diff
@@ -1,4 +1,233 @@
...
+     *
+     * @param {LegendRow} row The row to calculate bounds
+     * @param {LegendColumn} column The column to calculate bounds.
+     * @returns {Bounds}
+     */
+    public getBounds(row: LegendRow, column?: LegendColumn<any>) {
+      const indexOfRow = this._rows.indexOf(row);
```
themadcreator: can we avoid this `indexOf` linear scan if we pass in the row index instead? Will the caller already have that?

src/components/legend.ts
```diff
@@ -1,4 +1,233 @@
...
+      const rowYOffset = this._rows.slice(0, indexOfRow)
```
themadcreator: we're trading a memory copy for iteration, which is going to be slower in javascript than just scanning the whole array. Still, it seems like the faster thing would just be to do normal for-loop iteration with an early `break`.

themadcreator: Though this is probably premature optimization. Not like I profiled.

src/components/legend.ts
```diff
@@ -237,88 +514,107 @@ namespace Plottable.Components {
...
+        const rowBounds = table.getBounds(row);
+        const withinRow = within(p, rowBounds);
+
+        if (!withinRow) {
+          // the nearest entity isn't within this row, continue;
+          return entity;
+        }
+
+        return row.reduce((entity: Entity<Legend>[], column: LegendColumn<string>, columnIndex: number) => {
+          const columnBounds = table.getBounds(row, column);
```
themadcreator: `row` is irrelevant here, no? maybe i misunderstood your comment in `.getBounds`

test/components/legendTests.ts
```diff
@@ -3,6 +3,7 @@
 describe("Legend", () => {
   const ENTRY_SELECTOR = "." + Plottable.Components.Legend.LEGEND_ENTRY_CLASS;
   const SYMBOL_SELECTOR = "." + Plottable.Components.Legend.LEGEND_SYMBOL_CLASS;
+  const TEXT_LINE_SELECTOR = ".text-line";
```
themadcreator, line 6: all the other classes here are variables. Should this be?

themadcreator (changes requested):

src/components/legend.ts
hellochar: Yeah lets put this in MathUtils

hellochar (commented):
Code looks mostly fine.

Looks like the layout is a bit more spread out vertically -- do we know why that is?

| Old  | New |
| ------------- | ------------- |
| image | image |

There's more space between the second and third row.

-----

There's some craziness with how Legend behaves inside a Table. You should be able to control how much space e.g. a plot gets vs the legend by using Table.rowWeight or Table.columnWeight:

No column weights:
image

Plot column weight 3:
image

But it looks like develop has a bug where Legend only respects column weight if the Legend would've been rendered with text ellipses even if the other plots had no weight. If I change the previous screenshots to have a slightly less long label, then we get:

image

Irrespective of Plot column weight. This is where `maxWidth` would be most valuable, but if it was because of a bug to begin with, I'm now a little worried that `maxWidth` may not make sense at all. Thoughts on what to do with maxWidth and this bug?

src/components/legend.ts
CalvinFernandez: sure thing

src/components/legend.ts
CalvinFernandez: hmm good question -- rotated text should still be doable by the table (rotate the table :p ) I think even center-aligned legends are still a subset of a grid, just one with preference for changing margins such that everything is centered by row. Thoughts?

src/components/legend.ts
CalvinFernandez: lol yea the API is a little goofy i can change this if you think its too crazy

src/components/legend.ts
CalvinFernandez: sure

src/components/legend.ts
CalvinFernandez: extra infinity is the default value that will be returned if, say, `this.maxWidth` returns null .

src/components/legend.ts
CalvinFernandez: I can just fix that api

src/components/legend.ts
CalvinFernandez: ugh i know

src/components/legend.ts
themadcreator: Yeah, how about `getRowBounds` and `getColumnBounds`

src/components/legend.ts
themadcreator: But you have **two** `Infinity`s, one in the array, one as the second argument.

src/components/legend.ts
CalvinFernandez: nope just here

hellochar:
Behavior feels much better; lgtm after build failure!

hellochar (approved):

src/components/legend.ts
```diff
@@ -1,4 +1,190 @@
 namespace Plottable.Components {
+  /**
+   * The Legend's column representation. Stores position information
+   * as well as data
+   */
+  interface LegendColumn<T> {
```
themadcreator, line 6: Tight

src/components/legend.ts
themadcreator: Im still not sure about center alignment, but this is fine.

src/components/legend.ts
```diff
@@ -159,70 +395,72 @@ namespace Plottable.Components {
...
 
-      let availableWidthForEntries = Math.max(0, (availableWidth - this._padding));
+      const table = new LegendTable(width, height, this._padding);
+      const entryNames = this._colorScale.domain().slice().sort((a, b) => this._comparator(this._formatter(a), this._formatter(b)));
+
+      let row: LegendRow = new LegendRow();
+      table.addRow(row);
+      row.bottomPadding = this._rowBottomPadding;
+      entryNames.forEach((name, index) => {
+        if (row.columns.length / 2 === this.maxEntriesPerRow()) {
```
themadcreator, line 408: Are we guaranteed that the `length` will always be a multiple of 2?

src/components/legend.ts
```diff
@@ -1,4 +1,190 @@
...
+   /**
+     * Returns the bounds of the column relative to the parent and siblings of the
+     * column.
+     *
+     * @param {number} rowIndex The parent row containing the desired column.
+     * @param {number} columnIndex The column to calculate bounds.
+     * @returns {Bounds}
+     */
+    public getColumnBounds(rowIndex: number, columnIndex: number) {
+      const rowBounds = this.getRowBounds(rowIndex);
```
themadcreator, line 130: yeah, much nicer

themadcreator (approved):
Solid

src/components/legend.ts
CalvinFernandez: yep

PR #3178, opened 2017-01-23
Add max for time axis tick precision.

themadcreator:
Adds new property to time axes to limit how precise the
tick labels will go.

For example, setting the value of maxTimeIntervalPrecision
to "day" will prevent hours and minutes from showing
even if you are zoomed all the way in.

By default, not max is set.

### DIFF
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -1,14 +1,14 @@
 namespace Plottable {
 
-export namespace TimeInterval {
-  export var second = "second";
-  export var minute = "minute";
-  export var hour = "hour";
-  export var day = "day";
-  export var week = "week";
-  export var month = "month";
-  export var year = "year";
-};
+  export namespace TimeInterval {
+    export var second = "second";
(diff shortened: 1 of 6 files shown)


src/axes/timeAxis.ts
```diff
@@ -211,6 +222,43 @@ namespace Plottable.Axes {
...
+     * Gets the maximum TimeInterval precision
+     */
+    public maxTimeIntervalPrecision(): string;
+    /**
+     * Sets the maximum TimeInterval precision. This limits the display to not
+     * show time intervals above this precision. For example, if this is set to
+     * `TimeInterval.day` then no hours or minute ticks will be displayed in the
+     * axis.
+     *
+     * @param {TimeInterval} newPositions The positions for each tier. "bottom" and "center" are the only supported values.
```
hellochar: docs are off here

hellochar (commented):
Code looks good; just update docs

CalvinFernandez (approved):

Issue #3088, opened 2016-07-22
Per-label Formatting Option for plot.labelFormatter(d)


freedomflyer:
I've dug a lot, but haven't found a way to do custom, per-label formatting. As you see below, `labelFormatter` is only given one argument, `d`, so there's nothing else to key off of to intelligently format a label differently than any other label. However, in other methods such as `attr(d, i, dataset)`, I have that information.  Getting that information to labelFormatter seems trivial, and would be very useful in my use-case.

Is there a way around this? Otherwise, this is a feature request that seems fair. Something like `labelFormatter(d, i, dataset)`

``` ts
myPlot
.labelFormatter(d => {
    // All bars will be formatted only one way
    return myFormatter.display(d);
})
.attr('plot-index', (d, i, dataset) => {
    // Here, however, I have access to the dataset and index, so I can perform more custom logic
    return myAttrFunction(d, i, dataset);
})
```

hellochar:
There's an 80% ready PR at https://github.com/palantir/plottable/pull/3226 ; maybe we can just touch it up and get this in?

PR #3180, opened 2017-01-25
Provide datum, index, and dataset to label formatter

dallinregehr:
An attempt at implementing #3088

### DIFF
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -1,6 +1,7 @@
 namespace Plottable {
 
-  export type Formatter = (d: any, datum?: any, index?: number, dataset?: Dataset) => string;
+  export type Formatter = (d: any) => string;
+  export type LabelFormatter = (d: any, datum: any, index: number, dataset: Dataset) => string;
 }
 
 namespace Plottable.Formatters {
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -81,6 +81,15 @@ namespace Plottable.Formatters {
   }
 
   /**
+   * Creates a formatter for generating chart labels.
+   *
+   * @returns {LabelFormatter} A formatter for generating chart labels.
+   */
+  export function labelFormatter() {
+    return (d: any, datum: any, index: number, dataset: Dataset) => String(d);
+  }
+
+  /**
    * Creates a formatter for percentage values.
    * Multiplies the input by 100 and appends "%".
    *
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -1,7 +1,6 @@
 namespace Plottable {
 
-  export type Formatter = (d: any) => string;
-
+  export type Formatter = (d: any, datum?: any, index?: number, dataset?: Dataset) => string;
 }
 
 namespace Plottable.Formatters {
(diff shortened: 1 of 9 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @dallinregehr! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/core/formatters.ts
```diff
@@ -1,7 +1,6 @@
 namespace Plottable {
 
-  export type Formatter = (d: any) => string;
-
+  export type Formatter = (d: any, datum?: any, index?: number, dataset?: Dataset) => string;
```
CalvinFernandez: Nice! I'm a little concerned about the lack of type-safety afforded by including the parameters as optional arguments. 


Maybe we can include another `Formatter` type more tightly scoped to the components this change effects, namely `Pie` and `Bar` charts? Consider creating a new formatter type declared under the namespace `Plottable.Plots` called `LabelFormatter` or something which is of type `(d: any, datum: any, index: number, dataset: Dataset) => string;`. Then if you change the type of `_labelFormatter` to `LabelFormatter` in Pie and Bar, we can guarantee that datum, index and dataset will be passed to the labelFormatter callback for those charts while making no promises about the other `Formatter` instances.

dallinregehr: Good idea. I'll update shortly.

src/plots/barPlot.ts
```diff
@@ -212,17 +212,38 @@ namespace Plottable.Plots {
...
+    /**
+     * Sets the Formatter for the chart axis.
      *
      * @param {Formatter} formatter
      * @returns {Bar} The calling Bar Plot.
      */
-    public labelFormatter(formatter: Formatter): this;
-    public labelFormatter(formatter?: Formatter): any {
+    public axisLabelFormatter(formatter: Formatter): this;
+    public axisLabelFormatter(formatter?: Formatter): any {
```
CalvinFernandez, line 247: Ah we can't do this its an api-breaking change

CalvinFernandez (changes requested):
Hmm unfortunately we can't add an `axisLabelFormatter` endpoint and remove support for axis label formatting for those who've only defined a `labelFormatter`. 

The formatter call in stacked bar chart ( the thing thats messing up your current api ) is formatting the sum label that appears above the stack.
```
     9 < --- This label 
   [ 2 ]
   [ 3 ]
   [ 4 ]
```
One thing you could do to get around the api issue is to change `LabelFormatter` to `(d: any, datum: any[], index: number[], dataset: Dataset)` 


And then the stacked bar call would pass the formatting function the array of data points in the stack along with their indices. 


What are you thoughts on this?

adidahiya:
Ping on above questions @dallinregehr ^

Also, heads up, there are now a bunch of conflicts because we have migrated the codebase to use modules rather than namespaces.

dallinregehr:
Decided to open new PR (#3226) branched on migrated codebase to avoid messy conflict resolutions.

PR #3182, opened 2017-01-25
Add demo preview to plottable PRs

themadcreator:
### Adds demo previews to Plottable PR builds

* Uses new `circle-github-bot` npm package
* Vanilla HTML and CSS
* Uses native browser API for displaying example plots


hellochar (changes requested):
sweet! We already have a suite of good tests to check against in the quicktests folder; can we just expose those instead of needing to maintain a separate demo app?

hellochar:
Looks like some of the quicktests reference csv files with an absolute path (https://github.com/palantir/plottable/blob/develop/quicktests/overlaying/tests/realistic/baseball.js#L8) which is failing in circle; can we get those working? Other than that looks good

hellochar (approved):

PR #3189, opened 2017-01-30
Category Gridlines

themadcreator:
Continuation of #3067


themadcreator:
In particular: https://45-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/demo/index.html#gridlines

.gitignore
```diff
@@ -11,6 +11,7 @@ node_modules/
 local/
 quicktests/github_token.txt
 quicktests/overlaying/list_of_quicktests.json
+*.log
```
CalvinFernandez, line 14: whats this?

src/axes/categoryAxis.ts
```diff
@@ -117,22 +120,101 @@ namespace Plottable.Axes {
...
+    public tickMarkPosition(): ITickMarkPosition;
+    /**
+     * Sets  whether tick marks are drawn in the center of category or after it.
      *
-     * @param {Scales.Category} scale - The scale being downsampled. Defaults to this Axis' scale.
-     * @return {DownsampleInfo} an object holding the resultant domain and new stepWidth.
+     * @param {ITickMarkPosition} tickMarkPosition - "center" or "after"
+     * @returns {Category} The calling Category.
      */
-    public getDownsampleInfo(scale = <Scales.Category> this._scale): DownsampleInfo {
```
CalvinFernandez, line 130: Sigh this is technically an api break ...

src/axes/categoryAxis.ts
```diff
@@ -117,22 +120,101 @@ namespace Plottable.Axes {
     }
 
     protected _getTickValues() {
-      return this.getDownsampleInfo().domain;
+      return (this._scale as Scales.Category).getDownsampleInfo().domain;
```
CalvinFernandez: Should we just make the type of this._scale, Scales.Category? Any reason to leave it as Scale within categoryAxis?

src/axes/categoryAxis.ts
```diff
@@ -117,22 +120,101 @@ namespace Plottable.Axes {
...
+     * Gets the current tick mark position value.
+     */
+    public tickMarkPosition(): ITickMarkPosition;
+    /**
+     * Sets  whether tick marks are drawn in the center of category or after it.
      *
-     * @param {Scales.Category} scale - The scale being downsampled. Defaults to this Axis' scale.
-     * @return {DownsampleInfo} an object holding the resultant domain and new stepWidth.
+     * @param {ITickMarkPosition} tickMarkPosition - "center" or "after"
+     * @returns {Category} The calling Category.
```
CalvinFernandez, line 144: for consistency I suppose we can keep doing this but fyi we don't need the type info `{Typed}` as typedoc can infer stuff on its own.

src/axes/categoryAxis.ts
```diff
@@ -117,22 +120,101 @@ namespace Plottable.Axes {
...
+    public tickTextAlignment(tickTextAlignment: string): this;
+    public tickTextAlignment(tickTextAlignment?: string): any {
+      if (arguments.length === 0) {
+        return this._tickTextAlignment;
+      }
+
+      if (! tickTextAlignment) {
+        this._tickTextAlignment = null;
+      } else {
+        let v = tickTextAlignment.toLowerCase();
```
CalvinFernandez: nit: probably don't need to API hand-hold quite so much. The below error is clear enough imo

src/axes/categoryAxis.ts
```diff
@@ -117,22 +120,101 @@ namespace Plottable.Axes {
...
+     *     for -90 rotation, alignment must be "right". For "bottom"
+     *     orientation and 90 rotation, alignment must "right"; For
+     *     -90 rotation, alignment must be "left".
+     *
+     * @param {Number} tickTextPadding Padding in pixels.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextPadding(tickTextPadding: number): this;
+    public tickTextPadding(tickTextPadding?: number): any {
```
CalvinFernandez, line 220: The conditions around when this api will have an effect and when it won't are a bit confusing -- why do we need it? Can we simplify it?

src/axes/categoryAxis.ts
```diff
@@ -211,6 +293,53 @@ namespace Plottable.Axes {
...
+        if (this.orientation() === "top" && this.tickTextAlignment() === "left") {
+          pad.y = this.tickTextPadding();
+          translate.y = this.tickTextPadding();
+        } else if (this.orientation() === "bottom" && this.tickTextAlignment() === "right") {
+          pad.y = this.tickTextPadding();
+          translate.y = this.tickTextPadding() * -1;
+        }
+      }
+
+      return { pad: pad, translate: translate };
```
CalvinFernandez, line 349: why not just return translate and just make the space calculation take into account the possibility of negative values?

src/components/gridlines.ts
```diff
@@ -1,37 +1,119 @@
...
-      }
-      if (yScale != null && !(QuantitativeScale.prototype.isPrototypeOf(yScale))) {
-        throw new Error("yScale needs to inherit from Scale.QuantitativeScale");
-      }
+    constructor(
+        xScale: QuantitativeScale<any> | Plottable.Scales.Category,
+        yScale: QuantitativeScale<any> | Plottable.Scales.Category) {
+
+      Gridlines.validateScale(xScale, "xScale");
+      Gridlines.validateScale(yScale, "yScale");
```
CalvinFernandez, line 41: Thoughts on the defensive check here? Types aren't enough?

src/components/gridlines.ts
```diff
@@ -1,37 +1,119 @@
...
+
+    private _xScale: Scale<any, number>;
+    private _yScale: Scale<any, number>;
     private _xLinesContainer: d3.Selection<void>;
     private _yLinesContainer: d3.Selection<void>;
+    private _xTicks: () => ScaleDomainValues;
+    private _yTicks: () => ScaleDomainValues;
 
-    private _renderCallback: ScaleCallback<QuantitativeScale<any>>;
+    /**
```
CalvinFernandez: emptiness

src/components/gridlines.ts
```diff
@@ -67,10 +149,25 @@ namespace Plottable.Components {
       return this;
     }
 
+    private _mkTicks(scale: QuantitativeScale<any> | Plottable.Scales.Category): () => ScaleDomainValues {
+      if (QuantitativeScale.prototype.isPrototypeOf(scale)) {
+        return () => (<QuantitativeScale<any>>scale).ticks();
```
CalvinFernandez, line 151: consider an `.ticks` API for Category/Quantitative scale that you can call rather than the type check

src/components/gridlines.ts
```diff
@@ -1,37 +1,119 @@
...
 
+    /**
+     * Gets the xLinePosition value.
+     */
+    public xLinePosition(): ILinePosition;
+    /**
+     * Sets the xLinePosition value, which specifies whether lines are drawn in
+     * the center of category or after it.
+     *
+     * Valid options are "center" and "after". Note that this only makes sense
```
CalvinFernandez, line 70: Shouldn't these mirror the tick mark position on the axis?

CalvinFernandez (commented):
Looking good -- some overall questions. Main concern is an API break and a slightly confusing text padding api.

.gitignore
themadcreator: `npm-debug.log`, but also there aren't any .logs we should add, so just using the wildcard

src/components/gridlines.ts
themadcreator: I think that plottable has added various checks for when its used outside of a typescript environment to attempt to enforce our contracts for vanilla javascript users.

src/components/gridlines.ts
themadcreator: Well 1) they don't have to and 2) they are different components that don't know about eachother. The other option would be to add the setting to the Scale, but that doesn't really make sense to add a setting that is only used in this context to the scale.

src/axes/categoryAxis.ts
themadcreator: yeah, but ill keep this for consistency until we change everything

src/axes/categoryAxis.ts
themadcreator: I can add back referencing the scale's method

src/axes/categoryAxis.ts
themadcreator: agreed

src/axes/categoryAxis.ts
themadcreator: I don't really know, this was part of the initial PR i didn't touch.

src/components/gridlines.ts
themadcreator: i'll just remove since we have comments on the getter/setters

src/components/gridlines.ts
themadcreator: yeah that'd be nice. not now tho

src/axes/categoryAxis.ts
themadcreator: So im not 100% sure but i think the reason these needs to be separated are that padding will affect line wrapping and the "available space" but the translate applied after that and sets the proper position given the padding.

src/axes/categoryAxis.ts
CalvinFernandez: I have a really hard time reading the comment for this API. What does the "opposite edge of the axis" refer to? I'm guessing that it will intelligently pad based on the orientation and alignment of the label and the position of the axis? I'm guessing its probably really intuitive once you see it in action, but the comment is long and confusing. Do you think it's necessary to list every case?

src/axes/categoryAxis.ts
```diff
@@ -1,15 +1,7 @@
 namespace Plottable.Axes {
-  export interface DownsampleInfo {
-    domain: string[];
-    stepWidth: number;
-  }
+  export type ITickMarkPosition = "center" | "after";
```
hellochar, line 2: Why would people use "after"? I'd prefer a "between" (that puts ticks between the labels)

src/axes/categoryAxis.ts
```diff
@@ -117,22 +130,100 @@ namespace Plottable.Axes {
...
-      return {
-        domain: scale.domain().filter((d, i) => i % downsampleRatio === 0),
-        stepWidth: downsampleRatio * scale.stepWidth(),
-      };
+    public tickMarkPosition(tickMarkPosition: ITickMarkPosition): this;
+    public tickMarkPosition(tickMarkPosition?: ITickMarkPosition): any {
+      if (arguments.length === 0) {
+        return this._tickMarkPosition;
+      } else {
+        this._tickMarkPosition = tickMarkPosition;
```
hellochar, line 151: I don't know why other categoryAxis methods don't do this but we should add a .redraw() on prop changes

src/axes/categoryAxis.ts
```diff
@@ -117,22 +130,100 @@ namespace Plottable.Axes {
...
+     * Note that alignment is applied before rotation (see "{@link
+     * tickLabelAngle}").
+     *
+     * @param {String} tickTextAlignment One of "left", "right", "center",
+     * or a falsy value.
+     *
+     * @returns {Category} The calling Category.
+     */
+    public tickTextAlignment(tickTextAlignment: string): this;
+    public tickTextAlignment(tickTextAlignment?: string): any {
```
hellochar, line 174: nit: accept a string literal union instead of a generic string. In fact we already have http://plottablejs.org/docs/classes/plottable.components.alignment.html

src/axes/categoryAxis.ts
hellochar: Agreed with Calvin; not a fan of this API since it's is confusing and its use is special-casey (we should avoid properties that will only be applied when certain conditions are met). Also not sure what "opposite edge" means. Additionally, `Axis` already has properties `tickLabelPadding()` and `margin()` which sound like they have overlapping responsibilities

src/axes/categoryAxis.ts
hellochar: This is hard to maintain and ensure correctness. What is the goal here? I'd like to avoid this if possible

src/components/gridlines.ts
```diff
@@ -1,37 +1,116 @@
...
+
+  export type ILinePosition = "center" | "after";
+  type ScaleDomainValues = (string | number)[];
+
   export class Gridlines extends Component {
-    private _xScale: QuantitativeScale<any>;
-    private _yScale: QuantitativeScale<any>;
+    private static validateScale = (scale: any, which: string) => {
+      if (scale != null &&
+          ! QuantitativeScale.prototype.isPrototypeOf(scale) &&
```
hellochar, line 9: we use instanceof elsewhere; lets stick with that

src/components/gridlines.ts
```diff
@@ -67,10 +146,25 @@ namespace Plottable.Components {
       return this;
     }
 
+    private _mkTicks(scale: QuantitativeScale<any> | Plottable.Scales.Category): () => ScaleDomainValues {
```
hellochar, line 149: improve the name; what's `mk`?

src/components/gridlines.ts
hellochar: use instanceof, typescript will narrow the type inside

src/scales/categoryScale.ts
```diff
@@ -98,6 +112,22 @@ namespace Plottable.Scales {
...
     /**
+     * Take the scale and drop ticks at regular intervals such that the
+     * resultant ticks are all a reasonable minimum distance apart. Return the
+     * resultant ticks to render, as well as the new stepWidth between them.
+     *
+     * @return {DownsampleInfo} an object holding the resultant domain and new
+     * stepWidth.
+     */
+    public getDownsampleInfo(): DownsampleInfo {
+      const downsampleRatio = Math.ceil(Category._MINIMUM_WIDTH_PER_LABEL_PX / this.stepWidth());
```
hellochar, line 123: _MINIMUM_WIDTH_PER_LABEL_PX belongs in the Axis, not the Scale (scales shouldn't know anything about being rendered)

demo/demos/gridlines.ts
```diff
@@ -0,0 +1,125 @@
...
+    const table = new Plottable.Components.Table([
+            [new Plottable.Components.Group([grid, plot])],
+            [xAxis]
+        ]);
+    return table;
+}
+
+export const route = "gridlines";
+export const title = "Grid Lines";
+export function render(el: HTMLElement) {
```
hellochar, line 97: Quicktests already cover the use-case for "inspect a chart to see how it looks"; we should move this over there. Can you clarify what we should put in `demo`?

hellochar (changes requested):

hellochar:
image

bug: there should be a gridline before Human

image

downsampling makes no sense with "after" ticks; this chart looks like the single category is labelling the whole group of bars.

src/scales/categoryScale.ts
themadcreator: both the axis and the gridlines component need to know about this, and I dont want to repeat this calculation. what do you think?

src/components/gridlines.ts
themadcreator: i think the original author intended it to mean "make". I'll change to `_makeTicks`

src/axes/categoryAxis.ts
themadcreator: Re: the text padding API, I'm not sure what we want to do here. I didn't touch any of this since it came from the original community submission PR. Do we want this feature or not?

src/axes/categoryAxis.ts
themadcreator: In Adi's comment https://github.com/palantir/plottable/pull/3067#issuecomment-207103123 the image suggests that there are lines after each series, not just between. If we do go with "between", i'll want to remove the final gridline (should we also remove the final tick?

themadcreator:
Even if downsampling doesn't make sense with "after", we should still be able to do it.

src/axes/categoryAxis.ts
hellochar: What does it do? I don't know what the text padding API is for and I'm guessing it overlaps responsibilities with tickLabelPadding() and margin(); we should remove it so.

src/scales/categoryScale.ts
hellochar: Oh gridlines component... hmm... okay this is fine. Maybe in the future we should tie gridlines to a specific axis instead of a specific scale?

src/axes/categoryAxis.ts
hellochar: The rest of the discussion you linked settles on "between" though? My main concern with "after" is that it makes [this chart](https://cloud.githubusercontent.com/assets/1079495/22489635/998765fe-e7e6-11e6-9de9-790d939dd164.png) look bad

themadcreator:
From the reviews it sounds like category gridlines make sense but the padding API is a bit confusing. Also, due to the modularization changes, there are major conflicts here. So, I'm just going to close this PR for now and re-open a new PR with category gridlines only.

PR #3192, opened 2017-01-31
Add `stackingOrder` property to stacked bar and stacked area charts 

themadcreator:
By default, stacked plots are "bottomup", meaning for positive data, the
first series will be placed at the bottom of the scale and subsequent data
series will by stacked on top. This can be reveresed by setting stacking order
to "topdown".

This makes it easier to have a legend that matches the order of the stack since
legends will display the first series at the top of the list.

Add demo for different sort directions


CalvinFernandez:
Thoughts on just allowing users to specify a sorting lambda so they can stack however they want?

CalvinFernandez:
If you think this is better thought it looks great to me

themadcreator:
We can always later make the stackingOrder method take a type `"bottomup" | "topdown" | (fancyLambda) => here`

themadcreator:
but i think that this'll will solve the main issue with the least fuss

themadcreator:
demo https://50-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/demo/index.html#bars

CalvinFernandez (approved):

hellochar (approved):
Oh I see, so this would be the better fix for bp-charts' "reverse series/tooltips for stacked area" hack right?

themadcreator:
Yeah, exactly

PR #3199, opened 2017-02-02
Routable Quicktests

themadcreator:
The "overlaying" quicktests can't be individual linked to.

Moreover livereload resets your view so you have to re-select
the tests you were working on every time.

This commit adds a simple list of tests that can be linked
to with hashroutes so livereload updates will maintain
your place and you can work with tests one at a time.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,6 @@
 {
   "env": {
+    "es6": true,
     "browser": true,
     "jquery": true
   },
--- a/quicktests/dev/dev.js
+++ b/quicktests/dev/dev.js
@@ -1,5 +1,4 @@
 
-
 class Router {
     constructor() {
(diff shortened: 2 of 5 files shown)


CalvinFernandez (approved):

hellochar (approved):

.eslintrc
```diff
@@ -1,5 +1,6 @@
 {
   "env": {
+    "es6": true,
```
hellochar, line 3: What does this line do?

themadcreator: that enable es6 syntax to pass eslint rules (for example es6 classes, lambdas). eslint is only done on our quicktests, and since our quicktests use modern browsers in sauce and locally, this should be fine. Welcome to 2017!

PR #3201, opened 2017-02-02
Labels: Add 10px padding around bar labels

### LINKED ISSUE #3196: Bar labels should be more conservative about requesting space
Instead of drawing labels when text barely fits within a bar, I'd like labels to request 10px of extra padding on all sides before being drawn. This requires users to zoom in or create less dense charts in order to see labels. Currently, it is too easy to draw plots with too much visual complexity with labels.

Note that this is a small visual breaking change because some charts will now hide labels more often.

CalvinFernandez:
closes #3196

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -8,11 +8,12 @@ namespace Plottable.Plots {
   export class Bar<X, Y> extends XYPlot<X, Y> {
     public static ORIENTATION_VERTICAL = "vertical";
     public static ORIENTATION_HORIZONTAL = "horizontal";
+
     private static _BAR_WIDTH_RATIO = 0.95;
     private static _SINGLE_BAR_DIMENSION_RATIO = 0.4;
     private static _BAR_AREA_CLASS = "bar-area";
-    private static _LABEL_VERTICAL_PADDING = 5;
-    private static _LABEL_HORIZONTAL_PADDING = 5;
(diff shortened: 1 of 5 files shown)


hellochar (approved):

src/plots/barPlot.ts
```diff
@@ -584,7 +585,9 @@ namespace Plottable.Plots {
         };
         writer.write(text, containerWidth, containerHeight, writeOptions);
 
-        let tooWide = this._isVertical ? barWidth < measurement.width : barHeight < measurement.height;
+        let tooWide = this._isVertical
+          ? barWidth < (measurement.width + Bar._LABEL_PADDING * 2)
```
adidahiya, line 589: maybe a comment to help clarify that we want the padding applied on _both_ sides of the text? otherwise lgtm

adidahiya (approved):

src/plots/barPlot.ts
CalvinFernandez: word

adidahiya:
@CalvinFernandez does this work for stacked bar sum labels too?

CalvinFernandez:
no this doesn't affect the sum labels

adidahiya:
@CalvinFernandez can you handle that in a follow-up PR? I think the behavior should be consistent across those labels

Issue #3194, opened 2017-02-01
allow PanZoom on scales whose max < min


hellochar:
You can `scale.domain([10, 0])` to get a scale that goes in the "reverse" direction and almost all of Plottable handles it correctly, except PanZoomInteraction errors. We should make PanZoom support it

themadcreator:
Fixed!

PR #3203, opened 2017-02-02
Fix pan-zoom interaction for reversed domains and ranges

themadcreator:
Domains with max < min (like [-10, 0]) work just fine
with the pan-zoom interaction, but we had explicit tests
disabling this. The commit removes those conditions and
removes the unit tests that confirmed this.

However, there was an issue with reversed ranges since
all pan-zoom constraint calculations happen in range space.
This commit includes fixes for those calculations and
adds a quicktest that has every combination of x/y reversing
in scatter plots.

### DIFF
--- a/src/interactions/panZoomInteraction.ts
+++ b/src/interactions/panZoomInteraction.ts
@@ -81,6 +81,7 @@ namespace Plottable.Interactions {
         yScale.pan(this._constrainedTranslation(yScale, translateAmount.y));
       });
     }
+
     /**
      * Zooms the chart by a specified amount around a specific point
      *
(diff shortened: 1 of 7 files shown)


themadcreator:
Fixes #3194

themadcreator:
[Pan Zoom Reverse Quicktest](https://73-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/pan_zoom_reverse)

src/interactions/panZoomInteraction.ts
```diff
@@ -309,7 +321,7 @@ namespace Plottable.Interactions {
         const testMaxRange = zoomAt(currentMaxRange, zoomAmount, centerPoint);
 
         // move the center point to prevent max overflow, if necessary
-        if (testMaxRange > maxRange) {
+        if (testMaxRange > maxRange != reversed) {
```
CalvinFernandez, line 324: this is cool

CalvinFernandez (approved):

src/interactions/panZoomInteraction.ts
themadcreator: yep, a nice little XOR

PR #3206, opened 2017-02-03
add license header to source files

hellochar:
now that we release modules, all source files (that get compiled into
the /build folder) should have license headers.

### DIFF
--- a/license_header.txt
+++ b/license_header.txt
@@ -1,5 +1,5 @@
 /*!
 Plottable @VERSION (https://github.com/palantir/plottable)
-Copyright 2014-2017 Palantir Technologies
+Copyright 2014-present Palantir Technologies
 Licensed under MIT (https://github.com/palantir/plottable/blob/master/LICENSE)
 */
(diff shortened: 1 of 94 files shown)


license_header.txt
```diff
@@ -1,5 +1,5 @@
 /*!
 Plottable @VERSION (https://github.com/palantir/plottable)
-Copyright 2014-2017 Palantir Technologies
+Copyright 2014-present Palantir Technologies
```
CalvinFernandez, line 3: is this correct?

hellochar: Yeah I think "present" is an acceptable thing to put here; @themadcreator @adidahiya ?

themadcreator: Facebook uses this style in their react headers: e.g. https://cdnjs.cloudflare.com/ajax/libs/react/15.4.2/react.js

themadcreator (approved):
I kind of wanted:
```
/**
 * Copyright ...
 */
```
but this is fine

adidahiya (changes requested):
if we're making changes here, might as well do JSDoc license headers. http://usejsdoc.org/tags-license.html

hellochar:
@adidahiya @themadcreator how's that?

adidahiya (approved):

PR #3208, opened 2017-02-03
Fix stacked bar labels for date axes

CalvinFernandez:


adidahiya:
@CalvinFernandez can you describe what's going on here? What was the problem and its solution?

src/utils/stackingUtils.ts
```diff
@@ -20,7 +24,10 @@ export type IStackingOrder = "topdown" | "bottomup";
 
 // HACKHACK More accurate stacking result definition. Remove typeless stacking result
 // and replace with typed stacking result in 3.0.
-export type GenericStackingResult<D> = Utils.Map<Dataset, Utils.Map<D, StackedDatum>>;
+//
+// Maps compare Date keys using referential equality. So, for safety, we only want primitives
```
hellochar, line 28: Ugh that's annoying. The handling of Date vs number vs string in Plottable seems wonky and all over the place. Are we re-creating different `new Date()` objects for the same underlying data?

src/utils/stackingUtils.ts
```diff
@@ -86,28 +94,28 @@ export function stack(
...
+    stack.forEach((datum, key) => {
       // correctly handle negative bar stacks
       const maximalValue = Utils.Math.max([datum.offset + datum.value, datum.offset], datum.offset);
       const minimalValue = Utils.Math.min([datum.offset + datum.value, datum.offset], datum.offset);
 
       if (!maximumExtents.has(key)) {
-        maximumExtents.set(key, maximalValue);
-      } else if (maximumExtents.get(key) < maximalValue) {
-        maximumExtents.set(key, maximalValue);
+        maximumExtents.set(key, { extent: maximalValue, axisValue: datum.axisValue });
```
hellochar, line 110: How does this prevent the key from being a Date object?

hellochar (approved):
Looks fine for a fix, but bug feels symptomatic of an underlying issue

src/utils/stackingUtils.ts
CalvinFernandez: Wonky is a good way to describe it I didn't have time to dig into the specifics

PR #3209, opened 2017-02-03
[WIP] HTML Components

CalvinFernandez:


CalvinFernandez:
things are showing up now! 
image

CalvinFernandez:
The layout functionality now works

image


Interactions are still very broken

plottable.css
```diff
@@ -44,6 +44,15 @@ svg.plottable {
   pointer-events: visibleFill;
 }
 
+/* Root-level html components are relative, all others are absolute */
+div.plottable {
```
adidahiya: you shouldn't qualify class selectors with an element tag; it's not performant. I think you should use a different class to do this styling adjustment instead.

https://github.com/stylelint/stylelint/tree/master/lib/rules/selector-no-qualifying-type

quicktests/overlaying/overlaying.js
```diff
@@ -229,12 +229,12 @@ function loadAllQuickTests(quicktestsPaths, firstQTBranch, secondQTBranch){
       var className = "quicktest " + name;
       var div = d3.select("#results").append("div").attr("class", className);
       div.insert("label").text(name);
-      var firstsvg = div.append("div").attr("class", "first").append("svg").attr({width: svgWidth, height: svgHeight});
-      var secondsvg = div.append("div").attr("class", "second").append("svg").attr({width: svgWidth, height: svgHeight});
+      var firstdiv = div.append("div").attr("class", "first").append("div").attr({width: svgWidth, height: svgHeight});
```
adidahiya: I don't think you should use the "div" name to identify elements; instead use "element". the important change you're making here is a switch from `SVGElement` to `HTMLElement`; the tag itself is not so important (it could conceivably be a span, for instance).

this applies to all the quicktests

src/components/componentContainer.ts
```diff
@@ -1,27 +1,23 @@
-/**
- * Copyright 2014-present Palantir Technologies
- * @license MIT
- */
-
-import * as d3 from "d3";
-
-import { Component, ComponentCallback } from "./component";
+import { GenericComponentCallback, IComponent } from "./abstractComponent";
```
adidahiya: retain the copyright header

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,247 @@
+import * as d3 from "d3";
```
adidahiya, line 6: copyright header

src/npm-debug.log
```diff
@@ -0,0 +1,48 @@
+0 info it worked if it ends with ok
```
adidahiya: remove this file...

src/utils/clientToHTMLTranslator.ts
```diff
@@ -0,0 +1,35 @@
+import * as DOM from "./domUtils";
```
adidahiya: copyright header

src/utils/clientToHTMLTranslator.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Point } from "../core/interfaces";
+
+export class ClientToHTMLTranslator {
+  private static _TRANSLATOR_KEY = "__Plottable_ClientToHTMLTranslator";
+
+  /**
+   * Returns a Translator for the root of the component.
+   * If one already exists for the root, it will be returned; otherwise, a new one will be created.
+   */
+  public static getTranslator(component: HTMLComponent) {
```
adidahiya: add a return typedef for public APIs like this

src/utils/clientToHTMLTranslator.ts
```diff
@@ -0,0 +1,35 @@
...
+      measureElement.setAttribute("style", "opacity: 0; visibility: hidden; position: absolute; width: 1px; height: 1px;");
+      root.appendChild(measureElement);
+
+      translator = new Translator(new PlottableHTMLElement(measureElement));
+      (<any> root)[ClientToHTMLTranslator._TRANSLATOR_KEY] = translator;
+    }
+
+    return translator;
+  }
+}
```
adidahiya: EOF newline

src/utils/domUtils.ts
```diff
@@ -10,6 +10,19 @@ import { Range } from "../core/interfaces";
 let nativeMath: Math = (<any>window).Math;
 
 /**
+ * Returns whether the child is in fact a child of the parent
+ */
+export function contains(parent: Element, child: Element) {
```
adidahiya: add `:boolean` return typedef

src/utils/plottableElement.ts
```diff
@@ -0,0 +1,55 @@
+export interface IPlottableElement<D extends Element> {
```
adidahiya: copyright header

src/utils/plottableElement.ts
```diff
@@ -0,0 +1,55 @@
...
+  public setAttribute(name: string, value: string) {
+    this._element.setAttribute(name, value);
+    return this;
+  }
+
+  public abstract left(position: number): this;
+  public abstract top(position: number): this;
+}
+
+export class PlottableHTMLElement extends PlottableElement<HTMLElement> {
```
adidahiya: one class per file please

test/coverage.html
```diff
@@ -21,6 +21,7 @@
     <script src="../node_modules/sinon/lib/sinon/times_in_words.js"></script>
     <script src="../node_modules/sinon/lib/sinon/format.js"></script>
     <script src="../node_modules/sinon/lib/sinon/spy.js"></script>
+    <script src="../node_modules/sinon/lib/sinon/stub.js"></script>
```
adidahiya, line 24: oh dear, this is unfortunate

adidahiya (dismissed):

plottable.css
hellochar: @adidahiya could you provide some reading on that? I looked into this before but only found ~3 year old resources, along with other resources telling me not to pay attention to those old things since browsers have been improving perf

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+
+export interface IComponent<D> {
+  /**
+   * Attaches the Component as a child of a given html element.
+   *
+   * @param {D} selection.
+   * @returns {Component} The calling Component.
+   */
+  anchor(selection: D): this;
+  anchorHTML(selection: HTMLElement): this;
```
hellochar: why do we need both functions here?

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  /**
+   * Adds a callback to be called on anchoring the Component to the DOM.
+   * If the Component is already anchored, the callback is called immediately.
+   *
+   * @param {GenericComponentCallback} callback
+   * @return {Component}
+   */
+  onAnchor(callback: GenericComponentCallback<D>): this;
```
hellochar, line 34: nit: wonder if we even need this/what it's used for

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   * converts client coordinates to component coordinates and
+   * takes into account any transformations that have been applied
+   * to the element.
+   *
+   * @return {Utils.ITranslator}
+   */
+   translator(): Utils.Translator;
+}
+
+export abstract class AbstractComponent<D> implements IComponent<D> {
```
hellochar: can you explain again why we separate AbstractComponent from the IComponent interface

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   */
+  fixedHeight(): boolean;
+  /**
+   * Detaches a Component from the DOM. The Component can be reused.
+   *
+   * This should only be used if you plan on reusing the calling Component. Otherwise, use destroy().
+   *
+   * @returns The calling Component.
+   */
+  detach(): this;
```
hellochar, line 157: nit: do we know how people actually use this? This API always felt confusing to me

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   * The callback is identified by reference equality.
+   *
+   * @param {GenericComponentCallback} callback
+   * @return {IComponent} The calling Component.
+   */
+  offDetach(callback: GenericComponentCallback<D>): this;
+  /**
+   * Gets the parent ComponentContainer for this Component.
+   */
+  parent(): ComponentContainer;
```
hellochar: nit: can you doc what happens if there's no parent

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  /**
+   * Sets the parent ComponentContainer for this Component.
+   * An error will be thrown if the parent does not contain this Component.
+   * Adding a Component to a ComponentContainer should be done
+   * using the appropriate method on the ComponentContainer.
+   */
+  parent(parent: ComponentContainer): this;
+  parent(parent?: ComponentContainer): any;
+  /**
+   * @returns {Bounds} for the component in pixel space, where the topLeft
```
hellochar: nit: can you docs where the origin of pixel space is

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   */
+  height(): number;
+  /**
+   * Gets the origin of the Component relative to its parent.
+   *
+   * @return {Point}
+   */
+  origin(): Point;
+  /**
+   * Gets the origin of the Component relative to the root
```
hellochar: nit: in `bounds()` we established a "pixel space"; if `origin()` or `originToRoot()` is in the same "space" we should reuse terminology

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  root(): IComponent<any>;
+  /**
+   * Gets the translator for the component. A translator
+   * converts client coordinates to component coordinates and
+   * takes into account any transformations that have been applied
+   * to the element.
+   *
+   * @return {Utils.ITranslator}
+   */
+   translator(): Utils.Translator;
```
hellochar: why is this a top-level API? I don't know what the point of this is

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   */
+  originToRoot(): Point;
+  /**
+   * Gets the container holding the visual elements of the Component.
+   *
+   * Will return undefined if the Component has not been anchored.
+   *
+   * @return {IContent} content selection for the Component
+   */
+  content(): IContent;
```
hellochar: Why do we have an IContent interface vs. using d3 selecting html

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  /**
+   * Gets the top-level element of the component
```
hellochar: nit: docs should explain the difference between `element()` and `content()`

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+    return this;
+  }
+  /**
+   * Given available space in pixels, returns the minimum width and height this Component will need.
+   *
+   * @param {number} availableWidth
+   * @param {number} availableHeight
+   * @returns {SpaceRequest}
+   */
+  public requestedSpace(availableWidth: number, availableHeight: number) {
```
hellochar: nit: might make sense to leave this abstract now since it feels like every component should independently decide its space needs

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+   * Computes and sets the size, position, and alignment of the Component from the specified values.
+   * If no parameters are supplied and the Component is a root node,
+   * they are inferred from the size of the Component's element.
+   *
+   * @param {Point} [origin] Origin of the space offered to the Component.
+   * @param {number} [availableWidth] Available width in pixels.
+   * @param {number} [availableHeight] Available height in pixels.
+   * @returns {IComponent} The calling Component.
+   */
+  abstract computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number): this;
```
hellochar: Is there any gain from re-declare the props from IComponent like this?

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  /**
+   * Removes a given CSS class from the Component.
+   *
+   * @param {string} cssClass The CSS class to remove.
+   * @returns {IComponent} The calling Component.
+   */
+  removeClass(cssClass: string): this;
+  /**
+   * Checks if the Component has a fixed width or if it grows to fill available space.
+   * Returns false by default on the base Component class.
```
hellochar: this comment makes no sense in the context of a generic IComponent interface

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+    return this;
+  }
+  /**
+   * Gets the parent ComponentContainer for this Component.
+   */
+  public parent(): ComponentContainer;
+  /**
+   * Sets the parent ComponentContainer for this Component.
+   * An error will be thrown if the parent does not contain this Component.
+   * Adding a Component to a ComponentContainer should be done
```
hellochar: These docs imply users should never call this by themselves? If so we should make it private/protected

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+    }
+    this._parent = parent;
+    return this;
+  }
+  /**
+   * @returns {Bounds} for the component in pixel space, where the topLeft
+   * represents the component's minimum x and y values and the bottomRight represents
+   * the component's maximum x and y values.
+   */
+  public bounds() {
```
hellochar: We should remove bounds() as a top-level property from IComponent since it's just a computed property

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+    this.detach();
+  }
+  /**
+   * Gets the width of the Component in pixels.
+   */
+  public width() {
+    return this._width;
+  }
+  /**
+   * Gets the height of the Component in pixels.
```
hellochar: nit: do we need to re-docs all these properties from IComponent?

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+  public origin() {
+    const { x, y } = this._origin;
+    return { x, y };
+  }
+  /**
+   * Gets the origin of the Component relative to the root
+   *
+   * @return {Point}
+   */
+  public originToRoot() {
```
hellochar: same as `bounds`(), if this is a computed property it doesn't need its own entry in IComponent

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,579 @@
...
+    while (ancestor != null) {
+      let ancestorOrigin = ancestor.origin();
+      origin.x += ancestorOrigin.x;
+      origin.y += ancestorOrigin.y;
+      ancestor = ancestor.parent();
+    }
+    return origin;
+  }
+
+  public root() {
```
hellochar: same as `bounds`(), if this is a computed property it doesn't need its own entry in IComponent

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    if (this._element != null) {
+      // reattach existing element
+      selection.appendChild(this._element);
+    } else {
+      this._element = document.createElement("div");
+      selection.appendChild(this._element);
+      this._setup();
+    }
+
```
hellochar: nit: newlines

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (this._resizeHandler != null) {
+      this._resizeHandler(size);
+    }
+
+    return this;
+  }
+
+  public renderImmediately() {
+    return this;
```
hellochar, line 91: Why? Shouldn't subclasses implement this?

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    if (this._element == null) {
+      return this._cssClasses.has(cssClass);
+    } else {
+      return this._element.className.split(" ").indexOf(cssClass) !== -1;
+    }
+  }
+
+  public addClass(cssClass: string) {
```
hellochar: anyway to DRY this with the original component.ts? The code is 90% the same

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    if (this._element == null) {
+      this._cssClasses.delete(cssClass);
+    } else {
+      const classNames = this._element.className.split(" ");
+      const indexOfClass = classNames.indexOf(cssClass);
+      if (indexOfClass > -1) {
+        classNames.splice(indexOfClass, 1);
+        this._element.className = classNames.join(" ");
```
hellochar: can't we still use d3 to do all this?

src/utils/clientToSVGTranslator.ts
```diff
@@ -3,83 +3,35 @@
...
-  private _measureRect: SVGElement;
 
   /**
    * Returns the ClientToSVGTranslator for the <svg> containing elem.
    * If one already exists on that <svg>, it will be returned; otherwise, a new one will be created.
    */
-  public static getTranslator(elem: SVGElement): ClientToSVGTranslator {
-    let svg = DOM.boundingSVG(elem);
+  public static getTranslator(component: IComponent<any>): Translator {
+    let svg = DOM.boundingSVG(component.element().node() as SVGElement);
```
hellochar: shouldn't we make the component be `IComponent<SVGElement>` or somehting then? Or am I misunderstanding the type parameter

src/utils/plottableElement.ts
```diff
@@ -0,0 +1,55 @@
+export interface IPlottableElement<D extends Element> {
+    appendChild(newChild: Node): Node;
+    getBoundingClientRect(): ClientRect;
+    setAttribute(name: string, value: string): this;
+    left(position: number): this;
+    top(position: number): this;
+}
+
+export abstract class PlottableElement<D extends Element> implements IPlottableElement<D> {
```
hellochar: same question about separating class and interface

hellochar (changes requested):

src/components/htmlComponent.ts
CalvinFernandez: Sure, they can implement this if they want. Because this is an abstract method, the basic HTMLComponent has to do _something_ and it doesn't need to actually render itself.

src/utils/plottableElement.ts
CalvinFernandez: I like keeping classes and their related interface in the same file .. not sure if we want to enforce a pattern of separating them.

src/components/abstractComponent.ts
CalvinFernandez: Yea I'm not sure ... I think its for moving components around the hierarchy (without destroying them) .. but I don't know when it happens or why we need it.

src/components/abstractComponent.ts
CalvinFernandez: Sure the rationale behind having the two is purely for sanitary reasons. I like separating definition from implementation as much as possible. 1) It makes for a more flexible API in the future (say you have some new crazy component with methods that you don't want in abstract component) at very little up front cost. 2) It keeps other APIs around the codebase a little more readable. Rather than typing out AbstractComponent<D> everywhere, you just type IComponent<D>. Additionally, at least in my mind, reading something with ICompnent<D> as a parameter or property presents slightly less cognitive strain than AbstractComponent<D>

src/components/abstractComponent.ts
CalvinFernandez: Ugh this was me trying to come up with a way of associating the correct translator type with the component type. The goal was for a third party to be able to, given a generic IComponent, determine the offset of that component in screen space. Specifically the mouse dispatcher needs to be able to determine the offset of the component in order to dispatch interactions correctly 

https://github.com/palantir/plottable/pull/3209/files#diff-b5b8bc221836e19ab1bae0224ffd72c2R53 


But I think we talked about a solution for making the translator a one-size-fits-all top level `svg` that should render this method obsolete.

src/components/abstractComponent.ts
CalvinFernandez: this was a hack that allowed me to preserve back-comp for as long as possible while implementing. Let me see if I can remove this now and just officially change the anchoring behavior

src/components/abstractComponent.ts
CalvinFernandez: I didn't want to expose d3's api to end users. but I think its okay I'll just put that back

src/components/abstractComponent.ts
CalvinFernandez: what do you mean?

src/components/abstractComponent.ts
CalvinFernandez: Agreed, this is just copied from the old code

src/components/abstractComponent.ts
CalvinFernandez: nah don't think so

src/components/abstractComponent.ts
hellochar: The `computeLayout` method is already defined through `IComponent` interface so it should automatically turn into an abstract method on this class right?

CalvinFernandez: no it needs to be defined

plottable.css
```diff
@@ -45,7 +45,7 @@ svg.plottable {
 }
 
 /* Root-level html components are relative, all others are absolute */
-div.plottable {
+.root.plottable {
```
hellochar: ".root" is likely to conflict with 3rd party styles that might also style ".root" in different ways; how about `plottable-root`

src/components/abstractComponent.ts
```diff
@@ -206,45 +203,33 @@ export interface IComponent<D> {
    */
   origin(): Point;
   /**
-   * Gets the origin of the Component relative to the root
-   *
-   * @return {Point}
-   */
-  originToRoot(): Point;
```
hellochar: Doesn't bp-charts use this endpoint? Will we be able to replace it easily

src/components/abstractComponent.ts
```diff
@@ -502,78 +408,29 @@ export abstract class AbstractComponent<D> implements IComponent<D> {
     this._parent = parent;
     return this;
   }
-  /**
-   * @returns {Bounds} for the component in pixel space, where the topLeft
-   * represents the component's minimum x and y values and the bottomRight represents
-   * the component's maximum x and y values.
-   */
-  public bounds() {
```
hellochar: Sorry I guess my original comment wasn't clear; we can still expose these as top-level Component methods if they make sense, but we just don't need to expose them in the interface directly since they have an implementation already.

src/components/htmlComponent.ts
```diff
@@ -13,51 +17,42 @@ import * as RenderController from "../core/renderController";
 import * as Utils from "../utils";
 
 export class HTMLComponent extends AbstractComponent<HTMLElement> {
```
hellochar: still not sold on this separation between HTMLComponent and SVGComponent; the code's like 90% the same. IMO only the leaf components that actually render stuff (like a bar plot) should know anything about svg vs. html

hellochar (changes requested):
Still not sold on this SVG vs HTML thing so high up. That just sets us up for duplicate code, maintenance burden, unnecessary restrictions, and edge cases later on...

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,436 @@
...
+import { Point, SpaceRequest, Bounds } from "../core/interfaces";
+import * as RenderController from "../core/renderController";
+import * as Utils from "../utils";
+import { Set } from "../utils/set";
+
+export type GenericComponentCallback<D> = (component: IComponent<D>) => void;
+
+export type IResizeHandler = (size: { height: number, width: number }) => void;
+
+export interface IComponent<D> {
```
themadcreator, line 19: Why D not T?

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,436 @@
...
+    return this;
+  }
+
+  public offDetach(callback: GenericComponentCallback<D>) {
+    this._onDetachCallbacks.delete(callback);
+    return this;
+  }
+
+  public parent(): ComponentContainer;
+  public parent(parent: ComponentContainer): this;
```
themadcreator, line 400: Since `ComponentContainer` is an HTMLComponent, does this mean we can't nest SVG components inside other SVG components?

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,436 @@
...
+import * as d3 from "d3";
+
+import { ComponentContainer } from "./componentContainer";
+
+import { Point, SpaceRequest, Bounds } from "../core/interfaces";
+import * as RenderController from "../core/renderController";
+import * as Utils from "../utils";
+import { Set } from "../utils/set";
+
+export type GenericComponentCallback<D> = (component: IComponent<D>) => void;
```
themadcreator, line 15: "Generic..." is redundant. It's generic since it has a template type.

themadcreator: Also the callback function below is a function but uses the interface "I.." prefix. Make these consistent with eachother and other function definitions. I'm not sure if prefix or no prefix is better.

src/components/abstractComponent.ts
```diff
@@ -0,0 +1,436 @@
...
+    }
+    return this;
+  }
+
+  public onResize(resizeHandler: IResizeHandler) {
+    this._resizeHandler = resizeHandler;
+    return this;
+  }
+
+  abstract renderImmediately(): this;
```
themadcreator, line 293: Group the abstract methods at the top so that it's clear which parts an extending classes need to implement.

themadcreator: I'm surprised this isn't part of the method order linter rules

src/components/group.ts
```diff
@@ -1,16 +1,11 @@
-/**
- * Copyright 2014-present Palantir Technologies
- * @license MIT
- */
-
 import { SpaceRequest, Point } from "../core/interfaces";
```
themadcreator, line 1: Removed the copyright header?

src/components/abstractComponent.ts
themadcreator: Also please add a comment to this indicating what this template type is. Is it the type of the container? HTMLElement vs SVGElement? Various instances of this type elsewhere in this PR simply have `IComponent<any>` so it's hard to tell what this is for.

src/components/htmlComponent.ts
themadcreator: If this is a satisfactory implementation, why not just use this as the default impl in `AbstractComponent` and then just have your `SVGComponent` override it?

src/components/htmlComponent.ts
```diff
@@ -0,0 +1,198 @@
...
+  public renderImmediately() {
+    return this;
+  }
+
+  public redraw() {
+    if (this._isAnchored && this._isSetup) {
+      if (this.parent() == null) {
+        this._scheduleComputeLayout();
+      } else {
+        this.parent().redraw();
```
themadcreator, line 99: so redraws on any descendant will always redraw the root?

src/components/svgComponent.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  private _boxes: d3.Selection<void>[] = [];
+  private _boxContainer: d3.Selection<void>;
+  private _rootSVG: d3.Selection<void>;
+  private _clipPathID: string;
+  private static _SAFARI_EVENT_BACKING_CLASS = "safari-event-backing";
+
+  public constructor() {
+    super();
+    this._cssClasses.add("component");
```
themadcreator, line 31: So HTMLComponents shouldn't get the "component" class?

src/components/svgComponent.ts
```diff
@@ -0,0 +1,354 @@
...
+    /**
+     * The user has the option of passing an SVG element directly
+     * to the anchor method, in which case we will render directly
+     * to that SVG.
+     */
+    const isTopLevelSVG = (<Element> selection.node()).nodeName.toLowerCase() === "svg";
+    if (isTopLevelSVG) {
+      this._rootSVG = selection;
+    } else {
+      this._rootSVG = selection.append("svg");
```
themadcreator, line 55: maybe i don't understand how "anchoring" works but does this mean that if you try to "anchor" an `SVGComponent` to a `<g>` inside another SVG it will append a `<svg>` inside that `<g>`?

src/components/svgComponent.ts
```diff
@@ -0,0 +1,354 @@
...
+
+import { AbstractComponent, IComponent } from "./abstractComponent";
+import { Point, SpaceRequest, Bounds } from "../core/interfaces";
+import * as RenderController from "../core/renderController";
+import * as Utils from "../utils";
+
+// HACKHACK replace with GenericComponentCallback in 3.0
+export type ComponentCallback = (component: SVGComponent) => void;
+
+export class SVGComponent extends AbstractComponent<d3.Selection<void>> {
```
themadcreator: proper camel case is SvgComponent

src/components/svgComponent.ts
```diff
@@ -0,0 +1,354 @@
...
+    if (isTopLevelSVG) {
+      this._rootSVG = selection;
+    } else {
+      this._rootSVG = selection.append("svg");
+    }
+
+    // visible overflow for firefox https://stackoverflow.com/questions/5926986/why-does-firefox-appear-to-truncate-embedded-svgs
+    this._rootSVG.style("overflow", "visible");
+
+    // HACKHACK: Safari fails to register events on the <svg> itself
```
themadcreator, line 61: not your fault but OUCH

src/components/svgComponent.ts
```diff
@@ -0,0 +1,354 @@
...
+        // see http://www.w3.org/TR/CSS21/visudet.html#block-replaced-width
+        // and http://www.w3.org/TR/CSS21/visudet.html#inline-replaced-height
+        if (this._rootSVG.attr("width") == null) {
+          this._rootSVG.attr("width", "100%");
+        }
+        if (this._rootSVG.attr("height") == null) {
+          this._rootSVG.attr("height", "100%");
+        }
+
+        let elem: HTMLScriptElement = (<HTMLScriptElement> this._rootSVG.node());
```
themadcreator, line 153: Why is this cast to an `HTMLScriptElement`? Shouldn't this be a `SVGElement`, or actually a `SVGSVGElement`

src/components/table.ts
```diff
@@ -1,14 +1,9 @@
-/**
- * Copyright 2014-present Palantir Technologies
- * @license MIT
- */
-
 import * as d3 from "d3";
```
themadcreator, line 1: Removed copyright header

themadcreator (changes requested):
Overall this looks great. I like the abstraction you've pulled out and it looks like the replacements elsewhere in the code where quite minimal.

I did have some minor requests as well as some questions. I want little more information about the generic templates used.

src/components/htmlComponent.ts
themadcreator: How about consolidating the common stuff in ElementComponent?

src/components/svgComponent.ts
themadcreator: and HtmlComponent

src/components/htmlComponent.ts
CalvinFernandez: yep, this mirrors the current behavior

src/components/abstractComponent.ts
CalvinFernandez: Yes we can no longer do that

themadcreator: Why can't the parent just be an IComponent?

src/components/abstractComponent.ts
CalvinFernandez: well if we want behavior to exist on a component then it needs to exist on the interface as that's the definition of the component

src/components/htmlComponent.ts
CalvinFernandez: sure

src/components/svgComponent.ts
CalvinFernandez: yes exactly, however this change makes it so that you can't nest svgs within svgs any more

src/components/svgComponent.ts
CalvinFernandez: probably I think this may have been a copy paste bug

src/components/abstractComponent.ts
CalvinFernandez: its more specific to say componentcontainer since IComponents don't always contain children.

src/components/htmlComponent.ts
CalvinFernandez: Ideally, yea you'd just say I want to render this to this type of container, but Plottable's notion of anchoring and space measuring makes it necessary to divide, for now, the two concerns. If we get rid of anchoring and computeLayout then we can probably throw away a lot of this  code. I think chasing perfection in this corner of the code is going to be better solved by simplifying the anchor / setup behavior down the line. 


That said, there really isn't _that_ much duplication here, so i get that it isn't 100% perfect but it isn't too egregious. If we really want to throw away the duplicate code we need to revisit how anchoring is thought of

hellochar:
Closing in favor of #3248

PR #3210, opened 2017-02-04
Migrate build tasks from grunt to NPM scripts

hellochar:
* Move appending a license, version replacement, and watching to npm/webpack.
* consolidate building src and test into just "build" which does src and test at once; this should be faster and more robust

TODO
- [x] fix livereload
- [x] look into [multiple outputs](https://github.com/webpack/webpack/issues/1189) for the webpack configuration
    * filed https://github.com/palantir/plottable/issues/3224
- [x] use webpack-dev-server instead of connect (this should actually fix livereload, as well as fixing cache staleness in connect)


package.json
```diff
@@ -11,15 +11,14 @@
   "main": "build/src/index.js",
   "typings": "build/src/index.d.ts",
   "scripts": {
-    "build-src": "npm-run-all compile-typescript webpack-plottable",
-    "build-test": "npm-run-all compile-typescript-test webpack-test",
+    "build": "npm-run-all build:tsc build:webpack",
+      "build:tsc": "tsc -p test/",
```
themadcreator: build:tsc only builds test? shouldn't this be build:test?

plottable.js
```diff
@@ -1,9 +1,8 @@
 /*!
-Plottable 2.9.1 (https://github.com/palantir/plottable)
-Copyright 2014-2017 Palantir Technologies
-Licensed under MIT (https://github.com/palantir/plottable/blob/master/LICENSE)
-*/
-
+ * Plottable 2.9.1 (https://github.com/palantir/plottable)
+ * Copyright 2014-2017 Palantir Technologies
+ * Licensed under MIT (https://github.com/palantir/plottable/blob/master/LICENSE)
+ */
```
themadcreator, line 5: oh yay!

webpack.config.js
```diff
@@ -0,0 +1,115 @@
...
+   * (a) Getting the plottable.js file. This is either from bower or from a web url.
+   * (b) Referencing the Plottable object/code. This is either globally or by declaring
+   *     a dependency on a module named "plottable" (both in AMD and CommonJS).
+   */
+  {
+    entry: "./build/src/index.js",
+    output: {
+      filename: "plottable.js",
+      // adds the UMD header to allow export to AMD, commonJS, or global
+      libraryTarget: "umd",
```
themadcreator, line 97: bower might actually work with this

themadcreator: maybe we could have One Last Version for bower

themadcreator (approved):
Just some naming opinions, but overall looks good

package.json
CalvinFernandez: space funk

CalvinFernandez (dismissed):

package.json
hellochar: building test transitively builds src because test references src

hellochar: the indentation was intentional to show sub-tasks (e.g. `build:tsc` is part of top-level `build`)

PR #3218, opened 2017-02-06
make quicktests/dev chart fill the page

hellochar:
lets us tests the chats in a variety of dimensions quickly

Also add an example testing long tick labels


themadcreator (approved):
Yeah, looks good!

PR #3220, opened 2017-02-06
reduce memory leaks in barPlot

hellochar:
calling `Plot.entities()` creates new Plot.PlotEntity objects on each
invocation; repeated calls (e.g. in a mousemove callback) will quickly
accumulate memory and cause unnecessary GCs.

Address this problem in barPlot's entityNearest and entitiesIntersecting by avoiding this.entities().


themadcreator (approved):
This is fantastic!! I'm so happy we're fixing this

PR #3223, opened 2017-02-07
Category axis doesn't consider offscreen labels when requesting space

hellochar:
Category Axis will now only render/consider the ticks that will actually be drawn. This fixes a bug where pan/zoomed category axes might have a lot of unnecessary whitespace.


adidahiya:
screencaps / gifs? is this smooth now?

src/axes/categoryAxis.ts
```diff
@@ -134,10 +134,10 @@ export class Category extends Axis<string> {
    * @param {Scales.Category} scale - The scale being downsampled. Defaults to this Axis' scale.
    * @return {DownsampleInfo} an object holding the resultant domain and new stepWidth.
    */
-  public getDownsampleInfo(scale: Scales.Category = <Scales.Category> this._scale): DownsampleInfo {
+  public getDownsampleInfo(scale: Scales.Category = <Scales.Category> this._scale, domain = scale.invertRange()): DownsampleInfo {
```
themadcreator, line 142: I didn't know you could reference previous args in later default args. Neat

src/axes/categoryAxis.ts
```diff
@@ -370,7 +367,6 @@ export class Category extends Axis<string> {
     // hide ticks and labels that overflow the axis
     this._showAllTickMarks();
     this._showAllTickLabels();
-    this._hideOverflowingTickLabels();
```
themadcreator, line 378: Why don't we need this anymore?

src/scales/categoryScale.ts
```diff
@@ -50,6 +50,19 @@ export class Category extends Scale<string, number> implements TransformableScal
     this._outerPadding = Category._convertToPlottableOuterPadding(0.5, d3InnerPadding);
   }
 
+  /**
+   * Return a clone of this category scale without any included values providers.
```
themadcreator: I'm glad you're encapsulating this into a method. Can you include in this comment some info about what **IS** included in the partial "clone" of this scale.

themadcreator (changes requested):
Just a comment update request

hellochar:
There's some jank for how this behaves when pan/zoom and downsampling are enabled:

Zoom suddenly stops because the axis grows:


Layout jankily jumps around because the specific labels being drawn (and therefor the height) change while panning:


Thoughts?

helenkg:
@hellochar I would assume once we do vertical labels and truncate them at fixed height, the height of individual labels won't change when zooming and panning?

adidahiya:
> once we do vertical labels and truncate them at fixed height

@helenkg this still won't fix the problem because some labels might be at the full height & truncated while others are not close to the truncation limit

@hellochar in general I'm not a fan of the behavior shown in the above gifs; what if you waited until pan/zoom completed before redrawing the layout?

hellochar:
@adidahiya not a fan of tying re-compute to panzoom interactions since that adds a lot of complexity. Let me try debouncing requestedSpace() and see how that feels

adidahiya:
@hellochar can you elaborate on the code complexity just a bit? I think it could work out really well actually from a user's perspective.

Also curious if @pkwi has any thoughts

hellochar:
throttling doesn't work. Also, with maxWidth and 90 degree angles, it still has some jank:



@pkwi thoughts here?

hellochar:
@adidahiya here's "wait until pan is finished to recalculate":

ezgif com-video-to-gif 9

Definitely feels better. The zoom issue is still there but I think I'm okay with leaving that in until people complain.

themadcreator:
Okay, i think i finally get how this works. I was just missing the fact that your pass in the value of `.invertRange()` as the domain. Looks good, and this should work really well with sheared ticks

themadcreator (approved):

themadcreator:
I'm gonna merge this because I want to pull these changes in to test sheared ticks. If there are still more changes to address the "jank", that can be a followup PR

PR #3226, opened 2017-02-09
Provide datum and dataset to label formatter

dallinregehr:
Continuing the work of #3180 as a separate PR to avoid conflicts with large codebase migration.

I’ve dropped index as a param because axis extent labels have no concept of index. In practice all I need is the datum and associated dataset anyway.


hellochar (dismissed):
Looks fine to me, @CalvinFernandez ?

dallinregehr:
The initial commit didn't correctly handle stacked bar plot totals which are computed from multiple datasets. To handle this use case I've updated the dataset param to be a datasets array.

src/plots/barPlot.ts
```diff
@@ -500,7 +500,7 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
 
       let barWidth = attrToProjector["width"](d, i, dataset);
       let barHeight = attrToProjector["height"](d, i, dataset);
-      let text = this._labelFormatter(valueAccessor(d, i, dataset));
+      let text = this._labelFormatter(valueAccessor(d, i, dataset), d, [dataset]);
```
hellochar, line 503: What are the semantics of the last argument now? barPlot allows multiple datasets and will draw all of them; this allows you to e.g. color bars by dataset. IMO we could pass the whole dataset array, and/or just the dataset associated with this datum, but passing an array holding just the single dataset is confusing to me

src/plots/stackedBarPlot.ts
```diff
@@ -148,7 +148,7 @@ export class StackedBar<X, Y> extends Bar<X, Y> {
       if (maximum.extent !== baselineValue) {
         // only draw sums for values not at the baseline
 
-        const text = this.labelFormatter()(maximum.extent);
+        const text = this.labelFormatter()(maximum.extent, maximum.axisValue, maximum.datasets);
```
hellochar, line 151: similarly - is there a way to get the specific dataset associated with this datum now?

hellochar (changes requested):

src/plots/barPlot.ts
dallinregehr: That would be ideal, yes. An array of datasets is needed because of one use case: formatting the label of stacked bar chart totals. In that case the value is not tied to a single dataset but rather all the datasets that have a value at that axisValue. For example, if the chart is displaying various metrics grouped by state name we need to know all the datasets that were used to compute the total for each state name.

Because of this the API is constrained to be `Dataset[]`. Unfortunately most formatting use cases are for points on a single dataset so we see `[dataset]` in various places.

src/plots/stackedBarPlot.ts
dallinregehr: As pointed out above, the extent is computed using multiple datasets. I've attached them to the `StackExtent` but they could be moved to `GenericStackedDatum<D>` instead. In that case, above this line I'd have to add something like:
```typescript
let extentDatasets: Dataset[] = [];
this._stackingResult.forEach((stackedDatum, dataset) => {
    extentDatasets.push(stackedDatum.get(maximum.axisValue).dataset);
});
```

hellochar:
Hey @dallinregehr we're going to close this since there's been no movement, feel free to resubmit if you'd like to push this through. I'm fine with the third argument being an array of datasets too

PR #3228, opened 2017-02-10
Tick Label Shear

themadcreator:
Adds `tickLabelShearAngle` property to `CategoryAxis` to enable shearing of long tick label text.


themadcreator:
Check it out!

https://142-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/angle_text

hellochar (approved):

hellochar:
Looks like there's some moments where the axis labels are all disappearing before downsampling takes effect:

image

image

image

themadcreator:
I wonder if the default downsample threshold is just a little bit too small

hellochar:
Yeah... I wonder if we should rework how downsampling works to generically detect when the text isn't drawn, versus using 15px always. Filed #3229

PR #3230, opened 2017-02-14
remove measurer reset

hellochar:
Causes awful perf. The only "gain" was that Category Axis can auto-respond to e.g. dynamic changes in CSS (or font loading, for instance) but this is a tiny benefit whose cost is too great (poor performance on all bar/grid charts). Technically a breaking change since users who were relying on the live-css ability before will not be supported anymore (we should explore another way to tell Plottable "refresh the cache")

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -908,9 +908,6 @@ var Component = (function () {
     /**
      * Causes the Component to re-layout and render.
      *
-     * This function should be called when a CSS change has occured that could
-     * influence the layout of the Component, such as changing the font size.
-     *
      * @returns {Component} The calling Component.
      */
     Component.prototype.redraw = function () {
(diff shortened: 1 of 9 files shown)


hellochar:
So there's one issue here - which is that now we're breaking the semantics of `.redraw()`, whose docs say:
```
  /**
   * Causes the Component to re-layout and render.
   *
   * This function should be called when a CSS change has occured that could
   * influence the layout of the Component, such as changing the font size.
   * ...
```

But `.redraw()` gets used all over the place - in particular, it's used to redraw the Component whenever any of its properties have changed. I'll move the "updates as a result of changing the font size" semantics into a new method.

CalvinFernandez (approved):
Just the one concern over the semantics of the naming. If you're happy with it though i am as well.

themadcreator (approved):

PR #3231, opened 2017-02-14
fix circle; unbreak npm consumers

hellochar:
npm is broken right now because build/src/core/version.js includes a reference to a "__VERSION__" property that doesn't exist (it gets string replaced by webpack in standalone). We fix this by adding the actual
version number.

This also fixes the circleCI release script to actually compile the code (oops)


Gruntfile.js
```diff
@@ -191,7 +191,7 @@ module.exports = function(grunt) {
   grunt.registerTask("release:minor", ["bump:minor", "dist-compile", "gitcommit:version"]);
   grunt.registerTask("release:major", ["bump:major", "dist-compile", "gitcommit:version"]);
 
-  grunt.registerTask("dist-compile", ["test", "uglify", "compress"]);
+  grunt.registerTask("dist-compile", ["test", "exec:npm:sed-version", "uglify", "compress"]);
```
adidahiya, line 194: this is awkward, we have an NPM task which calls a grunt task which then calls an NPM task? can you think of a better way to organize this...

hellochar: I know, we're in a temporary bout of madness as we shift to npm as our build system. The real fix here is to just remove grunt completely but we're holding off in lieu of higher priority 3.0 stuff

themadcreator:
heh, eww, but sure

themadcreator (approved):

PR #3234, opened 2017-02-14
category axis downsampling accounts for shearing

### LINKED ISSUE #3229: category axis downsampling breaks with sheared labels
http://jsfiddle.net/hellochar/gbdq5fq9/

Looks like there's some moments where the axis labels are all disappearing before downsampling takes effect:

image

image

image

Instead of always being at 15px, downsampling could detect poor text conditions by looking at whether the rendered ticks are actually degenerate (e.g. they all render to empty, or the like).

A simple fix could be e.g. to scale the 15px threshold by (1 / cos(shear))

hellochar:
also add a slider to the angle_text example

Fixes #3229


CalvinFernandez (approved):

themadcreator (approved):
👍

PR #3235, opened 2017-02-15
Update package semver for release 3.0.0-beta.1

themadcreator:


hellochar:
Run `grunt dist-compile` to update plottable.js, plottable.min.js, and plottable.zip and we should be good!

adidahiya (approved):
do we still need bower.json?

themadcreator:
We've announced (in this version's release notes) our intentions to remove support but we should give the community time to respond. Also, I'm not sure bower currently works with async require.js (though maybe this was fixed by full module structure)

PR #3241, opened 2017-02-21
no need to invert points if scale is category

leebyp:
As shown in the GIFs below, we get the wrong `nearestEntity` when a scale is `Category`. In that case, we don't need to do any conversion from pixel coordinates to data coordinates.

Before:
broken_nearest_entity

After:
working_nearest_entity

### DIFF
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -1,4 +1,3 @@
-import { AccessorScaleBinding } from "./commons";
 /**
  * Copyright 2014-present Palantir Technologies
  * @license MIT
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -3,10 +3,10 @@
  * @license MIT
  */
 
-import { Category } from "../scales/categoryScale";
 import { Dataset } from "../core/dataset";
 import { Accessor, Point } from "../core/interfaces";
 import * as Scales from "../scales";
+import { Category } from "../scales/categoryScale";
 import { Scale, ScaleCallback } from "../scales/scale";
 import * as Utils from "../utils";
 
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -437,14 +437,14 @@ export class XYPlot<X, Y> extends Plot {
    * @return {Point} Returns the point represented in data coordinates
    */
   protected _invertPixelPoint(point: Point): Point {
-    const xScale = this.x();
-    const yScale = this.y();
+    const xScale = this.x().scale;
+    const yScale = this.y().scale;
 
-    if (Scales.isNotComparable(xScale.scale, yScale.scale)) {
+    if (!xScale.isComparable(yScale) || !yScale.isComparable(xScale)) {
       return point;
     }
 
-    return { x: xScale.scale.invertedTransformation(point.x), y: yScale.scale.invertedTransformation(point.y) };
+    return { x: xScale.invertedTransformation(point.x), y: yScale.invertedTransformation(point.y) };
   }
 
   protected _pixelPoint(datum: any, index: number, dataset: Dataset): Point {
--- a/src/plots/xyPlot.ts
+++ b/src/plots/xyPlot.ts
@@ -1,10 +1,12 @@
+import { AccessorScaleBinding } from "./commons";
 /**
  * Copyright 2014-present Palantir Technologies
  * @license MIT
  */
 
-import { Accessor, Point } from "../core/interfaces";
+import { Category } from "../scales/categoryScale";
 import { Dataset } from "../core/dataset";
+import { Accessor, Point } from "../core/interfaces";
 import * as Scales from "../scales";
 import { Scale, ScaleCallback } from "../scales/scale";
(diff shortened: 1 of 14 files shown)


CalvinFernandez:
Why does inverting the categorical scale cause this issue ?

leebyp:
@CalvinFernandez Conceptually the domain is always constant for a categorical scale, and so will be the default `[0, 1]`. This throws off the measurements when finding the shortest distance if the other scale is a quantitative one since we use the actual data values. Using the pixel count gives us the "close values" in the intuitive sense (shown in the GIF).

src/plots/xyPlot.ts
```diff
@@ -432,13 +433,18 @@ export class XYPlot<X, Y> extends Plot {
...
+   * (if neither scale is Category)
    * @param {Point} point Representation of the point in pixel coordinates
    * @return {Point} Returns the point represented in data coordinates
    */
   protected _invertPixelPoint(point: Point): Point {
     const xScale = this.x();
     const yScale = this.y();
 
+    if (xScale.scale instanceof Category || yScale.scale instanceof Category) {
+      return point;
```
hellochar: nit: would be good to make this more general; seems a bit special casey. Any ideas?

hellochar (approved):

hellochar (changes requested):
Actually sorry can you add a unit test for this

leebyp:
@hellochar realised that they're actually a lot more edge cases where the comparison doesn't make sense. Any thoughts on the new implementation? I don't like that we have this hardcoded place with a list of scales - any other ideas?

I've put in simple unit tests - do you think there's benefit in doing a full plot test in `xyPlotTests` too and calling `nearestEntity` from there?

src/scales/index.ts
```diff
@@ -100,3 +103,23 @@ export function isTransformable(scale: any): scale is TransformableScale {
   return (scale instanceof QuantitativeScale ||
   scale instanceof Category);
 }
+
+/*
+ * Whether scale should be comparable intuitively.
+ */
+export function isNotComparable(scale1: any, scale2: any): boolean {
```
hellochar: We shouldn't structure it like this; the methods on the scales themselves should define how the comparison works. How did the comparison work correctly pre-2.2.0? What did we change that introduced the bug? We should tackle it there instead

hellochar (changes requested):

src/scales/index.ts
leebyp: Is the updated more what you have in mind? The boolean pieces are this way to keep the scales not know about each other.

The bug itself was introduced as part of a hidden feature change in this PR: https://github.com/palantir/plottable/commit/f8945f075aef249347df45922961e54da8c02c8a#diff-813b01dbeb729f9cc4dce897dff649bdR93

I can also just revert those lines if we don't actually want that feature.

src/scales/scale.ts
```diff
@@ -200,4 +200,15 @@ export class Scale<D, R> {
     this._autoDomainIfAutomaticMode();
     return this;
   }
+
+  /**
+   * Whether Scale is comparable intuitively. eg. Quantitative vs Category, Time vs Time.
```
hellochar, line 205: Sorry, I'm not clear what this means - intuitively I wouldn't imagine Scales to be comparable at all since they're functions. How does comparability relate to "find the right data point on hover"?

Also, was there a unit test for this? This is the actual behavior we care about so I'd prefer it's tested (or somewhere else up the responsibility chain)

src/plots/xyPlot.ts
```diff
@@ -431,15 +432,19 @@ export class XYPlot<X, Y> extends Plot {
...
    * @param {Point} point Representation of the point in pixel coordinates
    * @return {Point} Returns the point represented in data coordinates
    */
   protected _invertPixelPoint(point: Point): Point {
-    const xScale = this.x();
-    const yScale = this.y();
+    const xScale = this.x().scale;
+    const yScale = this.y().scale;
+
+    if (!xScale.isComparable(yScale) || !yScale.isComparable(xScale)) {
```
hellochar, line 443: This is misleading now; invertPixelPoint will now not always give you a point in pixel space. Can we either (a) do this check when we call _invertPixelPoint instead or (b) rename/change the docs of this method to more accurately portray what it's trying to do

src/scales/scale.ts
leebyp: @hellochar Scales represent an underlying data type (numbers, time, categories...) which can be comparable. There wasn't any reasoning in the original PR but I can imagine if the xAxis is linear and the yAxis logarithmic, then the "closest data point" is no longer just the pixel count.

There wasn't a unit test for this, but I can write one with a full blown chart once we decide that this is the right format / functionality we're going for.

On further thought, I think we should actually just introduce a flag in `entityNearest` on whether to use data space or pixel space. There's no way we can tell if they're comparable without the user data, and we should just allow the option for both. Thoughts?

thelastnode:
By default, I think pixel space makes the most sense, since mouse
interactions are a pretty common use case.

hellochar (changes requested):

PR #3242, opened 2017-02-22
d3 v4

### LINKED ISSUE #3091: Upgrade to D3V4
Are there any plans to upgrade Plottable to use d3 version 4?

hellochar:
Upgrade to d3@4.5.0 and svg-typewriter@2.0.0. This will open the door for many improvements down the line and allows us to take advantage of all the changes in v4, including their render-to-canvas feature of d3-shape. 

* I've declared a "SimpleSelection" type which hides most of the type complexities of d3v4's Selection interface which has 4(!) type parameters. In the future it'd be good to remove these and replace them with proper Selection<> interfaces (which should remove the need to cast `<Element> selection.node()` everywhere.

* `LinePlot.interpolator` renamed to `curve`. Add some back-compat mapping to allow d3v3-style "named attributes" since I didn't know how else to expose these without asking our users to use d3 directly; would like input here

* `triangleUp` renamed to just `triangle`; `triangleDown` removed; added `wye` and `star` shapes

* Stop accessing selection's internals (e.g. `selection[0][0]`) in favor of using `.nodes()`. This may have adverse performance impacts; when this is true we should rethink the need to use `.nodes()` at all

* Copy/paste the source of the d3-selection-multi since we use `.attrs()` all over the place and have to make sure it monkey-patches the correct objects. See https://github.com/d3/d3-selection-multi/issues/11 for more info

Fixes #3091


src/animators/easingAnimator.ts
```diff
@@ -4,9 +4,29 @@
...
+import { AttributeToAppliedProjector, SimpleSelection } from "../core/interfaces";
+
+/**
+ * Converts an easing mode string like "sin-in-out" to the corresponding function
+ * in the d3-ease module.
+ *
+ * @param easingMode
+ */
+function easingFnMapping(easingMode: string) {
+  // sin, [in, out]
```
themadcreator: stray comment

src/animators/easingAnimator.ts
```diff
@@ -4,9 +4,29 @@
...
+  const words = easingMode.split("-");
+  const capitalCaseWords = words.map((w) => w[0].toUpperCase() + w.slice(1));
+  const methodName = `ease${capitalCaseWords.join("")}`;
+  let easingFn: (t: number) => number = (d3Ease as any)[methodName];
```
themadcreator: avert your eyes! This might deserve a // HACK comment

src/animators/nullAnimator.ts
```diff
@@ -16,8 +16,10 @@ export class Null implements Animator {
     return 0;
   }
 
-  public animate(selection: d3.Selection<any>, attrToAppliedProjector: AttributeToAppliedProjector) {
-    return selection.attr(attrToAppliedProjector);
+  public animate(selection: SimpleSelection<any>, attrToAppliedProjector: AttributeToAppliedProjector): SimpleSelection<any> {
+    // coerce possibly external d3 instance into our own instance of d3 so we can use d3-selection-multi
+    selection = d3.selectAll<d3.BaseType, void>(selection.nodes());
```
themadcreator: is this still necessary after the mbostock exchange?

src/axes/numericAxis.ts
```diff
@@ -84,11 +84,7 @@ export class Numeric extends Axis<number> {
     let domain = scale.domain();
     let min = domain[0] <= domain[1] ? domain[0] : domain[1];
     let max = domain[0] >= domain[1] ? domain[0] : domain[1];
-    if (min === domain[0]) {
-      return scale.ticks().filter((i: number) => i >= min && i <= max);
-    } else {
-      return scale.ticks().filter((i: number) => i >= min && i <= max).reverse();
```
themadcreator, line 90: do we no longer need to reverse?

src/axes/numericAxis.ts
```diff
@@ -342,14 +341,8 @@ export class Numeric extends Axis<number> {
...
       let nextRect = rects[i + interval];
-      if (this.isHorizontal()) {
-        if (currRect.right + padding >= nextRect.left) {
-          return false;
-        }
-      } else {
-        if (currRect.top - padding <= nextRect.bottom) {
-          return false;
-        }
+      if (Utils.DOM.clientRectsOverlap(currRect, nextRect)) {
```
themadcreator: im not sure this matches the same logic, especially it doesn't consider the padding.

src/components/componentContainer.ts
```diff
@@ -19,7 +20,9 @@ export class ComponentContainer extends Component {
     this._detachCallback = (component: Component) => this.remove(component);
   }
 
-  public anchor(selection: d3.Selection<void>) {
+  public anchor(selection: SimpleSelection<void>) {
+    // coerce possibly external d3 instance into our own instance of d3 so we can use d3-selection-multi
```
themadcreator: instead of copy+pasting this comment around, maybe we should just make a utils method for this? that might also help us track all the places where we do this sort of coercion.

src/drawers/drawer.ts
```diff
@@ -168,8 +177,8 @@ export class Drawer {
   /**
    * Returns the D3 selection corresponding to the datum with the specified index.
    */
-  public selectionForIndex(index: number): d3.Selection<any> {
-    return d3.select(this.selection()[0][index]);
+  public selectionForIndex(index: number): SimpleSelection<any> {
```
themadcreator, line 176: the previous implementation would validate the selection before returning. are we sure we're okay with losing that?

src/index.ts
```diff
@@ -1,3 +1,8 @@
+// HACKHACK d3-selection-multi doesn't play well with default "d3" package in a
+// bundler environment (e.g. webpack) - see https://github.com/d3/d3-selection-multi/issues/11
+// we add it manually to the default "d3" bundle
+import "./utils/addD3SelectionMulti";
```
themadcreator, line 4: ewww

src/plots/linePlot.ts
```diff
@@ -27,8 +27,42 @@ type EdgeIntersections = {
...
+  "step-before": d3.curveStepBefore,
+  "step-after": d3.curveStepAfter,
+  "basis": d3.curveBasis,
+  "basis-open": d3.curveBasisOpen,
+  "basis-closed": d3.curveBasisClosed,
+  "bundle": d3.curveBundle,
+  "cardinal": d3.curveCardinal,
+  "cardinal-open": d3.curveCardinalOpen,
+  "cardinal-closed": d3.curveCardinalClosed,
+  "monotone": d3.curveMonotoneX,
```
themadcreator: This is so much more sane than that easing function hackery above. Can we do something like this for easing functions?

src/plots/plot.ts
```diff
@@ -249,8 +253,8 @@ export class Plot extends Component {
    * Updates the extents associated with each attribute, then autodomains all scales the Plot uses.
    */
   protected _updateExtents() {
-    this._attrBindings.forEach((attr) => this._updateExtentsForAttr(attr));
-    this._propertyExtents.forEach((property) => this._updateExtentsForProperty(property));
+    this._attrBindings.each((_, attr) => this._updateExtentsForAttr(attr));
+    this._propertyExtents.each((_, property) => this._updateExtentsForProperty(property));
```
themadcreator, line 257: well there goes my hopes of `import * as _ from "lodash"`

src/scales/categoryScale.ts
```diff
@@ -89,14 +89,17 @@ export class Category extends Scale<string, number> implements TransformableScal
    * @returns {string[]}
    */
   public invertRange(range: [number, number] = this.range()): string[] {
-    const rangeBand = this._d3Scale.rangeBand();
+    const rangeBand = this._d3Scale.bandwidth();
     // offset the domain by half the rangeBand such that we consider the
     // center of the bars
-    const domainStartNormalized = this.invertedTransformation(range[0]) - rangeBand / 2;
```
themadcreator: what about this half rangeBand offset? If it's no longer needed maybe we can remove the comment above

src/utils/domUtils.ts
```diff
@@ -77,34 +77,43 @@ export function elementHeight(element: Element) {
...
+    return 0;
   }
+}
 
-  y = (y == null) ? 0 : y;
-  transformMatrix.translate[0] = x;
-  transformMatrix.translate[1] = y;
-  selection.attr("transform", transformMatrix.toString());
-  return selection;
+const SCALE_REGEX = /scale\(\s*([-+]?[0-9]*\.?[0-9]+)\s*,?\s*([-+]?[0-9]*\.?[0-9]+)\s*\)/;
```
themadcreator: minor nit: `[0-9]` can be replaced by `\d`

themadcreator: also, scientific notation values like 1e10 might technically be possible

themadcreator (changes requested):
This PR is heroic.

I have some questions about some matching functionality, but I can't WAIT to merge this =)

src/animators/nullAnimator.ts
hellochar: I *think* so, if consumers misconfigure webpack (which is likely...), we might hold two copies of d3, in which case we'll need to convert

src/axes/numericAxis.ts
hellochar: Yeah the .reverse() was only to cater to the `_hasOverlapWithInterval` method, which I've now refactored to be direction agnostic, so no reverse needed

src/axes/numericAxis.ts
hellochar: Oops you're right, they did this for a reason...

src/drawers/drawer.ts
hellochar: What do you mean validate?

src/plots/plot.ts
hellochar: welp

src/drawers/drawer.ts
themadcreator: Previously it called this.selection() which does the `!this._cachedSelectionValid` check. Now, you're accessing the cache directly, in which case the thing you get out of it might not be valid.

src/utils/coerceD3.ts
```diff
@@ -0,0 +1,28 @@
...
+import * as d3 from "d3";
+
+/**
+ * Coerce possibly external d3 instance into our own instance of d3 so we can use d3-selection-multi.
+ * See https://github.com/d3/d3-selection-multi/issues/11 for why we have to do this.
+ *
+ * Any public facing API that accepts a d3 selection should first pass that user-supplied selection
+ * through here - this ensures all selection objects that go through the Plottable codebase are "vetted".
+ */
+export function coerceExternalD3(externalD3Selection: d3.Selection<any, any, any, any>): SimpleSelection<any> {
```
themadcreator, line 12: awesome thank you

themadcreator (changes requested):
Okay changes look great thanks. But I still do have a question about the invalid cached selection.

src/drawers/drawer.ts
hellochar: good catch, I didn't even see that

src/plots/plot.ts
leebyp: i've seen people do `...each((_binding, property) => ...` to get the best of both worlds

themadcreator (approved):

PR #3245, opened 2017-02-24
Nearest entity data / pixel space option

leebyp:
As per the threads in #3241 
- There should just be a flag of which "space" to use to find the nearest entity instead of trying to second guess what the consumer wants.
- Because of the reasoning for changes in #3159 for pan/zoom interaction, we're constrained in entities only having their datum and not their pixel position anymore.
- There are a bunch of places where we were actually comparing pixel `Points` or data `Points` in calculations eg. in `_entityVisibleOnPlot` - passing around `transformPoint` fixes it up (and just defaults to an identity function)

EDIT: instead of pushing transform functions through to do what we want, I've made a couple of changes
- plots own the transform and take care of the closest entity math, keeping the `entityStore` a generic
- use `_pixelPoint` which we should be overwriting in extended classes

### DIFF
--- a/quicktests/overlaying/tests/functional/nearestEntity.js
+++ b/quicktests/overlaying/tests/functional/nearestEntity.js
@@ -30,14 +30,10 @@ function makeData() {
     ];
 }
 
-
-// area, bar, clustered, line, pie, rectangle, scatter, segment, stackedarea,
-// stackedbar, waterfall
 function run(svg, data, Plottable) {
     "use strict";
 
     // TODO: fix entityNearest for pie charts, createPie(data[0])
-
     var chart = new Plottable.Components.Table([
         [createPlot(new Plottable.Plots.Line(), data[0]), createClusteredBar(data[1])],
         [createPlot(new Plottable.Plots.Scatter(), data[0]), createSegmentPlot()],
--- a/quicktests/overlaying/tests/functional/nearestEntity.js
+++ b/quicktests/overlaying/tests/functional/nearestEntity.js
@@ -1,0 +1,131 @@
+function makeData() {
+    "use strict";
+    return [
+        // single data set
+        [
+            {x: "Amazon", y: 0.1},
+            {x: "Apple", y: 0.30},
+            {x: "Facebook", y: 0.9},
+            {x: "Google", y: 0.30},
+            {x: "Intel", y: 0.94},
+            {x: "Microsoft", y: 0.29},
+            {x: "Twitter", y: 0.30},
+        ],
+        // multiple data set
+        [
+            [{name: "jon", y: 1, type: "q1"}, {name: "dan", y: 2, type: "q1"}, {name: "zoo", y: 1, type: "q1"}],
+            [{name: "jon", y: 2, type: "q2"}, {name: "dan", y: 4, type: "q2"}, {name: "zoo", y: 2, type: "q2"}],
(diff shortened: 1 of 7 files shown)


thelastnode:
This is blocking us from upgrading from 2.2.0, would love to have it merged :)

leebyp:
looks like nearest entity on segment plots and pie plots are broken already and should be fixed in another PR

quicktests/overlaying/tests/functional/nearestEntity.js
```diff
@@ -0,0 +1,207 @@
...
+
+    chart.renderTo(svg);
+}
+
+function createPlot(plot, data) {
+    var xScale = new Plottable.Scales.Category();
+    var yScale = new Plottable.Scales.Linear();
+
+    plot.addDataset(new Plottable.Dataset(data))
+        .x(function (d) { return d.x; }, xScale)
```
themadcreator, line 51: weird space, also you could probably just use an arrow function here

quicktests/overlaying/tests/functional/nearestEntity.js
```diff
@@ -0,0 +1,207 @@
+function makeData() {
```
themadcreator, line 1: where are the actual tests here?

themadcreator: Or is this just for interactive testing in the browser?

src/plots/plot.ts
```diff
@@ -574,24 +573,38 @@ export class Plot extends Component {
...
    * @returns {Plots.PlotEntity} The nearest PlotEntity, or undefined if no {Plots.PlotEntity} can be found.
    */
   public entityNearest(queryPoint: Point, bounds = this.bounds()): Plots.PlotEntity {
-    const nearest = this._getEntityStore().entityNearest(queryPoint, (entity: Plots.LightweightPlotEntity) => {
-      return this._entityVisibleOnPlot(entity, bounds);
+    let closestDistanceSquared = Infinity;
+    let closestPointEntity: Plots.LightweightPlotEntity;
+    this._getEntityStore().forEach((entity) => {
+      if (this._entityVisibleOnPlot(entity, bounds) !== false) {
+        const entityPoint = this._pixelPoint(entity.datum, entity.index, entity.dataset);
```
themadcreator, line 589: Okay, i'd like the change how this works slightly.  Plots.LightweightPlotEntity already has a `position` prop. But if that is not a pixel position, what is it? There are no docs on the prop.

If `position` is computed incorrectly then we should fix that. If it's a separate kind of "Point" (like not in screen space), then i vote that we add another prop to the lightweight entity for `pixelPosition`. Example see xyplot's `_buildLightweightPlotEntities` method.

themadcreator (changes requested):

quicktests/overlaying/tests/functional/nearestEntity.js
leebyp: yeah these are for manual in browser testing

src/plots/plot.ts
leebyp: #3159 from @CalvinFernandez specifically shies away from storing `position` as the pixel position for pan / zoom interactions. Position likely doesn't mean anything anymore, will look into that.

I haven't looked into experimenting with changing the `lightweightplotentities` and how it interacts with the pan / zoom piece but can do.

themadcreator: Wait, doesn't this line indicate that position **SHOULD** be the pixel point? https://github.com/palantir/plottable/pull/3159/files#diff-47fd4a46971263fdadcbb8b6b09117c1R472

PR #3247, opened 2017-03-02
Remove unused touch code in double click interaction

### LINKED ISSUE #3246: DoubleClickInteractions sometimes doesn't fire onDoubleClick
Repro steps:
- set the DoubleClickInteraction on a plot
- add a callback to the handler
- click once
- without moving the mouse, double click
- the double click interaction doesnt fire

Source of bug:
- the double click interaction is distinguishing between click and doubleclick using an internal enum, and time independent logic which can become out of sync if you click in the same position.

leebyp:
Fixes #3246 

- The `_handleDblClick` method is only hooked up to the `dblclick` event - the `mouseup` `mousedown` logic is not actually doing anything. Any combinations of `Interactions.Click` and `Interactions.DoubleClick` should be done on the consumer side.
- `dblclick` can only get fired from a mouse - the touchevents logic that we have won't happen in practice.
- also, passing through the events to the callbacks

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -1392,11 +1392,7 @@ var Drawer = (function () {
         return this;
     };
     Drawer.prototype.selection = function () {
-        if (!this._cachedSelectionValid) {
-            this._cachedSelection = this.renderArea().selectAll(this.selector());
-            this._cachedSelectionNodes = this._cachedSelection.nodes();
-            this._cachedSelectionValid = true;
-        }
+        this.maybeRefreshCache();
         return this._cachedSelection;
     };
     /**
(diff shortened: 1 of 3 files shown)


hellochar:
So we'll still need to do the timeout stuff in bp-charts? I don't want that, I think this single vs double click thing will be a problem for everyone so I'd like to fix it at this level. Some ideas:

* add an "onClick" to DoubleClickInteraction and add logic (through a timeout or whatever) to *not* fire it on the double click
* do that same thing but move it to ClickInteraction (and delete DoubleClickInteraction)

I just don't want bp-charts to have two Interactions (since then we'll have to do hacky de-dupe event detection on our side)

leebyp:
@hellochar I want to keep the interactions separate precisely because solutions that I've explored are hacky in one way or another. Timeouts are generally the accepted solution and we can expose a param to set the value, but the delay for single clicks that this incurs is just plain awful.

The single click + double click interaction is an anti-pattern in itself and should be avoided, I'm wary of introducing this in the _lower level_ Plottable library. I feel like it should be on the consumer to solve that should they choose to go down this path. Some people are ok with the delay, others are ok with the single clicks firing.

hellochar:
The "delay" isn't "delay the first single-click until after the second click"; it should be "don't fire more single-click events after the first one for this amount of time". It sounds like there was only one solution, which is to add a delay, so imo it makes sense to add that to the Plottable logic itself

leebyp:
> it should be "don't fire more single-click events after the first one for this amount of time"

This means that we can no longer safely use the `dblclick` event because the delay there is a machine setting, which can be _longer_ than the time set in Plottable. At that point we're left with mouseup and mousedown events (to cater for touch interactions) and not firing double click events from `dblclick` doesn't seem right to me, especially in cases where single clicks aren't actioned on.

PR #3248, opened 2017-03-02
HTML Components first pass

hellochar:
## HTML components:

Allows Plottable to use HTML in its Components, unblocking canvas renderer, better layout behaviors, better overflow behavior, better line-wrapping text, and other features that HTML implements. The major changes are:

* Components must be rendered to HTML elements (no SVG anymore, breaking change)
* Component's `.element()` is now a div, and `.content()` is an SVG
* use absolute CSS positioning for layout (and anchor children to the `.element()` div)
* translator and dispatcher code from #3209 (might be a breaking API change)
* almost all test code that used to create SVGs now creates divs.
* remove all clipPath related code and replace it with `overflow: hidden`

## Next steps:

Immediate next step is to tackle canvas renderer. Beyond that, we should consider removing the following concepts; I suspect they're obsolete in an HTML world:

* foreground() and background() elements
* the Component's “box” API and related box-container stuff
* the .bounding-box


src/plots/plot.ts
```diff
@@ -67,7 +67,22 @@ export class Plot extends Component {
     this.animator(Plots.Animator.RESET, new Animators.Null());
   }
 
-  public anchor(selection: SimpleSelection<void>) {
+  private _canvas: d3.Selection<HTMLCanvasElement, any, any, any>;
+  private _createCanvas() {
+    const element = this.element();
+    this._canvas = element.append<HTMLCanvasElement>("canvas");
```
CalvinFernandez: I dislike how plot now manages its own drawing context -- what if i want to have other things draw to canvas as well?

src/components/component.ts
```diff
@@ -146,11 +194,11 @@ export class Component {
     });
     this._cssClasses = new Utils.Set<string>();
 
-    this._backgroundContainer = this._element.append("g").classed("background-container", true);
+    this._backgroundContainer = this._element.append("svg").classed("background-container", true);
```
CalvinFernandez, line 194: So all components have an svg element?

CalvinFernandez (commented):
My thoughts on this -- I think this approach sacrifices clarity of abstraction for some code de duplication, which i don't believe actually gains us any clarity. It feels like you're just squishing together the concept of svg plot and canvas plot. They're pretty different things. 

Specifically, I dislike that all components have an SVG element, regardless of where they may eventually render. 

I also dislike how this approach conflates the `plot` concept and the `component` concept by making the plot control its own rendering context. 

I'm also concerned that the implementation of future components ( like a legend rewrite based on html or axis rewrite ) will be made less-clear because the components aren't broken apart cleanly. 

I would prefer a cleaner abstraction of concepts despite the sacrifice of some code duplication.

src/plots/plot.ts
hellochar: Then those things can create canvases. Eventually we can de-duplicate canvas management code with a mixin or a separate class. This is no different than how existing Components manage their own drawing areas/d3 selections

src/components/component.ts
hellochar: Eventually I'd like to move these away from Component and have a separate thing own SVG specific concepts (either a mixin, or introduce a new idea of a RenderTarget).

If you're worried about unused DOM nodes, we can lazily instantiate/add these when they're used.

hellochar:
> My thoughts on this -- I think this approach sacrifices clarity of abstraction for some code de duplication, which i don't believe actually gains us any clarity. It feels like you're just squishing together the concept of svg plot and canvas plot. They're pretty different things.

Can you explain what the differences/incompatibilities are? I feel this abstraction is cleaner than having three versions of everything (IPlot, SVGPlot, CanvasPlot) whose implementations are 90% the same (this is a big red flag for me - it means that the two things are conceptually the same thing. But we exploded the entire codebase with this "two things that are conceptually the same thing"). 

> I also dislike how this approach conflates the plot concept and the component concept by making the plot control its own rendering context.

Part of today Plottable's `Component` is to let subclasses do whatever they want on `content`; Plots have always controlled their own rendering context since they build their own _renderArea upon which everything is placed

> I'm also concerned that the implementation of future components ( like a legend rewrite based on html or axis rewrite ) will be made less-clear because the components aren't broken apart cleanly.

Can you give me an example? Implementing HTML Legend seems pretty straightforward - just place a `<table>` (or equivalent) on the `.element()` and forget SVG entirely?

src/drawers/drawer.ts
```diff
@@ -92,6 +102,7 @@ export class Drawer {
    * @param{AppliedDrawStep} step The step, how data should be drawn.
    */
   private _drawStep(step: Drawers.AppliedDrawStep) {
+    console.log("Drawing", step.attrToAppliedProjector);
```
themadcreator: lint

src/drawers/drawer.ts
```diff
@@ -158,6 +169,39 @@ export class Drawer {
...
+    let delay = 0;
+    appliedDrawSteps.forEach((drawStep, i) => {
+      Utils.Window.setTimeout(() => this._canvasDraw(data, drawStep, context), delay);
+      delay += drawStep.animator.totalTime(data.length);
+    });
+
+    return this;
+  }
+
+  protected abstract _canvasDraw(data: any[], step: Drawers.AppliedDrawStep, context: CanvasRenderingContext2D): void;
```
themadcreator: Does this mean that we expect to produce both SVG and Canvas renderers for everything that extends Drawer?

src/plots/plot.ts
```diff
@@ -67,7 +67,22 @@ export class Plot extends Component {
...
+  }
+
+  public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
+    super.computeLayout(origin, availableWidth, availableHeight);
+    this._canvas.attrs({
```
themadcreator: this is mutating state, shouldn't compute just return values?

src/plots/plot.ts
```diff
@@ -80,6 +95,7 @@ export class Plot extends Component {
     super._setup();
     this._renderArea = this.content().append("g").classed("render-area", true);
     this.datasets().forEach((dataset) => this._createNodesForDataset(dataset));
+    this._createCanvas();
```
themadcreator: These last 3 lines are the crux of what might be what calvin dislikes about this approach. It feels weird to me that we have one "renderArea" object created from svg group inside of "content", which may or may not be used for rendering. Then, we create a canvas which may or may not be used for rendering. I feel like this should be better abstracted and that a plot should have a renderer that will choose to attach one or the other itself. The plot won't know whether or not the renderer is using and SVG or canvas.

src/plots/plot.ts
```diff
@@ -479,11 +495,14 @@ export class Plot extends Component {
     let dataToDraw = this._getDataToDraw();
     let drawers = this._getDrawersInOrder();
 
-    this.datasets().forEach((ds, i) => drawers[i].draw(dataToDraw.get(ds), drawSteps));
+    // this.datasets().forEach((ds, i) => drawers[i].draw(dataToDraw.get(ds), drawSteps));
+    //
+    // let times = this.datasets().map((ds, i) => drawers[i].totalDrawTime(dataToDraw.get(ds), drawSteps));
+    // let maxTime = Utils.Math.max(times, 0);
+    // this._additionalPaint(maxTime);
```
themadcreator: i don't know what "additionalPaint" is or why it has a time argument. Is this difficult to abstract?

src/plots/plot.ts
hellochar: Yeah good idea - how about this:

* Instead of having both a _renderArea SVG and a _canvas, we introduce a RenderArea class or interface which abstracts the renderArea and canvas. The RenderArea has two implementations: SVGRenderArea and CanvasRenderArea.
* Plot passes the RenderArea its _element() div and the RenderArea adds and owns the SVG/Canvas DOM elements
* We expose this API on Plot by adding `plot.renderArea("canvas") / plot.renderArea("svg")`, which we internally convert into the appropriate RenderArea class and muck the DOM appropriately
* In Plot._paint() we call appropriate methods on the RenderArea interface to draw appropriately

This way we get a separate renderer API while still keeping just one Plot hierarchy and minimal code-breakage for all our plots. Thoughts? @CalvinFernandez

CalvinFernandez:
> Can you explain what the differences/incompatibilities are? I feel this abstraction is cleaner than having three versions of everything (IPlot, SVGPlot, CanvasPlot) whose implementations are 90% the same (this is a big red flag for me - it means that the two things are conceptually the same thing. But we exploded the entire codebase with this "two things that are conceptually the same thing"). 

First, there are only two things: SVGPlot and CanvasPlot. IPlot is just the definition that ties Plots in Plottable together. The separation of concerns is what will make us able to -- without much fuss -- implement a CSSPlot or a WebGL plot or a ( whatever the nextgen of rendering contexts happens to be )Plot without having to completely rewrite the library. 

And yes, of course they are similar concepts. What I'm driving at here is that the concept of a Plot should be data-layer only, completely agnostic of the rendering context. This is why the 90% of the code that they share gets eventually broken (in the next PR) into its own part of the code entirely.  https://github.com/palantir/plottable/blob/fac724da30f85a0d2db11df0ae423c3137f363a3/src/plots/baseLinePlot.ts. The HTMLComponent/SVGComponent are just responsible for defining and handling the rendering contexts of the different plot types. The remainder of the additional code added is boiler-plate bloat caused by lack of multiple inheritance in typescript. 

We want the same thing here, which is a data abstraction that gets rendered to whatever context the user wants. What I want to see more of is a deliberate separation of data and rendering concepts in Plottable. The conflation of Component with the data layer is how we got into this mess in the first place and a more deliberate separation of components keeps things conceptually clean and easily extendable. 

> Part of today Plottable's Component is to let subclasses do whatever they want on content; Plots have always controlled their own rendering context since they build their own _renderArea upon which everything is placed

Sure they do this but its wrong. A drawer should be the tool that manipulates the render area, not the plot. The reason plots still manipulate the render area is to draw labels. The label drawing code is a complete nightmare and makes it harder to read the code that is specific to plots. We need to move the label drawing behavior out of the Plot code into its own location. 


More than anything else, I think this solution gets us what we want but doesn't force us to rethink the overall architecture of the code. Currently, Plot is a very convoluted concept and the code reflects that. This only makes the Plot harder to read.

hellochar:
> First, there are only two things: SVGPlot and CanvasPlot. IPlot is just the definition that ties Plots in Plottable together. The separation of concerns is what will make us able to -- without much fuss -- implement a CSSPlot or a WebGL plot or a ( whatever the nextgen of rendering contexts happens to be )Plot without having to completely rewrite the library.

If we're focusing on rendering contexts then we should abstract that out, instead of abstracting the Plot out.

> And yes, of course they are similar concepts. What I'm driving at here is that the concept of a Plot should be data-layer only, completely agnostic of the rendering context. This is why the 90% of the code that they share gets eventually broken (in the next PR) into its own part of the code entirely. https://github.com/palantir/plottable/blob/fac724da30f85a0d2db11df0ae423c3137f363a3/src/plots/baseLinePlot.ts. The HTMLComponent/SVGComponent are just responsible for defining and handling the rendering contexts of the different plot types. The remainder of the additional code added is boiler-plate bloat caused by lack of multiple inheritance in typescript.

This sounds like we're using inheritance to express a "has-a" relationship. We should say "a Plot has a renderer" -- and once we say that then it doesn't make sense to separate HTMLPlot vs SVGPlot. It's like using inheritance to create CategoricalBarPlot vs. LinearBarPlot.

> We want the same thing here, which is a data abstraction that gets rendered to whatever context the user wants. What I want to see more of is a deliberate separation of data and rendering concepts in Plottable. The conflation of Component with the data layer is how we got into this mess in the first place and a more deliberate separation of components keeps things conceptually clean and easily extendable.

Agreed that data and renderer should be separate; how does splitting HTML vs SVG component achieve that separation?

> > Part of today Plottable's Component is to let subclasses do whatever they want on content; Plots have always controlled their own rendering context since they build their own _renderArea upon which everything is placed
> Sure they do this but its wrong. A drawer should be the tool that manipulates the render area, not the plot. The reason plots still manipulate the render area is to draw labels. The label drawing code is a complete nightmare and makes it harder to read the code that is specific to plots. We need to move the label drawing behavior out of the Plot code into its own location.

I agree that label drawing is broken and should be moved to the Drawer; can you explain how that refactor becomes cleaner in your version vs this one?

I think the thing I'm settled on now is to introduce a separate RenderArea concept and have Plots talk to that - see https://github.com/palantir/plottable/pull/3248#discussion_r104039033 , thoughts on that?

src/plots/plot.ts
CalvinFernandez: Why is this better?

CalvinFernandez:
> This sounds like we're using inheritance to express a "has-a" relationship. We should say "a Plot has a renderer" -- and once we say that then it doesn't make sense to separate HTMLPlot vs SVGPlot. It's like using inheritance to create CategoricalBarPlot vs. LinearBarPlot.

Call them something else then? Semantics aside, the idea is to separate the context from the plot

> Agreed that data and renderer should be separate; how does splitting HTML vs SVG component achieve that separation?

Please look over the follow up pr.

hellochar:
I looked it over; this is my gist of the follow up pr:

* baseX does all the actual logic and is passed a DrawerFactory 
* (svg)X and canvasX are bare-bones classes who just delegate to the _plot; the only difference is passing different arguments to _createPlot -- either creating a CanvasDrawer or an SVGDrawer (and same idea with the entityadapter stuff).
* The svg classes also do more things that canvas doesn't do yet (e.g. svgStackedAreaPlot adds a _baseline svg line, which I imagine canvas would have to re-implement)

So:
* if svgX and canvasX are near carbon copies of each other and always delegate to _plot, why not just make _plot the thing that users interact with? The only difference between the two is _createPlot parameters, so we can add a `.renderer("canvas")` option into Plot itself to switch between the parameters. 
* SVG does things that canvas doesn't - this is bad! Instead of having svgStackedAreaPlot add _baseline, we should have the RenderArea interface support the concept of adding lines, and then the SVGRenderArea and CanvasRenderArea implement them appropriately. We want to have a consistent behavior (as much as possible) between the renderers.

Also typescript does support multiple inheritance (see mixins in ts2.2), but we shouldn't use inheritance where "ownership" is more appropriate.

src/plots/plot.ts
hellochar: less code, less complexity in the class hierarchy, we conceptually only need to start thinking about "svg vs canvas" down at the drawer level and the rest of the interfaces aren't affected

plottable.css
```diff
@@ -39,14 +39,52 @@
   background-color: #06cccc; /* ROBINS_EGG_BLUE */
 }
 
-svg.plottable {
```
themadcreator, line 42: 🎉

src/components/component.ts
```diff
@@ -16,16 +16,46 @@ export type ComponentCallback = (component: Component) => void;
...
+   */
   private _foregroundContainer: SimpleSelection<void>;
-  protected _clipPathEnabled = false;
+  /**
+   * Subclasses should set this to true in their constructor to prevent content from overflowing.
+   */
+  protected _overflowHidden = false;
   private _resizeHandler: IResizeHandler;
-  private _origin: Point = { x: 0, y: 0 }; // Origin of the coordinate space for the Component.
+  private _origin: Point = { x: 0, y: 0 };
```
themadcreator: Origin relative to what? The parent of `_element`? Some other top level plottable container?

src/components/component.ts
```diff
@@ -65,38 +122,23 @@ export class Component {
...
-      const safariBacking = this._rootSVG.select(`.${Component._SAFARI_EVENT_BACKING_CLASS}`);
-      if (safariBacking.empty()) {
-        this._rootSVG.append("rect").classed(Component._SAFARI_EVENT_BACKING_CLASS, true).attrs({
-          x: 0,
-          y: 0,
-          width: "100%",
-          height: "100%",
-        }).style("opacity", 0);
-      }
+    if (this.parent() == null) {
```
themadcreator: so `this.parent() == null` means it's what was previously called a top level component?

src/components/component.ts
```diff
@@ -65,38 +122,23 @@ export class Component {
...
     }
 
-    this._isTopLevelComponent = (<Node> selection.node()).nodeName.toLowerCase() === "svg";
-
-    if (this._isTopLevelComponent) {
-      // svg node gets the "plottable" CSS class
-      this._rootSVG = selection;
-      this._rootSVG.classed("plottable", true);
-      // visible overflow for firefox https://stackoverflow.com/questions/5926986/why-does-firefox-appear-to-truncate-embedded-svgs
-      this._rootSVG.style("overflow", "visible");
```
themadcreator, line 81: A notable difference in rendering. Might regress issues

src/components/component.ts
themadcreator: +1 for moving this out of here ASAP

src/components/component.ts
```diff
@@ -190,21 +233,10 @@ export class Component {
     if (origin == null || availableWidth == null || availableHeight == null) {
       if (this._element == null) {
         throw new Error("anchor() must be called before computeLayout()");
-      } else if (this._isTopLevelComponent) {
-        // we are the root node, retrieve height/width from root SVG
+      } else if (this._rootElement != null) {
```
themadcreator, line 239: same check as `this.parent() != null` above? Maybe encapsulate this in a method so we don't do different checks in different places?

src/components/component.ts
```diff
@@ -274,9 +311,6 @@ export class Component {
    * Component, Table, and Group; render them immediately with .renderTo() instead.
    */
   public renderImmediately() {
-    if (this._clipPathEnabled) {
-      this._updateClipPath();
```
themadcreator, line 278: so without `overflow: hidden` and no clip path, are we going to have overflow issues?

src/components/component.ts
```diff
@@ -287,7 +321,7 @@ export class Component {
    */
   public redraw() {
     if (this._isAnchored && this._isSetup) {
-      if (this._isTopLevelComponent) {
+      if (this.parent() == null) {
```
themadcreator: another check to encapsulate

src/components/component.ts
```diff
@@ -308,24 +342,27 @@ export class Component {
...
       } else {
         selection = coerceExternalD3(element);
       }
-      if (!selection.node() || (<Node> selection.node()).nodeName.toLowerCase() !== "svg") {
-        throw new Error("Plottable requires a valid SVG to renderTo");
+      if (!selection.node() || selection.node().nodeName == null) {
+        throw new Error("Plottable requires a valid Element to renderTo");
+      }
+      if (selection.node().nodeName === "svg") {
+        throw new Error("Plottable 3.x can only renderTo an HTML component; pass a div instead!");
```
themadcreator: Since we're never going to update this comment maybe say "For Plottable 3.x or later, ..."

src/components/component.ts
```diff
@@ -407,22 +444,6 @@ export class Component {
...
 
-  private _generateClipPath() {
-    // The clip path will prevent content from overflowing its Component space.
-    this._clipPathID = Utils.DOM.generateUniqueClipPathId();
-    let clipPathParent = this._boxContainer.append("clipPath").attr("id", this._clipPathID);
-    this._addBox("clip-rect", clipPathParent);
-    this._updateClipPath();
-  }
-
-  private _updateClipPath() {
```
themadcreator, line 418: good riddance

src/components/component.ts
```diff
@@ -643,7 +675,7 @@ export class Component {
   }
 
   /**
-   * Gets a Selection containing a <g> that holds the visual elements of the Component.
+   * Gets the SVG that holds the visual elements of the Component.
```
themadcreator, line 685: will all components have this SVG in the future?

src/components/componentContainer.ts
```diff
@@ -21,10 +21,10 @@ export class ComponentContainer extends Component {
     this._detachCallback = (component: Component) => this.remove(component);
   }
 
-  public anchor(selection: SimpleSelection<void>) {
+  public anchor(selection: d3.Selection<HTMLElement, any, any, any>) {
```
themadcreator, line 24: this is a super verbose type that is especially used by anchor. maybe do a `export type AnchorElement = d3.selection...`

src/drawers/drawer.ts
```diff
@@ -12,6 +12,16 @@ import { AttributeToProjector, AttributeToAppliedProjector, SimpleSelection } fr
...
 
+/**
+ * A Drawer is responsible for actually committing the DrawSteps to the DOM. You first pass a renderArea
+ * to the Drawer, which is the root DOM node holding all the drawing elements. Subclasses set an _svgElementName
+ * which is an HTML/SVG tag name. Then you call .draw() with the DrawSteps to draw, and the Drawer will draw
+ * to the DOM by clearing old DOM elements, adding new DOM elements, and then passing those DOM elements to
+ * the animator, which will set the appropriate attributes on the DOM.
+ *
+ * "Drawing" in Plottable really means "making the DOM elements and their attributes correctly reflect
+ * the data being passed in".
```
themadcreator, line 23: helpful comment

src/utils/transformAwareTranslator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { Point } from "../";
+import { SimpleSelection } from "../core/interfaces";
+import { Component } from "../components/component";
+
+const _TRANSLATOR_KEY = "__Plottable_ClientTranslator";
+
+export function getTranslator(component: Component): Translator {
+  // The Translator works by first calculating the offset to root of the chart and then calculating
+  // the offset from the component to the root. It is imperative that the measureElement
```
themadcreator: which variable is `measureElement`?

src/utils/transformAwareTranslator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  return translator;
+}
+
+/**
+ * Applies position as a style and attribute to the svg element
+ * as the position of the element varies by the type of parent.
+ * When nested within an SVG, the attribute position is respected.
+ * When nested within an HTML, the style position is respected.
+ */
```
themadcreator, line 38: iiinteresting

src/utils/transformAwareTranslator.ts
```diff
@@ -0,0 +1,96 @@
...
+   */
+  public computePosition(clientX: number, clientY: number): Point {
+    // get the origin
+    move(this._measurementElement, 0, 0);
+
+    let mrBCR = (this._measurementElement.node() as HTMLElement).getBoundingClientRect();
+    let origin = { x: mrBCR.left, y: mrBCR.top };
+
+    // calculate the scale
+    let sampleDistance = 100;
```
themadcreator: extract to SAMPLE_DISTANCE const

themadcreator (commented):
I've got some questions and some minor requests. I focused mostly on `src` and skimmed the rest

themadcreator (changes requested):
Oh i had some requests

src/components/component.ts
hellochar: Yeah I feel that we shouldn't overflow the container by default since it's not expected/standard behavior; users can opt-in to that behavior by setting `overflow: visible` on the root themselves. But I can see the backcompat argument as well. Thoughts?

src/components/component.ts
hellochar: Yeah there's three equivalences:

`this.parent() == null` <-> `this._rootElement != null` (once it's been rendered) <-> this is the top-level component. I'll add a method `isRoot()`?

Aside: we should do a pass over the terminology. I've seen "top level component", "top level element", "root element", "root component", as well as `component._element` and "root DOM node"

src/components/component.ts
hellochar: No because the SVG spec says SVG elements are `overflow: hidden` by default.

Actually, Plottable 2.x set `overflow: visible` explicitly, and then restricted specific components from overflowing (the Plots, GuidelineLayer and SelectionBoxLayer). I've replicated that same behavior by replacing `_clipPathEnabled` with `_overflowHidden` which applies the `overflow: hidden` style onto the **content()** when true; the net behavior should be the same

src/components/component.ts
hellochar: No. There's a few futures:

1. Component lazily adds `content()` into the DOM only when it's referenced
2. Component no longer has a notion of "content()"; subclasses using SVG add SVG things themselves (this plus removal of foreground/background/boxContainer should totally rid Component of all SVG related things)
3. SVG manipulation itself is defined in its own set of classes and concepts, and SVG subclasses talk to that.

We should reach (3) sooner or later

src/components/componentContainer.ts
hellochar: The `_element` and the `_rootElement` use this as well; agreed this is janky. In general we need to revisit the Selection types everywhere and I don't want to band-aid over the situation more than it already is (in which we use SimpleSelection in most places which throws all typing away).

src/utils/transformAwareTranslator.ts
hellochar: I actually did some poking around and I think we can get rid of this whole thing by using `event.offsetX`/`event.offsetY`: https://jsfiddle.net/hellochar/3soLpj9f/

src/components/component.ts
themadcreator: I think it's okay to remove because i agree that we shouldn't hide on default. However, certain components may expect it. So mostly I just want to call it out and keep an eye out for overflow issues after this merges.

src/components/component.ts
themadcreator: Yeah (3) is what makes most sense to me

src/components/componentContainer.ts
themadcreator: fair

themadcreator (approved):

PR #3252, opened 2017-03-06
Refactor click interaction into one

leebyp:
@hellochar collating the two interactions into one as previously discussed!


hellochar (changes requested):
Do a self-review of your code after you push it up 😉  (in this case, you could catch the deleted plottable.js file)

hellochar (approved):

PR #3256, opened 2017-03-07
Revert to calculating nearest entity in pixel space

leebyp:
@hellochar @themadcreator 

Much better fix compared to #3245 and #3241 
- Revert the `position` to be the coordinates of the data point in pixel space
- Removes the weird `position information in the data space` that we had before
- Calculate `position` as required so that we don't lose perf when panning / zooming
- Added visual tests (play around with it!)
- _Note: pie and segment plots for another PR_


quicktests/overlaying/tests/functional/nearestEntity.js
```diff
@@ -0,0 +1,220 @@
...
+    new Plottable.Interactions.PanZoom(xScale, null).attachTo(plot);
+
+    return new Plottable.Components.Table([
+        [yAxis, plot],
+        [null, xAxis],
+        [null, title],
+    ]);
+}
+
+function createPie(data) {
```
hellochar, line 193: Is this in the example? I didn't see a pie chart when playing around with it

src/plots/plot.ts
```diff
@@ -451,9 +451,13 @@ export class Plot extends Component {
           return;
         }
 
+        const plot = this;
         lightweightPlotEntities.push({
           datum,
-          position,
+          get position() {
+            // only calculate position when needing to improve pan zoom performance #3159
+            return plot._pixelPoint.call(plot, datum, datumIndex, dataset);
```
hellochar, line 459: niiice

hellochar (changes requested):
Awesome, great example too. One final thing - can you point your local blueprint-chart's Plottable dependency to this change and test the new tooltip perf on stacked_barincreasing, and compare it to the old perf? Just want to ensure we didn't regress

quicktests/overlaying/tests/functional/nearestEntity.js
leebyp: yah there's a `TODO` at the top of the run block to make pie and segment plots work with nearest entity: filed #3257

leebyp:
@hellochar checked out the other repo and there's no noticeable difference

quicktests/overlaying/tests/functional/nearestEntity.js
hellochar: Oh is nearestEntity just broken on pie and segment plots currently?

hellochar (approved):
Sounds good!

quicktests/overlaying/tests/functional/nearestEntity.js
leebyp: yah - the logic and calculations are very different than the problem here, so im gonna punt for another PR

src/plots/plot.ts
themadcreator: Cool work around for perf. This will likely not save us a huge amount since many methods up-convert "lightweight" entities to plot entities and i think that copies position over.

Still, it'll work great for interactions that compute with lightweight entities

Does this affect the serializability of lightweight plot entities vis a vis JSON.stringify?

src/plots/plot.ts
```diff
@@ -547,7 +551,7 @@ export class Plot extends Component {
   protected _lightweightPlotEntityToPlotEntity(entity: Plots.LightweightPlotEntity) {
     let plotEntity: Plots.PlotEntity = {
       datum: entity.datum,
-      position: this._pixelPoint(entity.datum, entity.index, entity.dataset),
+      position: entity.position,
```
themadcreator, line 554: yeah this. This will invoke the `get` right?

src/plots/scatterPlot.ts
```diff
@@ -55,8 +55,7 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
         lightweightPlotEntity.index,
         lightweightPlotEntity.dataset);
 
-      // convert diameter into data space to be on the same scale as the scatter point position
-      lightweightPlotEntity.diameter = this._invertedPixelSize({ x: diameter, y: diameter });
+      lightweightPlotEntity.diameter = { x: diameter, y: diameter };
```
themadcreator: huh? why does diameter have x and y? what does that mean geometrically?

themadcreator (approved):
So much minus in this PR! Love it!

Just a question about serializability, but overall looks good. Thanks for functional test.

src/plots/plot.ts
leebyp: just tried this out, yup we won't be able to use that in the future =(

this guy pops up:
`Uncaught TypeError: Converting circular structure to JSON`

src/plots/plot.ts
leebyp: yup

src/plots/scatterPlot.ts
leebyp: @themadcreator It's the pixel size of the scatter point. Left a comment and updated so that it's just a number. On checking some more, exposed another bug with the calculation itself and opened a separate ticket to track, #3268.

PR #3258, opened 2017-03-08
Upgrade  to Typesettable@3.0.2

themadcreator:
The new typesettable includes performance updates (removed superfluous clippaths) as well as built-in support for canvas.

Since the usage is still all SVG, all plots should be identical.


hellochar (approved):
Looks pretty straightforward!

Would it be difficult to add an HTML context to typesettable to typeset <span> elements or the like?

themadcreator:
Since we do manual positioning of the text SVG elements, we probably have all we need to do HTML rendering. I'll give it a shot.

Issue #3266, opened 2017-03-09
Canvas renderer for plots


hellochar:
Implement rendering Plot visual primitives to canvas to improve performance of our plots. Instead of having one SVG element per visual primitive, draw all of them to one canvas element.

Questions:

1. How will we style selected elements? In SVG we'd use `Plottable.Entity.selection`, but there is no SVG element in Canvas
    1. possible solution: Make `Plottable.Entity` a real class that represents the visual primitive (possibly backed by a d3.selection) and expose APIs on there that let users do whatever they previously did with the d3.selection (e.g. set attrs)
    2. audit the uses for `.selection` and see what people actually use it for
2. How will animators work? The api is to accept a selection of visual primitives and animate them (e.g. with d3.transition())
3. How will events be effected? Instead of the individual visual primitive being the target, all mouse/touch events on the plot area will be on the canvas element
    1. The current click callback exposes `(p: Point, event: MouseEvent | TouchEvent)`; these should still be fine, we'll just have to call out that the event will look a bit different

Plan to implement:
- [x] get an MVP canvas renderer for rectangle (optimized for grid use case)
- [x] add canvas rendering for Line plot (optimized for timeseries use case)

UPDATE: these two have been moved out of 3.0.0 and will be tackled on their own
- Refactor/abstract the concept of "rendering" and how plots "use" renderers (#3309)
- Refactor Plottable.Entity to be a real class that conceptually represents the visual primitive (#3301)

hellochar:
Another option: use d3.Selection as our "data layer" objects (so we still use .attr() to set properties on detached drawing primitive DOM elements, and then write a translator that takes a list of these elements and renders them onto Canvas).

Pros: users can still use .selection.attr() to style individual elements
Cons: we lose some amount of perf due to still holding DOM elements instead of simple JS objects

edit: jsfiddle comparing setting attrs on 90k rect elements: https://jsfiddle.net/u1wr93vw/2/ . On my machine the results are:
```
In DOM:
1751ms, 760ms, 1091ms, 683ms, 777ms, 758ms, 709ms, 762ms, 749ms, 771ms

Detached:
1060ms, 499ms, 480ms, 469ms, 464ms, 458ms, 407ms, 423ms, 418ms, 414ms
```

Note that this isn't the full perf picture: In DOM has extra cost around browser layout, style recomputation, and paint. Detached will have extra cost in doing the translation. Probably not worth it honestly

PR #3267, opened 2017-03-09
MVP canvas renderer for Rectangle Drawer

hellochar:
This PR introduces:

* API to set renderer on all plots: `plot.renderer("canvas")`, or `plot.renderer("svg")` (the default). Internally we've only modified the fixed-pipeline feature set of Plot to have both a _renderArea and a _canvas. It'd be nice to have a generic “Renderer” interface; we should tackle this later.

* API to set renderer on drawer. Drawers still need to implement the actual "draw a projector onto the canvas", as a totally separate codepath from the animator's .attrs. This area feels wrong but I'm not sure what the right abstraction/cut is here.

* Working implementation for RectangleDrawer, which powers Rectangle and Bar Plot

Things that are broken:
* No "selection" in entities() for a canvas plot (so you can't e.g. highlight the thing you've clicked on right now)
* No animations
* deferred rendering + canvas renderer is broken

Tackles first point in #3266


src/drawers/drawer.ts
```diff
@@ -43,23 +47,41 @@ export class Drawer {
...
    */
   public renderArea(area: SimpleSelection<void>): this;
   public renderArea(area?: SimpleSelection<void>): any {
-    if (area == null) {
+    if (area === undefined) {
       return this._renderArea;
     }
     area = coerceExternalD3(area);
     this._renderArea = area;
+    this._canvas = null;
```
themadcreator, line 84: this is a strange side effect. Even with a comment, this could be troublesome if the canvas is set to null but not removed from the DOM, no?

test/drawers/drawerTests.ts
```diff
@@ -104,6 +105,49 @@ describe("Drawers", () => {
       });
     });
 
+    describe("canvas drawing", () => {
+      it("setting the renderArea/canvas clears the other one", () => {
```
themadcreator, line 109: Okay, all this mutual exclusion and setting to null is really a strong signal that this should be abstracted into separate implementations for canvas and svg.

There are so many checks where you do one thing for canvas and one thing for SVG and they're littered all over the code. This should instead by handled by separate implementations of a common interface. So `Drawer` could have a `DrawerCanvasContext` and `DrawerSvgContext` or  `CanvasDrawer` and `SvgDrawer` that implement `IDrawerDrawStepper` or something.

themadcreator (changes requested):
Most of my thoughts on this revolve around the need to abstract away the mutual exclusion of drawing methods.

themadcreator (approved):
Still have major reservations about the lack of abstraction, but not going to block.

themadcreator:
See this side branch for one approach to abstracting the SVG/canvas rendering components https://github.com/palantir/plottable/pull/3271

hellochar:
Plan is: merge my thing now and then we PR the refactor next week into develop

Issue #3260, opened 2017-03-09
look into removing foreground() and background() elements


hellochar:
Component adds a foreground() and background() svg element to every Component. This are not used in our own code and only exist as a convenience for users to add custom visuals before/after the content. In an HTML world this may not make sense anymore (or at least they shouldn't be SVG specific).

Look into removing them: see if our users use it and how they use it.

adidahiya:
We use this right now to render tooltips on grid plots, but it should be easy enough to migrate to an HTML-based implementation.

danielbaak:
I've tried following the tooltip tutorial (http://plottablejs.org/tutorials/tooltips/) with plottable v3.1.1. It doesn't work.

apparently the API of .foreground() has changed?

hellochar:
Hey @danielbaak our website and docs aren't up to date for 3.1.1 yet - see #3342. We haven't modified the .foreground() api yet though; what exactly goes wrong?

zmandel:
@hellochar Im also seeing issues with using .foreground(). Ive spent lots of time debugging this but still cant pinpoint why all my custom drawing (lines, crosshairs) no longer work, they all use similar code to this crosshair example, as in var crosshairContainer = plot.foreground().append("g") in http://plottablejs.org/examples/finance/

hellochar:
@zmandel can you file a separate issue with a repro of what you're seeing? I'll take a look :)

zmandel:
@hellochar thank you! https://github.com/palantir/plottable/issues/3382

PR #3272, opened 2017-03-09
Prepare release v3.0.0-beta.3

leebyp:
This is the third in a series of beta releases for the next major revision of Plottable, v3.0.0. Please read on for a list of changes in beta.3, or track the v3.0.0 milestone at https://github.com/palantir/plottable/milestone/64.

# HTML Components

Previously, Components only understood SVG elements and had to be rendered to an SVG. This release allows Components to work with generic HTML elements, opening the way for more improvements such as canvas rendering. This introduces the following breaking changes:

## [ACTION REQUIRED] `.renderTo()` only accepts HTML elements

Components now must `.renderTo()` an HTML element, **not an `<svg>`**. All users will need to modify the element they are passing to `.renderTo()` (or `.anchor()`) appropriately. 

```html
old:
<svg width="300" height="300"></svg>

new:
<div style="width: 300px; height: 300px"></div>
```

In SVG you could use the `width=` and `height=` attributes; in HTML please use CSS to size the chart appropriately.

### Other changes

We no longer apply `overflow: visible` to the renderTo target.

`Component.originToSVG()` renamed to `Component.originToRoot()`.

Added `Component.root()` and `Component.isRoot()` methods.

Added `Component.element()` which returns the root HTML Element for that given component.

The internal DOM hierarchy has changed; existing CSS selectors targetting Plottable elements may no longer work. The gist of the changes are:
1. `content/foreground/background/box-container` are now `<svg>`s instead of `<g>`s
2. Table and Group now place their children on the `.component` element instead of on the `.content`.

# Canvas Renderer

This release adds preliminary support for rendering Rectangle and Bar Plot's visual primitives (the `<rect>`s) onto a single Canvas element rather than many SVG elements.

Plots now expose a `plot.renderer("svg" | "canvas")` method to tell the Plot to render to SVG or to Canvas. Currently:

* Only rectangle plot and bar plot are supported, supporting the `stroke-width`, `stroke`, `fill`, and `opacity` attributes.
* No animations are supported
* All `PlotEntity` objects (from calling `entities()`, `entitiesAt()`, `entitiesIn()`, or `entityNearest()`) returned from plots using the canvas renderer will have a null `selection` property (so e.g. changing the color of a clicked bar in Canvas is not possible)

# API Changes

### Click Interaction
The `DoubleClick` interaction has been combined with `Click` and internally controls overlap between the two events.
- callbacks are registered with `onDoubleClick` and unregistered with `offDoubleClick`
- on a `dblclick` event, the single click callback will not be called

### Dispatchers
`Dispatchers.Mouse.getDispatcher` and `Dispatchers.touch.getDispatcher` should now be passed Components instead of SVG elements.

`eventInsideSVG` renamed to `eventInside` and now must be passed the Component to test as the first argument.

### Considering removing `foreground()` and `background()`
We are considering removing the `Component.foreground()` and `Component.background()` from our public API since they're only used as a 3rd party hook just in case users would like to do further modification of the chart. Let us know your thoughts at #3260.

# Bugfixes

* Fix `Entity.position` sometimes not giving the pixel position. 
* Fix `nearestEntity` comparing entities in data space, which didn't make sense for categorical axes.

# Enhancements

* Performance of drawing labels has been improved across the board.


hellochar:
# HTML Components

Previously, Components only understood SVG elements and had to be rendered to an SVG. This release allows Components to work with generic HTML elements, opening the way for more improvements such as canvas rendering. This introduces the following breaking changes:

## [ACTION REQUIRED] `.renderTo()` only accepts HTML elements

Components now must `.renderTo()` an HTML element, **not an `<svg>`**. All users will need to modify the element they are passing to `.renderTo()` (or `.anchor()`) appropriately. 

```html
old:
<svg width="300" height="300"></svg>

new:
<div style="width: 300px; height: 300px"></div>
```

In SVG you could use the `width=` and `height=` attributes; in HTML please use CSS to size the chart appropriately.

### Other changes

We no longer apply `overflow: visible` to the renderTo target.

`Component.originToSVG()` renamed to `Component.originToRoot()`.

Added `Component.root()` and `Component.isRoot()` methods.

Added `Component.element()` which returns the root HTML Element for that given component.

The internal DOM hierarchy has changed; existing CSS selectors targetting Plottable elements may no longer work. The gist of the changes are:
1. `content/foreground/background/box-container` are now `<svg>`s instead of `<g>`s
2. Table and Group now place their children on the `.component` element instead of on the `.content`.

### Dispatchers

`Dispatchers.Mouse.getDispatcher` and `Dispatchers.touch.getDispatcher` should now be passed Components instead of SVG elements.

`eventInsideSVG` renamed to `eventInside` and now must be passed the Component to test as the first argument.

## Considering removing `foreground()` and `background()`

We are considering removing the `Component.foreground()` and `Component.background()` from our public API since they're only used as a 3rd party hook just in case users would like to do further modification of the chart. Let us know your thoughts at #3260.

hellochar (changes requested):
New development: i'm gonna merge canvas renderer for rectangle drawer first

hellochar (approved):

PR #3276, opened 2017-03-10
enable linting

hellochar:
turn on specific rules for now; would be good to get on tslint:latest eventually


src/axes/axis.ts
```diff
@@ -360,13 +359,13 @@ export class Axis<D> extends Component {
...
 
-    let isHorizontal = this.isHorizontal();
+    const isHorizontal = this.isHorizontal();
     bindElements(this._annotationContainer.select(".annotation-line-container"), "line", Axis.ANNOTATION_LINE_CLASS)
       .attrs({
-        x1: isHorizontal ? positionF : secondaryPosition,
-        x2: isHorizontal ? positionF : offsetF,
-        y1: isHorizontal ? secondaryPosition : positionF,
-        y2: isHorizontal ? offsetF : positionF,
+        "x1": isHorizontal ? positionF : secondaryPosition,
```
adidahiya, line 365: quotes don't seem needed here, is this an auto-fixer bug?

Gruntfile.js
```diff
@@ -206,7 +189,7 @@ module.exports = function(grunt) {
     grunt.task.run(["watch"]);
   });
 
-  grunt.registerTask("lint", ["parallelize:tslint", "jscs", "eslint"]);
+  grunt.registerTask("lint", ["jscs", "eslint"]);
```
adidahiya, line 192: heads up you can also drop eslint completely now (separate PR). tslint supports linting JS files and it's simpler to just have 1 configuration / fewer dependencies

src/axes/axis.ts
hellochar: this is autofixer modified code so... probably? 😆

PR #3279, opened 2017-03-11
MVP canvas renderer for Line Drawer

hellochar:
Implementation of canvas rendering for Line Drawer. Supports stroke-width, stroke, and opacity attributes. 

TODO would be nice to deduplicate the d3.line logic between svg and canvas

Perf of drawing lines is much better; when zoomed in, perf is now dominated by croppedRendering doing a linear pass through all datapoints. 

Addresses second point of #3266

### DIFF
--- a/src/drawers/lineDrawer.ts
+++ b/src/drawers/lineDrawer.ts
@@ -12,14 +12,14 @@ import { SimpleSelection } from "../core/interfaces";
 import { AppliedDrawStep } from "./index";
 
 export class Line extends Drawer {
-  private _d3LineFactory: (dataset: Dataset) => d3.Line<any>;
+  private _d3LineFactory: () => d3.Line<any>;
 
   /**
    * @param dataset
    * @param _d3LineFactory A callback that gives this Line Drawer a d3.Line object which will be
    * used to draw with.
    */
-  constructor(dataset: Dataset, d3LineFactory: (dataset: Dataset) => d3.Line<any>) {
+  constructor(dataset: Dataset, d3LineFactory: () => d3.Line<any>) {
     super(dataset);
     this._d3LineFactory = d3LineFactory;
     this._className = "line";
@@ -35,10 +35,17 @@ export class Line extends Drawer {
     return d3.select(this.selection().node());
   }
 
+  /**
+   *
+   * @param data Data to draw. The data will be passed through the line factory in order to get applied
+   * onto the canvas.
+   * @param step
+   * @private
+   */
   protected _drawStepCanvas(data: any[][], step: AppliedDrawStep) {
     const context = this.canvas().node().getContext("2d");
 
-    const d3Line = this._d3LineFactory(this._dataset);
+    const d3Line = this._d3LineFactory();
 
     const attrToAppliedProjector = step.attrToAppliedProjector;
     const resolvedAttrs = Object.keys(attrToAppliedProjector).reduce((obj, attrName) => {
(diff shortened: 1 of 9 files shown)


themadcreator:
Dude. Nice quicktest example. Holy cow.

src/drawers/lineDrawer.ts
```diff
@@ -26,4 +34,45 @@ export class Line extends Drawer {
...
+   *
+   * @param data Data to draw. The data will be passed through the line factory in order to get applied
+   * onto the canvas.
+   * @param step
+   * @private
+   */
+  protected _drawStepCanvas(data: any[][], step: AppliedDrawStep) {
+    const context = this.canvas().node().getContext("2d");
+
+    const d3Line = this._d3LineFactory();
```
themadcreator: why even bother saving this off as a var since you only call this once?

src/drawers/lineDrawer.ts
```diff
@@ -26,4 +34,45 @@ export class Line extends Drawer {
...
+    if (resolvedAttrs["stroke-width"]) {
+      context.lineWidth = parseFloat(resolvedAttrs["stroke-width"]);
+    }
+
+    if (resolvedAttrs["stroke"]) {
+      const strokeColor = d3.color(resolvedAttrs["stroke"]);
+      if (resolvedAttrs["opacity"]) {
+        strokeColor.opacity = resolvedAttrs["opacity"];
+      }
+      context.strokeStyle = strokeColor.rgb().toString();
```
themadcreator, line 72: should this be rgba?

src/drawers/lineDrawer.ts
```diff
@@ -26,4 +34,45 @@ export class Line extends Drawer {
...
+
+    const d3Line = this._d3LineFactory();
+
+    const attrToAppliedProjector = step.attrToAppliedProjector;
+    const resolvedAttrs = Object.keys(attrToAppliedProjector).reduce((obj, attrName) => {
+      obj[attrName] = attrToAppliedProjector[attrName](data, 0);
+      return obj;
+    }, {} as { [key: string]: any | number | string });
+
+    context.save();
```
themadcreator, line 56: do you think we should get in the habit of doing a `try { ... } finally { }` block when we do save/restore? We don't really recover from exceptions that well, and it's not the worst to mess up the context since all you're doing is creating a path and fill/stroke-ing it. Thoughts?

themadcreator (commented):
Questions

src/drawers/lineDrawer.ts
hellochar: What would be the goal there? I'd be wary of the try/catch swallowing underlying issues in the code. Also I believe code inside a try/catch can't be optimized as well so I'm worried about putting that in the core drawer code

src/drawers/lineDrawer.ts
hellochar: .rgb() will actually deceivingly output an `rgba()` string

hellochar:
@themadcreator any further concerns? Or can we 🚢

src/drawers/lineDrawer.ts
themadcreator: not a catch, just a finally. it would still bubble the error

src/drawers/lineDrawer.ts
themadcreator: oh, brother, ok

src/drawers/lineDrawer.ts
themadcreator: but, not really a blocker or anything

themadcreator (approved):
:shipit:

PR #3280, opened 2017-03-11
specify @types/d3 as a dependency

### LINKED ISSUE #3278: @types/d3 should be a normal dependency
See https://github.com/Microsoft/types-publisher/issues/81

We're a library that exposes d3 types in our public API, so typescript users should get `@types/d3` when they install Plottable. Otherwise, they will `npm install plottable`, and then get type errors when the compiler comes across d3 types in the Plottable public API.

hellochar:
Fixes #3278

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "write-file-webpack-plugin": "3.4.2"
   },
   "dependencies": {
-    "@types/d3": "4.5.0",
+    "@types/d3": "^4.5.0",
     "d3": "^4.5.0",
     "typesettable": "^3.0.2"
   }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -66,6 +65,7 @@
     "write-file-webpack-plugin": "3.4.2"
   },
   "dependencies": {
+    "@types/d3": "4.5.0",
```
adidahiya: This now needs to be a semver range so that users can pick up fixes to typings. You can lock it down in plottable development with yarn.lock

adidahiya (dismissed):

themadcreator (approved):

PR #3282, opened 2017-03-11
Move to yarn

hellochar:
which attaches yarn to the current dependencies

After this goes in we should **stop using npm for our day-to-day developer commands** -- instead, use the yarn commands: https://yarnpkg.com/en/docs/cli/

Note that circleCI is still using npm because it's a bit harder to get that working correctly (see https://circleci.com/docs/1.0/yarn/)


adidahiya:
> Note that circleCI is still using npm because it's a bit harder to get that working correctly

No, please don't do this. Your dev environment should mirror CI. What's so hard about getting yarn working on Circle? https://github.com/palantir/tslint-react/blob/master/circle.yml

themadcreator:
Agreed with @adidahiya , please update circle.yml to use yarn. Even **Documentalist** does this =) https://github.com/palantir/documentalist/blob/master/circle.yml

themadcreator (changes requested):
Want more yarn

hellochar:
Ok updated for circle (and also found some more stuff in the README/gruntfile)

hellochar:
Turns out we're on phantom 1.x which doesn't play well with yarn. The real problem here is that our testing infra is quite old so I'm going to first upgrade that (probably just move to Jest - see #3285) and then we should have no issue running yarn on circleCI. Sound good?

themadcreator:
Ah dang, too bad about yarn. Jest doesn't get use off of phantom, though does it? Jest doesn't include a headless browser. It's more like a mocha replacement, right? I think mocha is doing just fine as our unit test framework.

themadcreator (approved):

hellochar:
@themadcreator Jest has a JSDOM implementation that runs on node. It also has mocks, coverage, `describe/beforeEach/afterEach`, and expect built in. I didn't look into this before but it actually looks like you can't run Jest tests in the browser directly which might be more of a roadblock for us. We'll need to dig a bit deeper to decide if it's the right choice, and if not then we should upgrade mocha/karma/phantomJS and fix yarn-on-circle that way

adidahiya:
Unfortunately I doubt you can move to Jest completely for plottable. jsdom doesn't have a browser layout or CSS engine, which is crucial for testing chart layouts. So you'll have to upgrade to phantomjs 2 but you might be able to migrate some of the simpler browser-independent tests to Jest.

PR #3283, opened 2017-03-14
tickLabelAt method on axis

leebyp:
- similar to `Plot`'s `entitiesAt`
- does only measuring in the direction of the axis sounds reasonable?


src/axes/axis.ts
```diff
@@ -104,6 +104,33 @@ export class Axis<D> extends Component {
...
+    // only measure in the direction of the axis.
+    const pointValue = this.isHorizontal() ? point.x : point.y;
+    const tickValues = this._getTickValues();
+
+    const axisScale = this._scale;
+    let tickLabel: D;
+    this._tickLabelContainer.selectAll<SVGGElement, any>("." + Axis.TICK_LABEL_CLASS)
+      .each(function (data: D, index: number) {
+        const labelWidth = this.getBBox().width;
+        const scaledTick = axisScale.scale(tickValues[index]);
```
hellochar: What's the difference between `data` and tickValues[index]?

src/axes/axis.ts
```diff
@@ -104,6 +104,33 @@ export class Axis<D> extends Component {
...
+  public tickLabelAt(point: Point): D {
+    // only measure in the direction of the axis.
+    const pointValue = this.isHorizontal() ? point.x : point.y;
+    const tickValues = this._getTickValues();
+
+    const axisScale = this._scale;
+    let tickLabel: D;
+    this._tickLabelContainer.selectAll<SVGGElement, any>("." + Axis.TICK_LABEL_CLASS)
+      .each(function (data: D, index: number) {
+        const labelWidth = this.getBBox().width;
```
hellochar: this should be `height` for vertical axes right? Also, `getBBox()` is an expensive operation (e.g. on the order of milliseconds sometimes) so we should avoid it if possible

quicktests/overlaying/tests/functional/axisLabelAt.js
```diff
@@ -0,0 +1,50 @@
...
+  var chart = new Plottable.Components.Table([
+                                            [null,  title],
+                                            [yAxis, clusteredPlot],
+                                            [null,  xAxis]
+                                            ]);
+
+  [xAxis, yAxis].forEach((axis) => {
+    var clickInteraction = new Plottable.Interactions.Click();
+    clickInteraction.onClick((point) => {
+        const label = JSON.stringify(axis.tickLabelAt(point));
```
hellochar: This sequence of events feels very roundabout:

1. user clicks
2. our click listener (from mouseDispatcher) creates a "point" object from the `event.clientX` and `event.clientY`
3. axis.tickLabelAt iterates through all the tick label DOM elements and linear scans to find which one was clicked on
4. We get the data corresponding to that label

Why can't we do:

1. user clicks
2. We get the tick label DOM element we clicked on from `event.target`
3. We get the data corresponding to that label

Actually maybe we can add both? `tickLabelAt` and `tickLabelInteractedWith`?

src/axes/axis.ts
```diff
@@ -104,6 +104,33 @@ export class Axis<D> extends Component {
     this._scale.offUpdate(this._rescaleCallback);
   }
 
+  /**
+   * Gets the tick label data at a particular point.
+   *
+   * @param {Point} point
+   */
+  public tickLabelAt(point: Point): D {
```
hellochar: this is misnamed - we're not getting the tick label, we're getting the data value corresponding to that tick label

test/axes/axisTests.ts
```diff
@@ -655,4 +655,39 @@ describe("Axis", () => {
...
+    });
+
+    it("calculates only in the direction of the axis", () => {
+      assert.equal(axis.tickLabelAt({ x: 60, y: 0}), 20);
+      assert.equal(axis.tickLabelAt({ x: 60, y: 50}), 20);
+      assert.equal(axis.tickLabelAt({ x: 60, y: 100}), 20);
+    });
+
+    it("returns undefined if point is outside of labels", () => {
+      assert.isUndefined(axis.tickLabelAt({ x: 10, y: 10 }));
```
hellochar: It might be valuable to repurpose `tickLabelAt` to allow clicks outside of labels (so the user can click anywhere in the area, not specifically on the label text, to get the corresponding data)

src/axes/axis.ts
```diff
@@ -104,6 +104,33 @@ export class Axis<D> extends Component {
...
+    // only measure in the direction of the axis.
+    const pointValue = this.isHorizontal() ? point.x : point.y;
+    const tickValues = this._getTickValues();
+
+    const axisScale = this._scale;
+    let tickLabel: D;
+    this._tickLabelContainer.selectAll<SVGGElement, any>("." + Axis.TICK_LABEL_CLASS)
+      .each(function (data: D, index: number) {
```
hellochar: nit: we can do binary search to find it in log n; although if we can avoid doing `getBBox()` at all we can probably not care about the difference

hellochar (changes requested):

src/axes/axis.ts
leebyp: `tickLabelDataAt`?

src/axes/axis.ts
leebyp: Ah yup, it should be!

Yeah we have to measure the width of the label itself to see if the click is inside the text. We can avoid the `getBbox` on all the ticks with
- scale the `tickValue`s
- find the closest label
- only calculate `getBbox` on that tick

test/axes/axisTests.ts
leebyp: can play around with the idea

quicktests/overlaying/tests/functional/axisLabelAt.js
leebyp: The second way means that we don't need any changes to Plottable, but also directs implementations of interactions to go down the path of manually checking classes / traversing the DOM.

I don't want to add `tickLabelInteractedWith` because `Plot`s have no sense of interactions and events right now - unless we mean to change that?

leebyp: Just tried using `event.target`. When a label spans two lines, `event.target` points to the element for one of the lines. We can go through the DOM to join the text back, but that seems unsatisfactory.

test/axes/axisTests.ts
leebyp: because labels have a width and height, perhaps another method `getDataAt` would make sense to not convolute the two separate albeit similar concepts

leebyp:
ah my original logic goes out of the window with angled text - will rework...

quicktests/overlaying/tests/functional/axisLabelAt.js
hellochar: But can we do `d3.select(event.target).data()` to get the corresponding data?

PR #3286, opened 2017-03-15
remove box api

### LINKED ISSUE #3261: Remove Component's "box" API
Component has a “box” API that simply builds a <g> element whose width/height is always 100% (aka it auto-updates in response to layout changes).

It's currently only used for two things:
* the _boundingBox, which is used by (a) axes to cut off overflowing axis labels
* the .background-fill, see #1731 (used by 3rd party users to style the background using CSS)

These behavior are trivial in an HTML world; look into if we can get rid of it

hellochar:
Fixes #3261 

# Breaking changes

.bounding-box removed - use .width() and .height() instead, or use .element() to get the DOM element that bounds the component

.background-fill removed - was put in so CSS users could add a background color to the chart. Instead, style the .component

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -34,10 +34,6 @@ export class Component {
    */
   private _content: SimpleSelection<void>;
   /**
-   * Used by axes to hide cut off axis labels.
-   */
-  protected _boundingBox: SimpleSelection<void>;
-  /**
    * Place more objects just behind this Component's Content by appending them to the _backgroundContainer.
    */
   private _backgroundContainer: SimpleSelection<void>;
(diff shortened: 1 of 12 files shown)


src/components/component.ts
```diff
@@ -76,26 +72,6 @@ export class Component {
...
-   * .background-fill - for the background container (unclear what it's use is)
-   * .bounding-box - this._boundingBox
-   *
-   * boxes get their width/height attributes updated in computeLayout.
-   *
-   * I think this API is to make an idea of a "100% width/height" box that could be
-   * useful in a variety of situations. But enumerating the three usages of it, it
-   * doesn't look like it's being used very much.
-   *
-   * TODO possily remove in HTML world
```
themadcreator, line 91: 🎉

test/globalInitialization.ts
```diff
@@ -20,8 +20,7 @@ before(() => {
     // HACKHACK #2122
     window.Pixel_CloseTo_Requirement = 2;
   } else if (TestMethods.isIE()) {
-    const FP_EPSILON = 1e-5;
-    window.Pixel_CloseTo_Requirement = 1.5 + FP_EPSILON;
```
themadcreator, line 24: oh, brother

themadcreator (approved):
Cool with me

test/globalInitialization.ts
hellochar: I know... this whole thing feels wrong to me

PR #3293, opened 2017-03-16
Add high scale loading example for testing

themadcreator:


hellochar (approved):
Nice example! So definitely does look possible to do "lazy data loading" already. Can we test it with fakeLoadData() being async/loading after a delay?

PR #3294, opened 2017-03-16
support deferred rendering on canvas

hellochar:
Also set deferred rendering timeout to 200ms

Deferred rendering is an existing feature that replaces "When the Scale changes, render everything immediately" with "when the scale changes, translate/scale the rendering SVG by the offset amount, and debounce further scale changes before rendering everything". This PR adds this scale/translating behavior to the canvas element

ezgif com-video-to-gif 10

### DIFF
--- a/plottable.js
+++ b/plottable.js
@@ -78,7 +78,7 @@ return /******/ (function(modules) { // webpackBootstrap
 /******/ 	__webpack_require__.p = "";
 /******/
 /******/ 	// Load entry module and return exports
-/******/ 	return __webpack_require__(__webpack_require__.s = 124);
+/******/ 	return __webpack_require__(__webpack_require__.s = 125);
 /******/ })
 /************************************************************************/
 /******/ ([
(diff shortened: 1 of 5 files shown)


plottable.js
```diff
@@ -2300,20 +2310,31 @@ var XYPlot = (function (_super) {
...
         var _scalingX = 1;
+        // the Y scale that has occurred since the last time we rendered
         var _scalingY = 1;
+        // the X scale's domain the last time we rendered
         var _lastSeenDomainX = [null, null];
+        // the Y scale's domain the last time we rendered
         var _lastSeenDomainY = [null, null];
         var _timeoutReference = 0;
-        var _deferredRenderingTimeout = 500;
+        var _deferredRenderingTimeout = 200;
```
themadcreator: extract this to const or static

plottable.js
```diff
@@ -2300,20 +2310,31 @@ var XYPlot = (function (_super) {
...
                 return;
             }
-            _this._renderArea.attr("transform", "translate(" + _deltaX + ", " + _deltaY + ")" +
-                "scale(" + _scalingX + ", " + _scalingY + ")");
+            _this._renderArea.attr("transform", "translate(" + _deltaX + ", " + _deltaY + ") scale(" + _scalingX + ", " + _scalingY + ")");
+            if (_this._canvas != null) {
+                _this._canvas.style("transform", "translate(" + _deltaX + "px, " + _deltaY + "px) scale(" + _scalingX + ", " + _scalingY + ")");
+            }
             clearTimeout(_timeoutReference);
             _timeoutReference = setTimeout(function () {
```
themadcreator, line 2338: since this is defered, should we do extra null checks and stuff to make sure this chart hasn't been removed from the DOM?

src/plots/xyPlot.ts
```diff
@@ -39,22 +39,32 @@ export class XYPlot<X, Y> extends Plot {
...
     let _scalingX = 1;
+    // the Y scale that has occurred since the last time we rendered
     let _scalingY = 1;
+    // the X scale's domain the last time we rendered
     let _lastSeenDomainX: X[] = [null, null];
+    // the Y scale's domain the last time we rendered
     let _lastSeenDomainY: Y[] = [null, null];
     let _timeoutReference = 0;
-    const _deferredRenderingTimeout = 500;
+    const _deferredRenderingTimeout = 200;
```
themadcreator: Same constant. Should this be configurable?

themadcreator (changes requested):
Thanks for adding more comments

plottable.js
hellochar: A destroyed plot will have empty datasets so nothing will be drawn; a detached dataset should (arguably) still draw since it's expected to be reattached to something. I think we're okay here? We already do a `_renderArea == null` check when we register, so I think the worse that happens is we render to a detached element once

themadcreator (approved):
Good to go!

PR #3295, opened 2017-03-17
scale canvas by devicePixelRatio

### LINKED ISSUE #3273: canvas is blurry on high DPI displays
Retina display (note: should click into them to prevent github's image 100% style from distorting the image):
</img>

Non-retina display:
</img>

Probably some solution like https://www.html5rocks.com/en/tutorials/canvas/hidpi/

hellochar:
Fixes #3273 

FYI this decreases perf by devicePixelRatio^2 times (so e.g. rendering all of canvas_line is a little painful on the macbook)

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -152,10 +152,12 @@ export class Plot extends Component {
   public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
     super.computeLayout(origin, availableWidth, availableHeight);
     if (this._canvas != null) {
-      // update canvas width/height; this will also clear the canvas of any drawn elements so we should
+      // update canvas width/height taking into account retina displays.
+      // This will also clear the canvas of any drawn elements so we should
       // be sure not to computeLayout() without a render() in the future.
-      this._canvas.attr("width", this.width());
-      this._canvas.attr("height", this.height());
+      this._canvas.attr("width", this.width() * window.devicePixelRatio);
+      this._canvas.attr("height", this.height() * window.devicePixelRatio);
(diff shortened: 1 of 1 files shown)


src/plots/plot.ts
```diff
@@ -152,10 +152,12 @@ export class Plot extends Component {
...
     super.computeLayout(origin, availableWidth, availableHeight);
     if (this._canvas != null) {
-      // update canvas width/height; this will also clear the canvas of any drawn elements so we should
+      // update canvas width/height taking into account retina displays.
+      // This will also clear the canvas of any drawn elements so we should
       // be sure not to computeLayout() without a render() in the future.
-      this._canvas.attr("width", this.width());
-      this._canvas.attr("height", this.height());
+      this._canvas.attr("width", this.width() * window.devicePixelRatio);
+      this._canvas.attr("height", this.height() * window.devicePixelRatio);
```
themadcreator, line 159: don't we have to set the `<canvas style="width:...">` attributes to make sure the canvas is actually set to the right screen size?

themadcreator: If we don't, won't subsequent calls to this method double the width/height again?

themadcreator (commented):
??

src/plots/plot.ts
hellochar: The width/height is set by the css:
```css
.plottable .plot-canvas {
  width: 100%;
  height: 100%;
}
```

Or would you feel more comfortable setting it explicitly here?

themadcreator: so long as this doesn't double the width/height every call, i'm happy

themadcreator (approved):

Issue #3291, opened 2017-03-16
canvas scatter plot


hellochar:
We have canvas implementations for Line and Rectangle; support Scatter as well

hellochar:
See also #3266 for generic thinking around canvas renderers

themadcreator:
This has been released

PR #3300, opened 2017-03-20
Scatter canvas renderer

leebyp:
#3291 

Perf gains! 10k points in GIF (canvas scatter plot on the right), will upload a clearer one when my internet's better =(

purrrrrrf


src/drawers/symbolDrawer.ts
```diff
@@ -3,16 +3,69 @@
...
+      context.save();
+
+      const { x, y } = resolvedAttrs["transform"];
+      context.translate(x, y);
+
+      context.beginPath();
+      d3Symbol(datum, index, this._dataset, context);
+      context.closePath();
+
+      if (resolvedAttrs["fill"]) {
```
hellochar: DRY this coloring logic with the other drawers

src/plots/scatterPlot.ts
```diff
@@ -133,6 +133,43 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+
+    return propertyToProjectors;
+  }
+
+  protected _constructTransformProjector() {
+    const xProjector = Plot._scaledAccessor(this.x());
+    const yProjector = Plot._scaledAccessor(this.y());
+    return (datum: any, index: number, dataset: Dataset) => {
+      return this.renderer() === "svg"
+        ? "translate(" + xProjector(datum, index, dataset) + "," + yProjector(datum, index, dataset) + ")"
```
hellochar: We should avoid doing if/else checks on this.renderer if possible. The attributes that define a symbol should exist totally separate from how it's rendered. Can we store the x and y in the AttributeToProjector itself and have the SymbolDrawer decide how to interpret it?

src/core/symbolFactories.ts
```diff
@@ -9,36 +9,71 @@ import * as d3 from "d3";
  * A SymbolFactory is a function that takes in a symbolSize which is the edge length of the render area
  * and returns a string representing the 'd' attribute of the resultant 'path' element
  */
-export type SymbolFactory = (symbolSize: number) => string;
+export type SymbolFactory = (symbolSize: number, context?: CanvasRenderingContext2D) => string;
```
hellochar: This doesn't make sense from an API perspective - what are the semantics of the "context" argument? Who's supposed to use it? Also this won't always return a string anymore.

Actually this whole file never made sense - you invoked a function to return a new SymbolFactory, which you then invoked again with a symbolSize to return a path d string. What do you think?

src/plots/scatterPlot.ts
```diff
@@ -133,6 +133,43 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+    const sizeProjector = Plot._scaledAccessor(this.size());
+    return (datum: any, index: number, dataset: Dataset) => {
+      return symbolProjector(datum, index, dataset)(sizeProjector(datum, index, dataset));
+    };
+  }
+
+  protected _d3SymbolFactory() {
+    const symbolProjector = Plot._scaledAccessor(this.symbol());
+    const sizeProjector = Plot._scaledAccessor(this.size());
+    return (datum: any, index: number, dataset: Dataset, context: CanvasRenderingContext2D) => {
```
hellochar: This should be DRY-ed with _constructSymbolProjector I think. The implementation is exactly the same except we pass the `context` here, which doesn't make sense - why are we caring about the canvas rendering context in this place? We should try to keep setting/manipulating the "context" only in the drawer

hellochar (dismissed):

src/core/symbolFactories.ts
leebyp: I was confused by the same, it looks like the logic is there to wrap around the size from the accessor so that it's in pixel space for the user, the math of which is dependent on the symbol itself, so is intertwined into the factory. That's what the comment at the top is trying to get at.

src/drawers/drawer.ts
```diff
@@ -28,6 +28,10 @@ export class Drawer {
   private _renderArea: SimpleSelection<void>;
   private _canvas: d3.Selection<HTMLCanvasElement, any, any, any>;
 
+  // Attributes to style canvas context
+  protected static _CANVAS_CONTEXT_ATTRIBUTES = {
+    strokeWidth: "stroke-width", stroke: "stroke", opacity: "opacity", fill: "fill",
```
leebyp: using a map to ensure compatibility with the method below

src/plots/scatterPlot.ts
```diff
@@ -133,6 +134,31 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+    return propertyToProjectors;
+  }
+
+  // When generate is true, callback returns generated symbol, otherwise returns symbol generator
+  protected _constructSymbolGenerator(generate: boolean) {
+    const symbolProjector = Plot._scaledAccessor(this.symbol());
+    const sizeProjector = Plot._scaledAccessor(this.size());
+    return (datum: any, index: number, dataset: Dataset) => {
+      const generator = symbolProjector(datum, index, dataset)(sizeProjector(datum, index, dataset));
+      return generate ? generator(null) : generator;
```
leebyp: this is cleaner but bugs me that it returns two different types depending on the param

themadcreator:
### [100K SCATTER SYMBOLS](https://349-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#basic/canvas_scatter)

src/drawers/symbolDrawer.ts
```diff
@@ -4,15 +4,125 @@
...
+      return (
+        x + size >= 0 && x - size <= width &&
+        y + size >= 0 && y - size <= height
+      );
+    };
+
+    // lazilly create offscreen buffer of modest size
+    if (this.buffer == null) {
+      this.buffer = new CanvasBuffer(10, 10, 1);
+      this.buffer.ctx.translate(5, 5);
```
leebyp: why the translation?

src/drawers/symbolDrawer.ts
```diff
@@ -4,15 +4,125 @@
...
+    let prevSymbolGenerator: any = null;
+    let prevSymbolSize: number = null;
+    let skipped = 0;
+    for(let index = 0; index < data.length; index++) {
+      const datum = data[index];
+
+      // check symbol is in viewport
+      const attrs = this._resolveAttributesSubset(projector, attrKeys, datum, index);
+      const symbolSize = this.sizeProjector()(datum, index, this._dataset);
+      if (!intersectsCanvasBounds(attrs["x"], attrs["y"], symbolSize * 2)) {
```
leebyp: should this be `/ 2`? since the symbol is drawn from the center of the point

leebyp: This point is set to `(0,0)` in the data

screen shot 2017-03-22 at 13 40 27

src/drawers/symbolDrawer.ts
```diff
@@ -4,15 +4,125 @@
...
+      }
+
+      // check attributes and symbol type
+      const attrsSame = this._attributesSame(prevAttrs, attrs, styleKeys);
+      const symbolGenerator = this.symbolProjector()(datum, index, this._dataset);
+      if (attrsSame && prevSymbolSize == symbolSize && prevSymbolGenerator == symbolGenerator) {
+        skipped++;
+      } else {
+        // make room for bigger symbol if needed
+        if (2*symbolSize > this.buffer.pixelWidth || 2*symbolSize > this.buffer.pixelHeight) {
```
leebyp: this and in other place - `symbolSize` is the FULL width and height of the symbol

leebyp (commented):
@themadcreator not sure if ready for review yet but left some initial feedback

src/drawers/symbolDrawer.ts
themadcreator: this is centers the coordinates since symbols are drawn centered on (0, 0)

src/drawers/symbolDrawer.ts
themadcreator: Ah, okay, I'll go through and update my math

hellochar:
Just a quick note on the 100k scatter points - small incremental zooms makes the points jitter about back and forth. It looks like there's no subpixel/anti-aliased drawing. Is there some rounding thing going on?

themadcreator:
Yeah, we `Math.floor` the pixels because we are copying from a canvas buffer. If we don't do that, we're likely to hit aliasing effects due to pixel interpolated. The jitter is at most 1 pixel per symbol, which i think is reasonable.

adidahiya:
It's still pretty slow to zoom when you show 10s of thousands of scatter points. I agree that it's not a good plot too look at to glean useful information, but I feel like we should still do something to improve the UX. Are we collapsing any points that overlap significantly? I guess you'd still have to spend time computing lots of position data in that case... is the suggested approach for 100K points then to defer loading of plot data until it's needed (aggregate and "page" data outside of plottable)?

leebyp:
We loop over all the data points in #3303. We can experiment with the same, though here we have the size and shape in the equation. Do we also want to turn on `deferredRending` as we do in line plots?

themadcreator:
We could also later explore using some datastructures to do fast viewport queries on the data to prevent full linear iteration. But that wouldn't need to change the API at all, so that is safe to do later

leebyp:
@themadcreator @hellochar conflicts with the drawer re-design resolved!

src/drawers/canvasBuffer.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    public canvas: HTMLCanvasElement;
+    public ctx: CanvasRenderingContext2D;
+    public pixelWidth: number;
+    public pixelHeight: number;
+
+    constructor(
+        public screenWidth: number,
+        public screenHeight: number,
+        public devicePixelRatio = 1,
```
hellochar: nit: this value can change (e.g. if the user drags a window from a non-retina into a retina display); we have no way of detecting when it changes so we should always pull from `window.devicePixelRatio` when rendering

src/drawers/canvasDrawer.ts
```diff
@@ -51,3 +52,41 @@ export class CanvasDrawer implements IDrawer {
...
   }
 }
+
+export const ContextStyleAttrs = {
+  strokeWidth: "stroke-width", stroke: "stroke", opacity: "opacity", fill: "fill",
+};
+
+export function resolveAttributesSubsetWithStyles(projector: AttributeToAppliedProjector, extraKeys: string[], datum: any, index: number) {
+  const attrKeys = Object.keys(ContextStyleAttrs).concat(extraKeys);
+  const attrs: {[key: string]: any} = {};
```
hellochar, line 62: You can actually get statically typed attrs key objects using Typescript's mapped types: see http://www.typescriptlang.org/play/#src=const%20ContextStyleAttrs%20%3D%20%7B%0A%20%20strokeWidth%3A%20%22stroke-width%22%2C%20stroke%3A%20%22stroke%22%2C%20opacity%3A%20%22opacity%22%2C%20fill%3A%20%22fill%22%2C%0A%7D%3B%0A%0Afunction%20resolveAttributesSubsetWithStyles%3CExtraKeyLiterals%20extends%20string%3E(extraKeys%3A%20ExtraKeyLiterals%5B%5D%2C%20datum%3A%20any%2C%20index%3A%20number)%20%7B%0A%20%20type%20AttributeNames%20%3D%20(keyof%20typeof%20ContextStyleAttrs)%20%7C%20ExtraKeyLiterals%3B%0A%20%20const%20attrKeys%20%3D%0A%20%20%20%20Object.keys(ContextStyleAttrs).concat(extraKeys)%20as%20AttributeNames%5B%5D%3B%0A%20%20const%20attrs%3A%20%7B%20%5BK%20in%20AttributeNames%5D%3F%3A%20any%20%7D%20%3D%20%7B%7D%3B%0A%20%20return%20attrs%3B%0A%7D%0A%0Avar%20k%20%3D%20resolveAttributesSubsetWithStyles(%5B%22x%22%2C%20%22y%22%2C%20%22width%22%2C%20%22height%22%5D%2C%20%7B%7D%2C%200)%3B

src/drawers/symbolDrawer.ts
```diff
@@ -3,10 +3,95 @@
...
-  constructor() {
-    super("path", "symbol");
-  }
+    constructor() {
+        super("path", "symbol");
+    }
+}
+
+// for testing use of buffer
+export let buffer: CanvasBuffer;
```
hellochar: we create one global buffer? What if we have multiple scatter plots?

src/drawers/symbolDrawer.ts
```diff
@@ -3,10 +3,95 @@
...
+                styleContext(bufferCtx, attrs);
+
+                // save the values that are in the buffer
+                prevSymbolGenerator = symbolGenerator;
+                prevSymbolSize = symbolSize;
+                prevAttrs = attrs;
+            }
+
+            // blit the buffer to the canvas
+            buffer.blitCenter(context, attrs["x"], attrs["y"]);
```
hellochar, line 71: How much faster did the blitting actually make scatterplots?

src/drawers/symbolDrawer.ts
```diff
@@ -3,10 +3,95 @@
...
+        }
+    };
+}
+
+function attributesSame(prevAttrs: any, attrs: any, attrKeys: string[]) {
```
hellochar: nit: name this `isAttributeValuesEqual`

src/core/symbolFactories.ts
```diff
@@ -4,41 +4,56 @@
...
  * A SymbolFactory is a function that takes in a symbolSize which is the edge length of the render area
- * and returns a string representing the 'd' attribute of the resultant 'path' element
+ * and returns a d3 symbol generator.
  */
-export type SymbolFactory = (symbolSize: number) => string;
+export type SymbolFactory = (symbolSize: number) => d3Shape.Symbol<any, any>;
 
 export function circle(): SymbolFactory {
-  return (symbolSize: number) => d3.symbol().type(d3.symbolCircle).size(Math.PI * Math.pow(symbolSize / 2, 2))(null);
+  return (symbolSize: number) => {
```
hellochar, line 16: We're using d3 incorrectly here. We should create one d3.symbol() *per series*, not one per datapoint. We're creating 100k d3.symbol() objects and using it once per iteration each.

hellochar (changes requested):

src/drawers/symbolDrawer.ts
adidahiya: agreed, module-level singletons always look sketchy

src/drawers/symbolDrawer.ts
themadcreator: like from unusable to usable. A lot faster

src/drawers/symbolDrawer.ts
themadcreator: the buffer is only used for copying the symbol over and over. I didn't want to create buffers for every combination of symbol+color+size, so instead i just re-use this one buffer if all those are the same as previous

src/drawers/canvasBuffer.ts
```diff
@@ -30,7 +30,7 @@ export class CanvasBuffer {
     constructor(
         public screenWidth: number,
         public screenHeight: number,
-        public devicePixelRatio = 1,
+        public devicePixelRatio = window.devicePixelRatio,
```
hellochar: this isn't enough - we should recalculate the pixelWidth/pixelHeight whenever we use it I think? See https://github.com/palantir/plottable/blob/d93bdd897c972c065da4d8f5917869ba1d61d92b/src/plots/plot.ts#L205-L205

hellochar (changes requested):
Delete built plottable.js file plz. Also thoughts on how we're using SymbolFactory?

hellochar (approved):

PR #3303, opened 2017-03-21
Collapse x coordinates in canvas line drawer

themadcreator:
Adds functionality to the line drawer that will collapse very dense paths (that have a lot of line segments overlapping at the same x coordinate).

This enables canvas line rendering scale up to 100K points and beyond.


src/drawers/lineDrawer.ts
```diff
@@ -75,4 +90,103 @@ export class Line extends Drawer {
...
+  private _collapseLineGeometry(ctx: CanvasRenderingContext2D, data: any[]) {
+    const d3Line = this._d3LineFactory();
+    const xFn = d3Line.x();
+    const yFn = d3Line.y();
+
+    // move to valid start
+    let i = 0;
+    for (; i <= data.length; ++i) {
+      const d = data[i];
+      if (d != null) {
```
hellochar: use the function returned by `d3Line.defined()` to determine whether the data point is valid or not

src/drawers/lineDrawer.ts
```diff
@@ -75,4 +90,103 @@ export class Line extends Drawer {
...
+      data: any[],
+      iStart: number,
+      xFn: d3Projector,
+      yFn: d3Projector,
+      bucket: ILineBucket,
+    ) {
+
+    for (let i = iStart; i <= data.length; ++i) {
+      const d = data[i];
+      if (d == null) {
```
hellochar: same comment about using defined()

src/drawers/lineDrawer.ts
```diff
@@ -75,4 +90,103 @@ export class Line extends Drawer {
...
+        continue;
+      }
+      const x = Math.floor(xFn(d, i, data));
```
hellochar: this won't work if we apply a transform to the context correct?

src/drawers/lineDrawer.ts
```diff
@@ -75,4 +90,103 @@ export class Line extends Drawer {
...
+      const y = yFn(d, i, data);
+
+      if (bucket.prevX != null) {
```
hellochar: can we separate the bucketing logic into a data transformation that takes in an array of data and spits out a smaller array of data, and then feed that transformed data into d3Line?

LinePlot also already does some existing data transformations - see `_filterCroppedRendering` and `_filterDownsampling` on linePlot.ts. Would be good to consolidate all the data transformation stuff into one place.

Actually wait a minute, the _filterDownsampling logic looks suspiciously like bucketing by slope? Is there some duplicated responsibility there?

hellochar (commented):
Questions about where the data transformation logic lives

src/drawers/lineDrawer.ts
themadcreator: this will still be an effective way to bucket.

However, if we use a scale transform, the "buckets" won't abut probperly, but are we actually ever doing that?

src/drawers/lineDrawer.ts
themadcreator: To your first point i sincerely doubt we'd get any advantage to use d3.line and we'd just causing extra GC for no reason.

themadcreator: And to your other points about the stuff in lineplot, it looks like those are totally different than this method. the first just removes items that are not in the viewport (a good thing to do i think) the second will only remove data points that are on the same slope, like if you have a straight line from x,y (0, 0) -> (100, 100) with 100 datapoints, this method will draw a single line between the endpoints. it's such a minor win im curious if it's ever even used.

It's not the same logic as the stuff in the line drawer because the line may go up and down an arbitrary number of times in each bucket -- i.e. the slopes do not need to be equal.

hellochar: I don't care much about d3.line, I mostly just want all our data transformation to be logically contained to one location (or to at least say why we have "data transformation in the Plot" as well as "data transformation in the Drawer")

Cool thanks for looking into the `_filterDownsampling` method. If no one's even using it maybe we can just remove/deprecate it?

src/drawers/lineDrawer.ts
hellochar: We scale the context to fix the retina blurriness

src/drawers/lineDrawer.ts
```diff
@@ -56,5 +69,108 @@ export function makeLineCanvasDrawStep(d3LineFactory: () => d3.Line<any>): Canva
...
+ */
+function _collapseLineGeometry(d3Line: d3.Line<any>, ctx: CanvasRenderingContext2D, data: any[]) {
+  const xFn = d3Line.x();
+  const yFn = d3Line.y();
+
+  // move to valid start
+  let i = 0;
+  for (; i <= data.length; ++i) {
+    const d = data[i];
+    if (d != null) {
```
hellochar: use `d3Line.defined()`:
```
const xFn = d3Line.x();
const yFn = d3Line.y();
const definedFn = d3Line.defined();

...
if (definedFn(d, i, data)) {
  point
}
```

src/plots/linePlot.ts
```diff
@@ -226,8 +226,36 @@ export class Line<X> extends XYPlot<X, number> {
     return this;
   }
 
+  /**
+   * Gets if collapseDenseLines is enabled
+   *
+   * When collapseDenseLines is enabled, vertical or nearly vertical line
+   * segments that have the same floored x coordinate will be bucketed then
+   * drawn, drastically reducing the render time of dense line plots like
+   * timeseries. Only applies to the high performance "canvas" drawer.
```
hellochar, line 244: If we extract out the data-transformation logic we can use this in normal SVG lines as well

hellochar (changes requested):
Getting close

src/utils/mathUtils.ts
```diff
@@ -71,7 +71,7 @@ export function isNaN(n: any) {
  * Numbers represented as strings do not pass this function
  */
 export function isValidNumber(n: any) {
-  return typeof n === "number" && !isNaN(n) && isFinite(n);
+  return typeof n === "number" && n - n < 1;
```
hellochar, line 74: Is this different from before?

hellochar (approved):

src/utils/mathUtils.ts
themadcreator: same effect, much faster. it's in the inner loop of rendering so i just want to make sure we're squeeze as much perf out as possible

PR #3304, opened 2017-03-22
drawer api refactor

hellochar:
Create a SVGDrawer and a CanvasDrawer, and have Drawer just proxy between them.


src/drawers/drawer.ts
```diff
@@ -5,238 +5,130 @@
...
-    let delay = 0;
-    drawSteps.forEach((drawStep, i) => {
-      delay += drawStep.animator.totalTime(data.length);
-    });
-
-    return delay;
+  public remove() {
+    if (this._svgRootG != null) {
+      this._svgRootG.remove();
+    }
```
themadcreator: should this remove that canvas as well?

src/drawers/drawer.ts
```diff
@@ -5,238 +5,130 @@
...
-    if (this._renderArea != null) {
-      this._bindSelectionData(data);
-      this._cachedSelectionValid = false;
-
-      let delay = 0;
-      appliedDrawSteps.forEach((drawStep, i) => {
-        Utils.Window.setTimeout(() => this._drawStep(drawStep), delay);
-        delay += drawStep.animator.totalTime(data.length);
-      });
+    if (this._svgRootG != null) {
```
themadcreator: maybe this should null check the _svgDrawer and the one below can check _canvasDraw

src/drawers/drawer.ts
```diff
@@ -5,238 +5,130 @@
...
   }
 
   /**
-   * Removes the Drawer and its renderArea
+   * Use Canvas rendering from now on. This will remove the <g> element created by useSVG, if it exists.
    */
-  public remove() {
-    if (this.renderArea() != null) {
-      this.renderArea().remove();
+  public useCanvas(canvas: d3.Selection<HTMLCanvasElement, any, any, any>) {
```
themadcreator: given this mutual exclusion, maybe the HTMLElement (svg or canvas) should be managed by the respective drawer object.

for example, below, there is the .selection() method but that only applies to svg drawers that have a svg comeponent.

src/drawers/drawer.ts
themadcreator: Also another opportunity to delegate to a common interface between the drawer types

```typescript
interface IDrawer {
    remove: () => void;
    ...
}
```

src/drawers/drawer.ts
```diff
@@ -5,238 +5,130 @@
...
-      if (this._renderArea != null) {
-        this._cachedSelection = this.renderArea().selectAll(this.selector());
-        this._cachedSelectionNodes = this._cachedSelection.nodes();
-      } else {
-        this._cachedSelection = null;
-        this._cachedSelectionNodes = null;
-      }
-      this._cachedSelectionValid = true;
+  public selectionForIndex(index: number): SimpleSelection<any> {
+    if (this._svgRootG != null) {
```
themadcreator: more checks that could be put on the SVGDrawer interface only

src/drawers/lineDrawer.ts
```diff
@@ -5,56 +5,40 @@
...
+ * used to draw with.
+ *
+ * TODO put the d3.Line into the attrToAppliedProjector directly
+ */
+export function makeLineCanvasDrawStep(d3LineFactory: () => d3.Line<any>): CanvasDrawStep {
+  return (context: CanvasRenderingContext2D, data: any[][], attrToAppliedProjector: AttributeToAppliedProjector) => {
+    const d3Line = d3LineFactory();
     const resolvedAttrs = Object.keys(attrToAppliedProjector).reduce((obj, attrName) => {
-      obj[attrName] = attrToAppliedProjector[attrName](data, 0);
+      if (attrName !== "d") {
```
themadcreator, line 36: heh, i bet this saves some time!

src/drawers/rectangleDrawer.ts
```diff
@@ -5,52 +5,54 @@
...
+  data: any[],
+  attrToAppliedProjector: AttributeToAppliedProjector) => {
+  data.forEach((point, index) => {
+    const resolvedAttrs = Object.keys(attrToAppliedProjector).reduce((obj, attrName) => {
+      obj[attrName] = attrToAppliedProjector[attrName](point, index);
+      return obj;
+    }, {} as { [key: string]: any | number | string });
+
+    context.beginPath();
+    context.rect(resolvedAttrs["x"], resolvedAttrs["y"], resolvedAttrs["width"], resolvedAttrs["height"]);
```
themadcreator, line 30: i believe in the scatter plot PR ben has pulled this out into a method that might be used by multiple canvas renders since many of them will want to set the styles from attrs.

themadcreator (commented):
Okay, this is pretty tremendous! I'm excited about this refactor but i have some thoughts, lemme know what you think

hellochar:
@themadcreator I couldn't make the CanvasDrawer own creating/destroying the canvas element since we have one drawer per series, and 50 full-sized canvas elements is probably a bad idea. But I was able to build an interface for both of them which cleans stuff up, lemme know what you thnk!

src/drawers/rectangleDrawer.ts
leebyp: Hehe I toyed with this exact idea of extending from separate `SVGDrawer`s and `CanvasDrawer`s when I was implementing the scatter plot (though not as fleshed out of course...) but wanted to defer till after getting the MVP out. If we're close here, I'd be happy to rebase.

src/drawers/lineDrawer.ts
leebyp: related to the other thread of parsing out some of the "stylistic attribs", some differentiation of which attributes need to be evaluated in the draw steps is an easy perf win

src/drawers/lineDrawer.ts
```diff
@@ -5,56 +5,40 @@
...
-
-    const attrToAppliedProjector = step.attrToAppliedProjector;
+/**
+ * @param d3LineFactory A callback that gives this Line Drawer a d3.Line object which will be
+ * used to draw with.
+ *
+ * TODO put the d3.Line into the attrToAppliedProjector directly
+ */
+export function makeLineCanvasDrawStep(d3LineFactory: () => d3.Line<any>): CanvasDrawStep {
+  return (context: CanvasRenderingContext2D, data: any[][], attrToAppliedProjector: AttributeToAppliedProjector) => {
```
leebyp, line 33: for consistency in getting at what's going on here, make this also a `const` function as `RectangleCanvasDrawStep`?

src/drawers/lineDrawer.ts
leebyp: either in the drawer or drawsteps

src/drawers/canvasDrawer.ts
```diff
@@ -0,0 +1,53 @@
...
+  // public for testing
+  public getDrawStep() {
+    return this._drawStep;
+  }
+
+  public draw(data: any[], appliedDrawSteps: AppliedDrawStep[]) {
+    // don't support animations for now; just draw the last draw step immediately
+    const lastDrawStep = appliedDrawSteps[appliedDrawSteps.length - 1];
+    this._context.save();
+    this._drawStep(this._context, data, lastDrawStep.attrToAppliedProjector);
```
leebyp, line 38: do we still need the `setTimeout`? the svgDrawer is still using it

src/drawers/rectangleDrawer.ts
```diff
@@ -5,52 +5,49 @@
...
-
-  constructor(dataset: Dataset) {
-    super(dataset);
-    this._svgElementName = "rect";
+import { AttributeToAppliedProjector } from "../core/interfaces";
+import { CanvasDrawStep } from "./canvasDrawer";
+import { SVGDrawer } from "./svgDrawer";
+
+export class RectangleSVGDrawer extends SVGDrawer {
+  constructor(private _rootClassName: string = null) {
```
leebyp: `private _rootClassName: string = ""` will remove the need for the null checks?

src/drawers/lineDrawer.ts
hellochar: This is actually different from `RectangleCanvasDrawStep` though - that one is the same everywhere, but for line the draw logic changes depending on the d3LineFactory (although eventually we should remove that)

src/drawers/canvasDrawer.ts
hellochar: Funnily enough - `Window.Utils.setTimeout(..., 0)` just invokes the function directly, so this is exactly the same still. I'm not quite sure why we even need our own setTimeout

src/drawers/lineDrawer.ts
hellochar: Yep, eventually we should make the LineSVGDrawer put the "d" attribute, instead of the LinePlot creating the "d" attribute

src/drawers/drawer.ts
```diff
@@ -5,130 +5,96 @@
...
-function applyProjectorsWithDataset(attrToProjector: AttributeToProjector, dataset: Dataset): AttributeToAppliedProjector {
-  const modifiedAttrToProjector: AttributeToAppliedProjector = {};
-  Object.keys(attrToProjector).forEach((attr: string) => {
-    modifiedAttrToProjector[attr] =
-      (datum: any, index: number) => attrToProjector[attr](datum, index, dataset);
-  });
+/**
+ * Drawers draw data onto an output of some sort, usually a DOM element.
+ */
+export interface Drawer {
```
themadcreator: `Drawer` -> `IDrawer`

src/drawers/drawer.ts
```diff
@@ -5,238 +5,101 @@
...
-
-  protected _svgElementName: string;
-  protected _className: string;
-  protected _dataset: Dataset;
-
-  private _cachedSelectionValid = false;
-  private _cachedSelection: SimpleSelection<any>;
-  private _cachedSelectionNodes: d3.BaseType[];
-
+export interface Drawer {
```
themadcreator: `Drawer` -> `IDrawer`

src/drawers/drawer.ts
```diff
@@ -5,238 +5,101 @@
...
-    this._renderArea = area;
-    this._canvas = null;
-    this._cachedSelectionValid = false;
-    return this;
-  }
+  getVisualPrimitiveAtIndex(index: number): Element;
 
   /**
-   * Removes the Drawer and its renderArea
+   * Destroy this Drawer, remove all traces of it from the DOM, etc.
```
themadcreator: is this really true for canvas drawers?

src/drawers/drawer.ts
```diff
@@ -5,238 +5,101 @@
...
-        Utils.Window.setTimeout(() => this._drawStep(drawStep), delay);
-        delay += drawStep.animator.totalTime(data.length);
-      });
-    } else if (this._canvas != null) {
-      // don't support animations for now; just draw the last draw step immediately
-      const lastDrawStep = appliedDrawSteps[appliedDrawSteps.length - 1];
-      Utils.Window.setTimeout(() => this._drawStepCanvas(data, lastDrawStep), 0);
-    } else {
-      throw new Error("Drawer's canvas and renderArea are both null!");
+  public useCanvas(canvas: d3.Selection<HTMLCanvasElement, any, any, any>) {
```
themadcreator, line 74: maybe it'd still be possible share one canvas across drawers without keeping track of it in the component, but this should be fine for now

themadcreator (approved):
Okay, i'm cool with this but rename the interface if you're inclined

src/drawers/canvasDrawer.ts
leebyp: at the very least, we should remove from both or keep from both?

Issue #3190, opened 2017-01-31
Add entityNearest for axes labels


CalvinFernandez:
Enable users to determine which axis label was clicked on.

PR #3305, opened 2017-03-22
Tick label data

leebyp:
#3190 

Expose tick label data if given any descendent element, useful for deducing interactions with axis labels

We go for this implementation to avoid calculating the precise location of the labels, incurring a big performance hit. The calculations aren't simple since labels can be angled and truncated.


src/axes/axis.ts
```diff
@@ -105,6 +105,27 @@ export class Axis<D> extends Component {
     this._scale.offUpdate(this._rescaleCallback);
   }
 
+  /**
+   * Gets the tick label data on a element.
+   *
+   * @param {Element} element
+   */
+  public tickLabelDataOnElement(element: Element) {
+    if (element === null || element === undefined) { return undefined; }
```
hellochar: nit: you can do `== null` to catch both null and undefined (this is the only accepted use of ==)

src/axes/axis.ts
```diff
@@ -105,6 +105,27 @@ export class Axis<D> extends Component {
...
+
+    let tickLabel: Element;
+    // go up DOM tree to find tick label element in ancestor elements
+    while ((element !== undefined) && (element.classList) && (tickLabel === undefined)) {
+      if (element.classList.contains(Axis.TICK_LABEL_CLASS)) {
+        tickLabel = element;
+      } else {
+        element = element.parentNode as Element;
```
hellochar: will this sometimes not be an Element? If so, would that be a problem?

src/axes/axis.ts
```diff
@@ -105,6 +105,27 @@ export class Axis<D> extends Component {
...
+      }
+    }
+
+    return element === undefined ? element : d3.select(element).datum();
```
hellochar: nit: returning "element" to mean undefined is a little confusing; return `undefined` itself maybe?

test/axes/axisTests.ts
```diff
@@ -655,4 +655,58 @@ describe("Axis", () => {
...
+    function setup(domain: string[]) {
+      div = TestMethods.generateDiv();
+      scale = new Plottable.Scales.Category().domain(domain);
+      axis = new Plottable.Axes.Category(scale, "left");
+      const TICK_LABEL_MAX_WIDTH = 60;
+      axis.tickLabelMaxWidth(TICK_LABEL_MAX_WIDTH);
+      axis.renderTo(div);
+    };
+
+    function cleanup() {
```
hellochar: you can put this in an `afterEach()` - see https://mochajs.org/#hooks

src/axes/axis.ts
```diff
@@ -105,6 +105,27 @@ export class Axis<D> extends Component {
     this._scale.offUpdate(this._rescaleCallback);
   }
 
+  /**
+   * Gets the tick label data on a element.
```
hellochar: I'd add that the element should be a descendant of a drawn tick label element

test/axes/axisTests.ts
```diff
@@ -655,4 +655,58 @@ describe("Axis", () => {
     axis.destroy();
     div.remove();
   });
+
+  describe("tickLabelDataOnElement", () => {
+    let div: d3.Selection<HTMLDivElement, void | {}, any, any>;
+    let scale: Plottable.Scales.Category;
+    let axis: Plottable.Axes.Category;
+
+    function setup(domain: string[]) {
```
hellochar: you can probably put this in a `beforeEach()` by having just one domain of ["label1", "label2", "long..."]

hellochar (approved):
Looks good in general!

hellochar:
You can't browser select the label text anymore, this is important for bp-charts since our users want to be able to copy/paste a category around

hellochar (changes requested):
Need to be able to select labels

src/axes/axis.ts
leebyp: it's fine because we do the check for `element` and `classList` in the `while` statement

> parentNode is the parent of the current node. The parent of an element is an Element node, a Document node, or a DocumentFragment node.

if it gets to a Document node we want to stop, we want to stop the search

leebyp:
@hellochar oops, i put the zoom interaction on the wrong element in the example - fixed now!

hellochar (approved):

PR #3306, opened 2017-03-23
Turn on tslint

### LINKED ISSUE #2226: Improve TSLint
There are rules which are not enforced in tslint. Either open some PRs there or create some custom rules.

To begin with, exported types should end with a semicolon. As in 

``` typescript
export type AppliedDrawStep = {
  attrToAppliedProjector: AttributeToAppliedProjector;
  animator: Animators.Plot;
};
```

leebyp:
#3169 and while we're at it, let's just turn on tslint
Fixes #2226

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "grunt-git": "0.3.7",
     "grunt-jscs": "2.6.0",
     "grunt-saucelabs": "8.6.2",
-    "grunt-tslint": "^4.0.1",
     "jquery": "2.1.0",
     "load-grunt-tasks": "3.4.0",
     "mocha": "2.2.5",
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "grunt-git": "0.3.7",
     "grunt-jscs": "2.6.0",
     "grunt-saucelabs": "8.6.2",
+    "grunt-tslint": "^4.0.1",
     "jquery": "2.1.0",
     "load-grunt-tasks": "3.4.0",
     "mocha": "2.2.5",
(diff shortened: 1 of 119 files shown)


adidahiya (approved):
epic

adidahiya:
plan to use `tslint:latest` (and blacklist the rules that don't work yet) in another commit?

leebyp:
yup, though actually.... let's see what happens and how much effort it is if i turn that on since im changing most of the files anyway

leebyp:
ok the error list is scary 👎 , let's do it later

package.json
```diff
@@ -45,6 +45,7 @@
     "grunt-git": "0.3.7",
     "grunt-jscs": "2.6.0",
     "grunt-saucelabs": "8.6.2",
+    "grunt-tslint": "^4.0.1",
```
hellochar: We already `tslint` the src using npm scripts instead of using grunt (we'd like to move away from grunt in general; see #3210). I think you can just add a script to lint test files as well

hellochar (changes requested):
Awesome! Just quick comment about running it, but this looks great

package.json
leebyp: sweet, updated - and here i was, still using grunt to do everything...

tsconfig.json
```diff
@@ -12,6 +12,7 @@
       "target": "es5"
     },
     "include": [
-      "src/**/*"
+      "src/**/*",
+      "test/**/*"
```
hellochar, line 16: nit: we already ahve another tsconfig that does this: `test/tsconfig.json` (but the `outDir` is a bit different); might make sense to consolidate the two sometime in the future

hellochar (approved):

PR #3311, opened 2017-03-24
scrub ie9 related workarounds

### LINKED ISSUE #3243: Drop IE9 support; support IE11 and Edge only
Microsoft doesn't even support it anymore so neither should we

hellochar:
Fixes #3243 

changes work fine in ie11


src/utils/windowUtils.ts
```diff
@@ -15,13 +15,7 @@ export function warn(warning: string) {
     return;
```
adidahiya, line 15: we should remove `Configs` and instead check `process.env.NODE_ENV !== "production"`

adidahiya (commented):
nice

plottable.js
```diff
@@ -15948,14 +15948,7 @@ function warn(warning) {
         return;
     }
     /* tslint:disable:no-console */
```
leebyp: `// tslint:disable-next-line:no-console`

test/utils/windowUtilsTests.ts
```diff
@@ -1,51 +1,20 @@
...
-      it("uses console.warn if available, console.log otherwise", () => {
-        const originalWindowWarn = console.warn;
-        let receivedWarning: string = null;
-        const replacementWarn = (warning: string) => receivedWarning = warning;
-        console.warn = replacementWarn;
-
-        const expectedWarning = "BEEP BOOP";
-        Plottable.Utils.Window.warn(expectedWarning);
-        console.warn = originalWindowWarn;
-        assert.strictEqual(receivedWarning, expectedWarning, "console.warn was passed the expected message");
```
leebyp, line 17: why are we removing these tests?

hellochar: They tested behavior that was only possible in ie9 (a null `console` object)

src/utils/windowUtils.ts
hellochar: 👍  https://github.com/palantir/plottable/issues/3314

leebyp (approved):

PR #3312, opened 2017-03-25
publish develop snapshots on every commit to develop

### LINKED ISSUE #3308: Publish dev builds of Plottable to npm
Lets us iterate quicker by letting bp-charts use Plottable features as soon as they go in.

hellochar:
Fixes #3308 

Note: With our current release process, the following behavior happens:

1. We commit a "bump version release" for e.g. v3.2.0, **which triggers a dev snapshot**
2. That commit will get published as e.g. `3.1.0-6-g123abc` to `next`
3. Then, we create git tag v3.2.0, and publish `3.2.0` to `latest` (which is actually the exact same release as `3.1.0-6g123abc`).

This means every `latest` release will be mirrored as a `next` release with a different version number, which is not ideal. Thoughts? @themadcreator @giladgray @aidanns 

Also, on the next commit to develop, we'd publish `3.2.0-1-g-456def` to `next` -- this means `next` never gets `3.2.0`; they just jump from `3.1.0-6` to `3.2.0`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,11 +24,17 @@ test:
     - yarn test
 
 deployment:
+  publish-npm-next:
+    branch: develop
+    commands:
+      - echo -e "$NPM_USER\n$NPM_PASS\n$NPM_EMAIL" | npm login
+      - ./publishSnapshot.sh
+      - ./demo.js
   demo:
     branch: /.*/
     commands:
       - ./demo.js
-  npm:
+  publish-npm-latest:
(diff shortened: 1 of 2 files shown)


adidahiya:
> This means every latest release will be mirrored as a next release with a different version number, which is not ideal.

This sounds totally fine. Multiple releases off the same commit, as long as they all adhere to semver and are predictable, are fine.

adidahiya (approved):

PR #3316, opened 2017-03-27
Histogram Bar Plots

themadcreator:
Add barAlignment getter/setter for bar plots to
change how bars are aligned to the x accessor
coordinate.

Add barEnd accessor to bar plots which is used
to calculate individual bar widths as opposed
to using the same width for all bars. When
set the singular _barPixelWidth value is
not recomputed.

Add tests and update grunt-blanket-mocha so
that phantom runs on macOS Sierra


themadcreator:
Check out [some histograms](https://382-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#basic/histogram)

src/plots/barPlot.ts
```diff
@@ -113,6 +120,56 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
     return this;
   }
 
+  public barEnd(): Plots.TransformableAccessorScaleBinding<X, number>;
```
hellochar: Is this different enough from setting an `x2`/`y2` on a Rectangle Plot to warrant a new endpoint for bar plots? Feels like the API has a lot of overlap

src/plots/barPlot.ts
```diff
@@ -113,6 +120,56 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
...
+   * `"end"`, which determines the meaning of the accessor of the bar's scale
+   * coordinate (e.g. "x" for vertical bars).
+   *
+   * For example, a value of "start" means the x coordinate accessor sets the
+   * left hand side of the rect.
+   *
+   * The default value is "middle", which aligns to rect so that it centered on
+   * the x coordinate
+   */
+  public barAlignment(align?: BarAlignment): any {
```
hellochar, line 178: does this make sense on a categorical scale? I don't imagine we'd want "start" aligned category axes...

src/plots/barPlot.ts
```diff
@@ -662,11 +740,16 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
       return Math.abs(scaledBaseline - originalPositionFn(d, i, dataset));
     };
 
-    attrToProjector["width"] = this._isVertical ? widthF : heightF;
-    attrToProjector["height"] = this._isVertical ? heightF : widthF;
+    const gapF = attrToProjector["gap"];
```
hellochar, line 754: we should docs this `gap` thing or otherwise have proper first-class support for it

hellochar (changes requested):
Does Rectangle Plot fit the bill?

leebyp:
@hellochar @themadcreator re-rectangle plot, from tweaking the "realistic rectangle" quicktest slightly...

stacked histograms on the other hand, require a bit more effort in the code to repro

screen shot 2017-03-29 at 12 04 19

themadcreator:
Right, so this way we get stacked histograms for free. And you can still use a categorical axis and play with the barEnd.

image

The categorical axis in this example has a non-standard barEnd

src/plots/barPlot.ts
themadcreator: It *IS* supported, but whether or not it makes "sense" is debatable.

src/plots/barPlot.ts
themadcreator: There is definitely overlap insofar as you could make EVERY bar chart with a rectangle plot. But, implementing this in bar plot gives us stacked bar histograms for free.

src/plots/barPlot.ts
hellochar: I guess that's my point - users can do nonsensical things with the API. We're telling them "here's a combination that exists, but it's actually a trap because you'd never want to do it". It'd be better to structure the API such that it never happens to begin with imo. But that does sound unreasonably hard to accomplish in this circumstance.

hellochar:
Ok stacked histograms convinces me the endpoint makes sense. I'm not a fan of "barEnd/barAlignment" as the API name though. `barEnd` suggests there's a `barStart` as well - but it's actually called `.x`. But wait - `.x` is actually the bar *center*, unless you set a `.barAlignment`. If you do set `.barAlignment`, `.barEnd` is interpreted differently as well? Maybe `barWidth` would make more sense

themadcreator:
The problem with `barWidth` is that it's not possible to easily compute in data space. Both the `x` and `y` accessor are in data space, then projected through their respective scales. I wanted `barEnd` to be consistent with that, so it is also in data space and then projected through the scale.

The real problem is that the primary axis can be `x` or `y` depending on the vertical/horizontal state.

src/plots/barPlot.ts
```diff
@@ -113,6 +120,56 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
     return this;
   }
 
+  public barEnd(): Plots.TransformableAccessorScaleBinding<X, number>;
+  public barEnd(end: number | Accessor<number>): this;
+  public barEnd(end: X | Accessor<X>, scale: Scale<X, number>): this;
```
hellochar: barEnd should always use the same scale as `.x` right? We should follow RectanglePlot's `.x2` here (it doesn't accept a scale)

hellochar:
I won't block on the API semantics since users need this but I'm not convinced it makes sense... Would you ever want a `barAligment("middle")` when you've set `barEnd`? Maybe we should make barAlignment internal and say "when barEnd is specified, .x becomes the start instead of the center".

Blocking things:
* add docs to barEnd
* make barEnd use `.x`'s scale

hellochar (approved):
Nice comment!

PR #3317, opened 2017-03-28
Remove plottable.js and plottable.min.js

### LINKED ISSUE #3212: Don't commit built plottable.js file
Remove the built file from our committed files since it adds noise to PRs, causes git conflicts, etc.

This will break existing users who pull us in through a script tag. To unbreak them, we should publish every dev build (#3308), which will include the built .js file, and recommend users reference that through https://unpkg.com/.

handrail: replace plottable.js with a script that emits a console warning that tells people how to upgrade.

hellochar:
Removes plottable.js and plottable.min.js from the git repo. Fixes #3212 

### rawgit.com no longer supported
rawgit.com will no longer work as a CDN for newer versions of Plottable. In most cases there's an upgrade path to unpkg or npm:

1. If on a specific version, use //unpkg.com/plottable@<version>/plottable.js
2. If on the develop branch, use //unpkg.com/plottable@next/plottable.js
3. If on a specific commit, you won't break because plottable.js will still exist in the git history
4. If on a different branch/anything else, you can dig up the artifact in circleCI, but this isn't supported.

Basically every existing jsfiddles will break, although most were already broken because of 3.x changes.

### cdnjs.com probably broken too
People referencing us through cdnjs could also suffer - there's some auto-git-commit hook to update cdnjs that I couldn't find. But cdnjs is probably obsolete anyways, so I imagine this isn't a big deal

### Change in expectation/messaging
Breaking people is expected. We are saying our main supported channel for getting plottable is npm. unpkg works as well because it's built around npm.

Update overlaying/ to reflect this.


webpack.config.js
```diff
@@ -63,7 +63,7 @@ module.exports = [
    * User adds two script tags to their html page:
    *
    *   <script src="//cdnjs.cloudflare.com/ajax/libs/d3/4.5.0/d3.js"></script>
```
adidahiya, line 65: might as well switch to unpkg for d3 as well? or is that not an option

adidahiya (approved):

webpack.config.js
hellochar: this is just pseudocode simulating what a user might do; totally possible for them to pull in d3 through unpkg or anywhere else

themadcreator (approved):
So happy about this. I see you updated the overlaying tests as well. I don't think any of us are testing with that are we? Now that we have actual PR previews maybe we should just remove it.

leebyp (approved):

adidahiya:
I agree with @themadcreator, we can 🔪  overlaying tests now that we have historical build previews. Can be a separate PR.

hellochar:
@themadcreator I actually do use it 😉  gives a one pager that ensures nothing has visually regressed

PR #3318, opened 2017-03-29
Build plottable.min.js with webpack

### LINKED ISSUE #3315: plottable.min.js doesn't have Copyright header
https://unpkg.com/plottable@3.0.0-beta.4/plottable.min.js

hellochar:
Use webpack to build plottable.min.js, removing old grunt code and opening the door for more enhancements later (sourcemaps, dev logging that gets wiped in prod, removing comments)

Fixes #3315


webpackConfig/build-min.js
```diff
@@ -0,0 +1,17 @@
+const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
```
adidahiya: let's rename this to `config/webpack.config.prod.js`, it's more consistent with other projects

webpackConfig/build-min.js
```diff
@@ -0,0 +1,17 @@
...
+const webpackMerge = require("webpack-merge");
```
adidahiya: is this to do deep merging of objects? I don't love the idea of pulling in another random dependency for this

webpackConfig/build.js
```diff
@@ -0,0 +1,79 @@
+const webpack = require("webpack");
```
adidahiya, line 1: this is the base config right? let's call it `config/webpack.config.js`

webpackConfig/build.js
```diff
@@ -0,0 +1,79 @@
...
+const WriteFilePlugin = require("write-file-webpack-plugin");
```
adidahiya: why do you need this? are entry points + output not sufficient?

webpackConfig/test.js
```diff
@@ -0,0 +1,39 @@
+const webpack = require("webpack");
```
adidahiya, line 1: `config/webpack.config.test.js`

package.json
```diff
@@ -53,8 +50,11 @@
     "sinon": "^2.1.0",
     "tslint": "4.5.1",
     "typescript": "2.1.5",
+    "uglify-js": "^2.8.18",
+    "uglifyjs-webpack-plugin": "^0.3.1",
```
adidahiya: webpack includes minification out of the box... are you sure you need this?

adidahiya (dismissed):

webpackConfig/build.js
hellochar: The test config is totally separate, and we have a real `/webpack.config.js` that runs both `test` and `build` - see https://github.com/palantir/plottable/pull/3318/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR5 for why

webpackConfig/build-min.js
hellochar: They use it throughout the webpack docs: https://webpack.js.org/guides/production-build/#the-manual-way-configuring-webpack-for-multiple-environments and it has features to do "smart" merging of specific webpack properties which might be useful in the future

adidahiya: alright fair enough

webpackConfig/build.js
hellochar: this wrote plottable.js to disk but is unneeded now, good catch

adidahiya (approved):

PR #3319, opened 2017-03-29
Configurable bar labels position

leebyp:
Allow positioning of labels on bar charts to be configurable with `.labelsEnabled(boolean, position)` where position can be `start`, `middle`, `end`, `outside`, and is relative to the baseline. Defaults to `end`.

_Note:_ I can't quite figure out why we have a difference between `labelContainerOrigin` and `labelOrigin` so have re-used the existing code as guidance.


leebyp:
[Positioned labels](https://394-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#basic/label_bars)

themadcreator (changes requested):
_drawLabel is now like 200 lines long. Can you please extract those x/y positioning functions? I know they modify scope variables, but that method is just too huge.

src/plots/barPlot.ts
```diff
@@ -525,69 +534,126 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
...
         } else if (barY < 0) {
           effectiveBarHeight = barY + barHeight;
         }
         const offset = Bar._LABEL_PADDING;
-        showLabelOnBar = measurement.height + 2 * offset <= effectiveBarHeight;
+        showLabelOnBar = (measurement.height + 2 * offset <= effectiveBarHeight)
+          && (this._labelsPosition !== LabelsPosition.outside);
+
+        const placeLabel = (position: "top" | "middle" | "bottom" | "outside-top" | "outside-bottom") => {
+          switch (position) {
```
hellochar: DRY the vertical and horizontal cases please, and move the label drawing logic into its own method imo

hellochar (changes requested):
I thought in the stacked case the designs called for per-stack labels to be centered in the stack, plus a total outside the stack, whereas this puts all of them on top? And then for stacks with only one stack, or normal bar charts (or clustered), we just show the value of the bar outside the bar?

leebyp:
@hellochar the positioning of the labels feels like a logical configuration point we can leave to the user

leebyp:
Extracting out the huge `drawLabel` to more specific methods. Some repetition in variables within the methods, but this avoid any subtle mutations and is now much more obvious which variables are being used in the calculations.

hellochar (approved):

hellochar:
@themadcreator wanna take a quick pass at this? Looks quite close so it'd be nice to get in for 3.1.0

themadcreator (approved):

PR #3325, opened 2017-04-06
Much faster entityNearest using a quadtree

### LINKED ISSUE #3307: Performance regressions on XYPlot.entityNearest
From gitter:

>  I have an XY chart with multiple datasets, and implement a tooltip. with beta.4 the performance started to get janky. Doing some profiling the hotspot looks like that in beta.2 the call to XYPlot.entityNearest was taking 0.5ms but with beta.4 the function is gone, and Plot.entityNearest takes 189ms (on same dataset)!!

themadcreator:
fixes #3307

### DIFF
--- a/src/plots/linePlot.ts
+++ b/src/plots/linePlot.ts
@@ -451,7 +451,7 @@ export class Line<X> extends XYPlot<X, number> {
 
     const chartBounds = this.bounds();
     this.entities().forEach((entity) => {
-      if (!this._entityVisibleOnPlot(entity, chartBounds)) {
+      if (!Utils.Math.within(entity.position, chartBounds)) {
         return;
       }
       const xDist = Math.abs(queryPoint.x - entity.position.x);
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -164,7 +164,7 @@ export class Plot extends Component {
     selection = coerceExternalD3(selection);
     super.anchor(selection);
     this._dataChanged = true;
-    this._cachedEntityStore = undefined;
+    this._resetEntityStore();
     this._updateExtents();
     return this;
   }
@@ -244,7 +244,7 @@ export class Plot extends Component {
   protected _onDatasetUpdate() {
     this._updateExtents();
     this._dataChanged = true;
-    this._cachedEntityStore = undefined;
+    this._resetEntityStore();
     this.render();
   }
 
@@ -388,6 +388,7 @@ export class Plot extends Component {
    * Updates the extents associated with each attribute, then autodomains all scales the Plot uses.
    */
   protected _updateExtents() {
+    this._resetEntityStore();
     this._attrBindings.each((_, attr) => this._updateExtentsForAttr(attr));
     this._propertyExtents.each((_, property) => this._updateExtentsForProperty(property));
     this._scales().forEach((scale) => scale.addIncludedValuesProvider(this._includedValuesProvider));
@@ -700,7 +701,7 @@ export class Plot extends Component {
    * @return {Plots.PlotEntity[]}
    */
   public entities(datasets?: Dataset[]): Plots.IPlotEntity[] {
-    return this._getEntityStore(datasets).map((entity) => this._lightweightPlotEntityToPlotEntity(entity));
+    return this._getEntityStore(datasets).entities().map((entity) => this._lightweightPlotEntityToPlotEntity(entity));
   }
 
   /**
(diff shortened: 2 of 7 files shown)


themadcreator:
This is SOOO much faster, but there are a couple more things to do

- [ ] Don't rebuild quadtree while panning
- [x] Fix tests (???)

leebyp:
Note that for the 100k plot, the "rebuilding" of the tree is taking a good 400ms

EDIT: BAM! take out the getter which we don't need anymore because we nuke the cache, and it goes down to 40ms!

src/plots/plot.ts
```diff
@@ -388,6 +388,7 @@ export class Plot extends Component {
    * Updates the extents associated with each attribute, then autodomains all scales the Plot uses.
    */
   protected _updateExtents() {
+    this._resetEntityStore();
```
leebyp, line 398: why is this needed here now? isn't all we need done through the `scale.onUpdate`?

src/utils/entityStore.ts
```diff
@@ -3,95 +3,90 @@
...
+export class EntityStore<T extends IPositionedEntity> implements IEntityStore<T> {
   private _entities: T[];
+  private _tree: d3.Quadtree<T>;
 
   constructor() {
     this._entities = [];
+    this._tree = d3.quadtree<T>()
+      // int casting for faster computation. losing sub-pixel precision here is
+      // of no concern.
+      .x((d) => d.position.x | 0)
```
hellochar: nit: prefer Math.floor() to cast to int. Also do we know that casting to an int actually improves perf significantly?

src/utils/entityStore.ts
```diff
@@ -3,95 +3,90 @@
...
+    this._entities.push(...entities);
+    this._tree.addAll(entities);
   }
 
   /**
-   * Iterates through array of of entities and computes the closest point using
-   * the standard Euclidean distance formula.
+   * Looks up the nearest data point using the internal quadtree.
+   *
+   * Note that for performance reasons, the optional filter is **DESTRUCTIVE**
```
hellochar: So calling `entityNearest` mutates the quadtree? Is the expectation that this is only called once per call to addAll?

hellochar (approved):
Had some questions but looks good in general

leebyp (changes requested):
blocking for panning (and zoom?) throttling - unless the removal of the getter makes this fine

src/plots/plot.ts
themadcreator: Yeah, we can probably remove it but previously we were clearing the cache here and i didn't want to miss it if it was needed. Since "clearing the cache" is just setting to `undefined` it's not gonna be slow

src/utils/entityStore.ts
themadcreator: It does, which is why i made this comment. There are no guarantees about who will call entityNearest from the outside, but unfortunately creating a copy of the tree every time is quite expensive. What would be ideal is if we could just get the quadtree itself to do the filtering, but that's not how it's implemented

src/utils/entityStore.ts
themadcreator: Alright, i guess math.floor isnt so bad:

image

src/utils/entityStore.ts
themadcreator: Making a copy means this method takes 10-20msec, whereas without the copy it take <1msec. Thoughts?

src/plots/plot.ts
leebyp: we weren't clearing it here originally

themadcreator:
Remaining tasks are 
- [x] defer resetting entitystore on scale update
- [x] use bounds at entity store build time not nearest query time to avoid destructive update

src/plots/linePlot.ts
```diff
@@ -451,7 +451,7 @@ export class Line<X> extends XYPlot<X, number> {
 
     const chartBounds = this.bounds();
     this.entities().forEach((entity) => {
-      if (!this._entityVisibleOnPlot(entity, chartBounds)) {
+      if (!Utils.Math.within(entity.position, chartBounds)) {
```
leebyp, line 454: oh i guess we can just replace `this._entityVisibleOnPlot` with the math util in `plot.ts` too

src/utils/entityStore.ts
```diff
@@ -3,95 +3,94 @@
...
 }
 
 /**
- * Array-backed implementation of {EntityStore}
+ * Implementation of {IEntityStore} that uses an array for easy iteration as
+ * well as a quad tree for fast nearest-point queries.
+ *
+ * Note that if the position of your entities changes, you MUST rebuild the
+ * entity store for the `entityNearest` method to work since the quadtree does
+ * not know that its nodes have moved.
```
leebyp, line 58: `and may need to re-filter on the bounds`

leebyp (approved):

src/utils/entityStore.ts
themadcreator: yeah but im pretty clear about that in the comment on .addAll and entityNearest

src/plots/linePlot.ts
themadcreator: yeah i did that which is why i also had to change this here

PR #3332, opened 2017-04-26
axis sets its own size

hellochar:
Modify Axis to always be the width/height that it wants; Axes tell the outside what height they are.
This allows blueprint-chart to put the Axis in document flow by removing absolute positioning.

reduce the non_canvas_scatter number of points for better perf while looking through overlaying.

Fix a bad search/replace of let -> const in the tests

Refactor DrawStep to remove a circular dependency

### DIFF
--- a/src/axes/categoryAxis.ts
+++ b/src/axes/categoryAxis.ts
@@ -6,11 +6,11 @@
 import * as d3 from "d3";
 import * as Typesetter from "typesettable";
 
+import { Component } from "../components/component";
 import { Point, SimpleSelection, SpaceRequest } from "../core/interfaces";
 import * as Scales from "../scales";
 import * as Utils from "../utils";
 import { Axis, AxisOrientation } from "./axis";
-import { Component } from "../components/component";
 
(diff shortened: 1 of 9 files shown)


src/axes/categoryAxis.ts
```diff
@@ -142,6 +143,12 @@ export class Category extends Axis<string> {
     return this.getDownsampleInfo().domain;
   }
 
+  protected _sizeFromOffer(availableWidth: number, availableHeight: number) {
```
themadcreator, line 146: what? i don't get why we're doing this here vs just calling super. i see this comment, but if we override this will angled tick labels still request enough space to fit?

themadcreator (commented):
questions about why we're jumping the super call hierarchy

themadcreator (approved):

src/axes/categoryAxis.ts
hellochar: Spoke offline - without this hack, category axis layout gets borked. In the interest of speed we're going to just keep this in for now

PR #3336, opened 2017-04-27
preparing release 3.1.0

hellochar:
# v3.1.0

# New Features
* Scatter Plot now supports the Canvas renderer. Simply set `.renderer("canvas")` on a Scatter Plot. We further optimize scatterplot canvas drawing by culling out-of-viewport elements and image-blitting the same symbol.
* Added `Component.setBounds(width, height, originX, originY)` to allow exact placement of a Component, overriding any previous layout.
* Line Plot now has a `.collapseDenseLinesEnabled(true)` feature which improves performance of drawing dense vertical lines (e.g. 100k timeseries points) by bucketing lines that would be drawn on the same pixel. This may affect the look of the chart for semi-transparent lines.
* Bar Plot labels can now be positioned to be the `"start"`, `"middle"`, `"end"`, or `"outside"` of the bar. Pass it as the second argument to labelsEnabled: `barPlot.labelsEnabled(true, "middle")`.

# Improvements
* Plottable is now releasing dev builds on every commit into develop. Get new changes immediately by pulling from the `"next"` dist-tag on NPM: `yarn add "plottable@next"`.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "plottable",
   "description": "A modular charting library built on D3",
-  "version": "3.0.0",
+  "version": "3.1.0",
   "main": [
     "plottable.js",
     "plottable.css"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "plottable",
   "description": "A modular charting library built on D3",
-  "version": "3.0.0",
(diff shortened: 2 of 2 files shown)


themadcreator (approved):
🔥 🚢 🔥

PR #3337, opened 2017-04-28
Fix dist script

themadcreator:


package.json
```diff
@@ -12,10 +12,11 @@
   "typings": "build/src/index.d.ts",
   "scripts": {
     "build": "npm-run-all -p lint:tslint build:tsc -s build:webpack",
-    "build:tsc": "tsc -p test/",
+    "build:tsc": "tsc -p .",
```
hellochar, line 15: Building test files already builds the src (since the src/ tree is a subset of the test/) and isn't much slower. Can we keep building test/?

adidahiya: if you intend to compile them all at once, can you switch to only having one tsconfig project? that would make this more straightforward

adidahiya (approved):

Issue #3338, opened 2017-05-04
__VERSION__ is not defined


clarkmoody:
Upgraded to `3.1.0` just now, and when I `require('plottable')` in my project, I get `__VERSION__` not defined.

The browser points me to `core/version.js` that contains the offending line. Perhaps the `sed-version` task did not run properly for this build?

I'm using watchify 3.8.0 and browserify 13.1.1

hellochar:
@clarkmoody Oof, thanks for the bug report, we really should have an automated test for that. Will get this fixed immediately...

clarkmoody:
Thanks! New version `3.3.1` works well.

PR #3340, opened 2017-05-08
Sed version for distributable. Update version in package for patch release.

themadcreator:


themadcreator:
Fixes #3341
Fixes #3338

hellochar (approved):
I wonder, can we test this stuff as part of our build somehow?

themadcreator:
We really should test all release modules and artifacts. But how to test those without releasing?

hellochar:
Our unit tests didn't catch this P0 because we don't have any that tests the module loading. Maybe we could add a dumb nodejs test:

```js
// test-module-loading.js
var Plottable = require("plottable");
```

and run it with our build step:

```
// package.json
  "test-node": "node test-module-loading.js"
```

PR #3343, opened 2017-05-09
Update README.md

hellochar:
Users need to `yarn build` before `yarn start`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ If you run into any problems using Plottable, please let us know. We want Plotta
 
 - Clone the repo
 - Install local dependencies `yarn install`
+- Run `yarn build` to build the dependencies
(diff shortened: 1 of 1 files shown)


adidahiya (dismissed):

README.md
```diff
@@ -39,6 +39,7 @@ If you run into any problems using Plottable, please let us know. We want Plotta
 
 - Clone the repo
 - Install local dependencies `yarn install`
+- Run `yarn build` to build the dependencies
```
adidahiya, line 42: which part of this is actually required? the webpack part?

hellochar: This creates `/build/index.ts`. `yarn watch` only watches, it doesn't build

themadcreator (approved):

PR #3344, opened 2017-05-10
firefox text size estimations

acolombi:
@hellochar check this out

### DIFF
--- a/plottable.css
+++ b/plottable.css
@@ -104,6 +104,10 @@
   transform-origin: 0px 0px 0px;
 }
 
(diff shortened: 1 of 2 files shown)


hellochar:
@themadcreator we should probably add this in typesettable; filed https://github.com/palantir/typesettable/issues/68

hellochar (approved):

hellochar:
Hmm circle never built this for some reason... @themadcreator I'm just gonna merge if you're cool with that

PR #3356, opened 2017-06-02
Pass datum, index, and dataset to labelFormatter

### LINKED ISSUE #3088: Per-label Formatting Option for plot.labelFormatter(d)

hellochar:
Fixes #3088 

Introduce a `DatumFormatter`, separate from normal `Formatter`. DatumFormatter is passed the value (like Formatter), as well as the `datum`, `index`, and `dataset` of the datum being labelled. Use this for bar and pie plots.

Loosely based on #3226, thanks @dallinregehr for the existing work!

### DIFF
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -4,8 +4,20 @@
  */
 
 import * as d3 from "d3";
+import { Dataset } from "./dataset";
 
-export type Formatter = (d: any) => string;
+/**
+ * A basic formatter function that will be passed a value to format (e.g. the tick value for axes).
+ * A Formatter should return the formatted string representation of the input value.
+ */
+export type Formatter = (value: any) => string;
+
+/**
+ * A formatter function that will be passed the value to format as well as the underlying datum, index, and
+ * dataset that the value came from. Datum-backed visual elements that display labels, such as Bar Plot bars
+ * or Pie Plot sectors, will use this formatter.
+ */
+export type DatumFormatter = (value: any, datum: any, index: number, dataset: Dataset) => string;
 
 interface IPredicatedFormat {
   specifier: string;
--- a/src/core/formatters.ts
+++ b/src/core/formatters.ts
@@ -9,6 +9,9 @@ import { Dataset } from "./dataset";
 /**
  * A basic formatter function that will be passed a value to format (e.g. the tick value for axes).
  * A Formatter should return the formatted string representation of the input value.
+ *
+ * The value will be different for each use of the formatter - check the documentation of individual
+ * methods to see what will be passed into the formatter.
  */
 export type Formatter = (value: any) => string;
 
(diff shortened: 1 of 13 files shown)


src/plots/stackedBarPlot.ts
```diff
@@ -151,7 +151,7 @@ export class StackedBar<X, Y> extends Bar<X, Y> {
       if (maximum.extent !== baselineValue) {
         // only draw sums for values not at the baseline
 
-        const text = this.labelFormatter()(maximum.extent);
+        const text = this.labelFormatter()(maximum.extent, undefined, undefined, undefined);
```
themadcreator: oh, eww. This seems to not break the contract of the datumformatter and so now the formatter needs to check whether or not it is formatting an actual data value.

themadcreator: Can we not provide any data here? not even the dataset obj?

hellochar: Yeah it's gross. A single stack is built from an array of datums, each one coming from a different dataset. How do we reconcile the two?

themadcreator: reconciliation is hard. we could do an array of datums, but that still means they have to check the args. could we use an additional formatter?? `extremaFormatter`, but even just providing of the datums at random seems slightly better. It would probably make sense to use whatever dataset is at the top of the stack

hellochar: resolution: expose a separate "extrema" formatter

src/core/formatters.ts
```diff
@@ -4,8 +4,20 @@
...
+ * A basic formatter function that will be passed a value to format (e.g. the tick value for axes).
+ * A Formatter should return the formatted string representation of the input value.
+ */
+export type Formatter = (value: any) => string;
+
+/**
+ * A formatter function that will be passed the value to format as well as the underlying datum, index, and
+ * dataset that the value came from. Datum-backed visual elements that display labels, such as Bar Plot bars
+ * or Pie Plot sectors, will use this formatter.
+ */
```
themadcreator: What is `value`? isn't that the same as `datum`? or is it the result of `scale(datum)` or something?

themadcreator: the standard callback is like (datum, index, dataset)

themadcreator (changes requested):
Want comment to explain what `value` is going to be in various plot types

themadcreator (approved):

PR #3358, opened 2017-06-04
Allow null in Table constructor type.

devnev:
The documentation and implementation currently permit null, but the actual constructor type requires all cells to be valid components.

### DIFF
--- a/src/components/table.ts
+++ b/src/components/table.ts
@@ -50,10 +50,10 @@ export class Table extends ComponentContainer {
    * ```
    *
    * @constructor
-   * @param {Component[][]} [rows=[]] A 2-D array of Components to be added to the Table.
+   * @param {(Component|null)[][]} [rows=[]] A 2-D array of Components to be added to the Table.
    *   null can be used if a cell is empty.
    */
-  constructor(rows: Component[][] = []) {
+  constructor(rows: (Component|null)[][] = []) {
     super();
     this.addClass("table");
     rows.forEach((row, rowIndex) => {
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @devnev! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/components/table.ts
```diff
@@ -50,10 +50,10 @@ export class Table extends ComponentContainer {
    * ```
    *
    * @constructor
-   * @param {Component[][]} [rows=[]] A 2-D array of Components to be added to the Table.
+   * @param {(Component|null)[][]} [rows=[]] A 2-D array of Components to be added to the Table.
    *   null can be used if a cell is empty.
    */
-  constructor(rows: Component[][] = []) {
+  constructor(rows: (Component|null)[][] = []) {
```
hellochar: Lets also allow `undefined`

hellochar (changes requested):
Much welcomed PR, thanks for submitting!

hellochar:
Went ahead and fixed it in Github, cheers :)

devnev:
Hey @hellochar, thanks for the quick turnaround. Worth noting that the implementation does not check for `undefined`, only for `null`, so the new type specification doesn't match the implementation. The documentation also mentions `null can be used if a cell is empty.`, but doesn't specify the behaviour for `undefined`.

hellochar:
Absolutely correct, good eye 😉  we consciously allow `undefined` by using double-equals (`== null`) to be more user friendly/robust, but it's not documented since we don't want to necessarily encourage it

PR #3359, opened 2017-06-06
legend perf improvement

### LINKED ISSUE #3089: calculating legend layout during draw is extremely slow
Calling redraw on a legend with 431 items locks the browser for 9 seconds. Culprit seems to be [Legend._calculateLayoutInfo](https://github.com/palantir/plottable/blob/develop/src/components/legend.ts#L162-L191)

screen shot 2016-07-23 at 5 44 34 pm

Seems like an easy performance win to speed up draw times. Especially important since the responsive strategy for plottable is to redraw on resize events!

hellochar:
Just use CacheMeasurer everywhere

Fixes #3089

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -2,6 +2,11 @@
   "rules": {
     "align": [true, "parameters", "statements"],
     "arrow-parens": true,
+    "ban": [
+      true,
+      "eval",
+      { "name": ["Typesettable", "Measurer"], "message": "Use CacheMeasurer instead" }
+    ],
     "interface-name": [true, "always-prefix"],
     "ordered-imports": true,
     "no-boolean-literal-compare": true,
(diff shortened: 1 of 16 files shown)


themadcreator:
I think I had originally gone through and replaced every since measurer with a cachemeasurer, but some of the plots didn't match after that. How thoroughly did you review the existing dev examples?

hellochar:
Not closely. Do you remember what the issue was?

tslint.json
```diff
@@ -2,6 +2,11 @@
   "rules": {
     "align": [true, "parameters", "statements"],
     "arrow-parens": true,
+    "ban": [
+      true,
+      "eval",
+      { "name": ["Typesettable", "Measurer"], "message": "Use CacheMeasurer instead" }
```
themadcreator, line 8: heh. 👍

themadcreator (approved):
Cool, LGTM

PR #3360, opened 2017-06-06
[WIP] category axis word breaking

hellochar:
For certain labels this could look a lot better.

There's an outstanding bug where large single words that can't be broken will hide the entire label completely. We should resolve that before making this a first-class API


themadcreator (approved):

themadcreator (changes requested):
Blocking on fix for https://github.com/palantir/typesettable/issues/69

themadcreator:
Closing for now

PR #3365, opened 2017-06-10
Create ISSUE_TEMPLATE.md

hellochar:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
+<!--
+
+Thanks for your interest in Plottable. For quickest turnaround on bugs, please create a live example
+(jsfiddle, codepen, etc) that reproduces the faulty behavior.
+
+Is this a support question? Please ask on the Gitter chat channel: gitter.im/palantir/plottable
```
adidahiya, line 6: are you sure you want to prefer gitter over stack overflow? maybe mention both

adidahiya (approved):

.github/ISSUE_TEMPLATE.md
hellochar: Yeah our gitter is much more active; we have like one Plottable SO question per three months (so keeping a record of the problem/fix isn't even that useful since the APIs are in flux)

PR #3369, opened 2017-06-22
Deferred Rendering Fixes

themadcreator:
Refactored deferred rendering in xyplot mostly into new `DeferredRenderer` class.

Added `postScale` projector to `.x()` and `.y()` setters to allow transformation in pixel coordinates while still storing the scale on the accessor binding.

Buffer canvas content while resizing to prevent janky redraws while resizing.

Make dataset changes use deferred rendering callback.

Exposed deferred render callback as `renderLowPriority`.

### DIFF
--- a/src/plots/deferredRenderer.ts
+++ b/src/plots/deferredRenderer.ts
@@ -1,16 +1,23 @@
+/**
+ * Copyright 2017-present Palantir Technologies
+ * @license MIT
+ */
 
 import { Scale } from "../scales/scale";
 
+/**
+ * Stores the deferred transformation state for a single scale.
+ */
 class DomainTransform<T> {
-  public translate = 0;
   public scale = 0;
+  public translate = 0;
+
   private cachedDomain: T[] = [null, null];
   private lastSeenDomain: T[] = [null, null];
-  private renderCallback: () => void;
 
   public reset() {
-    this.translate = 0;
     this.scale = 1;
+    this.translate = 0;
     this.cachedDomain = this.lastSeenDomain;
   }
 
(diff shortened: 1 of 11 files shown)


src/plots/deferredRenderer.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Call `setDomains` when deferred rendering is initially enabled to fix the
+ * current domain values.
+ *
+ * Call `updateDomains` when scale domains change, which uses the domain to
+ * compute CSS-tyle transform parameters passed to `applyTransformCallback`,
+ * mimicking the result of a full re-render. After a deferred timeout, invoke
+ * `applyTransformCallback` again with an identity transform and finally invoke
+ * `renderCallback`, which should actually redraw the plot.
+ */
+export class DeferredRenderer<X, Y> {
```
hellochar, line 49: 💯

src/plots/plot.ts
```diff
@@ -157,14 +162,14 @@ export class Plot extends Component {
...
-    this._renderCallback = (scale) => this.render();
+    this._renderCallback = () => this.render();
     this._onDatasetUpdateCallback = () => this._onDatasetUpdate();
     this._propertyBindings = d3.map<Plots.IAccessorScaleBinding<any, any>>();
     this._propertyExtents = d3.map<any[]>();
     const mainAnimator = new Animators.Easing().maxTotalDuration(Plot._ANIMATION_MAX_DURATION);
     this.animator(Plots.Animator.MAIN, mainAnimator);
     this.animator(Plots.Animator.RESET, new Animators.Null());
-    this._deferredResetEntityStore = Utils.Window.debounce(Plot._DEFERRED_RENDERING_DELAY, this._resetEntityStore);
+    this._deferredResetEntityStore = Utils.Window.debounce(DeferredRenderer.DEFERRED_RENDERING_DELAY, this._resetEntityStore);
```
hellochar, line 172: this kinda hints that the entity store belongs in the renderer as well (later refactor?)

src/plots/plot.ts
```diff
@@ -199,6 +204,16 @@ export class Plot extends Component {
   public setBounds(width: number, height: number, originX?: number, originY?: number) {
     super.setBounds(width, height, originX, originY);
     if (this._canvas != null) {
+      if (this._bufferCanvas && !this._bufferCanvasValid) {
```
hellochar, line 207: why do we need a buffer canvas?

hellochar (changes requested):
Awesome! can we confirm the bufferCanvas doesn't regress the window.devicePixelRatio bug?

src/plots/plot.ts
themadcreator: so, as soon as you change the size of a canvas with css or whatever, it clears the contents. we have some drag-resizable charts in various apps, and what was happening is that it would immediately do a full redraw while you drag-resize. This buffer canvas stores the current canvas contents when the resize starts and redraws it into the resized canvas. Eventually, the deferred rendering callback will trigger and do a full-rez redraw. If deferred rendering is disabled, the buffer copy will be overwritten immediately by a full redraw.

themadcreator: Demonstration:
update-after-resize

themadcreator:
It uses the exact same size as the original canvas, so there should be no devicepixelratio issues.

hellochar (approved):

PR #3370, opened 2017-06-22
Fix drawLabels with barAlignment

hellochar:
Previously stackedBar's drawLabels didn't take into account barAlignment. It now does. Refactored/cleaned code to more accurately reflect what's going on.

Use the term bar "thickness" instead of bar "width" where appropriate. width is specifically associated with the x direction but it was being used to represent the thickness (which is meaningful in both x and y).

Also consolidate the terms "primary" and "secondary" into bar "length" and bar "position". Add methods to get the "length" binding and "position" bindings.

PR #3371, opened 2017-06-23
Bumping minor version for release

themadcreator:


themadcreator:
Release Notes:

## Bug Fixes

#### Fixed several issues with deferred rendering (#3369)

Adds new post-scaled callback `IRangeProjector` with the following signature
```typescript
(value: T, datum: any, index: number, dataset: Dataset): T;
```
This new callback can be passed to `.x()` and `.y()` accessor methods on xy plots to manipulate pixel position after the scale is applied. This allows the `Scale` to also be stored on the accessor binding.

Adds buffered canvas rendering to plots so that resizes of the canvas don't clear the canvas view. Combined with deferred rendering, this greatly reduce janky resizing of charts.

Adds `renderLowPriority` callback to plots.

#### Fixed drawLabels with barAlignment (#3370)

Previously stackedBar's drawLabels didn't take into account barAlignment. It now does.

hellochar (approved):

PR #3372, opened 2017-06-26
Add gridlines between ticks

CalvinFernandez:
Change adds `betweenX` and `betweenY` methods to the gridlines api
which each toggle rendering of x and y gridlines between ticks,
respectively. Additionally, gridlines can now be displayed for
oridinal scales.

### DIFF
--- a/quicktests/overlaying/tests/basic/gridlines.js
+++ b/quicktests/overlaying/tests/basic/gridlines.js
@@ -59,26 +59,6 @@ function run(svg, data, Plottable) {
         [plotGroup2, plotGroup3]
     ]);
 
-    /*
-    var xScale = new Plottable.Scales.Category().domain(data.map(d => d.x));
-    var xAxis = new Plottable.Axes.Category(xScale, "bottom");
-    xAxis.tickLabelMaxLines(1);
-    var yScale = new Plottable.Scales.Linear().domain([0, 20]);
-    var yAxis = new Plottable.Axes.Numeric(yScale, "left");
-    var gridlines = new Plottable.Components.Gridlines(xScale, yScale).betweenX(true);
-
-    var plot = new Plottable.Plots.Bar();
-    plot.x(function(d) { return d.x }, xScale);
-    plot.y(function(d) { return d.y }, yScale);
-    plot.addDataset(new Plottable.Dataset(data));
-
-    var plotGroup = new Plottable.Components.Group([plot, gridlines]);
-    var table = new Plottable.Components.Table([
-        [yAxis, plotGroup],
-        [null,  xAxis]
-    ]);
-
-    */
     table.renderTo(svg);
     window.addEventListener("resize", () => table.redraw());
 }
--- a/quicktests/overlaying/tests/basic/gridlines.js
+++ b/quicktests/overlaying/tests/basic/gridlines.js
@@ -1,0 +1,84 @@
+function makeData() {
+    "use strict";
+    const rockTypes = [
+            "pegmatite",
+            "dolomite",
+            "quartzite",
+            "limestone",
+            "diorite",
+            "serpentite",
+            "travertine",
+            "lapis lazuli",
+            "marble",
+            "coquina",
+            "syenite",
+            "dunite",
+            "coal",
+            "flint",
+            "chalk",
+            "hornfels",
+            "skarn",
+            "greenschist"
+        ];
+    return rockTypes.map(function(rockType, index) { return { x: rockType, y: index + 3 } });
+}
+
+function run(svg, data, Plottable) {
+    "use strict";
+
+    var xScale = new Plottable.Scales.Category().domain(data.map(d => d.x));
+    var yScale = new Plottable.Scales.Linear().domain([0, 100]);
+    var xAxis = new Plottable.Axes.Category(xScale, "bottom");
+    var yAxis = new Plottable.Axes.Numeric(yScale, "left");
+    var gridlines = new Plottable.Components.Gridlines(xScale, yScale).betweenX(true).betweenY(true);
+    var plotGroup = new Plottable.Components.Table([[yAxis, gridlines], [null, xAxis]]);
+
+    var xScale1 = new Plottable.Scales.Linear().domain([0, 100]);
+    var yScale1 = new Plottable.Scales.Category().domain(data.map(d => d.x));
+    var xAxis1 = new Plottable.Axes.Numeric(xScale1, "bottom");
+    var yAxis1 = new Plottable.Axes.Category(yScale1, "left");
+    var gridlines1 = new Plottable.Components.Gridlines(xScale1, yScale1).betweenX(true).betweenY(true);
+    var plotGroup1 = new Plottable.Components.Table([[yAxis1, gridlines1], [null, xAxis1]]);
+
+    var xScale2 = new Plottable.Scales.Time().domain([new Date("1/1/1"), new Date("2/2/2")]);
+    var yScale2 = new Plottable.Scales.Category().domain(data.map(d => d.x));
+    var xAxis2 = new Plottable.Axes.Time(xScale2, "bottom");
+    var yAxis2 = new Plottable.Axes.Category(yScale2, "left");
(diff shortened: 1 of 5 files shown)


quicktests/overlaying/tests/basic/gridlines.js
```diff
@@ -0,0 +1,84 @@
...
+    var yAxis3 = new Plottable.Axes.Category(yScale3, "left");
+    var gridlines3 = new Plottable.Components.Gridlines(xScale3, yScale3).betweenX(true);
+    var plotGroup3 = new Plottable.Components.Table([[yAxis3, gridlines3], [null, xAxis3]]);
+
+    var table = new Plottable.Components.Table([
+        [plotGroup1, plotGroup],
+        [plotGroup2, plotGroup3]
+    ]);
+
+    /*
```
themadcreator: delete this unused comment

src/components/gridlines.ts
```diff
@@ -4,31 +4,53 @@
  */
 
 import { Point, SimpleSelection } from "../core/interfaces";
-import { QuantitativeScale } from "../scales/quantitativeScale";
-import { IScaleCallback } from "../scales/scale";
+import { IScaleCallback, Scale } from "../scales/scale";
 
 import { Component } from "./component";
 
+function getScaledValue(
```
hellochar: docs plz, this looks like it's returning a tick generator function that has internal "previousPosition" state? `getScaledValue` doesn't really convey that

src/components/gridlines.ts
```diff
@@ -4,31 +4,53 @@
...
+    const position = scale.scale(tickVal);
+
+    if (!between) {
+      return position;
+    }
+
+    let gridPosition: number;
+
+    if (previousPosition !== undefined) {
+      gridPosition = previousPosition + (position - previousPosition) / 2;
```
themadcreator: this assumes that this method will be called in sorted order but i don't think that's necessarily true. can't the ticks be out of order?

src/components/gridlines.ts
```diff
@@ -42,6 +64,36 @@ export class Gridlines extends Component {
...
+
+    if (_betweenX !== this._betweenX) {
+      this._betweenX = _betweenX;
+      this.render();
+    }
+
+    return this;
+  }
+
+  public betweenY(_betweenY: boolean): this;
```
hellochar: jsdoc this plz

src/components/gridlines.ts
```diff
@@ -42,6 +64,36 @@ export class Gridlines extends Component {
     }
   }
 
+  public betweenX(_betweenX: boolean): this;
```
hellochar: 1. I think between-tick-lines are called "minor gridlines"; how about an API like `minorLinesEnabled(true, false)` that enables/disables it on both axes?
1. jsdoc this plz

src/components/gridlines.ts
themadcreator: also, halfway between ticks isn't necessarily correct. what if the bar charts define a separate barwidth function. imagine a really fat bar next to a really skinny bar. the gridline between their centers isn't right. it needs to be in the middle of the space BETWEEN the bars.

themadcreator: furthermore, bar charts can also be aligned differently. in "histogram" mode the bar's left side is aligned to its x domain coordinate. this may be okay since you're passing the tick value through the scale, though.

themadcreator (changes requested):
I don't think the simple average of prev and current coordinate for "between" is sufficient for all charts.

src/components/gridlines.ts
```diff
@@ -79,13 +131,19 @@ export class Gridlines extends Component {
...
+
       const xLinesUpdate = this._xLinesContainer.selectAll("line").data(xTicks);
       const xLines = xLinesUpdate.enter().append("line").merge(xLinesUpdate);
-      xLines.attr("x1", getScaledXValue)
+      xLines.attr("x1", getScaledValue(this._xScale, between, initialXValue))
         .attr("y1", 0)
-        .attr("x2", getScaledXValue)
+        .attr("x2", getScaledValue(this._xScale, between, initialXValue))
         .attr("y2", this.height())
         .classed("zeroline", (t: number) => t === 0);
```
hellochar: can we add a `gridlines-minor` CSS class for minor gridlines?

src/scales/scale.ts
```diff
@@ -105,6 +105,15 @@ export class Scale<D, R> {
   }
 
   /**
+   * Gets an array of tick values spanning the domain.
+   *
+   * @returns {D[]}
+   */
+  public ticks(): D[] {
+    return this.domain();
```
hellochar, line 113: why not use `this.domain()`?

hellochar:
@pkwi for design input

hellochar (changes requested):

src/components/gridlines.ts
themadcreator: `minor` isn't quite what he's going for i think. `minor` implies there are two sets: `major` and `minor`. but with `between`, that means take the normal grid lines and move them elsewhere

hellochar: oh as in re-interpret the "main" gridlines to be between the ticks? Hmm I see, that makes sense for Category axis I think but maybe not linear. I think we can all agree on "gridlines do between ticks for Category", maybe we only implement that for now and leave the trickiness of simple averaging for later

src/components/gridlines.ts
CalvinFernandez: oh wow they can be called out of order?

yea you're right I forgot that you could do variable step sizes.

CalvinFernandez:
@hellochar @themadcreator haha good to be back in the land of  ❌ ❌ ❌ ❌   😄 


@themadcreator I agree that the simple average isn't sufficient. It looks like the old implementation also used the average, however. https://github.com/palantir/plottable/pull/3189/files#diff-653dc28c2857e38d0e1b192291c6eba2R184 . Am I missing something with the way that was implemented?

src/scales/scale.ts
CalvinFernandez: I like the shared api of `ticks` better than type checking in the gridlines code.

src/components/gridlines.ts
CalvinFernandez: yes if we all agree that gridlines-minor is what we want to call this

src/components/gridlines.ts
CalvinFernandez: Oops yea I can jsdoc whichever api name we land on

src/components/gridlines.ts
CalvinFernandez: yea sure, this was just a small refactor of an existing method. Ill fix it

src/components/gridlines.ts
CalvinFernandez: It can make sense for linear as well! For example, it enables inset tick labels with a grid, like this

image

hellochar: Sounds good, fine with `between`

hellochar:
Things I'd like to see:

1. docs for API
2. CSS class denoting between gridlines
3. better docs/naming for `getScaledValue`

Looks good other than that!

src/components/gridlines.ts
```diff
@@ -79,31 +171,34 @@ export class Gridlines extends Component {
...
   private _redrawXLines() {
     if (this._xScale) {
-      const xTicks = this._xScale.ticks();
-      const getScaledXValue = (tickVal: number) => this._xScale.scale(tickVal);
+      const between = this.betweenX();
+      const xTicks = this._xScale.ticks().slice(between ? 1 : 0);
       const xLinesUpdate = this._xLinesContainer.selectAll("line").data(xTicks);
       const xLines = xLinesUpdate.enter().append("line").merge(xLinesUpdate);
-      xLines.attr("x1", getScaledXValue)
+      xLines.attr("x1", gridPositionFactory(this._xScale, between, this._xScale.ticks()))
```
themadcreator, line 178: I don't know how expensive `.ticks()` is but you already called it above, so you may as well re-use that

themadcreator (approved):
Minor request, otherwise looks good

hellochar (approved):

hellochar:
merged, thanks @CalvinFernandez !

PR #3374, opened 2017-06-27
use a <rect> inside an svg to implement transform detector

### LINKED ISSUE #3355: interaction measurer blocks all interaction on the page in iPad
3.0.0's refactor to Plottable's interaction manager now make it block all touch inputs on the page in iPad, so e.g. tapping buttons, textfield, etc. are all disabled. Probably should be fixed ASAP

@sixinli  for sa

hellochar:
rects do not count as DOM modification in Safari which allows Safari to
properly fire click events on tap

random docs

Fixes #3355

### DIFF
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -6,6 +6,14 @@
 import * as Utils from "../utils";
 
 export class Dispatcher {
+  /**
+   * Subclasses set these in constructor. Then, these
+   * get attached to the global document in _connect
+   *
+   * eventname is a DOM event name like "mouseup", "touchstart", etc.
+   * The callback is simply registered to the event callback with bubbling.
+   *
+   */
   protected _eventToProcessingFunction: { [eventName: string]: (e: Event) => any; } = {};
   private _eventNameToCallbackSet: { [eventName: string]: Utils.CallbackSet<Function>; } = {};
   private _connected = false;
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -12,7 +12,6 @@ export class Dispatcher {
    *
    * eventname is a DOM event name like "mouseup", "touchstart", etc.
    * The callback is simply registered to the event callback with bubbling.
-   *
    */
   protected _eventToProcessingFunction: { [eventName: string]: (e: Event) => any; } = {};
   private _eventNameToCallbackSet: { [eventName: string]: Utils.CallbackSet<Function>; } = {};
--- a/src/utils/transformAwareTranslator.ts
+++ b/src/utils/transformAwareTranslator.ts
@@ -32,14 +32,12 @@ export class Translator {
     this._measurer = rootContent.append<SVGRectElement>("rect")
         .classed("measurer", true)
         .styles({
-          border: "1px solid red",
-          fill: "none !important",
-          height: "1px",
           opacity: 0,
-          "pointer-events": "none",
-          position: "absolute",
-          stroke: "none !important",
-          width: "1px",
+          visibility: "hidden",
+        }).attrs({
+          height: 1,
+          "stroke-width": 0,
+          width: 1,
         });
   }
 
--- a/src/utils/transformAwareTranslator.ts
+++ b/src/utils/transformAwareTranslator.ts
@@ -23,6 +23,12 @@ export function getTranslator(component: Component): Translator {
   return translator;
 }
 
+ /**
+ * The translator implements CSS transform aware event measuring. When a Component is rendered an element that
+  * has a css3 transform applied, mouse/touch interactions will be totally misaligned with the actual chart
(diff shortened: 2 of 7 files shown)


src/utils/transformAwareTranslator.ts
```diff
@@ -12,41 +12,33 @@ export function getTranslator(component: Component): Translator {
...
+  constructor(rootContent: SimpleSelection<void>) {
+    // Use a "rect" element inside the root content node to resolve https://github.com/palantir/plottable/issues/3355
+    this._measurer = rootContent.append<SVGRectElement>("rect")
+        .classed("measurer", true)
+        .styles({
+          opacity: 0,
+          visibility: "hidden",
+        }).attrs({
+          height: 1,
+          "stroke-width": 0,
```
themadcreator, line 45: does d3 support the camel case convention of using `strokeWidth`?

test/dispatchers/touchDispatcherTests.ts
```diff
@@ -52,6 +52,7 @@ describe("Dispatchers", () => {
         div = TestMethods.generateDiv(DIV_WIDTH, DIV_HEIGHT);
         const component = new Plottable.Component();
         component.anchor(div);
+        component.computeLayout();
```
themadcreator, line 55: how were the tests failing that required you to add this? will users of plottable be required to add this to get their touch/mouse events to work?

test/utils/transformAwareTranslatorTests.ts
```diff
@@ -7,73 +7,33 @@ import * as Plottable from "../../src";
...
-        const t1 = Plottable.Utils.getTranslator(component);
-        assert.isNotNull(t1, "created a new Translator on a <svg>");
-        const t2 = Plottable.Utils.getTranslator(component);
-        assert.strictEqual(t1, t2, "returned the existing Translator if called again with same <svg>");
-
-        div.remove();
-      });
-
-    it("converts points to html-space correctly", () => {
-        const div = TestMethods.generateDiv();
```
themadcreator, line 24: was this test no longer relevant or just broken?

src/dispatchers/dispatcher.ts
```diff
@@ -6,6 +6,14 @@
 import * as Utils from "../utils";
 
 export class Dispatcher {
+  /**
+   * Subclasses set these in constructor. Then, these
+   * get attached to the global document in _connect
+   *
+   * eventname is a DOM event name like "mouseup", "touchstart", etc.
+   * The callback is simply registered to the event callback with bubbling.
+   *
```
themadcreator: remove empty line

src/utils/transformAwareTranslator.ts
```diff
@@ -12,41 +12,33 @@ export function getTranslator(component: Component): Translator {
...
 export class Translator {
   private static SAMPLE_DISTANCE = 100;
-  private _measurementElement: SimpleSelection<void>;
-
-  constructor(measurementElement: SimpleSelection<void>) {
-    this._measurementElement = measurementElement;
+  private _measurer: d3.Selection<SVGRectElement, any, any, any>;
+
+  constructor(rootContent: SimpleSelection<void>) {
+    // Use a "rect" element inside the root content node to resolve https://github.com/palantir/plottable/issues/3355
```
themadcreator, line 37: this comment is helpful but i would like a comment on the class explaining what the `Translator` does (i.e. append a `<rect>`) and why that's necessary for browser support.

Especially because constructing a new instance has DOM side-effects.

themadcreator (changes requested):
Nits, questions, small requests

src/utils/transformAwareTranslator.ts
hellochar: Nope, that's a react special thing; `attrs` passes things directly to `domElement.setAttribute`

test/dispatchers/touchDispatcherTests.ts
hellochar: 1. This is only to fix phantomjs - when a component i s anchored but not computeLayout-ed, it has `position: absolute` but no width/height/left/top properties. phantom treats this as a 0x0 div and forces all DOM elements, including the transform measurer, to be "inside" that 0x0 box, so getBoundingClientRect() doesn't change and our measuring algorithm fails. But chrome/ie/ff all do the "right" thing here so it doesn't really matter.
1. no one anchors without computeLayout; this is a leftover of Plottable's API but we should eventually just merge these (everyone just uses .renderTo() which does both immediately)

Users of plottable don't have to change anything

test/utils/transformAwareTranslatorTests.ts
hellochar: No longer relevant; we don't have notions of "html-space" vs "svg-space" (it's just "Component space")

test/dispatchers/touchDispatcherTests.ts
themadcreator: Roger that

themadcreator (approved):
Nice comments

PR #3375, opened 2017-06-28
Use clientwidth/height for screenspace equivalent coordinates

themadcreator:


themadcreator:
Before:
browser-zoom-bad-sm

After:
browser-zoom-good-sm

hellochar (approved):

PR #3379, opened 2017-07-21
Canvas all the things

themadcreator:
Add area canvas renderer.
Drawer-based entity bounds for selection and tooltips.

### DIFF
--- a/src/core/interfaces.ts
+++ b/src/core/interfaces.ts
@@ -101,7 +101,7 @@ export interface IEntityBounds {
  * An object representing a data-backed visual entity inside a Component.
  */
 export interface IEntity<C extends Component> {
-  bounds?: IEntityBounds; // TODO make non-optional
+  bounds: IEntityBounds;
   component: C;
   datum: any;
   position: Point;
--- a/src/core/interfaces.ts
+++ b/src/core/interfaces.ts
@@ -88,11 +88,22 @@ export type Bounds = {
 };
 
 /**
+ * Client bounds
+ */
+export interface IEntityBounds {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+}
+
+/**
  * An object representing a data-backed visual entity inside a Component.
  */
 export interface IEntity<C extends Component> {
+  bounds?: IEntityBounds; // TODO make non-optional
+  component: C;
   datum: any;
   position: Point;
   selection: SimpleSelection<any>;
-  component: C;
 }
--- a/src/drawers/areaDrawer.ts
+++ b/src/drawers/areaDrawer.ts
@@ -35,12 +35,12 @@ export function makeAreaCanvasDrawStep(
       "fill",
       "opacity",
       "stroke-width",
-      "area-fill-opacity",
-      "area-stroke-opacity",
+      "fill-opacity",
+      "stroke-opacity",
     ], data[0], 0);
     const baseOpacity = style["opacity"] != null ? style["opacity"] : 1;
-    const fillOpacity = style["area-fill-opacity"] != null ? style["area-fill-opacity"] : DEFAULT_AREA_FILL_OPACITY;
-    const strokeOpacity = style["area-stroke-opacity"] != null ? style["area-stroke-opacity"] : DEFAULT_AREA_STROKE_OPACITY;
+    const fillOpacity = style["fill-opacity"] != null ? style["fill-opacity"] : DEFAULT_AREA_FILL_OPACITY;
+    const strokeOpacity = style["stroke-opacity"] != null ? style["stroke-opacity"] : DEFAULT_AREA_STROKE_OPACITY;
 
     const d3Area = d3AreaFactory();
     context.save();
--- a/src/drawers/areaDrawer.ts
+++ b/src/drawers/areaDrawer.ts
@@ -4,7 +4,6 @@
  */
 
 import { AttributeToAppliedProjector, SimpleSelection } from "../core/interfaces";
-import * as Utils from "../utils";
 import { CanvasDrawStep, renderArea, resolveAttributes } from "./canvasDrawer";
 import { SVGDrawer } from "./svgDrawer";
 
@@ -26,15 +25,9 @@ export class AreaSVGDrawer extends SVGDrawer {
 
 const AREA_FILL_ATTRS = [ "fill", "opacity", "fill-opacity" ];
 
-const DEFAULT_AREA_FILL_STYLE = {
-  opacity: 1,
-  "fill-opacity": 0.25,
-};
-
 export function makeAreaCanvasDrawStep(d3AreaFactory: () => d3.Area<any>): CanvasDrawStep {
   return (context: CanvasRenderingContext2D, data: any[][], projector: AttributeToAppliedProjector) => {
     const attrs = resolveAttributes(projector, AREA_FILL_ATTRS, data[0], 0);
-    const fillStyle = Utils.Object.assign({}, DEFAULT_AREA_FILL_STYLE, attrs);
-    renderArea(context, d3AreaFactory(), data[0], fillStyle);
+    renderArea(context, d3AreaFactory(), data[0], attrs);
   };
 }
--- a/src/drawers/areaDrawer.ts
+++ b/src/drawers/areaDrawer.ts
@@ -28,15 +28,15 @@ export function makeAreaCanvasDrawStep(
     d3LineFactory: () => d3.Line<any>,
 ): CanvasDrawStep {
   return (context: CanvasRenderingContext2D, data: any[][], attrToAppliedProjector: AttributeToAppliedProjector) => {
+    const areaStyle = resolveAttributes(attrToAppliedProjector, ["fill", "opacity", "stroke-width"], data[0], 0);
+    const opacity = areaStyle["opacity"] ? areaStyle["opacity"] : 1;
     const d3Area = d3AreaFactory();
     context.save();
     context.beginPath();
     d3Area.context(context);
     d3Area(data[0]);
     context.lineJoin = "round";
-    const areaStyle = resolveAttributes(attrToAppliedProjector, ["fill", "opacity"], data[0], 0);
-    areaStyle['opacity'] = areaStyle['opacity'] ? areaStyle['opacity'] : 1;
-    areaStyle['opacity'] *= 0.25;
+    areaStyle["opacity"] = opacity * 0.25;
     styleContext(context, areaStyle);
     context.restore();
 
@@ -46,10 +46,10 @@ export function makeAreaCanvasDrawStep(
     d3Line.context(context);
     d3Line(data[0]);
     context.lineJoin = "round";
-    const lineStyle = resolveAttributes(attrToAppliedProjector, ["stroke", "opacity"], data[0], 0);
-    lineStyle['opacity'] = lineStyle['opacity'] ? lineStyle['opacity'] : 1;
-    lineStyle['opacity'] *= 0.8;
-    styleContext(context, lineStyle);
+    areaStyle["stroke"] = areaStyle["fill"];
+    delete areaStyle["fill"];
+    areaStyle["opacity"] = opacity * 0.8;
+    styleContext(context, areaStyle);
     context.restore();
   };
 }
--- a/src/drawers/areaDrawer.ts
+++ b/src/drawers/areaDrawer.ts
@@ -3,7 +3,8 @@
  * @license MIT
  */
 
-import { SimpleSelection } from "../core/interfaces";
+import { AttributeToAppliedProjector, SimpleSelection } from "../core/interfaces";
+import { CanvasDrawStep, resolveAttributes, styleContext } from "./canvasDrawer";
 import { SVGDrawer } from "./svgDrawer";
 
 export class AreaSVGDrawer extends SVGDrawer {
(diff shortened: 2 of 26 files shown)


src/drawers/areaDrawer.ts
```diff
@@ -21,3 +22,34 @@ export class AreaSVGDrawer extends SVGDrawer {
...
+  return (context: CanvasRenderingContext2D, data: any[][], attrToAppliedProjector: AttributeToAppliedProjector) => {
+    const d3Area = d3AreaFactory();
+    context.save();
+    context.beginPath();
+    d3Area.context(context);
+    d3Area(data[0]);
+    context.lineJoin = "round";
+    const areaStyle = resolveAttributes(attrToAppliedProjector, ["fill", "opacity"], data[0], 0);
+    areaStyle['opacity'] = areaStyle['opacity'] ? areaStyle['opacity'] : 1;
+    areaStyle['opacity'] *= 0.25;
```
hellochar: why are we multiplying by 0.25?

src/drawers/areaDrawer.ts
```diff
@@ -21,3 +22,34 @@ export class AreaSVGDrawer extends SVGDrawer {
...
+
+    const d3Line = d3LineFactory();
+    context.save();
+    context.beginPath();
+    d3Line.context(context);
+    d3Line(data[0]);
+    context.lineJoin = "round";
+    const lineStyle = resolveAttributes(attrToAppliedProjector, ["stroke", "opacity"], data[0], 0);
+    lineStyle['opacity'] = lineStyle['opacity'] ? lineStyle['opacity'] : 1;
+    lineStyle['opacity'] *= 0.8;
```
hellochar: why are we multiplying by 0.8?

src/drawers/areaDrawer.ts
```diff
@@ -21,3 +22,34 @@ export class AreaSVGDrawer extends SVGDrawer {
...
+    d3Area.context(context);
+    d3Area(data[0]);
+    context.lineJoin = "round";
+    const areaStyle = resolveAttributes(attrToAppliedProjector, ["fill", "opacity"], data[0], 0);
+    areaStyle['opacity'] = areaStyle['opacity'] ? areaStyle['opacity'] : 1;
+    areaStyle['opacity'] *= 0.25;
+    styleContext(context, areaStyle);
+    context.restore();
+
+    const d3Line = d3LineFactory();
```
hellochar: We should call lineCanvasDrawStep here instead of copy/paste. The goal should be composition of draw steps

src/drawers/canvasDrawer.ts
```diff
@@ -73,19 +86,27 @@ export function styleContext(context: CanvasRenderingContext2D, attrs: {[key: st
...
   }
+
+  if (attrs[ContextStyleAttrs.opacity]) {
+    context.globalAlpha = parseFloat(attrs[ContextStyleAttrs.opacity]);
+  } else {
+    context.globalAlpha = 1;
+  }
+
+  if (attrs[ContextStyleAttrs.strokeWidth]) {
+    context.lineWidth = parseFloat(attrs[ContextStyleAttrs.strokeWidth]);
```
hellochar: this is the same code as line 86

src/drawers/canvasDrawer.ts
```diff
@@ -73,19 +86,27 @@ export function styleContext(context: CanvasRenderingContext2D, attrs: {[key: st
   if (attrs[ContextStyleAttrs.strokeWidth]) {
     context.lineWidth = parseFloat(attrs[ContextStyleAttrs.strokeWidth]);
   }
+
+  if (attrs[ContextStyleAttrs.opacity]) {
+    context.globalAlpha = parseFloat(attrs[ContextStyleAttrs.opacity]);
```
hellochar: Why globalAlpha vs. setting the stroke/fill opacity?

src/plots/areaPlot.ts
```diff
@@ -147,10 +147,40 @@ export class Area<X> extends Line<X> {
...
-    return new ProxyDrawer(() => new AreaSVGDrawer(), () => {
-      warn("canvas renderer not implemented on Area Plot!");
-    });
+  protected _createDrawer(dataset: Dataset) {
+    return new ProxyDrawer(
+      () => new AreaSVGDrawer(),
+      (ctx) => {
+        return new Drawers.CanvasDrawer(ctx, makeAreaCanvasDrawStep(
+          () => {
+            const xProjector = Plot._scaledAccessor(this.x());
```
hellochar, line 160: this looks copy-pasted from somewhere, we should DRY it?

src/core/interfaces.ts
```diff
@@ -88,11 +88,22 @@ export type Bounds = {
 };
 
 /**
+ * Client bounds
+ */
+export interface IEntityBounds {
```
hellochar, line 93: We should not have both a `Bounds` type (on line 85) and this interface; lets DRY it

src/core/interfaces.ts
```diff
@@ -88,11 +88,22 @@ export type Bounds = {
...
+  y: number;
+  width: number;
+  height: number;
+}
+
+/**
  * An object representing a data-backed visual entity inside a Component.
  */
 export interface IEntity<C extends Component> {
+  bounds?: IEntityBounds; // TODO make non-optional
```
hellochar: This API feels very specific to rects and implementing tooltips for rects. Could we instead expose the resolved attrs associated with this Entity?

src/drawers/areaDrawer.ts
```diff
@@ -21,3 +22,34 @@ export class AreaSVGDrawer extends SVGDrawer {
...
+): CanvasDrawStep {
+  return (context: CanvasRenderingContext2D, data: any[][], attrToAppliedProjector: AttributeToAppliedProjector) => {
+    const d3Area = d3AreaFactory();
+    context.save();
+    context.beginPath();
+    d3Area.context(context);
+    d3Area(data[0]);
+    context.lineJoin = "round";
+    const areaStyle = resolveAttributes(attrToAppliedProjector, ["fill", "opacity"], data[0], 0);
+    areaStyle['opacity'] = areaStyle['opacity'] ? areaStyle['opacity'] : 1;
```
hellochar: Don't we handle not having an opacity property in `styleContext` anyways?

src/drawers/canvasDrawer.ts
```diff
@@ -20,6 +20,9 @@ export type CanvasDrawStep = (
  * This class is immutable (but has internal state) and shouldn't be extended.
  */
 export class CanvasDrawer implements IDrawer {
+  protected _lastDrawnData: any[];
+  protected _lastProjector: AttributeToAppliedProjector;
```
hellochar: Would avoid adding state here -- stateful drawers complicate thing (such as how we now need a drawerFactory in ProxyDrawer). This state belongs in the Plot itself I think.

hellochar (changes requested):

src/drawers/areaDrawer.ts
themadcreator: so the area's opacity can be set externally to display selection. This mimics what CSS was doing because it was using the SVG fill-opacity attribute in plottable as well as CSS opacity from bp-chart to further reduce opacity when another series is selected.

src/drawers/areaDrawer.ts
themadcreator: they handle opacity slightly differently

src/drawers/canvasDrawer.ts
themadcreator: this is the simplest way to allow the drawer to report the bounds of the element. this is necessary for entityNearest to work with canvas drawers

src/drawers/canvasDrawer.ts
themadcreator: whoops

src/drawers/areaDrawer.ts
themadcreator: also, canvas drawers only has one drawstep

src/drawers/areaDrawer.ts
themadcreator: well yeah but i need a default to multiply opacity to get different opacities for line and area paths

src/drawers/areaDrawer.ts
themadcreator: same reason as above

src/drawers/areaDrawer.ts
hellochar: the opacity should be configurable and bp-chart should set it to 0.25; we shouldn't bake in 0.25 opacity into our core drawing code

src/drawers/areaDrawer.ts
hellochar: Can we generalize/abstract the code to achieve composition? I'm really not a fan of copy/paste here

src/plots/areaPlot.ts
themadcreator: 👍

src/core/interfaces.ts
themadcreator: This is replacing `SvgElement.getBBox()` with a compatible call implemented by the canvas drawer. Sadly `Bounds` has a different interface that doesn't match. We can't change the DOM's API so I think we still need this.

src/core/interfaces.ts
themadcreator: this is very close to the resolved attrs, just a subset including "x", "y", "width", and "height". It's used for .entityNearest and .entityIntersecting for tooltips and selections.  It is specific to entities, which is why it's here.

src/drawers/canvasDrawer.ts
hellochar: Adding more state to canvasDrawer is a hack imo. Plottable has a big issue where our interface's responsibilities aren't well defined or respected, and this causes all sorts of broken behavior, illogical code, perf bugs, etc. We already have three representations of "drawn data": IPlotEntity, attrToProjector, and VisualPrimitive DOM elements. Consolidating this will give us 100x the return.

With that in mind, I think there are two things that we should do here:

1. The Plot calls `resolveAttributesSubsetWithStyles`, store that, and throw that into the Drawers - this keeps the Drawers stateless and puts the idea of a "last draw" only within Plot.

2. We refactor IDrawer's `getVisualPrimitives` to return our own `VisualPrimitive` interface (instead of a DOM Element), which has the bounds (again, preferably it has the resolved attrs directly so it's not rect specific), and use that everywhere.

Thoughts? This ask is obviously outside the scope of this PR but I want us to have a plan to fix this before we add more hack

themadcreator: The reason i did it this way was to avoid too much unnecessary duplicate computation.

The SVG drawer doesn't need to invoke `resolveAttributesSubsetWithStyles` here -- it already does it when it renders and it'll just use the element's BBox. Furthermore, the various canvas drawers use `resolveAttributesSubsetWithStyles` in different ways or not at all. So, this approach allows us to only compute the bounds lazily. And even though we are storing this state, it's a shallow copy--whereas precomputing with `resolveAttributesSubsetWithStyles` has a perf penalty.

I agree that there are several improvements we can make across the various places that use entity bounds. I agree we should consolidate. I think both drawers (svg & canvas) as well as the plot code that iterates over entities could benefit from a refactor here.

This change was made without refactoring mainly for expediency.

src/plots/areaPlot.ts
```diff
@@ -119,6 +118,10 @@ export class Area<X> extends Line<X> {
   }
 
   protected _additionalPaint() {
+    if (this.renderer() === "canvas") {
```
hellochar: We should avoid if/elses on the renderer. Why are we doing this?

src/plots/areaPlot.ts
```diff
@@ -230,13 +221,19 @@ export class Area<X> extends Line<X> {
...
+      Plot._scaledAccessor(this.x()),
       Plot._scaledAccessor(this.y()),
-      Plot._scaledAccessor(this.y0()));
+      Plot._scaledAccessor(this.y0()),
+    );
     return propertyToProjectors;
   }
 
   public selections(datasets = this.datasets()): SimpleSelection<any> {
+    if (this.renderer() === "canvas") {
```
hellochar, line 229: same question about if/elses on the renderer

hellochar (changes requested):

src/plots/areaPlot.ts
themadcreator: the area plot refers to an additional SVG drawstep for the line. this is not needed when using canvas rendering. so, i just shortcircuit in these cases.

src/plots/areaPlot.ts
themadcreator: super.selections does the same, but this needs to short-circuit as well.

themadcreator: otherwise it fails to return nodes and causes and exception

src/plots/areaPlot.ts
themadcreator: without this, it adds the SVG on top of the canvas

hellochar: We should configure the additional drawstep to use canvas to draw the line (that achieves the same goal as the "draw step composition") and delete the copy/pasted code in area renderer. This could fix the `selections()` method too

src/drawers/areaDrawer.ts
```diff
@@ -21,3 +23,18 @@ export class AreaSVGDrawer extends SVGDrawer {
     return super.getVisualPrimitiveAtIndex(0);
   }
 }
+
+const AREA_FILL_ATTRS = [ "fill", "opacity", "fill-opacity" ];
+
+const DEFAULT_AREA_FILL_STYLE = {
+  opacity: 1,
+  "fill-opacity": 0.25,
```
hellochar: We shouldn't need this; barPlot already sets:

```
    this.attr("fill-opacity", 0.25);
```

in the constructor

tslint.json
```diff
@@ -9,7 +9,7 @@
     ],
     "interface-name": [true, "always-prefix"],
     "ordered-imports": true,
-    "no-boolean-literal-compare": true,
+    "no-boolean-literal-compare": false,
```
hellochar: nit: can we delete this line?

src/drawers/canvasDrawer.ts
```diff
@@ -20,6 +20,9 @@ export type CanvasDrawStep = (
  * This class is immutable (but has internal state) and shouldn't be extended.
  */
 export class CanvasDrawer implements IDrawer {
+  protected _lastDrawnData: any[];
```
hellochar: i thought we removed this?

themadcreator: oh yeah, we don't need this, will remvoe.

hellochar (approved):

PR #3380, opened 2017-07-21
only call barPlot's updateBarPixelWidth on computeLayout

### LINKED ISSUE #3219: bar and stackedbar plots are slow
Sorry if this is already posted. My problem is that the drawing of bar plots, especially stackedbar are extremely slow compared to e.g. line. Created a JSFiddle to demonstrate this:

https://jsfiddle.net/BaEigner/17yq0zne/

It draws 3 different charts:

- one with two bar plots mimicking the stackedbars
- one with a simple stackedbar plot
- and one with two line plots for performance comparison

Performance for each is measured:

- the cost of plot creation
- the render time
- and the total time taken

It draws 100 points, but you can change it in the first line. So the result are:

My laptop (2015 15" RMBP, OS X El Capitane, Chrome): 
1.1 Bar Plot 1: 29 ms
1.2 Bar Plot 2: 53 ms
1.3 Bar Render: 453 ms
**1.0 Bar Total: 563 ms**
2.1 StackedBar Plot: 300 ms
2.2 StackedBar Render: 782 ms
**2.0 StackedBar Total: 1160 ms**
3.1 Line Plot 1: 2 ms
3.2 Line Plot 2: 1 ms
3.3 Line Render: 27 ms
**3.0 Line Total: 35 ms**

Our webserver (Windows Server 2012, 4 core Xenon @2,6GHz, Chrome -- but it not really counts...):
1.1 Bar Plot 1: 126 ms
1.2 Bar Plot 2: 217 ms
1.3 Bar Render: 1663 ms
**1.0 Bar Total: 2113 ms**
2.1 StackedBar Plot: 900 ms
2.2 StackedBar Render: 3982 ms
**2.0 StackedBar Total: 5210 ms**
3.1 Line Plot 1: 6 ms
3.2 Line Plot 2: 6 ms
3.3 Line Render: 86 ms
**3.0 Line Total: 116 ms** 

A random windows laptop at the company (Chrome): 
1.1 Bar Plot 1: 38 ms
1.2 Bar Plot 2: 62 ms
1.3 Bar Render: 614 ms
**1.0 Bar Total: 752 ms**
2.1 StackedBar Plot: 350 ms
2.2 StackedBar Render: 1052 ms
**2.0 StackedBar Total: 1500 ms**
3.1 Line Plot 1: 2 ms
3.2 Line Plot 2: 1 ms
3.3 Line Render: 24 ms
**3.0 Line Total: 29 ms**

Are there any workarounds, fixes or anything to make the bars faster? If you draw one of them it's let's say, okay, but in some cases we have to draw 10-20 at a time.

hellochar:
BarPlot.render used to call _updateBarPixelWidth and _updateExtents (because the extents depend on the bar pixel width) which would send Plottable into a loop:

BarPlot.render -> updateExtents -> Scale.autoDomain -> Scale.dispatchUpdate -> Plot._renderCallback -> BarPlot.render

This results in unnecessary computation and slowdown.

The reason .render() called _updateBarPixelWidth to begin with is because the bar pixel width is computed based off several inputs and detecting when the input has updated is not always easy. It's pretty much a rats nest of knowing when the scale domain, scale range, other scale properties, plot dimensions, or data, has changed. Putting _updateBarPixelWidth in .render was a sort of catch-all that made sure it was always "up to date", but wastefully so. This PR does the opposite - it removes the catch all to reduce wasted computation.

A side effect is that _updateBarPixelWidth may now be stale/out of date (aka we're introducing some regressions). We should tackle this in an upcoming PR (the groundwork for this change is why computeBarPixelThickness has been refactored).

I've also deleted ModifiedLog Bar Plot tests since the behavior of "the pixel width affects the extent" is really tricky to get right here, and it doesn't affect the perf use case.

Before this change: many_bars took **3.01s**
After this change: many_bars took **0.39s**

Fixes #3219

### DIFF
--- a/quicktests/overlaying/tests/realistic/many_bars.js
+++ b/quicktests/overlaying/tests/realistic/many_bars.js
@@ -1,0 +1,42 @@
+
+function makeData() {
+    "use strict";
+    var JAN_01_2012 = new Date(2012, 0, 0).valueOf();
+    var MILLIS_IN_DAY = 1000 * 60 * 60 * 24;
+    var data1 = [];
+    var data2 = [];
+    var data3 = [];
+    for (var i = 0; i < 1000; i++) {
+        data1.push({
+            x: new Date(JAN_01_2012 + MILLIS_IN_DAY * i),
+            y: Math.random() * 50 - 25,
+        });
+        data2.push({
+            x: new Date(JAN_01_2012 + MILLIS_IN_DAY * i),
+            y: Math.random() * 50 - 25,
+        });
+        data3.push({
+            x: new Date(JAN_01_2012 + MILLIS_IN_DAY * i),
+            y: Math.random() * 50 - 25,
+        });
+    }
+    return [data1, data2, data3];
+}
+
+function run(div, datas, Plottable) {
+    "use strict";
+
+    var xScale = new Plottable.Scales.Time();
+    var yScale = new Plottable.Scales.Linear();
+    var colorScale = new Plottable.Scales.Color();
+
+    var ds1 = new Plottable.Dataset(datas[0]).metadata(1);
+    var ds2 = new Plottable.Dataset(datas[1]).metadata(2);
+    var ds3 = new Plottable.Dataset(datas[2]).metadata(3);
+    var plot = new Plottable.Plots.StackedBar()
+        .datasets([ds1, ds2, ds3])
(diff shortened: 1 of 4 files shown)


themadcreator (approved):
Pretty straightforward. Let's get this party started

Issue #3302, opened 2017-03-20
Improve caching story


hellochar:
The code exhibits poor behavior around caching:

* Different attempts/implementations at caching
* Other concepts that aren't called caching per-se but actually are
* Cache invalidation that doesn't catch all inputs
* Unnecessary cache recomputation which results in thrown away work
* Not caching computing properties that probably should be cached

We need to think deeply and figure out a good story around how to solve caching - the basic concepts and tools that we build our code around. I think getting this right will improve perf, code maintainability, make clearer semantics, and catch bugs.

Issue #2949, opened 2015-11-07
[Plot] Adding a Dataset is not idempotent


jtlan:
An added `Dataset` is first removed if it's already in the `Plot`, causing multiple updates. It would probably be more efficient to to no-op if the `Dataset` is already in the `Plot`.

PR #3383, opened 2017-07-25
memoize bar pixel thickness, POC memoizing and Signature API

hellochar:
Improve perf of _computeBarPixelThickness by memoizing gets to it, reducing intermediate wasted computes (intermediate computes are computations that are called before renderImmediately).

Implementing memoization in Plottable is complicated by mutable scales and datasets. To resolve this we write a recursive Signature interface that holds an immutable snapshot of whatever state the scale/data was in at the time. Then on memoized function invocation we sign the new inputs and compare the signatures to decide if we should recompute.

Unfortunately signing and comparison is itself a performance hit; in tests they took ~0.3ms per get (which, for 3k bars, is 900ms of overhead), which is unacceptable. To get around this, we notice that almost all of these gets happen inside Plot.renderImmediately (aka getting the thickness of the bar to actually render the bar). Thus we introduce the "doLocked" method that momentarily bypasses sign/comparison logic and simply returns the cached value.

This paves the way for more perf improvements as we refactor other code to using the sign and memoize API - e.g. #2495, #3117, #2949

Starts tackling #3302

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -71,8 +71,6 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
    * If so, use the _barPixelThickness property to access the thickness.
    */
   private _fixedBarPixelThickness = true;
-  private _barPixelThickness = 0;
-  private _updateBarPixelThicknessCallback: () => void;
 
   /**
    * A Bar Plot draws bars growing out from a baseline to some value
@@ -89,15 +87,13 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
     this._isVertical = orientation === "vertical";
     this.animator("baseline", new Animators.Null());
     this.attr("fill", new Scales.Color().range()[0]);
-    this.attr(Bar._BAR_THICKNESS_KEY, () => this._barPixelThickness);
+    this.attr(Bar._BAR_THICKNESS_KEY, () => this._barPixelThickness());
     this._labelConfig = new Utils.Map<Dataset, LabelConfig>();
     this._baselineValueProvider = () => [this.baselineValue()];
-    this._updateBarPixelThicknessCallback = () => this._updateBarPixelThickness();
   }
 
   public computeLayout(origin?: Point, availableWidth?: number, availableHeight?: number) {
     super.computeLayout(origin, availableWidth, availableHeight);
-    this._updateBarPixelThickness();
     this._updateExtents();
     return this;
   }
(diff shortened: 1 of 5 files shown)


src/plots/barPlot.ts
```diff
@@ -848,22 +844,24 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
...
             this.position(),
             this.datasets(),
             this._isVertical ? this.width() : this.height(),
         );
       } else {
-        this._barPixelThickness = 0;
+        return 0;
       }
+    } else {
+      // throw new Error("called _barPixelThickness when thickness isn't fixed!");
```
themadcreator, line 863: do you want this? if so, uncomment, if not, remove.

test/plots/barPlotTests.ts
```diff
@@ -212,7 +212,7 @@ describe("Plots", () => {
       });
 
       describe(`auto bar width calculation when ${orientation}`, () => {
-        const scaleTypes = ["Linear", "Time"];
+        const scaleTypes = ["Linear", "ModifiedLog", "Time"];
```
themadcreator, line 215: the return!

src/plots/barPlot.ts
```diff
@@ -948,9 +945,182 @@ function computeBarPixelThickness(
...
+   *
+   * Be sure to force the memoization to the value
+   * you want before calling this!
+   *
+   * @param fn
+   */
+  doLocked<T>(fn: () => T): T;
+};
+
+function memoize<F extends Function>(compute: F): LockedMemoize<F> {
```
themadcreator, line 1099: move these classes and methods to its own file

src/plots/barPlot.ts
```diff
@@ -948,9 +945,182 @@ function computeBarPixelThickness(
...
+  memoizeFn.doLocked = function<T>(cb: () => T) {
+    if (locked) {
+      throw new Error("Locking an already locked memoize function!");
+    }
+    locked = true;
+    const retVal = cb.apply(this);
+    locked = false;
+    return retVal;
+  };
+  return memoizeFn;
```
themadcreator, line 1125: this is a lot of new code that is fairly complex. would be amazing to get any sort of test for it.

themadcreator (commented):
Should we block release until this goes in? We could always do a 3.4.2

src/plots/barPlot.ts
hellochar: Fixed in #3385

src/plots/barPlot.ts
hellochar: Fixed in #3385

src/plots/barPlot.ts
hellochar: Fixed in #3385

themadcreator (approved):

PR #3385, opened 2017-07-27
memoize compute extents

hellochar:
Memoize calls to computing the plot extents (which does a linear pass through the data), improving perf. This also lets us delete a bunch of _update methods and move away from the "update" paradigm


src/memoize/memoize.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright 2014-present Palantir Technologies
```
themadcreator, line 2: technically i think these should be 2017-present

themadcreator (approved):

src/memoize/memThunk.ts
```diff
@@ -0,0 +1,55 @@
...
+    arg2: Thunk<I2>,
+    arg3: Thunk<I3>,
+    compute: (this: void, arg1: I1, arg2: I2, arg3: I3) => O,
+): MemoizedFunction<Thunk<O>>;
+export function memThunk<I1, I2, I3, I4, O>(
+    arg1: Thunk<I1>,
+    arg2: Thunk<I2>,
+    arg3: Thunk<I3>,
+    arg4: Thunk<I4>,
+    compute: (this: void, arg1: I1, arg2: I2, arg3: I3, arg4: I4) => O,
```
themadcreator, line 31: you know what we need? variadic template types. dang.

src/memoize/signature.ts
```diff
@@ -0,0 +1,178 @@
...
+export class ReferenceSignature extends Signature {
+    constructor(private ref: any) {
+        super();
+    }
+
+    isSignatureDifferent(other: ReferenceSignature) {
+        return this.ref !== other.ref;
+    }
+}
+
```
themadcreator, line 145: `export type ISignatureRecord = Record<string, Signature>;`

src/memoize/signature.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Base signature. Subclasses should implement isSignatureDifferent. All classes
+ * should be immutable.
+ *
+ * Users should only call `isDifferent`, not `isSignatureDifferent`.
+ */
+export abstract class Signature {
+    public isDifferent(other: Signature): boolean {
+        // debugger;
```
themadcreator: stray debugger

src/memoize/signature.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Generic signature factory - pass any value and get a signature for it.
+ *
+ * Datasets and Scales are handled specially - see their respective signing methods.
+ *
+ * If the input is already a signature, simply return it.
+ *
+ * @param a
+ * @returns {Signature}
+ */
+export function sign<T>(a: T): Signature {
```
themadcreator: since you're not using T anywhere else, isn't this as good as `any` ?

src/memoize/memoize.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Turn on perf logging. Useful for debugging.
+     */
+    logPerformance(): this;
+}
+
+export type MemoizedFunction<F extends Function> = F & IMemoizedFunctionMethods;
+
+export function memoize<F extends Function>(compute: F): MemoizedFunction<F> {
```
themadcreator: alright, so this is the actual memoize function that will be the main first thing people read. i'd like the comment to tell me:

how to use it (short code example).
what sort of pitfalls there are (vis a vis signatures). a warning that arguments that are not shallow comparable may need a custom signature.
also,  what memoization locking is and why you would want to use it.

quicktests/overlaying/tests/realistic/scatter_bar.js
```diff
@@ -25,7 +25,7 @@ function run(svg, data, Plottable) {
   var bars = new Plottable.Plots.Rectangle()
     .addDataset(ds1)
     .x(function (d) { return d.overall; }, xScale)
-    .x2(function(d) { return d.leadership; }, xScale)
+    .x2(function(d) { return d.leadership; })
```
themadcreator, line 28: why was this necessary? doesn't calling `.x2` here re-use the `xScale` from the call to `.x` above anyway?

src/memoize/memoize.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (locked) {
+            throw new Error("Locking an already locked memoize function!");
+        }
+        locked = true;
+        const retVal = cb.apply(this);
+        locked = false;
+        return retVal;
+    };
+
+    memoizeFn.logPerformance = function() {
```
themadcreator: since this is debugging only, i doubt you'd care to turn it off, but it might be worth taking an optional argument.

themadcreator: `.logPerformance(logit = true)`

src/plots/plot.ts
```diff
@@ -437,56 +434,49 @@ export class Plot extends Component {
...
-
-    if (scale == null) {
-      return [];
-    }
-
-    let data = dataset.data();
-    if (filter != null) {
-      data = data.filter((d, i) => filter(d, i, dataset));
+  protected getExtentsForAttr(attr: string) {
+    if (this._attrExtents[attr] == null) {
```
themadcreator, line 465: does this ever need to be overrided or deleted once it's defined?

src/plots/plot.ts
```diff
@@ -837,7 +827,7 @@ export class Plot extends Component {
     return {};
   }
 
-  protected static _scaledAccessor<D, R>(binding: Plots.IAccessorScaleBinding<D, R>) {
+  public static _scaledAccessor<D, R>(binding: Plots.IAccessorScaleBinding<D, R>) {
```
themadcreator: public _methodName ? we should rename without underscore. or can we get away with protected?

src/plots/plot.ts
```diff
@@ -859,3 +849,24 @@ export class Plot extends Component {
     return this._animate && this._dataChanged;
   }
 }
+
+function _computeExtent(
```
themadcreator: no need to underscore for unexported functions, i think

themadcreator: or do we do that elsewhere in plottable?

src/plots/stackedAreaPlot.ts
```diff
@@ -19,8 +20,23 @@ import { Plot } from "./plot";
...
 export class StackedArea<X> extends Area<X> {
   private _stackingOrder: Utils.Stacking.IStackingOrder;
-  private _stackingResult: Utils.Stacking.StackingResult;
-  private _stackedExtent: number[];
+  private _stackingResult: Thunk<Utils.Stacking.StackingResult> = memThunk(
+      () => this.datasets(),
+      () => this.x().accessor,
+      () => this.y().accessor,
+      () => this._stackingOrder,
+      (datasets, keyAccessor, valueAccessor, stackingOrder) => {
```
themadcreator, line 28: this is basically reselect's API. should we just us reselect?

src/plots/stackedBarPlot.ts
```diff
@@ -7,12 +7,11 @@ import * as Typesettable from "typesettable";
 
 import { Dataset } from "../core/dataset";
 import { Formatter, identity } from "../core/formatters";
-import { Bounds, IAccessor, Point, SimpleSelection } from "../core/interfaces";
-import { Scale } from "../scales/scale";
+import { Bounds, Point, SimpleSelection } from "../core/interfaces";
+import { memThunk, Thunk } from "../memoize/index";
```
themadcreator: don't think you need the `/index` ?

test/memoize/memoizeTests.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as sinon from "sinon";
+
+import * as Plottable from "../../src";
+
+import { Dataset } from "../../src/core/dataset";
+import { memoize } from "../../src/memoize/memoize";
+import { IAccessorScaleBinding } from "../../src/plots/commons";
+import { Plot } from "../../src/plots/plot";
+
+describe("memoize()", () => {
```
themadcreator, line 11: 💯

src/plots/stackedAreaPlot.ts
themadcreator: i guess it won't quite work if we don't do immutable objects

themadcreator (changes requested):
some minor requests and questions

quicktests/overlaying/tests/realistic/scatter_bar.js
hellochar: The second argument to `x2` is now the `postScale` function; this was breaking the examples

src/plots/plot.ts
hellochar: Nope, this should be defined the same way for all attrs (and will always be defined that way)

src/plots/stackedAreaPlot.ts
hellochar: Yeah we basically rewrote reselect

quicktests/overlaying/tests/realistic/scatter_bar.js
themadcreator: oh, okay

src/plots/stackedAreaPlot.ts
themadcreator: rereselect

themadcreator (approved):

PR #3387, opened 2017-08-18
Improve MouseDispatcher perf

themadcreator:
- Use MouseEvent.offsetX/Y instead of TransformAwareTranslator
- Attach listener to component root instead of document

### DIFF
--- a/src/dispatchers/mouseDispatcher.ts
+++ b/src/dispatchers/mouseDispatcher.ts
@@ -14,7 +14,6 @@ export type MouseCallback = (p: Point, event: MouseEvent) => void;
 
 export class Mouse extends Dispatcher {
   private static _DISPATCHER_KEY = "__Plottable_Dispatcher_Mouse";
-  private _translator: Utils.Translator;
   private _lastMousePosition: Point;
   private static _MOUSEOVER_EVENT_NAME = "mouseover";
   private static _MOUSEMOVE_EVENT_NAME = "mousemove";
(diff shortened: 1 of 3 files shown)


themadcreator:
In a page with many charts, we see a major performance improvement over previous approach.

### Old & Busted
tat-bad

### New Hotness
image

src/dispatchers/mouseDispatcher.ts
```diff
@@ -191,11 +188,11 @@ export class Mouse extends Dispatcher {
       throw new Error("Invalid scope '" + scope + "', must be 'element' or 'page'");
     }
     if (scope === "page" || this.eventInside(component, event)) {
-      const newMousePosition = this._translator.computePosition(event.clientX, event.clientY);
-      if (newMousePosition != null) {
-        this._lastMousePosition = newMousePosition;
-        this._callCallbacksForEvent(eventName, this.lastMousePosition(), event);
-      }
+      const origin = component.originToRoot();
```
hellochar, line 191: should we just delete the Translator? Is it used elsewhere?

hellochar (approved):

src/dispatchers/mouseDispatcher.ts
themadcreator: Yeah, i'll delete

themadcreator: can't delete because it's used for touch events. see follow up PR for perf update there.

Issue #3345, opened 2017-05-13
PointerInteraction doesn't respect the Component's translated bounds


hellochar:
From gitter:

> The pointer interaction translates the point to component space (LinePlot). p: {x:53,y:140}, translatedP: {x: 2.98, y:140}. This makes sense. Then in retrieving the entities they are mapped via _lightweightPlotEntityToPlotEntity and their real ??component?? position is calculated. e.g. my first entity is pos: (x: 10, y: 97) where this pos should be relative to the LinePlot right? not to the whole table.
https://github.com/palantir/plottable/blob/v3.0.0/src/plots/xyPlot.ts#L409
Then in https://github.com/palantir/plottable/blob/v3.1.0/src/plots/linePlot.ts#L488 the point is checked if it is within bounds, but these bounds are relative to the entire Table, not the Line. my bounds.topLeft is (43,0) because of the y-axis

Repro: ![hover missing points](https://cloud.githubusercontent.com/assets/1079495/26020773/d0a3df04-3737-11e7-80dc-99b9c1bbc64d.gif)

1. Open interaction_hover_scatter
2. Mouse around on the bottom-left corner. You may have to refresh the page a few times to see the random data get within the threshold of ignored points

leebyp:
same issue as #3268 for more details

hellochar:
Closing this as dupe of #3268

Issue #3268, opened 2017-03-09
scatterPlot  _entityVisibleOnPlot is incorrect


leebyp:
(ScatterPlot case)
- `DomUtils.intersectsBBox` should have `bbox.x + bbox.width <= xRange.max` and similarly for `yRange`
- the `bounds` passed in are calculated from the origin of the plot **including** the axis, but the entity position excludes the axis

Update:
This is a bigger problem with all plots. The problem is more obvious when trying to use an interaction on a plotB in following group. The entities are relative to plotB's own origin, whereas bounds are relative to the parent.
```
[ [plotA],
  [plotB] ]
```

PR #3388, opened 2017-08-22
Incorrectly excluded entities not considered for nearest entity in interactions #3345 #3268

alechill:
Fixes / relates to https://github.com/palantir/plottable/issues/3345 & https://github.com/palantir/plottable/issues/3268

Plot entity store creation passes incorrect bounds to the entity store (plot's parent space rather than from plot's origin itself), which then can exclude visible entities from being considered for nearest calculation. 

The only other call to `addAll` entities does not pass bounds info, so other than performance reasons is not absolutely necessary and precedence is already set here to not always be needed for perf, so originally I removed it, which fixed the use case, however it seems to be desired for certain plot types (eg scatter) as per the tests, so have done a simple mapping of bounds to become relative to plot's own origin for this particular use case, and also applied it to the missing one

This fix is simple and localised - it doesn't address the wider issue, but for us and presumably for others, this is actually a blocker currently as it is very noticeable by users - for example when showing tooltips, or selecting data points

For the disparity between whether the bounds of a plot in general are to be used in plot or parent space I felt before I could do anything more meaningful to help other than this quick fix, I'd need some input first as it...

1) has API design decisions - bounds in various different places now have different pixel space meanings but how you want to convey that to the user is not my decision as could be a breaking change (ie for use in other methods where bounds are passed in signature - like `_entitityIsVisibleOnPlot` - could default it to use the localised version but that may have repercussions)
2) has too wide an impact on parts of the lib I have not enough knowledge of (ie panning/zooming or how interpolation from the different spaces needs to be done to work with scales etc)

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -12,6 +12,7 @@ import { Dataset, DatasetCallback } from "../core/dataset";
 import {
   AttributeToAppliedProjector,
   AttributeToProjector,
+  Bounds,
   IAccessor,
   IRangeProjector,
   Point,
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @alechill! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/plots/plot.ts
```diff
@@ -828,11 +828,9 @@ export class Plot extends Component {
    * or undefined if no {Plots.PlotEntity} can be found.
    *
    * @param {Point} queryPoint
-   * @param {bounds} Bounds The bounding box within which to search. By default, bounds is the bounds of
-   * the chart, relative to the parent.
    * @returns {Plots.PlotEntity} The nearest PlotEntity, or undefined if no {Plots.PlotEntity} can be found.
    */
-  public entityNearest(queryPoint: Point, bounds = this.bounds()): Plots.IPlotEntity {
```
alechill, line 835: This was never used and gave false hope that it could be provided to hack around this interaction issue externally

If the change was made to pass it along to `_getEntityStore` it would be subject to the same discrepency of what pixel space it is provided in and whether that needs interpolating automatically or up to the user - hence i thought that more of a design decision for you

For these reasons i chose to remove for now instead to avoid confusion

alechill:
I can't fix the build failure as I'm not sure where the github auth token env var is supposed to be coming from

themadcreator:
Thank you for this quality contribution! We get few of these so our build system isn't configured properly. I'm setting up the build to avoid the failure (which is unrelated to your changes).

Stay tuned.

themadcreator:
@alechill can you please merge the latest develop update to trigger a rebuild?

alechill:
@themadcreator thanks, all done

hellochar:
Thanks very much @alechill for the thoughtful PR. I think we can approve this as is and figure out the larger design issues as people run into them. @themadcreator thoughts?

hellochar (approved):

themadcreator (approved):
I appreciate you taking the time to explain your reasoning. The `bounds` clipping in the entity indeed a simple performance optimization, but also to prevent tooltips/selection for entitites outside the plot's view. This is especially likely when panzoom interactions are enabled.

I had a similar feeling about the inconsistency of using a default `bounds` only some of the time, as well as confusion around the coordinate spaces.

Thanks again for this fix!

themadcreator:
[Dev Preview](https://624-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/index.html)

PR #3389, opened 2017-08-24
Reform TransformAwareTranslator for touch events

themadcreator:
Since touch events have `clientX`/`clientY` properties but no `offsetX`/`offsetY` properties, we still need to use something like `TransformAwareTranslator`.

The previous `TransformAwareTranslator` would add, move, and measure a pseudo hidden component, causing awful DOM layout thrashing.

In the new approach,  we compute the a cumulative CSS3 transform matrix of the root element ancestors up to `<body>` and apply it directly to touch event `clientX`/`clientY` values.

### DIFF
--- a/quicktests/overlaying/tests/interactions/css_transforms.js
+++ b/quicktests/overlaying/tests/interactions/css_transforms.js
@@ -1,7 +1,7 @@
 function makeData() {
 }
 
-function run(container, data, Plottable) {
+function run(container) {
   "use strict";
 
   container
@@ -56,7 +56,7 @@ function debugPlot(child) {
     .x(({ x }) => x, xScale)
     .y(({ y }) => y, yScale);
 
-  const pointer = new Plottable.Interactions.Pointer()
+  new Plottable.Interactions.Pointer()
     .onPointerMove((p) => {
       var update = datasetData.slice();
       update.push({
@@ -69,7 +69,7 @@ function debugPlot(child) {
     })
     .attachTo(plot);
 
-  const chart = new Plottable.Components.Table([
+  new Plottable.Components.Table([
     [null, title],
     [yAxis, plot],
     [null, xAxis],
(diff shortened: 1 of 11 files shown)


themadcreator:
[The new css transform preview test](https://598-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/css_transforms) (don't forget to emulate a touch device)

src/utils/mathUtils.ts
```diff
@@ -119,3 +140,101 @@ export function within(p: Point, bounds: Bounds) {
...
+  const determinant = a[0] * a[3] - a[1] * a[2];
+  if (determinant === 0) {
+    throw new Error("singular matrix");
+  }
+  const inverseDeterminant = 1 / determinant;
+  return [
+    inverseDeterminant * a[3],
+    inverseDeterminant * -a[1],
+    inverseDeterminant * -a[2],
+    inverseDeterminant * a[0],
```
hellochar, line 224: 💯 🔥 💯 🔥 💯 🔥

src/utils/transformAwareTranslator.ts
```diff
@@ -1,103 +1,65 @@
-import * as d3 from "d3";
+/**
+ * Copyright 2017-present Palantir Technologies
```
hellochar, line 2: Why only copyright 2017-present?

quicktests/overlaying/tests/interactions/css_transforms.js
```diff
@@ -0,0 +1,110 @@
+function makeData() {
+}
+
+function run(container) {
+  "use strict";
+
```
hellochar, line 6: can we add some text to the example that says "emulate a touch device"?

hellochar (approved):
So to be clear:

1. mouse events look at "offsetX/offsetY" of the mouse event which gives the chart-local coordinates
2. touch events walk the matrix tree, to also give the chart-local coordinates (because offsetX doesn't exist on touch events)

right?

src/utils/transformAwareTranslator.ts
themadcreator: because the copyright didn't exist before and this is the first time i'm adding it.

themadcreator: i suppose i could retroactively date, but i usually just use the current year

PR #3391, opened 2017-08-24
preparing release 3.5.0

hellochar:
# Improvements
- Plot rendering performance across the board has been improved by removing unnecessary computation.
- Rewritten mouse and touch event position computation. Supports CSS3 transforms while avoiding the previous approach's DOM thrashing, improving the performance of interactive Charts for both desktop and mobile. 
- Performance of Bar, ClusteredBar, and StackedBar rendering has been drastically improved by removing unnecessary computation.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "plottable",
   "description": "A modular charting library built on D3",
-  "version": "3.4.1",
+  "version": "3.5.0",
   "main": [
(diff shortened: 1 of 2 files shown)


themadcreator:
- Rewritten mouse and touch event position computation. Supports CSS3 transforms while avoiding the previous approach's DOM thrashing.

themadcreator (approved):

PR #3394, opened 2017-08-25
Bring back TransformAwareTranslator for MouseDispatcher

themadcreator:
Now that the translator doesn't thrash the DOM, we can
use it once again to handle the clientX/clientY coordinate
conversion for components.

Using offsetX/offsetY was producing incorrect results
when the event target was outside the plot.


themadcreator:
[the css transform preview](https://618-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/css_transforms)

hellochar (approved):

PR #3397, opened 2017-08-29
Fix TransformAwareTranslator transform

themadcreator:
The previous implementation walked the `HTMLElement` ancestry
using the `element.offsetParent` chain. However, some elements
in the acenstry may contain scroll offsets that are not
taken into account with this approach.

This new implementation walks the `HTMLElement` ancestry using
`element.parentElement` to include all scroll offsets.
Client/offset offsets are added only for offsetParent elements.

No longer reverse order of elements in the acnestry. Instead
we reverse the order of matrix multiplications.

Removed origin addition in dispatchers since that is taken into account
by offset chain.

Updated unit test stub.


themadcreator:
[full css transforms demo](https://627-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/css_transforms)

[codepen demo](https://codepen.io/TheMadCreator/pen/GvYwpE)

package.json
```diff
@@ -18,7 +18,8 @@
     "check_version": "echo ${npm_package_version} $(npm view ${npm_package_name} version) | awk '{ if ($1 == $2) { exit 1 }}'",
     "dist": "npm-run-all clean build:tsc dist:sed-version",
     "dist:sed-version": "sed -i.bak -e \"s/exports.version = __VERSION__/exports.version = \\\"${npm_package_version}\\\"/\" build/src/core/version.js",
-    "lint": "npm run lint:tslint",
+    "lint": "npm run lint:tslint && npm run lint:eslint",
+    "lint:eslint": "grunt eslint",
```
themadcreator: just for convenience, since i've forgotten to run eslint for dev preview code before

src/dispatchers/mouseDispatcher.ts
```diff
@@ -191,10 +191,7 @@ export class Mouse extends Dispatcher {
     }
     if (scope === "page" || this.eventInside(component, event)) {
       const position = this._translator.computePosition(event.clientX, event.clientY);
-      const origin = component.originToRoot();
-      const x = position.x + origin.x;
-      const y = position.y + origin.y;
```
themadcreator, line 196: this offset is now taken into account by `computePosition`

src/utils/transformAwareTranslator.ts
```diff
@@ -14,7 +14,7 @@ const _TRANSLATOR_KEY = "__Plottable_ClientTranslator";
  * Returns a singleton-ized `Translator` instance associated with the component.
  */
 export function getTranslator(component: Component): Translator {
-  const rootElement = component.element().node() as HTMLElement;
+  const rootElement = component.root().rootElement().node() as HTMLElement;
```
themadcreator, line 17: this matches the `Translator.isEventInside` code

src/utils/mathUtils.ts
```diff
@@ -155,24 +156,31 @@ export function within(p: Point, bounds: Bounds) {
...
-      const midX = elem.clientWidth/2;
-      const midY = elem.clientHeight/2;
-      transform = premultiplyTranslate([-midX, -midY], transform);
-      transform = multiplyMatrix(invertMatrix(elementTransform), transform);
-      transform = premultiplyTranslate([midX, midY], transform);
+      const midX = elem.clientWidth / 2;
+      const midY = elem.clientHeight / 2;
+      transform = multiplyTranslate(transform, [midX, midY]);
+      transform = multiplyMatrix(transform, invertMatrix(elementTransform));
+      transform = multiplyTranslate(transform, [-midX, -midY]);
```
themadcreator, line 170: note that we swap the order of arguments and the order of calls because we now iterate over the ancestry starting from the deepest child up to `<body>` (hence the removal of `elems.reverse()` above. we swapped the ancestry iteration order so we can keep track of the `element.offsetParent`

hellochar (approved):

package.json
```diff
@@ -18,7 +18,8 @@
     "check_version": "echo ${npm_package_version} $(npm view ${npm_package_name} version) | awk '{ if ($1 == $2) { exit 1 }}'",
     "dist": "npm-run-all clean build:tsc dist:sed-version",
     "dist:sed-version": "sed -i.bak -e \"s/exports.version = __VERSION__/exports.version = \\\"${npm_package_version}\\\"/\" build/src/core/version.js",
-    "lint": "npm run lint:tslint",
+    "lint": "npm run lint:tslint && npm run lint:eslint",
```
hellochar: you can use `npm-run-all` for this: `npm-run-all lint:tslint lint:eslint`

themadcreator: oh yeah i guess we do already depend on it

PR #3398, opened 2017-08-30
Release/3.5.2

hellochar:
There's one more thing left to do to finish removing bower: unregister ourselves from the registry. I've filed https://github.com/bower/registry/issues/268 to track.

# v3.5.2

# Bug Fixes
- Fix a regression in mouse and touch interactions on Charts where the page was scrolled.

# Other Changes
- We've officially stopped publishing to Bower. Please use npm or unpkg to consume Plottable.


adidahiya:
why do we need to unregister in the bower registry?

themadcreator (approved):
So this is the end for bower, eh?

PR #3403, opened 2017-09-07
Fast entity bounds queries

themadcreator:
- Add R-Tree and supporting unit tests.
- Use rtree to efficiently test bounds for entity overlap.
- Add API methods for `entitiesInBounds` and as well as x- and y- only versions.
- Switch to using more standard `IEntityBounds`.
- Add more tests.
- Add proper bounds for scatter plot and rect plot entities.
- Fix r-tree split strategy balancing and verify with test.
- Comments.

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -17,6 +17,7 @@ import {
   IEntityBounds,
   IRangeProjector,
   Point,
+  Range,
   SimpleSelection,
 } from "../core/interfaces";
 import * as Drawers from "../drawers";
@@ -108,7 +109,7 @@ export class Plot extends Component {
    * all the dots to size 0), and then they do the "main" animation into the correct visualization
    * (e.g. scatter plot dots grow to their specified size).
    */
-  private _animators: {[animator: string]: IAnimator} = {};
+  private _animators: { [animator: string]: IAnimator } = {};
 
   /**
    * Callback that triggers when any scale that's bound to this plot Updates.
@@ -283,7 +284,7 @@ export class Plot extends Component {
   protected _createDrawer(dataset: Dataset): ProxyDrawer {
     return new ProxyDrawer(
       () => new SVGDrawer("path", ""),
-      (ctx) => new CanvasDrawer(ctx, () => {}),
+      (ctx) => new CanvasDrawer(ctx, () => { }),
     );
   }
 
@@ -326,8 +327,10 @@ export class Plot extends Component {
    * @returns {Plot} The calling Plot.
    */
   public attr<A>(attr: string, attrValue: A | IAccessor<A>, scale: Scale<A, number | string>): this;
-  public attr<A>(attr: string, attrValue?: number | string | IAccessor<number> | IAccessor<string> | A | IAccessor<A>,
-                 scale?: Scale<A, number | string>): any {
+  public attr<A>(
+    attr: string, attrValue?: number | string | IAccessor<number> | IAccessor<string> | A | IAccessor<A>,
+    scale?: Scale<A, number | string>,
+  ): any {
     if (attrValue == null) {
       return this._attrBindings.get(attr);
     }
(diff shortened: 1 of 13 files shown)


src/plots/plot.ts
```diff
@@ -855,6 +865,41 @@ export class Plot extends Component {
...
+      return undefined;
+    }
+    return found.map((entity) => this._lightweightPlotEntityToPlotEntity(entity));
+  }
+
+  /**
+   * Returns the entites whose bounding boxes overlap the parameter on the
+   * x-axis.
+   */
+  public entitiesInXRange(queryBounds: IEntityBounds): Plots.IPlotEntity[] {
```
hellochar, line 923: I'd prefer not to have separate `range/xRange/yRange` endpoints. We can do `XRange` by passing in `Infinity` for the `y/height` on the generic Bounds object. Maybe it'd make more sense for `Bounds` to be two `Range` objects

src/plots/plot.ts
```diff
@@ -855,6 +865,41 @@ export class Plot extends Component {
     return nearest === undefined ? undefined : this._lightweightPlotEntityToPlotEntity(nearest);
   }
 
+  /**
+   * Returns the entites whose bounding boxes overlap the parameter.
+   */
+  public entitiesInBounds(queryBounds: IEntityBounds): Plots.IPlotEntity[] {
```
hellochar: Nit: to be clear - this ignores the `y/height` properties on IEntityBounds? This API feels inconsistent with the existing `entitiesIn(xRangeOrBounds: Range | Bounds, yRange?: Range): IPlotEntity[] {` that bar, scatter, line, and rectangle have.

I think part of the issue is that we have overlapping responsibilities in our interfaces:

```ts
export type Range = {
  min: number;
  max: number;
};

/**
 * A location in pixel-space.
 */
export type Point = {
  x: number;
  y: number;
};

/**
 * The corners of a box.
 */
export type Bounds = {
  topLeft: Point;
  bottomRight: Point;
};

/**
 * Client bounds
 */
export interface IEntityBounds {
  x: number;
  y: number;
  width: number;
  height: number;
}
```

An `IEntityBounds` and a `Bounds` are semantically equivalent to each other, as well as equivalent to two `Range`s.

src/utils/entityStore.ts
```diff
@@ -60,36 +85,58 @@ export interface IEntityStore<T extends IPositionedEntity> {
 export class EntityStore<T extends IPositionedEntity> implements IEntityStore<T> {
   private _entities: T[];
   private _tree: d3.Quadtree<T>;
+  private _rtree: RTree<T>;
 
   constructor() {
     this._entities = [];
+    this._rtree = new RTree<T>();
     this._tree = d3.quadtree<T>()
```
hellochar, line 93: nit: we can probably implement "nearest entity to point" with the rtree and get rid of the quadtree completely

src/utils/rTree.ts
```diff
@@ -0,0 +1,363 @@
...
+/**
+ * There are several strategies for splitting nodes that contain overlapping
+ * regions. By default we use `SplitStrategyLinear` which minimizes the change
+ * in node bounding box area.
+ */
+const DEFAULT_SPLIT_STRATEGY: IRTreeSplitStrategy = new SplitStrategyLinear();
+
+/**
+ * R-Tree is a multidimensional spatial region tree. It stores entries that have
+ * arbitrarily overlapping bounding boxes and supports efficient point and
```
hellochar, line 28: nit: can we document runtime and space complexity to be more precise with what "efficient" means

hellochar (approved):
Nothing blocking, nice rtree impl :)

nit: a quicktest for this would be nice!

src/plots/plot.ts
themadcreator: very true. i'm not sure why the original bounds was structured as two points, but i've always found it awkward.

src/utils/entityStore.ts
themadcreator: it's not exactly the same. the quadtree structure allows us to guarantee that no other point is closer based on the bounding box of the tree node. r-trees don't make the same guarantees, but they are great for querying overlapping bounds!

src/plots/plot.ts
themadcreator: nah, that'd just add another semantically equiv interface.

I think different endpoints is really helpful. It allows the user of the API to know that it'll "do what it says on the tin" as opposed to having to think about how the same method will re-interpret their arguments at runtime.

src/plots/plot.ts
themadcreator: I'll take a look at replacing the existing entitiesIn methods to defer to the relevant entitystore method

themadcreator:
Now that entitiesIn uses the entitystore methods, the dragbox quicktest should demonstrate this feature. [this one](https://657-16149854-gh.circle-artifacts.com/0/home/ubuntu/plottable/quicktests/dev/index.html#interactions/dragbox)

PR #3404, opened 2017-09-11
doLocked on entityNearest

hellochar:
reverts perf regression on entityNearest

also emit sourcemap for plottable.js

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -399,7 +399,7 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
   /**
    * Returns the PlotEntity nearest to the query point according to the following algorithm:
    *   - If the query point is inside a bar, returns the PlotEntity for that bar.
-   *   - Otherwise, gets the nearest PlotEntity by the primary direction (X for vertical, Y for horizontal),
+   *   - Otherwise, gets the nearest PlotEntity by the positional direction (X for vertical, Y for horizontal),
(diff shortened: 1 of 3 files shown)


src/plots/barPlot.ts
```diff
@@ -399,80 +399,81 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
...
+      const originalPositionProjector = Plot._scaledAccessor(this.length());
+      const scaledBaseline = this.length().scale.scale(this.baselineValue());
+      const plotPointFactory = (datum: any, index: number, dataset: Dataset, rect: IEntityBounds) => {
+        return this._pixelPointBar(originalPositionProjector(datum, index, dataset), scaledBaseline, rect);
+      };
+
+      let minPrimaryDist = Infinity;
+      let minSecondaryDist = Infinity;
+      let closest: ILightweightPlotEntity;
+      this._getEntityStore().entities().forEach((entity: ILightweightPlotEntity) => {
```
themadcreator, line 434: this is just the type of full list iteration we should avoid. good news: we can use the entity store to do this! with the new methods on entity store we can do `entityStore.entitiesInBounds` and then `entityStore.entityNearest`

themadcreator: or maybe not quite possible if you only do 1 direction and not the other.

themadcreator (approved):
If possible to re-use entity store, let's do that, otherwise this is good to go

PR #3405, opened 2017-09-12
Prepare release 3.5.4

themadcreator:
Notes:
# v3.5.4

### Performance Enhancements

- (#3403) Fast entity bounds queries. Added new methods on all plots for efficient bounds-overlap queries for plot entities.
  - `entitiesInBounds`
  - `entitiesInXRange`
  - `entitiesInYRange`
  - Use these methods directly or use the existing `entitiesIn` method that takes either a single `Bounds` object or two `Range` objects. In both of these cases, they are converted to `IEntityBounds` and then applied to `entitiesInBounds`.

- (#3404) Fix perf regression of `entityNearest` in bar plots.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "plottable",
   "description": "A modular charting library built on D3",
-  "version": "3.5.3",
+  "version": "3.5.4",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


hellochar:
We should explain how `entitiesInBounds` is different from `entitiesIn` and give a usage example

hellochar (changes requested):
see comment

hellochar (approved):

PR #3408, opened 2017-09-20
Don't short-circuit `_includedValuesProvider` for unanchored components

themadcreator:
The lines that were removed prevented `_includedValuesProvider` from working when a component was not yet anchored. This may be a minor performance optimization but it is not necessary.

It prevents extents calculations from working. Furthermore, the calculations in this method depend only on data space, not pixel space, so anchoring is irrelevant.


hellochar:
Why didn't the first commit work? Can we fix the unit tests instead of introducing this `ignoreAttachState` arg (which feels a bit hacky)?

hellochar (approved):

PR #3409, opened 2017-09-22
don't require label padding to draw bar labels; improve perf

hellochar:
improves data density.

drastically improve bar plot perf by performing viewbox culling on data before
rendering


src/core/version.ts
```diff
@@ -11,4 +11,4 @@ declare var __VERSION__: string;
  * real version number during the dist phase for for npm module publishing. Modifying this line should
  * be accompanied by modifying the "sed-version" task in package.json accordingly.
  */
-export let version = __VERSION__;
```
themadcreator: i don't think should be updated

themadcreator (changes requested):

themadcreator (approved):
approved once version is fixed

PR #3412, opened 2017-10-11
Fix PanZoomInteraction constraints

hellochar:
|| should've been an &&

### DIFF
--- a/quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
+++ b/quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
@@ -29,6 +29,14 @@ function run(svg, data, Plottable) {
       .setMinMaxDomainValuesTo(xScale)
       .setMinMaxDomainValuesTo(yScale);
 
+    new Plottable.Interactions.PanZoom(xScale, null)
+        .attachTo(xAxis)
+        .setMinMaxDomainValuesTo(xScale);
+
+    new Plottable.Interactions.PanZoom(null, yScale)
+        .attachTo(yAxis)
+        .setMinMaxDomainValuesTo(yScale);
+
     return table;
   }
 
--- a/src/interactions/panZoomConstraints.ts
+++ b/src/interactions/panZoomConstraints.ts
@@ -62,20 +62,8 @@ function constrainZoomExtents(
 }
 
 /**
- * We want to avoid scale.zoom(zoomAmount, centerPoint) returning a domain that's
- * "out of bounds", defined as:
- *
- * 1. Math.min(domain[0], domain[1]) < minDomainValue, or
- * 2. Math.max(domain[0], domain[1]) > maxDomainValue
- *
- * There's four cases to cover and what to do in each case:
- *
- * 1. Not out of bounds. Don't change zoomAmount or centerPoint.
- * 2. Out of bounds only on lower end. Don't change zoomAmount, just shift the centerPoint such that the domain min
- *    lines up with the min bounds. Be careful - you might hit case 4 after this.
- * 3. Out of bounds only on upper end. Do some as step 2.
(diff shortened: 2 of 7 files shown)


hellochar:
Hmm this is breaking on x axis panzoom...

quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
```diff
@@ -29,6 +29,14 @@ function run(svg, data, Plottable) {
       .setMinMaxDomainValuesTo(xScale)
       .setMinMaxDomainValuesTo(yScale);
 
+    new Plottable.Interactions.PanZoom(xScale, null)
+        .attachTo(xAxis)
+        .setMinMaxDomainValuesTo(xScale);
+
+    new Plottable.Interactions.PanZoom(null, yScale)
+        .attachTo(yAxis)
+        .setMinMaxDomainValuesTo(yScale);
```
themadcreator, line 38: Trying out this preview, i see this behavior:

constrained-zoom-2

If you zoom an axes, you can use the plot to zoom all the way out. It imposes a constraint that doesn't match the scale extents.

themadcreator: The behavior that would be most desirable, IMO, is to have the x axis zoom out while the y axis stays fixed. That is, constraining one axis doesn't necessarily constrain the other axis. Also what about multiple independent axes?

src/interactions/panZoomConstraints.ts
```diff
@@ -0,0 +1,176 @@
...
+    return { centerPoint, zoomAmount };
+  }
+
+  const reversed = isRangeReversed(scale);
+  // if no constraints set, we're done
+  if (minDomainValue == null && maxDomainValue == null) {
+    return { centerPoint, zoomAmount };
+  }
+
+  minDomainValue = minDomainValue == null ? -Infinity : minDomainValue;
```
themadcreator: won't `-Infinity` be wrong if the domain is reversed? maybe verify this by removing the max constraint from the pan_zoom_reverse preview

themadcreator: otherwise, i like this approach to clean up this function

themadcreator (commented):
Questions

quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
hellochar: Yeah we had discussed this - I decided against it because users can already implement that behavior by creating separate PanZoomInteractions and attaching all of them to the same plot. If we make that the behavior of PZI, then we no longer support the users who want the current "constrain all scales" behavior.

src/interactions/panZoomInteraction.ts
```diff
@@ -140,25 +133,44 @@ export class PanZoom extends Interaction {
    * By default, `centerValue` is the center of the x and y range of each scale.
    */
   public zoom(zoomAmount: number, centerValue?: Point) {
```
themadcreator: nit: given this is a functional change, it might be wise to hedge against upsetting users with an optional third arg to disable the constraint.

quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
themadcreator: That may be possible for plottable users, but not for our react component. If we are going to use this axis zoom ourselves, won't our users have the same problem with no way to fix it?

themadcreator (approved):
not gonna block but i still think the locked axis constraint is gonna be an issue later

quicktests/overlaying/tests/interactions/pan_zoom_reverse.js
hellochar: We can do the same with our React Component, I was experimenting with it. We just rebuild an array of PZIs instead of a single one when the props change

PR #3413, opened 2017-10-12
Allow sparse data arrays in canvas renderers

themadcreator:
Plottable sometimes hides data values if a datum is
invalid, downsampled, or off screen. Filtering these before
rendering leads to an inaccurate data index in accessors.
For example, a index-based selection state accessor
would produce incorrect results.

In this PR, we instead set "hidden" data values to null,
and maintain the same data array length.

Using so-called "sparse" data arrays that might contain
nulls allows us to accurately maintain the correct index
for each datum while iterating.

Canvas renderers can take advantage of this, but SVG
renderers using d3 selections do not support null
values, so using the index argument in those accessors
may lead to incorrect results. To avoid npe issues, we
filter out nulls for SVG renderers.

### DIFF
--- a/src/drawers/rectangleDrawer.ts
+++ b/src/drawers/rectangleDrawer.ts
@@ -27,6 +27,9 @@ export const RectangleCanvasDrawStep: CanvasDrawStep = (
     projector: AttributeToAppliedProjector) => {
   context.save();
   data.forEach((datum, index) => {
+    if (datum == null) {
+      return;
+    }
     const attrs = resolveAttributesSubsetWithStyles(projector, RECT_ATTRS, datum, index);
     context.beginPath();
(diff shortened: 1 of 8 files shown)


hellochar (approved):

themadcreator:
As noted, this does not resolve the index misalignment for SVG renders.

For future reference, these approaches might be used to fix this issue completely:

1. In `getDataToDraw` *we* call the .attr functions and resolve all the attrs directly (like how the canvas drawers do already), and then we use those “resolved attribute objects” (which I’m calling Entitys) as the d3 data, and we don’t have to pass any .attrs to d3 anymore.
2. Overhaul renderer architecture so that we do not need to call accessor callbacks multiple times. In the course of doing so, we may use a strategy like in (1).

PR #3415, opened 2017-10-17
Fix deferred rendering

themadcreator:
The primary cause of incorrect deferred rendering results was that
the deferred renderer did not know when a component was rendered
except when it triggered the render. Occassionally, direct calls
to render would occur and the deferred renderer did not update
the css transform. This PR refactors the code so that the
component's own render method will reset the deferred renderer to
guarantee that it resets the css transform.

- Add `renderLowPriority` method to all components to make typing easier
- Extract transform method from constructor for legibility
- Switch to transformable scales to support category axis deferred render

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -296,6 +296,10 @@ export class Component {
     return this;
   }
 
+  public renderLowPriority() {
+    return this.render();
+  }
+
   private _scheduleComputeLayout() {
     if (this._isAnchored && this._isSetup) {
       RenderController.registerToComputeLayoutAndRender(this);
--- a/src/plots/deferredRenderer.ts
+++ b/src/plots/deferredRenderer.ts
@@ -3,17 +3,17 @@
  * @license MIT
  */
 
-import { Scale } from "../scales/scale";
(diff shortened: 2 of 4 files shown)


src/components/component.ts
```diff
@@ -296,6 +296,10 @@ export class Component {
     return this;
   }
 
+  public renderLowPriority() {
+    return this.render();
```
hellochar, line 300: What does low priority mean? Why does putting this on Component make typings easier?

hellochar (approved):

src/components/component.ts
themadcreator: It's what you call instead of render to use the deferred rendering codepath (when available). When writing the deferred renderer originally, i didn't want to mess with the existing uses of `.render()` so i introduced `.renderLowPriority` on `Plottable.Plot` which triggers the deferred callbacks. This change just pushes the method up the `Plottable.Component`

hellochar: Hmm, but deferred rendering only exists on Plot, so I worry users will get confused by the difference (especially since there's no discernible distinction).

PR #3416, opened 2017-10-19
Respect opacity for fill and fill opacity

kevinschaich:
Previously opacity set by "rgba(0, 0, 0, x)" would get overwritten by getFillOpacity() which returns 1 by default. Change behavior to set the opacity to the product of the two, which mimics chromes native implementation.

### DIFF
--- a/src/drawers/canvasDrawer.ts
+++ b/src/drawers/canvasDrawer.ts
@@ -132,14 +132,14 @@ export function renderPathWithStyle(context: CanvasRenderingContext2D, style: Re
   if (style["stroke"]) {
     context.lineWidth = getStrokeWidth(style);
     const strokeColor = d3.color(style["stroke"]);
-    strokeColor.opacity = getStrokeOpacity(style);
+    strokeColor.opacity *= getStrokeOpacity(style);
     context.strokeStyle = strokeColor.toString();
     context.stroke();
   }
 
   if (style["fill"]) {
     const fillColor = d3.color(style["fill"]);
-    fillColor.opacity = getFillOpacity(style);
+    fillColor.opacity *= getFillOpacity(style);
     context.fillStyle = fillColor.toString();
     context.fill();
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @kevinschaich! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/drawers/canvasDrawer.ts
```diff
@@ -132,14 +132,14 @@ export function renderPathWithStyle(context: CanvasRenderingContext2D, style: Re
   if (style["stroke"]) {
     context.lineWidth = getStrokeWidth(style);
     const strokeColor = d3.color(style["stroke"]);
-    strokeColor.opacity = getStrokeOpacity(style);
```
themadcreator, line 135: are we guaranteed that `strokeColor.opacity` will be a valid numerical value? This only works if opacity is a number [0, 1]. But can't it be defined like `"50%"`?

kevinschaich: I think `stroke-opacity` and `fill-opacity` function the same way as `opacity`, they should just be a float between 0-1:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-opacity
hsla/rgba require it to be a float as well:
https://www.w3.org/TR/2011/REC-css3-color-20110607/#hsla-color
https://www.w3.org/TR/2011/REC-css3-color-20110607/#rgba-color

CalvinFernandez: Also d3's implementation of color only allows for numerical valued opacity. 


https://github.com/d3/d3-color/blob/master/src/color.js

hellochar (approved):

PR #3422, opened 2017-11-16
Add labels to scatterplots

cmeyer63:
If a data point includes a "label" field it will be rendered centered at that points (x,y) coordinates on top of the symbol. In the event the symbol's diameter is smaller than the rendered text, the text will be rendered below the point.

### DIFF
--- a/src/plots/scatterPlot.ts
+++ b/src/plots/scatterPlot.ts
@@ -1,3 +1,9 @@
+import { DatumFormatter } from './../core/formatters';
+import { SimpleSelection } from './../core/interfaces';
+import * as Typesettable from "typesettable";
+import * as Formatters from "../core/formatters";
+
+
 /**
  * Copyright 2014-present Palantir Technologies
  * @license MIT
@@ -15,6 +21,7 @@ import * as Animators from "../animators";
 import * as Drawers from "../drawers";
 import * as Scales from "../scales";
 import * as Utils from "../utils";
+import { makeEnum } from "../utils/makeEnum";
 import * as Plots from "./";
 import { IAccessorScaleBinding, ILightweightPlotEntity, IPlotEntity, ITransformableAccessorScaleBinding } from "./";
 import { Plot } from "./plot";
@@ -23,11 +30,27 @@ import { XYPlot } from "./xyPlot";
 export interface ILightweightScatterPlotEntity extends ILightweightPlotEntity {
   // size of the entity in pixel space
   diameter: number;
+  label?: string;
 }
+
+type LabelConfig = {
+  labelArea: SimpleSelection<void>;
+  measurer: Typesettable.CacheMeasurer;
+  writer: Typesettable.Writer;
+};
+export const LabelsPosition = makeEnum(["start", "middle", "end", "outside"]);
+
 
 export class Scatter<X, Y> extends XYPlot<X, Y> {
   private static _SIZE_KEY = "size";
   private static _SYMBOL_KEY = "symbol";
+  
+  // label stuff
+  protected static _LABEL_AREA_CLASS = "bar-label-text-area";
+  private _labelConfig: Utils.Map<Dataset, LabelConfig>;
+  private _labelFormatter: DatumFormatter = Formatters.identity();
+  private _labelsEnabled = true;
+  private _labelsPosition = LabelsPosition.end;
 
   /**
    * A Scatter Plot draws a symbol at each data point.
@@ -47,6 +70,7 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
     this.size(6);
     const circleSymbolFactory = SymbolFactories.circle();
     this.symbol(() => circleSymbolFactory);
+    this._labelConfig = new Utils.Map<Dataset, LabelConfig>();
   }
 
   protected _buildLightweightPlotEntities(datasets: Dataset[]) {
@@ -59,6 +83,9 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
         lightweightPlotEntity.dataset);
 
       lightweightPlotEntity.diameter = diameter;
+
+      lightweightPlotEntity.label = "NO_LABEL";
+
       return lightweightPlotEntity;
     });
   }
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @cmeyer63! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/plots/scatterPlot.ts
```diff
@@ -205,4 +222,105 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+    if (labelConfig != null) {
+      labelConfig.labelArea.remove();
+      this._labelConfig.delete(dataset);
+    }
+  }
+
+  protected _additionalPaint(time: number) {
+    this.datasets().forEach((dataset) => this._labelConfig.get(dataset).labelArea.selectAll("g").remove());
+    if (this._labelsEnabled) {
+      Utils.Window.setTimeout(() => this._drawLabels(), time);
```
themadcreator, line 272: why is this done on the next stackframe instead of immediately?

src/plots/scatterPlot.ts
```diff
@@ -25,10 +28,23 @@ export interface ILightweightScatterPlotEntity extends ILightweightPlotEntity {
...
+  measurer: Typesettable.CacheMeasurer;
+  writer: Typesettable.Writer;
+};
+
 export class Scatter<X, Y> extends XYPlot<X, Y> {
   private static _SIZE_KEY = "size";
   private static _SYMBOL_KEY = "symbol";
 
+  // label stuff
+  protected static _LABEL_AREA_CLASS = "bar-label-text-area";
```
themadcreator: should be `"scatter-..."` instead of `"bar-..."` class name. Or, if there are multi-selectors that use `"bar-label-text-area"`, then perhaps a better common name is `"plot-label-text-area"`.

src/plots/scatterPlot.ts
```diff
@@ -205,4 +222,105 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+    };
+  }
+
+  private _createLabelContainer(
+    labelArea: SimpleSelection<void>, labelContainerOrigin: Point, labelOrigin: Point, measurement: Typesettable.IDimensions) {
+
+    const labelContainer = labelArea.append("g").attr("transform", `translate(${labelContainerOrigin.x}, ${labelContainerOrigin.y})`);
+    labelContainer.classed("on-bar-label", true);
+    labelContainer.classed("dark-label");
+
```
themadcreator: aren't the above two lines equivalent? if so, why the different syntax?

src/plots/scatterPlot.ts
```diff
@@ -205,4 +222,105 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+        x: "center",
+        y: "center",
+      },
+    };
+  }
+
+  private _createLabelContainer(
+    labelArea: SimpleSelection<void>, labelContainerOrigin: Point, labelOrigin: Point, measurement: Typesettable.IDimensions) {
+
+    const labelContainer = labelArea.append("g").attr("transform", `translate(${labelContainerOrigin.x}, ${labelContainerOrigin.y})`);
```
themadcreator: style nit: wrap this line at the `.attr`

src/plots/scatterPlot.ts
```diff
@@ -205,4 +222,105 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+      dataToDraw.get(dataset).forEach((datum, index) => {
+        if (datum == null) {
+          return;
+        }
+        this._drawLabel(datum, index, dataset, attrToProjector);
+      });
+    });
+  }
+
+  private _drawLabel(datum: any, index: number, dataset: Dataset, attrToProjector: AttributeToProjector) {
```
themadcreator, line 289: note: this approach is svg only, whereas we support both svg and canvas rendering for scatter plots. So, adding labels may induce a significant performance penalty on high scale scatter plots.

src/plots/scatterPlot.ts
```diff
@@ -205,4 +222,105 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
...
+      },
+      alignment: {
+        x: "center",
+        y: "center",
+      },
+    };
+  }
+
+  private _createLabelContainer(
+    labelArea: SimpleSelection<void>, labelContainerOrigin: Point, labelOrigin: Point, measurement: Typesettable.IDimensions) {
```
themadcreator: style nit: switch to one arg per line for readability.

src/plots/scatterPlot.ts
```diff
@@ -25,10 +28,23 @@ export interface ILightweightScatterPlotEntity extends ILightweightPlotEntity {
...
   private static _SIZE_KEY = "size";
   private static _SYMBOL_KEY = "symbol";
 
+  // label stuff
+  protected static _LABEL_AREA_CLASS = "bar-label-text-area";
+  private _labelConfig: Utils.Map<Dataset, LabelConfig>;
+  private _labelFormatter: Formatters.DatumFormatter = Formatters.identity();
+
+  protected static _LABEL_MARGIN_FROM_BUBBLE = 15;
+  private _labelsEnabled = true;
```
themadcreator: add getter/setter methods for this

themadcreator:
image

themadcreator (changes requested):

src/plots/scatterPlot.ts
themadcreator: using separate names may help in applying custom colors for labels

src/plots/scatterPlot.ts
themadcreator: This won't block the PR, though. We already use svg text labels for canvas bar charts.

src/plots/scatterPlot.ts
themadcreator: Actually it looks like the second call is a no-op since it doesn't alter the label but rather returns a boolean indicating whether or not the selection has that class name.

https://github.com/d3/d3-selection/blob/master/README.md#selection_classed

src/plots/scatterPlot.ts
cmeyer63: This is how labels are done in Bar Plot, from what I can guess it's because it allows labels to be drawn after all of the bars/scatter symbols are done?

src/plots/scatterPlot.ts
cmeyer63: Typesettable does do canvas text rendering AFAIK so maybe that could be in a future PR

themadcreator: It does, as a matter of fact. I added the capability a while ago, which you can see here: https://115-25238947-gh.circle-artifacts.com/0/home/ubuntu/typesettable/preview/index.html

No worries though, it's not that important.

themadcreator (approved):

PR #3429, opened 2017-12-07
Upgrade circle.yml to use node 8.8.1 in builds

themadcreator:


circle.yml
```diff
@@ -11,14 +11,14 @@ general:
...
-    version: 6.1.0
   environment:
     PATH: "${PATH}:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin"
+  node:
+    version: 8.8.1
 
 dependencies:
   override:
-    - yarn
+    - yarn install --pure-lockfile
```
adidahiya: no, prefer the following:

```yml
- yarn install
- echo "Checking if yarn.lock changed..." && git diff --exit-code
```

hellochar: So we're saying "all your deps should be kept up to date to upstream or else we'll fail the build"? Not a huge fan since I feel that puts unnecessary burden on every PR; I'd prefer doing manual upgrades every once in a while.

adidahiya: that's not what `yarn install` does. `yarn upgrade` bumps the actual versions in the lockfile. `yarn install` just tries to get to a stable state given the current lockfile and the package.json file. if and only if it notices discrepancies, it makes edits to the lockfile, which should fail your build.

my approach is a quick and easy way to fail fast when package.json is updated but the developer forgot to update the lockfile.

hellochar: How about frozen-lockfile: https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile ?

adidahiya: I haven't tried it but if it does the same thing I proposed, that seems fine.

PR #3432, opened 2018-01-04
Configurable UTC/local-time

waltonseymour:
This PR allows for `Plottable.Formatters.time` to display dates in UTC in addition to local time with the addition of an optional boolean `useUTC`.

This flag is also passed down from the `TimeAxis` constructor which also contains the additional `useUTC` flag.

For us, this solves a business requirement of displaying time in UTC without having to modify the underlying Date objects, which has a number of drawbacks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "main": "build/src/index.js",
   "typings": "build/src/index.d.ts",
   "scripts": {
+    "prepare": "npm run dist",
     "build": "npm-run-all -p lint:tslint build:tsc -s build:webpack",
     "build:tsc": "tsc -p .",
     "build:webpack": "webpack",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   "main": "build/src/index.js",
   "typings": "build/src/index.d.ts",
   "scripts": {
-    "prepare": "npm run dist",
     "build": "npm-run-all -p lint:tslint build:tsc -s build:webpack",
     "build:tsc": "tsc -p .",
     "build:webpack": "webpack",
--- a/src/axes/timeAxis.ts
+++ b/src/axes/timeAxis.ts
@@ -80,115 +80,118 @@ export class Time extends Axis<Date> {
     [TimeInterval.year]: 6,
   };
 
-  private static _DEFAULT_TIME_AXIS_CONFIGURATIONS: TimeAxisConfiguration[] = [
-    [
-      { interval: TimeInterval.second, step: 1, formatter: Formatters.time("%I:%M:%S %p") },
-      { interval: TimeInterval.day, step: 1, formatter: Formatters.time("%B %e, %Y") },
-    ],
-    [
-      { interval: TimeInterval.second, step: 5, formatter: Formatters.time("%I:%M:%S %p") },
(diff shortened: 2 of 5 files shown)


src/core/formatters.ts
```diff
@@ -244,10 +244,14 @@ export function multiTime() {
  * List of directives can be found on: https://github.com/mbostock/d3/wiki/Time-Formatting#format
  *
  * @param {string} [specifier] The specifier for the formatter.
+ * @param {boolean} [useUTC] Displays time in UTC if true, local time if false.
  *
  * @returns {Formatter} A formatter for time/date values.
  */
-export function time(specifier: string): Formatter {
+export function time(specifier: string, useUTC?: boolean): Formatter {
```
themadcreator: Please add the default value for this optional argument: `useUTC = DEFAULT_USE_UTC` and above, `const DEFAULT_USE_UTC = false;` (also update comment to indicate default it to not use utc)

package.json
```diff
@@ -11,6 +11,7 @@
   "main": "build/src/index.js",
   "typings": "build/src/index.d.ts",
   "scripts": {
+    "prepare": "npm run dist",
```
themadcreator: I'm not sure we want this. Even if we do, this isn't the PR for it, so please remove.

themadcreator (changes requested):

themadcreator:
Other than my comments, this changes looks good to me. Thanks!

package.json
waltonseymour: removed

waltonseymour:
I've updated the PR to address comments.
I've also corrected an issue with `tickInterval` where the ticks would be generated with local time while the labels would reflect UTC time.

src/axes/timeAxis.ts
```diff
@@ -80,115 +80,120 @@ export class Time extends Axis<Date> {
...
+      [
+        { interval: TimeInterval.year, step: 500, formatter: formatter("%Y") },
+      ],
+      [
+        { interval: TimeInterval.year, step: 1000, formatter: formatter("%Y") },
+      ],
+    ];
+  }
+
+  private _useUTC: boolean = false;
```
themadcreator: I believe this default will actually be overridden with `undefined` if the constructor is invoked without the third parameter.

when the object instance is first created, `this._useUTC` is initialized with `false`, but then the constructor runs, which explicitly overrides `this._useUTC` with the optional argument. You can fix this by simply moving this default to the constructor signature like `(..., useUTC: boolean = false)`. Or better yet, `(..., useUTC: boolean = DEFAULT_USE_UTC)` and put `const DEFAULT_USE_UTC = false;` at the top of this file.

themadcreator: Though, since you're already defining the default at a lower level, you could just not initialize `this._useUTC`

themadcreator (changes requested):
Really close, just one minor thing

src/axes/timeAxis.ts
waltonseymour: Given that default values are set at `time` and `tickInterval`, I think it's fine to leave it potentially undefined. I've removed the initialized value for `_useUTC`

themadcreator (approved):

PR #3439, opened 2018-02-08
Upgrade and Update

themadcreator:
Upgrade to typescript 2.3.4
Switch to es2015 module export
Upgrade jquery to 3.0.0 for security purposes


tsconfig.json
```diff
@@ -3,12 +3,15 @@
       "declaration": true,
       "forceConsistentCasingInFileNames": true,
       "importHelpers": true,
-      "module": "commonjs",
+      "module": "es2015",
```
adidahiya, line 6: this is a breaking change. although webpack 3 can handle es2015 modules fine, users may be using other tools that can't. if you want to make this change, it needs to be additive. blueprint publishes both commonjs and es2015 modules in the same NPM package.

adidahiya (changes requested):

tsconfig.json
themadcreator: Okay, we don't need to do this for the jquery removal anyway, so I'm going to separate these steps.

PR #3441, opened 2018-02-13
Upgrade typescript@^2.7

themadcreator:


package.json
```diff
@@ -53,7 +54,7 @@
     "requirejs": "2.1.18",
     "sinon": "^2.1.0",
     "tslint": "4.5.1",
-    "typescript": "2.1.5",
+    "typescript": "^2.7",
```
adidahiya: use a tilde range `~` since there are breaking changes between minor versions of this package.

adidahiya (changes requested):
looks good otherwise

adidahiya (approved):

PR #3445, opened 2018-02-22
Use locked value of bar thickness in more cases.

dcervelli:


themadcreator:
There is a subtle issue here. Bar plots don't always have equal-width bars. For example histogram plots have arbitrary start/end coordinates for each bar.

So, there are a couple ways to make sure we're doing this correctly, 1) change `_entitiesIntersecting` to use `super.entitiesInBounds` which uses the entityStore's r-tree for ~log(n) intersection lookups, or only use the `doLocked` invocation when `this._fixedBarPixelThickness` is `true`

themadcreator (approved):
Works for me

PR #3449, opened 2018-03-22
Perf: Drastically increase bar plot hover perf

themadcreator:
- add "nearest" query capability to rtree
- replace slow bar plot nearest logic with rtree version
- remove now-unused d3 quadtree


themadcreator:
Example of improved perf on a scatter plot with 100K maybe-overlapping entities.
scatter_hover
image

leebyp (approved):

PR #3450, opened 2018-03-23
Remove gap from bar thickness if width of bar is below threshold

themadcreator:
This prevents bars from dissappearing at just the right
resolution and zoom level due to pixel sampling moire

### DIFF
--- a/src/plots/barPlot.ts
+++ b/src/plots/barPlot.ts
@@ -51,6 +51,7 @@ export type BarAlignment = keyof typeof BarAlignment;
 
 export class Bar<X, Y> extends XYPlot<X, Y> {
   public static _BAR_THICKNESS_RATIO = 0.95;
+  public static _BAR_GAPLESS_THRESHOLD_PX = 3;
   public static _SINGLE_BAR_DIMENSION_RATIO = 0.4;
   private static _BAR_AREA_CLASS = "bar-area";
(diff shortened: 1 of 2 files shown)


themadcreator:
### Before
pbar-before

### After
pbar-after

src/plots/barPlot.ts
```diff
@@ -751,7 +752,11 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
     const thicknessF = attrToProjector[Bar._BAR_THICKNESS_KEY];
     const gapF = attrToProjector["gap"];
     const thicknessMinusGap = gapF == null ? thicknessF : (d: any, i: number, dataset: Dataset) => {
-      return thicknessF(d, i, dataset) - gapF(d, i, dataset);
+      const thick = thicknessF(d, i, dataset);
+      // only subtract gap if bars are at least 2 pixels wide, otherwise canvas
+      // interpolation can cause bars to become invisible due to subpixel
+      // sampling
```
giladgray, line 758: tiny nit: put this word on the previous line. shorter file!

giladgray (approved):

src/plots/barPlot.ts
themadcreator: i autowrap comments to 80 chars for readability

PR #3459, opened 2018-04-04
Normalize keys instead of using .toString to avoid NPE

zerovox:
@themadcreator ran into an NPE where the domain contains null (and/or undefined) with Stacked Area charts


zerovox:
@giladgray care to take a look?

zerovox:
@giladgray care to take a look?

giladgray (approved):
Looks fine but I don't actually know this code.

PR #3465, opened 2018-04-30
Expose y offset for stacked area plots

jscheiny:
This exposes a `yOffset` method (happy to change name) on the `StackedArea` plot which allows users to retrieve the offset of a given data point on a specific dataset for the purpose of coordinating the display of data across plots that share datasets but some of which are not stacked.

screen shot 2018-04-30 at 3 24 03 pm

### DIFF
--- a/src/plots/stackedAreaPlot.ts
+++ b/src/plots/stackedAreaPlot.ts
@@ -129,7 +129,7 @@ export class StackedArea<X> extends Area<X> {
     if (datasetStackingResult == null) {
       return undefined;
     }
-    const result = datasetStackingResult.get(`${x}`);
+    const result = datasetStackingResult.get(String(x));
     if (result == null) {
       return undefined;
     }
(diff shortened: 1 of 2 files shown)


src/plots/stackedAreaPlot.ts
```diff
@@ -114,6 +114,28 @@ export class StackedArea<X> extends Area<X> {
...
+  public yOffset(dataset: Dataset, x: any): number {
+    const stackingResult = this._stackingResult();
+    if (stackingResult == null) {
+      return undefined;
+    }
+    const datasetStackingResult = stackingResult.get(dataset);
+    if (datasetStackingResult == null) {
+      return undefined;
+    }
+    const result = datasetStackingResult.get(`${x}`);
```
themadcreator: Using string interpolation here looks a little awkward? Isn't this the same as `.get(x)` or for typing purposes you could also do the very standard `.get(String(x))`

themadcreator (approved):
Optional nit, otherwise good to go

src/plots/stackedAreaPlot.ts
jscheiny: It's necessary since the keys of this map are strings and the user might reasonable pass numbers for numeric axes. Changed to `.get(String(x))`

PR #3467, opened 2018-05-03
Add 0x0 size check to canvas

themadcreator:
Prevents error from being thrown when we try to
invoke context.drawImage with an argument
that is a 0x0 canvas.

Instead we just skip the drawImage.

### DIFF
--- a/src/plots/plot.ts
+++ b/src/plots/plot.ts
@@ -230,7 +230,7 @@ export class Plot extends Component {
         if (btx) {
           const originalCanvas = this._canvas.node();
           if (originalCanvas.width > 0 && originalCanvas.height > 0 && btx.canvas.width > 0 && btx.canvas.height > 0) {
-            btx.drawImage(this._canvas.node(), 0, 0);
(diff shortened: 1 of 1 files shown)


cmslewis (approved):
Worth a warning of some kind?

themadcreator:
Yeah, i was considering. Might as well

src/plots/plot.ts
```diff
@@ -228,7 +228,14 @@ export class Plot extends Component {
...
         // for headless test compat (jest)
         if (btx) {
-          btx.drawImage(this._canvas.node(), 0, 0);
+          const originalCanvas = this._canvas.node();
+          if (originalCanvas.width > 0 && originalCanvas.height > 0) {
+            if (btx.canvas.width > 0 && btx.canvas.height > 0) {
+              btx.drawImage(originalCanvas, 0, 0);
+            }
+          } else {
+            console.warn("Failed to fill buffer canvas with with 0x0 canvas");
```
cmslewis, line 237: @themadcreator FYI there's a duplicate word here: `with with`

PR #3468, opened 2018-05-07
Render canvas line plot with multiple styles

waltonseymour:
Enables the ability for canvas lineDrawer to utilize different styles in the same plot.
This is especially useful for changing opacity/stroke for segments along a plot.
When drawing between two points that produce different styles, it chooses the first style arbitrarily.

### DIFF
--- a/src/drawers/canvasDrawer.ts
+++ b/src/drawers/canvasDrawer.ts
@@ -65,6 +65,27 @@ export const ContextStyleAttrs = [
   "stroke-dasharray",
 ];
 
+/**
+ * attributesEqual performs a shallow equality check on two attributes objects.
+ * @param a
+ * @param b
+ * @returns true if a and b are equavalent false otherwise
+ */
+export function attributesEqual(a: Record<string, any>, b: Record<string, any>): boolean {
+  if (a === b) {
+    // literally the same object
+    return true;
+  }
+
+  for (const attrKey of Object.keys(a)) {
+    if (!b.hasOwnProperty(attrKey) || a[attrKey] !== b[attrKey]) {
+      return false;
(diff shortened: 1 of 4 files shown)


themadcreator:
Hey thanks for this contribution. I'm still reviewing the performance implications of the new iteration in the line renderer. At high scale, this can have a significant impact on perf. Thanks for your patience

src/plots/linePlot.ts
```diff
@@ -417,7 +417,7 @@ export class Line<X> extends XYPlot<X, number> {
       }
       const projector = attrToProjector[attribute];
       attrToProjector[attribute] = (data: any[], i: number, dataset: Dataset) =>
-        data.length > 0 ? projector(data[0], i, dataset) : null;
```
themadcreator, line 420: im not sure this is right. data is an array of arrays. as you can see above we use `data[0]` everywhere to refer to the line data

src/drawers/canvasDrawer.ts
```diff
@@ -65,6 +65,27 @@ export const ContextStyleAttrs = [
...
+ * @returns true if a and b are equavalent false otherwise
+ */
+export function attributesEqual(a: Record<string, any>, b: Record<string, any>): boolean {
+  if (a === b) {
+    // literally the same object
+    return true;
+  }
+
+  for (const attrKey of Object.keys(a)) {
+    if (!b.hasOwnProperty(attrKey) || a[attrKey] !== b[attrKey]) {
```
themadcreator, line 81: note that this strict equal means `null != undefined` is that intentional?

themadcreator (commented):
questions

PR #3469, opened 2018-05-30
For-each loop optimization and memoize normalizeKey

philcchen:
Optimization around for-each loops (https://jsperf.com/fast-array-foreach ) for a comparison. 

Memoization of `normalizeKey` function, would like a sanity check to make sure that `normalizeKey` never gets called with a mutable object. 

Note: the memoization here isn't necessarily a runtime optimization, but a garbage collection optimization/trade-off

Was seeing excessive GC (especially by Firefox) when running without the memoized keys in production environment

### DIFF
--- a/src/drawers/rectangleDrawer.ts
+++ b/src/drawers/rectangleDrawer.ts
@@ -30,7 +30,7 @@ export const RectangleCanvasDrawStep: CanvasDrawStep = (
   for (let index = 0; index < dataLen; index++ ) {
     const datum = data[index];
     if (datum == null) {
-      return;
+      continue;
     }
     const attrs = resolveAttributesSubsetWithStyles(projector, RECT_ATTRS, datum, index);
     context.beginPath();
--- a/src/drawers/rectangleDrawer.ts
+++ b/src/drawers/rectangleDrawer.ts
@@ -26,7 +26,9 @@ export const RectangleCanvasDrawStep: CanvasDrawStep = (
     data: any[],
     projector: AttributeToAppliedProjector) => {
   context.save();
-  data.forEach((datum, index) => {
+  const dataLen = data.length;
+  for (let index = 0; index < dataLen; index++ ) {
+    const datum = data[index];
     if (datum == null) {
       return;
     }
@@ -34,7 +36,7 @@ export const RectangleCanvasDrawStep: CanvasDrawStep = (
     context.beginPath();
     context.rect(attrs["x"], attrs["y"], attrs["width"], attrs["height"]);
     renderPathWithStyle(context, attrs);
-  });
+  }
   context.restore();
 };
 
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -604,7 +604,7 @@ export class Pie extends Plot {
 
       let value = this.sectorValue().accessor(datum, datumIndex, dataset);
       if (!Utils.Math.isValidNumber(value)) {
-        return;
+        continue;
       }
       value = this._labelFormatter(value, datum, datumIndex, dataset);
       const measurement = measurer.measure(value);
--- a/src/plots/piePlot.ts
+++ b/src/plots/piePlot.ts
@@ -598,7 +598,10 @@ export class Pie extends Plot {
     const writer = new Typesettable.Writer(measurer, context);
     const dataset = this.datasets()[0];
     const data = this._getDataToDraw().get(dataset);
-    data.forEach((datum, datumIndex) => {
+    const dataLen = data.length;
+    for (let datumIndex = 0; datumIndex < dataLen; datumIndex++) {
+      const datum = data[datumIndex];
+
       let value = this.sectorValue().accessor(datum, datumIndex, dataset);
       if (!Utils.Math.isValidNumber(value)) {
         return;
@@ -647,6 +650,6 @@ export class Pie extends Plot {
         xAlign: "center",
         yAlign: "center",
       }, g.node());
-    });
+    }
   }
 }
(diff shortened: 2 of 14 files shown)


src/utils/stackingUtils.ts
```diff
@@ -173,6 +183,6 @@ export function stackedExtent(stackingResult: StackingResult, keyAccessor: IAcce
  * @param {any} key The key to be normalized
  * @return {string} The stringified key
  */
-export function normalizeKey(key: any) {
+export const normalizeKey = lMemoize((key: any) => {
```
philcchen: @themadcreator I _believe_ that any `key` passed in here should be immutable, but given it's typed as `any`, would want a sanity check?

themadcreator: I'm not sure this memoization will actually help because in order to create the cache key and store the value, it will convert the argument into a `Map` key which might be just as slow as the `String(key)` below

src/utils/stackingUtils.ts
```diff
@@ -11,6 +11,8 @@ import { IAccessor } from "../core/interfaces";
 import * as Utils from "./";
 import { makeEnum } from "./makeEnum";
 
+const lMemoize = require("lodash.memoize");
```
themadcreator: even though this keeps the library size smaller, i think i'd prefer to use the normal import syntax. Unfortunately the modular lodash packages don't support `import` statements for some reason.

themadcreator: There are some instructions on how to get treeshaking to work with lodash, but i couldn't get it to work

https://medium.com/@martin_hotell/tree-shake-lodash-with-webpack-jest-and-typescript-2734fa13b5cd

src/drawers/rectangleDrawer.ts
```diff
@@ -26,15 +26,17 @@ export const RectangleCanvasDrawStep: CanvasDrawStep = (
     data: any[],
     projector: AttributeToAppliedProjector) => {
   context.save();
-  data.forEach((datum, index) => {
+  const dataLen = data.length;
+  for (let index = 0; index < dataLen; index++ ) {
+    const datum = data[index];
     if (datum == null) {
       return;
```
themadcreator: `return` -> `continue`

src/plots/piePlot.ts
```diff
@@ -598,7 +598,10 @@ export class Pie extends Plot {
...
     const dataset = this.datasets()[0];
     const data = this._getDataToDraw().get(dataset);
-    data.forEach((datum, datumIndex) => {
+    const dataLen = data.length;
+    for (let datumIndex = 0; datumIndex < dataLen; datumIndex++) {
+      const datum = data[datumIndex];
+
       let value = this.sectorValue().accessor(datum, datumIndex, dataset);
       if (!Utils.Math.isValidNumber(value)) {
         return;
```
themadcreator: `return` -> `continue`

src/plots/plot.ts
```diff
@@ -703,7 +703,10 @@ export class Plot extends Component {
...
       let validDatumIndex = 0;
 
-      dataset.data().forEach((datum: any, datumIndex: number) => {
+      const data = dataset.data();
+      const dataLen = data.length;
+      for (let datumIndex = 0; datumIndex < dataLen; datumIndex++) {
+        const datum = data[datumIndex];
         const position = this._pixelPoint(datum, datumIndex, dataset);
         if (Utils.Math.isNaN(position.x) || Utils.Math.isNaN(position.y)) {
           return;
```
themadcreator: `return` -> `continue`

src/plots/rectanglePlot.ts
```diff
@@ -408,7 +408,10 @@ export class Rectangle<X, Y> extends XYPlot<X, Y> {
     const yMin = Math.min.apply(null, yRange);
     const yMax = Math.max.apply(null, yRange);
     const data = dataToDraw.get(dataset);
-    data.forEach((datum, datumIndex) => {
+    const dataLength = data.length;
+    for (let datumIndex = 0; datumIndex < dataLength; datumIndex++) {
+      const datum = data[datumIndex];
+
       if (datum == null) {
```
themadcreator, line 415: `return` -> `continue`

src/plots/scatterPlot.ts
```diff
@@ -277,12 +277,15 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
     const dataToDraw = this._getDataToDraw();
     const attrToProjector = this._getAttrToProjector();
     this.datasets().forEach((dataset) => {
-      dataToDraw.get(dataset).forEach((datum, index) => {
+      const data = dataToDraw.get(dataset);
+      const dataLen = data.length;
+      for (let index = 0; index < dataLen; index++) {
+        const datum = data[index];
         if (datum == null) {
           return;
```
themadcreator: `return` -> `continue`

src/utils/stackingUtils.ts
```diff
@@ -153,13 +160,16 @@ export function stackedExtents<D>(stackingResult: GenericStackingResult<D>): {
 export function stackedExtent(stackingResult: StackingResult, keyAccessor: IAccessor<any>, filter: IAccessor<boolean>) {
   const extents: number[] = [];
   stackingResult.forEach((stackedDatumMap: Utils.Map<string, StackedDatum>, dataset: Dataset) => {
-    dataset.data().forEach((datum, index) => {
+    const data = dataset.data();
+    const dataLen = data.length;
+    for (let index = 0; index < dataLen; index++) {
+      const datum = data[index];
       if (filter != null && !filter(datum, index, dataset)) {
         return;
```
themadcreator: `return` -> `continue`

themadcreator (changes requested):

src/plots/rectanglePlot.ts
philcchen: (addressed, but diff doesn't recognize)

themadcreator (approved):

PR #3470, opened 2018-05-30
Optimize Rendering Inner Loop

themadcreator:
- Move concats out of inner loop for resolving attributes.
- Memoize AttrToProjector lookups.
- Increased bar count in canvas_bars quicktest to 100K to more easily see perf changes.
- Custom memoize cache implementation for faster 2D index lookup.

Initial local testing indicates 12 - 21% speedup in render.


themadcreator:
Also added feature flag (default false) for this memoization

PR #3486, opened 2019-03-25
Convert circle config to v2

themadcreator:


.circleci/config.yml
```diff
@@ -0,0 +1,83 @@
...
+    steps:
+      - checkout
+      - attach_workspace:
+          at: '.'
+      - restore_cache: *restore_cache
+      - run: ./demo.js || true
+      - persist_to_workspace: *persist_to_workspace
+
+  dist:
+    <<: *setup_env
```
lnunno, line 76: Does `yarn dist` publish to npm also?

lnunno (approved):

.circleci/config.yml
themadcreator: no, i think it just generates the built js and css

PR #3558, opened 2020-10-16
Use entityNearest() rather than entityNearestByXThenY() for LinePlot.entitiesAt()

mcintyret:
Usinig `entityNearestByXThenY` causes some weird behaviour - if the cursor is marginally closer in the x domain to another point than the one the user intended to click on, that point will be the one that gets selected - even if it is hundreds of pixels away in the y domain.

Even worse, the tooltip-on-hover codepath just uses `entityNearest()`, so the intuitive nearest point has a tooltip, but then clicking selects a different point.

I went through the git history and couldn't see any reason why the "by x then y" approach was taken, but may well be missing something.

Hovering showing the nearest point:
image

After clicking at that exact same cursor location:
image

After this PR, the selected point is the expected one.

### DIFF
--- a/src/plots/linePlot.ts
+++ b/src/plots/linePlot.ts
@@ -1,4 +1,4 @@
-/**
+  /**
  * Copyright 2014-present Palantir Technologies
  * @license MIT
  */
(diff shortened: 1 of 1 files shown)


src/plots/linePlot.ts
```diff
@@ -424,46 +424,14 @@ export class Line<X> extends XYPlot<X, number> {
...
-    const entity = this.entityNearestByXThenY(point);
+    const entity = this.entityNearest(point);
     if (entity != null) {
       return [entity];
     } else {
       return [];
     }
   }
 
-  /**
```
themadcreator, line 435: Maybe we shouldn't delete this because technically this would be a public API break and require a major version bump

themadcreator (changes requested):
Let's leave the "unused" method since it's public

PR #3584, opened 2021-10-06
Support custom font sizes

michael-yx-wu:
Add support for custom label font size in the following places:
- [x] Bar
- [x] StackedBar
- [x] Scatter
- [x] Axes (Category, Numeric, Time)

cc @nifdee @billytrend

## Testing

SVG measures distance to text elements using the text baseline instead of the top-left corner like it does with most other elements. I've taken care to redraw or reset the `typesettable` measurer's cache where applicable to make subsequent renders draw the text in the right place. When playing with the quicktests, please double check that the label is positioned reasonably as font size changes i.e. it doesn't crowd out other elements, it's hidden when there is no space to draw, etc.

⚠️ The blueprint-bot's quicktest link is broken, so I recommend testing by checking out the repo and running `yarn start` and navigate to the links below. Gifs show what you should _expect_ to see. I recommend using node `8.10.0` with this repo, just to be consistent with the CI environment.

### Bar
http://localhost:9999/quicktests/dev/index.html#functional/bar_plot_label_font_size

### Scatter
http://localhost:9999/quicktests/dev/index.html#functional/scatter_plot_label_font_size
Something weird is happening w/ the scatter plot labels. Unclear why the size "stutters" when increasing font size. Note that internally, we do not use scatter plot labels.

### Category axis
http://localhost:9999/quicktests/dev/index.html#functional/category_axis_label_font_size

### Numeric axis
http://localhost:9999/quicktests/dev/index.html#functional/numeric_axis_label_font_size

### Time axis
http://localhost:9999/quicktests/dev/index.html#functional/time_axis_label_font_size


src/plots/stackedBarPlot.ts
```diff
@@ -106,9 +106,35 @@ export class StackedBar<X, Y> extends Bar<X, Y> {
...
+     }
+   }
+
   protected _setup() {
     super._setup();
-    this._labelArea = this._renderArea.append("g").classed(Bar._LABEL_AREA_CLASS, true);
+    this._labelArea = this._renderArea
+      .append("g")
+      .classed(Bar._LABEL_AREA_CLASS, true)
+      .classed(`label-${this._labelsFontSize}`, true);
```
styu: is there no way to dynamically style the font vs hardcoding css classes with specific font sizes?

michael-yx-wu: we need to set the font size for _children_ `<text>` elements so setting font-size with inline-styles doesn't work. this requirement comes from the `typesettable` measurement implementation which creates `<text>` elements. we could perhaps extend the `typesettable#AbstractMeasurer` and implement our own in plottable that forcibly inherits the font size from the parent, but we'd need to also re-implement the caching behavior (it caches dimensions of strings it's recently measured)

src/plots/barPlot.ts
```diff
@@ -73,6 +74,7 @@ export class Bar<X, Y> extends XYPlot<X, Y> {
   private _labelFormatter: DatumFormatter = Formatters.identity();
   private _labelsEnabled = false;
   private _labelsPosition = LabelsPosition.end;
+  protected _labelsFontSize: number = Label._DEFAULT_FONT_SIZE_PX;
```
styu: if there is only a range of font sizes supported due to needing styles for each pixel size (from your comment response below that sounds like the case?), is it actually right to type as `number` vs something like `12 | 13 | 14 | 15 ...` (gross I know)

michael-yx-wu: good point!

src/plots/stackedBarPlot.ts
```diff
@@ -106,9 +107,38 @@ export class StackedBar<X, Y> extends Bar<X, Y> {
     }
   }
 
+  /**
+   * Get the label font size in px.
+   */
```
michael-yx-wu: Since this extends `Bar`, I think I can remove these comments?

plottable.css
```diff
@@ -112,11 +112,84 @@
...
+
+.plottable .tick-label-container.label-16 text,
+.plottable .bar-label-text-area.label-16 text,
+.plottable .scatter-label-text-area.label-16 text {
+  font-size: 16px;
+}
+
+.plottable .tick-label-container.label-17 text,
+.plottable .bar-label-text-area.label-17 text,
+.plottable .scatter-label-text-area.label-13 text {
```
Nifdee: This should be 17, which would probably explain your "stutter" issue
```suggestion
.plottable .scatter-label-text-area.label-17 text {
```

src/axes/axis.ts
```diff
@@ -218,6 +243,21 @@ export class Axis<D> extends Component {
...
+        // clearing to remove outdated font-size classes
+        .attr("class", null)
+        .classed(`${Axis.TICK_LABEL_CLASS}-container`, true)
+        .classed(`label-${this._tickLabelFontSize}`, true);
+    } else {
+      this._tickLabelContainer = this.content()
+        .append("g")
+        .classed(`${Axis.TICK_LABEL_CLASS}-container`, true)
+        .classed(`label-${this._tickLabelFontSize}`, true);
+    }
```
Nifdee: Could we have the classes bits only once? 
```suggestion
    if (this._tickLabelContainer == null) {
      this._tickLabelContainer = this.content().append("g");
    }
    this._tickLabelContainer
      // clearing to remove outdated font-size classes
      .attr("class", null)
      .classed(`${Axis.TICK_LABEL_CLASS}-container`, true)
      .classed(`label-${this._tickLabelFontSize}`, true);
```

Nifdee (approved):
Tested locally, seems to work fine.

plottable.css
michael-yx-wu: Doh 🤦

styu (approved):

PR #3586, opened 2021-10-18
Bump version to 3.11.0 for release

michael-yx-wu:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "plottable",
   "description": "A modular charting library built on D3",
-  "version": "3.9.0",
+  "version": "3.11.0",
```
adidahiya, line 4: why 3.11 and not 3.10?

michael-yx-wu: I tried to release 3.10.0 but the publish failed because I didn't bump this version number. 3.11.0 since the 3.10.0 tag 
already exists now

styu (approved):

PR #3589, opened 2021-11-16
Use lodash-es and upgrade to TS@4.4

michael-yx-wu:


src/drawers/symbolDrawer.ts
```diff
@@ -56,7 +56,7 @@ export function makeSymbolCanvasDrawStep(
 
             // check attributes and symbol type
             const attrsSame = isAttributeValuesEqual(prevAttrs, attrs, ContextStyleAttrs);
-            const symbolGenerator = symbolAccessor(datum, index, this._dataset);
+            const symbolGenerator = symbolAccessor(datum, index, dataset);
```
michael-yx-wu, line 59: I wonder why this ever worked, since the `this` in a browser context refers to the `window` object and doesn't make any sense here. Maybe a transpilation bug that is surfacing now that we've upgraded typescript?

michael-yx-wu:
A lot of casting to `SVGRect`, `DOMRect`, `ClientRect`, etc. because newer definitions for these require us to define things like the `x`, `y`, and `toJSON` properties. Given things were working fine before, I think it's _okay_ to just forcibly cast

src/memoize/memoizeProjectors.ts
```diff
@@ -1,5 +1,5 @@
-
-import { MapCache, memoize as lodashMemoize } from "lodash";
+import { MapCache } from "lodash";
```
styu: is there no equivalent in lodash-es or even an interface we can copy over? else maybe this should be like this instead, since we no longer depend on lodash
```suggestion
import type { MapCache } from "lodash";
```

src/plots/rectanglePlot.ts
```diff
@@ -276,7 +276,7 @@ export class Rectangle<X, Y> extends XYPlot<X, Y> {
       y: attrToProjector["y"](datum, index, dataset),
       width: attrToProjector["width"](datum, index, dataset),
       height: attrToProjector["height"](datum, index, dataset),
-    };
+    } as SVGRect;
```
styu: i think it would be more status quo if you changed the following line to `Pick<DOMRect, "x" | "y" | "width" | "height">` instead

https://github.com/palantir/plottable/blob/2db30e420a54075b8e555aa9db3731b05bfd20a9/src/utils/domUtils.ts#L222

src/plots/scatterPlot.ts
```diff
@@ -194,7 +194,7 @@ export class Scatter<X, Y> extends XYPlot<X, Y> {
       y: entity.position.y - entity.diameter / 2,
       width: entity.diameter,
       height: entity.diameter,
-    };
+    } as DOMRect;
```
styu: same as above

src/utils/domUtils.ts
```diff
@@ -33,7 +33,7 @@ export function elementBBox(element: SimpleSelection<any>) {
   try {
     bbox = (<any> element.node()).getBBox();
   } catch (err) {
-    bbox = { x: 0, y: 0, width: 0, height: 0 };
+    bbox = { x: 0, y: 0, width: 0, height: 0 } as SVGRect;
```
styu: maybe the return type of the function should use `Pick` similar to the other suggestion

src/utils/stackingUtils.ts
```diff
@@ -8,7 +8,8 @@ import * as d3 from "d3";
 import { Dataset } from "../core/dataset";
 import { IAccessor } from "../core/interfaces";
 
-import { memoize, MemoizedFunction } from "lodash";
+import { MemoizedFunction } from "lodash";
```
styu:
```suggestion
import type { MemoizedFunction } from "lodash";
```

src/scales/interpolatedColorScale.ts
```diff
@@ -96,8 +96,8 @@ export class InterpolatedColor extends Scale<number, string> {
   /**
    * Generates the converted QuantitativeScale.
    */
-  private _d3InterpolatedScale() {
-    return this._colorScale.range([0, 1]).interpolate(this._interpolateColors());
+  private _d3InterpolatedScale(): supportedScale {
```
adidahiya: nit: I'd suggest renaming the local type def to `SupportedScale` in this file

```suggestion
  private _d3InterpolatedScale(): SupportedScale {
```

src/memoize/memoizeProjectors.ts
michael-yx-wu: Unfortunately, it's only defined in `@types/lodash`: https://github.com/DefinitelyTyped/DefinitelyTyped/search?q=MapCache

yarn.lock
```diff
@@ -4470,6 +4477,11 @@ locate-path@^2.0.0:
     p-locate "^2.0.0"
     path-exists "^3.0.0"
 
+lodash-es@^4.17.15:
+  version "4.17.21"
+  resolved "https://registry.yarnpkg.com/lodash-es/-/lodash-es-4.17.21.tgz#43e626c46e6591b7750beb2b50117390c609e3ee"
+  integrity sha512-mKnC+QJ9pWVzv+C4/U3rRsHapFfHvQFoFB92e52xeyGMcX6/OlIl78je1u8vePzYZSkkogMPJ2yjxxsb89cxyw==
+
```
michael-yx-wu: shoot, I will pint this to 4.17.15 to match w/ the internal version

yarn.lock
```diff
@@ -4470,6 +4477,11 @@ locate-path@^2.0.0:
     p-locate "^2.0.0"
     path-exists "^3.0.0"
 
+lodash-es@^4.17.15:
+  version "4.17.15"
```
michael-yx-wu, line 4481: Matches our internal version

adidahiya (approved):

PR #3592, opened 2021-11-19
Use separate lodash modules per function

michael-yx-wu:
Treeshaking in downstream libs not possible because we are targeting
commonjs. Not changing to esnext since it might be a breaking change.

### DIFF
--- a/src/memoize/signature.ts
+++ b/src/memoize/signature.ts
@@ -13,7 +13,7 @@
  * We must hand-write a signature for each custom class we wish to support.
  */
 
-import * as isPlainObject from "is-plain-object";
+import isPlainObject from "is-plain-object";
 
 import { Dataset } from "../core/dataset";
 import { Scale } from "../scales/scale";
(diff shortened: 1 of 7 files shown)


michael-yx-wu:
kick bot

src/memoize/signature.ts
```diff
@@ -13,7 +13,7 @@
  * We must hand-write a signature for each custom class we wish to support.
  */
 
-import * as isPlainObject from "is-plain-object";
+import isPlainObject from "is-plain-object";
```
michael-yx-wu, line 16: alternatively, could've kept the `import * as isPlainObject` and replaced `isPlainObject(obj)` with `isPlainObject.default(obj)`, but this seemed cleaner

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "compilerOptions": {
       "declaration": true,
+      "esModuleInterop": true,
```
styu: hm... i would prefer we not do this and make the `import * as` statements instead. This will probably break upstream consumers who are not yet using `esModuleInterop`

michael-yx-wu:
Closing in favor of https://github.com/palantir/plottable/pull/3594

PR #3605, opened 2023-01-05
chore: rename NPM auth env variable

adidahiya:


derekjobst (approved):
Issued a new NPM publishing token and added to CircleCI env variables as `NPM_AUTH_TOKEN`

PR #3619, opened 2026-09-15
Add a positionable label to GuideLineLayer (#3526)

### LINKED ISSUE #3526: Unable to align the text of a GuideLineLayer to anything other than left
## Bug report

I've got a bar graph using a GuideLineLayer in a similar way to the Space Race example at http://plottablejs.org/examples/spacerace/.

I have the same fault as can be seen with the example where when I hover over the last item (First Man on the Moon) as the label is aligned to the left of the gridline the text cannot be displayed due to lack of room.

The documentation refers to xAlignment which takes the arguments "left", "centre" and "right"which I assumed refers to the placement of the label.

However any change to either xAlignment or yAlignment has no effect and the label is always shown at the bottom left of the gridline regardless of these values.

My plan was to detect where the mouse is hovering over the graph in the onPointerMove and change the xAlignment accordingly.

Live example URL: (unable to get jsfiddle to work but this issue can be easily seen using the demo at http://plottablejs.org/examples/spacerace/)

Steps to repro:
1. Create a Plottable.Components.GuideLineLayer
1. Set the xAlignment to "centre"
1. Allow the graph to be drawn normally

### Expected behavior
The text label of the date for the gridline should be centre to the guideline

### Actual behavior
The text label of the date for the gridline is always bottom left regardless of what xAlignment, or yAlignment is set to. however calling the getter for these methods returns what I have set the value.

### Possible Solution
The text label should respect the alignment and update accordingly.

### Context
Ensuring that the date label can be seen regardless of where on the graph the mouse is currently hovering over.

### Environment

- Plottable version: 3.8.6
- Browser name/version: Google Chrome Version 75.0.3770.100 (Official Build) (64-bit)
- OS name/version: MacOS 10.14.5

ZayanKhan-12:
Fixes #3526.

## The problem

The issue reports that `xAlignment()`/`yAlignment()` on a `GuideLineLayer` have no effect on
its text, and that the text is stuck at the bottom left.

Two things are going on:

1. **`GuideLineLayer` has no text of its own.** It renders a single `<line>` and nothing else,
   so the label in the [Space Race example](http://plottablejs.org/examples/spacerace/) is a
   separate `Component` the caller composes next to it. There is no supported way to attach a
   label to the guide line.
2. **`xAlignment()`/`yAlignment()` could never have done this.** They place a `Component`
   inside the box its *parent* offers it. `GuideLineLayer._sizeFromOffer()` returns the whole
   offer, so those accessors are provably no-ops on this component — and even on a component
   that does not fill its offer, they move the component, never its contents. The getters
   returning the value that was set (as the issue notes) is expected; it just does not
   correspond to anything on screen.

So the alignment accessors aren't broken — the feature the reporter wanted is missing.

## The change

Give `GuideLineLayer` a real label, with its own placement accessors:

```js
guideLine
  .label("First Man on the Moon")   // null or "" removes it
  .labelPadding(6)                  // px from the line, and from the layer edge
  .labelXAlignment("left")          // "left" | "center" | "right"
  .labelYAlignment("top");          // "top" | "center" | "bottom"
```

The two axes mean different things depending on orientation, which keeps them symmetric:

|                   | vertical guide line             | horizontal guide line           |
| ----------------- | ------------------------------- | ------------------------------- |
| `labelXAlignment` | which **side** of the line      | slides **along** the line       |
| `labelYAlignment` | slides **along** the line       | which **side** of the line      |

*Side* means `"left"` ends one padding before the line, `"right"` starts one padding after it,
and `"center"` straddles it. *Along* means `"top"`/`"left"` sits one padding in from that edge
of the layer, `"bottom"`/`"right"` one padding in from the opposite edge, and `"center"` is
centred between them.

This directly supports what the issue wanted to do — flipping the label to the other side of
the line as the pointer nears the edge of the plot, so it stops getting clipped:

```js
guideLine.labelXAlignment(xScale.scale(d.year) > guideLine.width() / 2 ? "left" : "right");
```

I deliberately did **not** repurpose `xAlignment()`/`yAlignment()` for this. They are inherited
public API with a defined meaning, and quietly changing what they do in one subclass would
break anyone placing a `GuideLineLayer` in a `Table` or `Group`. Instead their doc comments now
state that they position the component within its parent, do nothing for a component that takes
all the space it is offered, and never move a component's own contents — which is the
distinction that caused the confusion in the first place.

## Compatibility

Additive. `label()` defaults to `null` and a layer without one renders exactly as before — the
label container is an empty `<g>`. `DragLineLayer` inherits the feature for free.

## Testing

`grunt test-ci` (compile + tslint + eslint + jscs + the full suite) passes: **1218 specs, 0
failures**, up from 1202 on `develop`.

16 new specs in `test/components/guideLineLayerTests.ts` cover the accessors (defaults,
chaining, case-insensitivity, and the rejection of bad padding/alignments), and assert exact
label geometry for every alignment in both orientations. The geometry specs measure the text
with `typesettable` in a throwaway element carrying the same CSS classes as the real label
container, so the expected coordinates are derived independently of the implementation rather
than read back from it. I sanity-checked they are not vacuous by mutating the placement
arithmetic — each mutation fails the specs that cover it.

There is also a spec pinning the behaviour the issue ran into: `xAlignment()`/`yAlignment()`
must leave the label where it is, and one covering a layer that has a label but no position
yet, where the label is held back rather than written to an invalid transform.

New quicktest at `quicktests/overlaying/tests/interactions/guideline_labels.js` demonstrates
the edge-flip, verified to render headlessly.

## Note on the second commit

`Add CLAUDE.md` is unrelated to the fix and is separated out so it can be dropped with a single
`git rebase` if you would rather not carry that file in-tree. It documents the build/test
commands, the component lifecycle, and repo conventions. One finding from writing it that may
be worth a look independently: `yarn test` maps to `grunt test-local`, which runs PhantomJS
against the already-built `test/tests.js` **without compiling first**, so it can report a pass
against stale code. `grunt test` and `grunt test-ci` both compile first and are unaffected.


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3620, opened 2026-09-15
Stop the cumulative transform walk at a fixed-position ancestor (#3437)

### LINKED ISSUE #3437: Static position on container messes up which plot point is hovered
Hovering over plot points can hover over the wrong point if there is a static container somewhere in the  chart's ancestors. 

The fix is to check if the ancestor statically positioned in `getCumulativeTransform()`. Something functionally equivalent to:

```js
        var styles = window.getComputedStyle(elem);
        if (styles.getPropertyValue("position") == "static") {
            return transform;
        }
```

ZayanKhan-12:
Fixes #3437.

## What's actually going on

`getCumulativeTransform()` walks from a Component's root element up to `<html>`, accumulating
each ancestor's scroll offset and, at each `offsetParent` boundary, its
`offsetLeft`/`offsetTop`. That is correct only while every ancestor genuinely moves the
element.

A **`position: fixed`** ancestor breaks both halves of that assumption:

1. It is laid out against the viewport, so its `offsetLeft`/`offsetTop` are *already* the
   client coordinates of its border box — nothing above it can move it.
2. Its `offsetParent` is `null`. That makes the `offsetParent === null` branch — which is only
   meant to fire on the first iteration — match again on `<body>`, restarting the offsetParent
   chain. The walk then subtracts `<body>`'s offsets and, at `<html>`, adds the **document's
   scroll offset**.

So pointer coordinates drifted by exactly the scroll position, and hover picked the wrong
entity — further off the more the page was scrolled. Measured directly:

```
fixed ancestor, page scrolled 250px
  root.getBoundingClientRect().top  = 56     <- truth
  getCumulativeTransform() gives    = 306    <- off by exactly the 250px scroll
```

## The change

Stop the walk once an element's computed position is `fixed`.

## On the suggested fix in the issue

The issue proposes bailing out when an ancestor is `position: static`. I did not apply that,
because it makes things worse rather than better — a static ancestor is the *normal* case, so
the walk stops almost immediately and never reaches the real `offsetParent`. I measured it
rather than assuming (methodology below):

| build | configurations wrong |
| --- | --- |
| `develop` today | 180 / 420 |
| with `return` on `position: static` (as suggested in the issue) | **420 / 420** |
| this PR | **0 / 420** |

The report was pointing at the right function, and `static` containers *are* present in the
broken cases — they're what typically sits between the chart and the fixed container — but they
aren't the trigger. The trigger is the fixed ancestor above them.

## How I verified it

I enumerated every combination of one and two wrapper elements drawn from
`{static, relative, absolute, fixed} × {plain, padding, margin, border, scrolling}`, built each
chain in headless Chrome, and probed a known client point inside the root element, comparing
`applyTransform(getCumulativeTransform(root), probe)` against ground truth from
`getBoundingClientRect()`. Run with the document both scrolled and unscrolled.

- 420 configurations, **180 wrong before, 0 after**.
- **Every** failing configuration contained a `position: fixed` ancestor, and **no**
  configuration without one was ever wrong — which is what localised the bug.
- Re-ran with `position: sticky` added to the space (650 configurations, 0 failures), and
  separately confirmed a genuinely stuck sticky element (`offsetTop` 600 while scrolled 600,
  `bcrTop` 0) still resolves correctly. Sticky is unaffected either way.

## Tests

6 new specs in `test/utils/mathUtilsTests.ts` under `Utils.Math.getCumulativeTransform`, three
of them regression guards for the ordinary cases (static ancestor, positioned ancestor,
scrolling ancestor) so the fix can't quietly break them.

The load-bearing one is `does not apply the document scroll offset`, which fails on `develop`
with exactly the reported symptom:

```
1) does not apply the document scroll offset:
   the document scroll offset is not added to a viewport-positioned element (y):
   expected 519 to be close to 19 +/- 0.5
```

`grunt test-ci` (compile + tslint + eslint + jscs + suite) passes: **1208 specs, 0 failures**,
up from 1202 on `develop`.

## Performance

This runs for every pointer event, so it should not get more expensive. The computed style is
now read once per ancestor and shared between the transform and position lookups, so the number
of `getComputedStyle` calls is unchanged — and the walk now visits fewer ancestors, since it
stops at the fixed element instead of continuing to `<html>`. `getElementTransform()` keeps its
signature and behaviour; it just delegates to the new `getStyleTransform()`.

## No quicktest

The README asks for one, and I'd normally add it, but a `position: fixed` container is
viewport-relative by definition — it would escape its cell and float over the whole quicktest
gallery, breaking every other test on the page. The unit tests cover the regression instead.
Happy to add one behind a separate standalone page if you'd prefer.

## Note on the second commit

`Add CLAUDE.md` is unrelated to the fix and separated so it can be dropped with a single
`git rebase`. It is byte-identical to the file proposed in #3619 (same blob SHA), so the two
PRs will not conflict — whichever lands first makes the other commit a no-op.


palantirtech:
Thanks for your interest in [palantir/plottable](https://github.com/palantir/plottable), @ZayanKhan-12! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

